PR #1, opened 2018-05-08
Integrate with travis-ci

EdDev:


.travis.yml
```diff
@@ -0,0 +1,12 @@
+sudo: required
```
phoracek: no need for this

.travis.yml
```diff
@@ -0,0 +1,12 @@
...
+dist: trusty
```
phoracek: i don't think this matters with tox

.travis.yml
EdDev: But we will need it for integration tests (as it will interact with NM). Do you suggest changing it only when we will need it?

.travis.yml
EdDev: All others projects are using it, not 100% if it is correct or not.
Why do you think we do not need it?
https://docs.travis-ci.com/user/reference/overview/#Virtualisation-Environment-vs-Operating-System

.travis.yml
phoracek: Let's add it once we fail with forbidden.

.travis.yml
phoracek: With sudo: required and dist: trusty we will get Ubuntu 14.04 VM, that sounds like good solution for NM testing, but it will not work (IMO), NM on Ubuntu 14 would be really outdated and API will be different with missing features and bugs.

I suggest we use sudo: false dist: thrusty to get a container and do proper NM testing either in a VM or run it on Jenkins.

EdDev: Ok, so this line stays the same anyway.
Regarding the NM version, we will have to solve this somehow, but I hope we can do it on travis.

tyll:
Does it maybe make more sense to use CentOS CI instead of travis to get the right NM version? https://dustymabe.com/2016/01/23/the-centos-ci-infrastructure-a-getting-started-guide/

And for integration tests, we might be able to use this, too:
https://github.com/linux-system-roles/test-harness

EdDev:
I am not familiar with CentOS CI, but this is an issue we should address after the PoC. Currently, travis will be used for unit tests and triggering coverage. I added a PR for it.

.travis.yml
```diff
@@ -0,0 +1,12 @@
...
+language: python
+python:
+  - 2.7
+  - 3.6
+
+install:
+  - pip install tox
+
+script:
+  - tox -e flake8,pylint,check-py27,check-py36
```
phoracek: tox alone is enough

.travis.yml
```diff
@@ -0,0 +1,12 @@
+sudo: false
+dist: trusty
+language: python
+python:
+  - 2.7
+  - 3.6
+
+install:
+  - pip install tox
```
phoracek: please add pbr as another dep for #3

.travis.yml
EdDev: Done

PR #5, opened 2018-05-13
Introduce a basic schema with validation tests

EdDev:
A basic schema is introduced, defined in json-schema[1] and implemented
by jsonschema[2] python implementation.

The full/partial definition for the nmstate schema will be introduced in
a seperate patch.

[1] http://json-schema.org
[2] https://github.com/Julian/jsonschema

### DIFF
--- a/libnmstate/schema.py
+++ b/libnmstate/schema.py
@@ -19,5 +19,13 @@
 #
 from __future__ import absolute_import
 
-from . import validator
-validator
+import json
+import pkgutil
+
+
+def load(schema_name):
+    return json.loads(
+        pkgutil.get_data('libnmstate', 'schemas/' + schema_name + '.json'))
+
+
+ifaces_schema = load('interfaces')
--- a/libnmstate/validator.py
+++ b/libnmstate/validator.py
@@ -19,5 +19,10 @@
 #
 from __future__ import absolute_import
 
-from . import validator
-validator
+import jsonschema as js
+
+from . import schema
+
+
+def verify(data, validation_schema=schema.ifaces_schema):
+    js.validate(data, validation_schema)
--- a/tests/lib/schema_validation_test.py
+++ b/tests/lib/schema_validation_test.py
@@ -19,5 +19,40 @@
 #
 from __future__ import absolute_import
 
-from . import validator
-validator
+import pytest
+
+import jsonschema as js
+
+import libnmstate
+
+
+def test_valid_basic_instance():
+    data = {
+        'interfaces': [
+            {
+                'name': 'lo',
+                'description': 'Loopback Interface',
+                'type': 'ethernet',
+                'state': 'down'
+            }
+        ]
+    }
+
+    libnmstate.validator.verify(data)
+
+
+def test_invalid_basic_instance():
+    data = {
+        'interfaces': [
+            {
+                'name': 'lo',
+                'description': 'Loopback Interface',
+                'type': 'ethernet',
+                'state': 'bad-state'
+            }
+        ]
+    }
+
+    with pytest.raises(js.ValidationError) as err:
+        libnmstate.validator.verify(data)
+        assert 'bad-state' in err.value.args[0]
--- a/tox.ini
+++ b/tox.ini
@@ -42,6 +42,8 @@ commands =
 [testenv:pylint]
 skip_install = true
 deps =
+    -rrequirements.txt
+    pytest==3.5.1
     pylint==1.8.4
 commands =
     pylint \
--- a/libnmstate/schemas/interfaces.json
+++ b/libnmstate/schemas/interfaces.json
@@ -1,0 +1,30 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+
+    "title": "Interfaces tree",
+    "type": "object",
+    "properties": {
+        "interfaces": {
+            "title": "Interfaces",
+            "type": "array",
+            "items": {
+                "title": "Interface",
+                "type": "object",
+                "properties": {
+                    "name": { "type": "string" },
+                    "description": { "type": "string" },
+                    "type": {
+                        "type": "string",
+                        "enum": ["ethernet", "bond", "vlan"]
+                    },
+                    "state": {
+                        "type": "string",
+                        "enum": ["absent", "up", "down", "unknown"]
+                    }
+                },
+                "required": ["name"]
+            }
+        }
+    },
+    "required": ["interfaces"]
+}
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,3 +4,5 @@
 #packageB!=4.4,>=4.0
 #packageC>=3.0.0
 #packageD!=0.13.0,<0.14,>=0.12.0
+
+jsonschema


src/libnmstate/schemas/interfaces.json
```diff
@@ -0,0 +1,30 @@
...
+            "type": "array",
+            "items": {
+                "title": "Interface",
+                "type": "object",
+                "properties": {
+                    "name": { "type": "string" },
+                    "description": { "type": "string" },
+                    "type": {
+                        "type": "string",
+                        "enum": ["ethernet", "bond", "vlan"]
```
phoracek: type should be required too

src/libnmstate/schemas/interfaces.json
```diff
@@ -0,0 +1,30 @@
...
+                    },
+                    "state": {
+                        "type": "string",
+                        "enum": ["absent", "up", "down", "unknown"]
```
phoracek: is this read only? if not we might want to define a default for it (i don't think it would affect validation, just for documentation purpose, so there is always a state passed to underlying app)

src/libnmstate/validator.py
```diff
@@ -0,0 +1,28 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+#
+# Refer to the README and COPYING files for full details of the license
+#
+from __future__ import absolute_import
+
+import jsonschema as js
+
+from . import schema
```
phoracek: it keeps schemaS, maybe rename it to schemas, so we don't need that ugly _schema parameter

tests/lib/schema_validation_test.py
```diff
@@ -0,0 +1,58 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+#
+# Refer to the README and COPYING files for full details of the license
+#
+from __future__ import absolute_import
+
+import pytest
+
+import jsonschema as js
```
phoracek: could be in the same group as pytest, both are external packages

tests/lib/schema_validation_test.py
```diff
@@ -0,0 +1,58 @@
...
+
+import libnmstate
+
+
+def test_valid_basic_instance():
+    data = {
+        "interfaces": [
```
phoracek: use single quotes '

tox.ini
```diff
@@ -14,6 +14,7 @@ setenv =
 deps =
     pytest-cov
     pytest==3.5.1
+    jsonschema
```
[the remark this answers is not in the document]
EdDev: Sure, after we merge that PR I will rebase and update.

tox.ini
```diff
@@ -50,6 +51,7 @@ commands =
 setenv =
     {[base]setenv}
 deps =
+    jsonschema
```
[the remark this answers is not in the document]
EdDev: Sure, after we merge that PR I will rebase and update.

src/libnmstate/schema.py
```diff
@@ -0,0 +1,34 @@
...
+# Refer to the README and COPYING files for full details of the license
+#
+from __future__ import absolute_import
+
+import json
+import os
+
+
+def load(schema_name):
+    self_path = os.path.dirname(__file__)
```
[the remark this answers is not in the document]
EdDev: Done

src/libnmstate/schemas/interfaces.json
```diff
@@ -0,0 +1,30 @@
+{
```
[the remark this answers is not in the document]
EdDev: I agree, I think it will go in that direction.
For now, it is hard to "see" what is the most suitable way to split it. We need some mass to see pros/cons.
Lets do it in steps.
Importing/Referencing it is supposed to be very easy.

src/libnmstate/schemas/interfaces.json
```diff
@@ -0,0 +1,30 @@
...
+    "$schema": "http://json-schema.org/draft-04/schema#",
+
+    "title": "Interfaces tree",
+    "type": "object",
+    "properties": {
+        "interfaces": {
+            "title": "Interfaces",
```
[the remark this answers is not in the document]
EdDev: This patch is just a schema foo, lets discuss it in the next PR which will include all interface common properties.

src/libnmstate/schemas/interfaces.json
```diff
@@ -0,0 +1,30 @@
...
+            "type": "array",
+            "items": {
+                "title": "Interface",
```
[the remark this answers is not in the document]
EdDev: Same, next PR.

src/libnmstate/schemas/interfaces.json
EdDev: Nope, it actually does not need to be required: Given an interface, a user may want to change its MTU without knowing of what type it is. Sounds reasonable (for now at least) to me.
Even true for removing an interface.
Anyway, the discussion fits the next PR.

src/libnmstate/schemas/interfaces.json
EdDev: The client may not be interested to change the state.

src/libnmstate/validator.py
EdDev: It is not a container, so I do not like the plural option. I don't have anything better, please propose something else.

tests/lib/schema_validation_test.py
EdDev: One is for testing and one is 3rd party helper here.

tests/lib/schema_validation_test.py
EdDev: Done

libnmstate/schema.py
```diff
@@ -0,0 +1,31 @@
...
+#
+# Refer to the README and COPYING files for full details of the license
+#
+from __future__ import absolute_import
+
+import json
+import pkgutil
+
+
+def load(schema_name):
```
phoracek: Why do we want to load sub-schemas? What would be a scenario for that? I imagine we get data with full description on input, we validate against the whole schema and that's it. Is it here because of we want to allow plugins to register for specific sub-schema?

libnmstate/validator.py
```diff
@@ -0,0 +1,28 @@
...
+# Refer to the README and COPYING files for full details of the license
+#
+from __future__ import absolute_import
+
+import jsonschema as js
+
+from . import schema
+
+
+def verify(data, _schema=schema.ifaces_schema):
```
phoracek: maybe schema_definition or description instead of _schema?

libnmstate/schema.py
EdDev: I do not have context here to your comment. We load a schema (or a list of them in the future, not sure yet), and can validate against it.
With the json-schema, one schema can reference others.

libnmstate/validator.py
EdDev: I will call it validation_schema.

libnmstate/schema.py
phoracek: In my opinion we should have main schema with the object that references other types (defined in other schema files). These other schema files will not contain the top level object, only definitions, therefore they cannot be used stand-alone for validation.

My question is, do we have a use case for validating subtrees? If not, we can use other schema files only for definitions.

EdDev: This is the current state, unless we find a different need, I guess we will stick with it.
The interfaces schema is currently the root schema, we may change it when the routing subtree will be introduced.

phoracek: I see now, thanks. All good then.

PR #12, opened 2018-05-23
Control interface state (up/down)

EdDev:


libnmstate/netapplier.py
```diff
@@ -0,0 +1,49 @@
...
+
+def _apply_ifaces_state(state):
+    client = nmclient.client()
+
+    for iface_state in state['interfaces']:
+        nmdev = client.get_device_by_iface(iface_state['name'])
+        if iface_state['state'] == 'up':
+            if nmdev.get_state() == nmclient.NM.DeviceState.ACTIVATED:
+                continue
+            client.activate_connection_async(device=nmdev)
```
phoracek: what if there is no connection for the interface? don't we need to create one?

libnmstate/netapplier.py
```diff
@@ -0,0 +1,49 @@
...
+        elif iface_state['state'] == 'down':
+            active_connection = nmdev.get_active_connection()
+            if active_connection:
+                client.deactivate_connection_async(active_connection)
+        else:
+            raise UnsupportedIfStateError(iface_state)
+
+
+class UnsupportedIfStateError(Exception):
```
phoracek: UnsupportedIfaceStateError

nmstatectl/nmstatectl.py
```diff
@@ -22,24 +22,40 @@
...
+    parser_show = subparsers.add_parser('show', help='Show network state')
     parser_show.set_defaults(func=show)
 
 
-def show():
+def setup_subcommand_set(subparsers):
+    parser_set = subparsers.add_parser('set', help='Set network state')
+    parser_set.add_argument('-f', '--file',
+                            help='File containing desired state')
+    parser_set.set_defaults(func=apply)
```
phoracek: why don't you use the same name "set", but "apply" instead?

tests/lib/netapplier_test.py
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+
+@mock.patch.object(netapplier.nmclient, 'NM')
+@mock.patch.object(netapplier.nmclient, 'client')
+class TestDevStateChange(object):
+    def test_apply_iface_state_up_to_up(self, mk_client, mk_nm, config):
+        mock_get_device_by_iface = mk_client.return_value.get_device_by_iface
+        mock_get_device_by_iface.return_value = MockNmDevice(devstate=UP)
+        mk_nm.DeviceState.ACTIVATED = UP
```
phoracek: why do you always set this to UP?

libnmstate/netapplier.py
EdDev: We assume it has one, which for the majority of the device types it is a valid assumption. We may need to change this assumption for a more generic solution. For now it is focused on the minimum PoC scenario we aim for.

libnmstate/netapplier.py
EdDev: I prefer to keep it short.

nmstatectl/nmstatectl.py
EdDev: Because "set" is reserved for dicts.

tests/lib/netapplier_test.py
EdDev: We mock here ACTIVATED and give it the value of UP. We consider activated as up, all the rest as down.

libnmstate/netapplier.py
phoracek: But then I get UnsupportedIfStateError out of blue, what's the "If"

PR #17, opened 2018-06-01
Add initial yaml support

tyll:
This was my initial idea to add yaml support.


nmstatectl/nmstatectl.py
```diff
@@ -50,12 +54,20 @@ def setup_subcommand_set(subparsers):
...
+                         separators=(',', ': ')))
 
 
 def apply(args):
     with open(args.file) as f:
-        state = json.load(f)
+        try:
+            state = json.load(f)
+        except ValueError:
+            f.seek(0)
```
phoracek: We want to add stdin support, right? Can we seek 0 from stdin? If not, it might be better to load input and than use loads.

tyll: thx, not sure it is a priority but seeking from stdin, so here is an update

nmstatectl/nmstatectl.py
```diff
@@ -50,12 +54,20 @@ def setup_subcommand_set(subparsers):
...
+        print(json.dumps(state, indent=4, sort_keys=True,
+                         separators=(',', ': ')))
 
 
 def apply(args):
     with open(args.file) as f:
-        state = json.load(f)
-        netapplier.apply(state)
+        statedata = f.read()
+    try:
```
EdDev: I find this problematic, because it shadows a bad json file.
So either the "set" should also have a "--yaml" flag, or attempt to guess the file type from the file extension.

nmstatectl/nmstatectl.py
```diff
@@ -50,12 +54,20 @@ def setup_subcommand_set(subparsers):
 
 
 def show(args):
-    print(json.dumps(netinfo.show(), indent=4, sort_keys=True,
-                     separators=(',', ': ')))
+    state = netinfo.show()
+    if args.yaml:
+        print(yaml.dump(state, default_flow_style=False))
+    else:
+        print(json.dumps(state, indent=4, sort_keys=True,
```
EdDev: It will also be nice to have a --json flag, to output a "pure" json format (for callers to consume).

nmstatectl/nmstatectl.py
tyll: I changed it to try only of each types depending on the first character.

nmstatectl/nmstatectl.py
tyll: I guess you mean no pretty printing with "pure". Callers can consume the pretty-printed format AFAIK, however I changed it to do this only on ttys. `nmstateclt show | cat` will show you the raw format. Is this what you meant? I did not squash it, yet, so you can review the changes.

nmstatectl/nmstatectl.py
```diff
@@ -56,18 +57,26 @@ def setup_subcommand_set(subparsers):
...
 
 def apply(args):
-    with open(args.file) as f:
-        statedata = f.read()
-    try:
+    with open(args.file) as statefile:
+        statedata = statefile.read()
+
+    # JSON dictionaries start with a curly brace
+    if statedata[0] == '{':
```
EdDev: If the file is empty it will raise in an non descriptive manner.
You need to handle that edge case (len(statedata) == 0).

nmstatectl/nmstatectl.py
EdDev: Nice.

nmstatectl/nmstatectl.py
EdDev: Other networking cli tools (ovs-vsctl for example) print a raw json output (non pretty as you said).
Not sure if it matters to a consumer, and I would expect it to use the lib and not the ctl anyway.
Your solution is nice, but in the code it looks ugly (too many if/s, too much options ...) lets take it out and assume the callers can consume it as you mentioned.

nmstatectl/nmstatectl.py
tyll: Good catch, should it be an error? And how would you like it to be reported? I guess non-zero returncode (for example 1) and a message to stderr "ERROR: statefile is empty". Or should be considert this as nothing needs to be done and just exit?

EdDev: Like all the other errors... raise. And better raise a dedicated error (EmptyFileError).
When raising, the argument can be the file path:
'''
class EmptyFileError(Exception):
    pass

raise EmptyFileError(file)
'''

tyll: Are you sure? This is in the tool and not the library, IMHO an exception is not a good user error message in this case. And is it really an error? If it is getting data from stdin and the other tool that pipes its output to nmstatectl does not have a state to apply, getting no data sounds like a valid scenario.

EdDev: You will have to catch all exceptions then: (missing file, no permission to open, etc..)
Then log it to stderr and so.

I guess you could solve it by raising at the error location, but translating it at the main entry point (from an exception to a stderr print/log in a generic fashion).
BTW, it is improper to exit the program in multiple places.

tyll: Perfect is the enemy of good. Ideally, all exceptions would be caught, however, if we just want exceptions to be shown to the user, `len(statedata) == 0` does not need to be caught... IMHO it is still better to not raise an exception when it is meant for the end user even when there are other exceptions that could be raised.

EdDev: There are several errors we could expect to see: Missing file, empty file, corrupted json/yaml file and so on. Which one you prefer the user to see as an exception and which one you want the user to just get an stderr output? What is the rule here?
We do not need perfect here, we just need consistency.
As an example, you could say that a bug in the code should throw an exception while an expected error in the input should just print the error ad exit with a non zero code.
The easiest path I guess is to throw.
An upgrade to that is to serialize the error using stderr for specific exceptions.
Lets just be consistent as much as we can.

PR #18, opened 2018-06-01
Do not require "-f/--file" to specify state

tyll:
Since the network state file is always requirer/not optional, specify it
as an argument, not as an option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,14 @@ nmstatectl show
 
 Change to desired state:
 ```shell
-nmstatectl set --file <desired-state.json>
+nmstatectl set desired-state.json
+```
+
+`nmstatectl` will also read from stdin when no file is specified:
+
+
+```shell
+nmstatectl set < desired-state.json
 ```
 
 Desired/Current state example:
@@ -67,6 +74,14 @@ Desired/Current state example:
     ]
 }
 ```
+
+The state is also supported as YAML, to get the current state in YAML format:
+
+```shell
+nmstatectl show --yaml
+```
+
+The `set` command accepts both YAML and JSON.
 
 ## Supported Interfaces:
 - bond
--- a/nmstatectl/nmstatectl.py
+++ b/nmstatectl/nmstatectl.py
@@ -19,6 +19,8 @@ from __future__ import absolute_import
 import argparse
 import json
 import logging
+import os
+import sys
 
 import yaml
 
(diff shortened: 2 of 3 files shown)


EdDev:
I was thinking we could use stdin as an alternative to --file, don't you think it is worth it?

tyll:
We can add that `-` as a file would  make it read from stdin, not sure if it is useful for the PoC. Usually you can use /dev/stdin in bash to specify stdin anyhow.

phoracek:
@tyll I would go with the `--f -`, it's somehow expected.

tyll:
I believe there are different expectations. `-f` is also used to force things somethings and since the state file is not optional for the set command, it should not be an option IMHO, see for example `sort`, `cut` or `cat`.

EdDev:
The reason that --file <file name> looks better is due to the explicit naming of the argument.
Perhaps we could specify something like this:
- "--file" is optional, if it is not specified, a default desired-state.json is expected to exist. That fits with most config files pattern.
- We could change the name to something like "-s , --desired-state-file".

Will this fit better?

tyll:
If the arguments should be explicitly named, then it would need to be `nmstatectl --action set --state foo.state`. I am not convinced, that making it an option, i.e. requiring any `--string` makes it better. And I also do not see this in other tools. For example xrdb is used to load the X ressources into an X session (just like nmstatectl is used to load the network state) and the command line would be `xrdb filename`. Since xrdb needs to file to read from, it does not make sense to make the filename an option.

Also I would not call the state a configuration file for nmstate, this would be a file that describes how nmstate behaves not a file that is used by nmstate.

EdDev:
Not all arguments, only the file one.
The comparison to the configuration was only to give an example of how a default state file could be used. Why do we need to require a file path to be given? That is not minimalistic.
If you have strong opinions on this, I will take the PR in, but I would feel more comfortable with another unix cli geek giving an opinion on this.
@dankenigsberg ,   @thom311 ?

tyll:
Good point about the minimalist example, I guess we could just read from stdin when no file is presented, just as `cat` or `grep` does (when the file is omitted). However, I guess endless hours where wasted for people to wait for `grep foo` when they meant `grep foo *`.

tyll:
ping?

EdDev:
Could you please update the README now?
Will be nice to see all the options on how to use it now in the commit message.

Specifically, is this supported?
cat desired-state.json | nmstatectl set

tyll:
I update the README, squashed/rebased the commits and mentioned all possibilities in the README.

PR #19, opened 2018-06-01
Initial IP settings support

tyll:
Add initial support to get and configure IP address. This needs nmstate to run the GLib MainLoop so that the async() methods to activate the connections are actually called. Inspecting /etc/sysconfig/network-script  shows here, that IP settings are changed as expected.

### DIFF
--- a/libnmstate/netapplier.py
+++ b/libnmstate/netapplier.py
@@ -18,8 +18,11 @@
 # Refer to the README and COPYING files for full details of the license
 #
 
+import socket
+
 from libnmstate import nmclient
 from libnmstate import validator
+from libnmstate.nmclient import NM
 
 
 def apply(desired_state):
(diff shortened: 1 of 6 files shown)


tyll:
This is how I made it work:
https://github.com/tyll/nmstate/commit/97cb145d27ed9195fe6e2bb2f6c3735f4f0eb0f7

It still needs an externally generated NM configuration for the interface.

Any idea why tox fails in Travis? it works locally...

tests/lib/mock_nm.py
```diff
@@ -0,0 +1,38 @@
...
+    def get_ip4_config(self):
+        return None
+
+    def get_connection(self):
+        return MockNmConnection()
+
+
+class MockNmDevice(object):
+    def __init__(self, devstate=DOWN, active_connection=MockNmConnection(),
+                 iface="lo"):
```
phoracek: single quotes please, the same for others unless there is a single quote inside the string

libnmstate/netapplier.py
```diff
@@ -33,12 +36,31 @@ def _apply_ifaces_state(state):
 
     for iface_state in state['interfaces']:
         nmdev = client.get_device_by_iface(iface_state['name'])
+        active_connection = nmdev.get_active_connection()
+
         if iface_state['state'] == 'up':
-            if nmdev.get_state() == nmclient.NM.DeviceState.ACTIVATED:
-                continue
-            client.activate_connection_async(device=nmdev)
+            ip4state = iface_state.get("ip")
```
phoracek: please keep the IP logic in separate functions, this would get really messy once we add support for more features

libnmstate/netinfo.py
```diff
@@ -45,17 +46,36 @@ def interfaces():
 def devices():
     client = nmclient.client()
 
-    devs = client.get_devices()
+    devlist = []
+    for device in client.get_devices():
+        ip4info = {}
```
phoracek: same here, please try to split IP logic and the rest, split it into more functions

tests/lib/mock_nm.py
tyll: This is a new rule for me, where does it come from?

EdDev:
#20 suggests a specific structure for the code.
Lets try to agree on the structure first so it will be simpler to extend the functionality (adding types and options).

tests/lib/mock_nm.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/python3 -tt
```
EdDev: I like the test helper module.
But what are these two first lines?
And we need to have the licensing statement here.

tyll: Good catch, I fixed this.

tyll:
I split this into #27, #28, #29  and it should address your comments

PR #20, opened 2018-06-03
Bond support

EdDev:


libnmstate/netapplier.py
```diff
@@ -48,5 +73,41 @@ def _apply_iface_admin_state(client, iface_state, nmdev):
...
 
 
-class UnsupportedIfaceStateError(Exception):
-    pass
+def _index_by_name(ifaces_state):
+    return {iface['name']: iface for iface in ifaces_state}
+
+
+def _build_connection_profile(iface_desired_state):
+    CON_PROFILE_PREFIX = 'nms-'
```
tyll: It is a good idea to discuss a naming scheme for profiles. IMHO just using the interface name as the profile name should suffice. Also we should maybe define some expectations about the systems that we are managing with nmstate, e.g. require that there is only one profile per device on the system, it is named after the device and it is limited to the device.

libnmstate/netapplier.py
```diff
@@ -48,5 +73,41 @@ def _apply_iface_admin_state(client, iface_state, nmdev):
...
+    if iface_type == 'bond':
+        bond_conf = iface_desired_state['link-aggregation']
+        bond_opts = _translate_to_nmbond_options(bond_conf)
+
+        settings.append(nm.bond.create_settings(bond_opts))
+
+    return nm.connection.create_profile(settings)
+
+
+def _translate_to_nmtype(type):
```
tyll: This could just be a dictionary:
```python
NM_TYPES = {'bond': nmclient.NM.SETTING_BOND_SETTING_NAME}
```

and you can access it with
```
NM_TYPES.get(type, 'unknown')
```

A wrapper to allow lookups in both directions will probably make sens to allow to translyte nmtypes back to nmstate types.

libnmstate/netapplier.py
```diff
@@ -18,20 +18,45 @@
...
+    ]
+    edit_ifaces = [
+        ifaces_desired_state[name] for name in
+        six.viewkeys(ifaces_desired_state) & six.viewkeys(ifaces_current_state)
+    ]
+
+    new_con_profiles = [_build_connection_profile(iface_desired_state)
+                        for iface_desired_state in add_ifaces]
+    for connection_profile in new_con_profiles:
+        client.add_connection_async(connection_profile, save_to_disk=True)
```
tyll: AFAIU this will not do anything as long as there is no GLib.MainLoop running. @thom311 can probably confirm this.

thom311: Gdbus is a multithreadded D-Bus library, meaning, it uses a separate thread to send/receive D-Bus messages. (I am not sure, but) I think if you just issue asynchronous requests, without ever iterating the main-loop, the requests will still be sent out (at least, if the process lives long enough for the other thread to process the send-queue).

Having said that, I think when asynchronous requests are made, it generally is advised to also listen and handle the response (if only to determine success/failure, and to synchronize with the completion of the request).

thom311: btw, for many D-Bus requests, NetworkManager needs to authenticate the caller, which generally requires that the process is still around. That means, while you can fire-and-forget style send a D-Bus request and exit (the process), NetworkManager will then be unable to learn the user-id of the process, and fail with permission-denied. For that reason, for many/most D-Bus requests against NetworkManager, you need to keep your process alive long enough. A `sleep` would work, but better of course is to wait for the asynchronous response.

libnmstate/netinfo.py
```diff
@@ -30,43 +31,66 @@ def show():
...
-            'type': resolve_nm_dev_type(dev.get_type_description()),
-            'state': resolve_nm_dev_state(dev.get_state()),
-        }
-        for dev in devs
-    ]
+def devices():
+    devlist = []
+    for dev in nm.device.list_devices():
+        devinfo = _devinfo_common(dev)
+        devinfo.update(_devinfo_bond(dev, devinfo))
```
tyll: IMHO if there are several methods that return the same object that they get as an parameter, it makes more sense to put them into a class.

libnmstate/netinfo.py
```diff
@@ -30,43 +31,66 @@ def show():
...
+
 def resolve_nm_dev_state(nm_state):
     if nm_state == nmclient.NM.DeviceState.ACTIVATED:
         return 'up'
     return 'down'
 
 
 def resolve_nm_dev_type(nm_type):
-    if nm_type != 'ethernet':
+    supported_types = (
```
tyll: The types should probably be a module constant.

libnmstate/netapplier.py
EdDev: In the design document, an option to target multiple system wide profiles is suggested.
We will probably not implement it any time soon (unless a scenario will present itself), but we should leave it as an option.
At the moment I have no objection to leaving it as you suggested, defining that the active/running profile has no prefix.
I will change it.

libnmstate/netapplier.py
EdDev: Yes, good idea. I was planning to have "translators", requiring the provider to register to the map with the type and handler function. But that will fit a later stage.
For now, I can work with your suggested approach.

libnmstate/netapplier.py
EdDev: It somehow works fine on my machine.
As we discussed already, if we can operate without the mainloop for now, I prefer to postpone handling this after the first PoC due to time restraints.

libnmstate/netinfo.py
EdDev: In this specific case, it implies that all info generators must be placed under the same "roof" and I see no good enough reason why that is needed.
I hope to see each type register to netinfo, with the relevant handler, making it simple for an additional type to be added without touching the core code.

libnmstate/netapplier.py
EdDev: Added the translator, as we need this shared between netinfo and the netapplier.

libnmstate/netapplier.py
tyll: It did not work here with the code already in master that also uses an async method. Addind the MainLoop helped.

EdDev:
TODO:  (on the last commit)
- Move the translator functionality from netapplier to the nm.translator.
- *_settings functions should change to *_setting

libnmstate/netapplier.py
EdDev: Do you mean the code from this PR?
Does a sleep works for you?

libnmstate/nm/translator.py
```diff
@@ -0,0 +1,66 @@
...
+    bond_opts = {'mode': bond_conf['mode']}
+    bond_opts.update(bond_conf.get('options', {}))
+    return bond_opts
+
+
+def _swap_dict_keyval(dictionary):
+    return {val: key for key, val in six.viewitems(dictionary)}
+
+
+_API2NM_IFACE_TYPES = None
```
phoracek: How is it a constant if you change it via global?

libnmstate/netapplier.py
```diff
@@ -18,32 +18,80 @@
...
-def _apply_ifaces_state(state):
+def _apply_ifaces_state(desired_state, current_state):
     client = nmclient.client()
 
-    for iface_state in state['interfaces']:
-        nmdev = client.get_device_by_iface(iface_state['name'])
-        if iface_state['state'] == 'up':
-            if nmdev.get_state() == nmclient.NM.DeviceState.ACTIVATED:
-                continue
+    ifaces_desired_state = _index_by_name(desired_state['interfaces'])
```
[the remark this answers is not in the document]
EdDev: Yes, it makes sense. I will change it, but note that this does not mean we will not need to change it back. There are always exceptions hiding out there.

libnmstate/netapplier.py
```diff
@@ -18,32 +18,80 @@
...
+    ifaces_current_state = _index_by_name(current_state['interfaces'])
+    add_ifaces = [
+        ifaces_desired_state[name] for name in
+        six.viewkeys(ifaces_desired_state) - six.viewkeys(ifaces_current_state)
+    ]
+    edit_ifaces = [
+        ifaces_desired_state[name] for name in
+        six.viewkeys(ifaces_desired_state) & six.viewkeys(ifaces_current_state)
+    ]
+
```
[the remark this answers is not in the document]
EdDev: We have only two options: add a new iface or edit an iface. Editing also includes removal.
The state is edited from up/down to absent.

libnmstate/netapplier.py
```diff
@@ -18,32 +18,80 @@
...
+        active_connection = nmdev.get_active_connection()
+        if active_connection:
+            client.deactivate_connection_async(active_connection)
+    else:
+        raise UnsupportedIfaceStateError(iface_state)
 
 
-class UnsupportedIfaceStateError(Exception):
-    pass
+def _index_by_name(ifaces_state):
```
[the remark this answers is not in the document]
EdDev: I do not understand the question.

libnmstate/nm/bond.py
```diff
@@ -0,0 +1,46 @@
...
+
+from libnmstate import nmclient
+
+from . import connection
+
+
+def create_setting(options):
+    bond_setting = nmclient.NM.SettingBond.new()
+    for option_name, option_value in six.viewitems(options):
+        success = bond_setting.add_option(option_name, option_value)
```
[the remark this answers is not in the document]
EdDev: What do you mean? bond_settings is a libnm object.

libnmstate/netapplier.py
```diff
@@ -28,21 +28,29 @@ def apply(desired_state):
...
-            active_connection = nmdev.get_active_connection()
-            if active_connection:
-                client.deactivate_connection_async(active_connection)
-        else:
-            raise UnsupportedIfaceStateError(iface_state)
+    elif iface_state['state'] == 'down':
+        active_connection = nmdev.get_active_connection()
+        if active_connection:
+            client.deactivate_connection_async(active_connection)
+    elif iface_state['state'] == 'absent':
```
EdDev: This is from #24

libnmstate/netapplier.py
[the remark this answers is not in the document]
tyll: I believe @phoracek relates to the name that contains "index". I guess what you mean is _get_by_name().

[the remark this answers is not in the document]
tyll: Another possible name would be `_interface_by_name()`.

PR #22, opened 2018-06-08
Refresh nmclient after changing state

tyll:
This is required to be able to have a chance to see changes after
applying a new state:

```python
    previous_state = netinfo.show()
    netapplier.apply(state)
    import time; time.sleep(10)
    current_state = netinfo.show()
```

Without this, `current_state` and `previous_state` will be the same as
long as `netapplier.aplly()` does not run `GLib.MainLoop`.


EdDev:
What causes a refresh in case the mainloop is used?

libnmstate/netapplier.py
```diff
@@ -26,6 +26,7 @@ def apply(desired_state):
     validator.verify(desired_state)
 
     _apply_ifaces_state(desired_state)
+    nmclient.client(refresh=True)
```
EdDev: How about just adding the refresh on netapplier and netinfo call to the client? It is temporary anyway.

tyll:
The main loop makes the client process DBus messages with the new information.

libnmstate/netapplier.py
tyll: I am not sure that I understand you, can you rephrase your suggestion?

EdDev: In both netapplier and netinfo modules, we have an existing call to fetch the client: client = nmclient.client()
I think it will be clearer for the time being to add there the "refresh", instead of here, because the call to client is expected to return the client object, and not using the returned value is weird.

tyll: Do you mean to always use a fresh nmclient instead of refreshing it only when there were changes?

EdDev: Partially yes: To get a fresh client on each entrance to netinfo and netapplier.
If there will be another call to client (in the same modules or others), it will not refresh.

tyll: I changed it, but I guess now the caching of the client does not make sense anymore.

PR #23, opened 2018-06-08
Role backend

tyll:
The first commit contains a PoC to use the Network Linux System Role as a backend. It is not nice but it works. I added the commits to report the IP state from my other branch and afaics it works better than the original code. Here is some test data for manual tests:

```
sudo ip link add nm type veth peer name peernm
cat > 'down-test.json' <<EOF
{
    "interfaces": [
        {
            "name": "nm",
            "state": "down",
            "type": "ethernet"
        }
    ]
}
EOF
cat > 'test-up.json' <<EOF
{
    "interfaces": [
        {
            "ip": {
                "enabled": true,
                "addresses":
                [
                    {
                        "ip": "192.0.72.8",
                        "prefix-length": 24
                    }
                ]
            },
            "name": "nm",
            "state": "up",
            "type": "ethernet"
        }
    ]
}
EOF

```

### DIFF
--- a/libnmstate/netapplier.py
+++ b/libnmstate/netapplier.py
@@ -18,8 +18,12 @@
 # Refer to the README and COPYING files for full details of the license
 #
 
-from libnmstate import nmclient
 from libnmstate import validator
+
+from network_connections import ArgValidator_ListConnections
+from network_connections import Cmd
+from network_connections import LogLevel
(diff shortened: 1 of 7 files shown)


tyll:
One issue with this is that it will remove all IP addresses when just state=up is used without specifying the IP addresses.

PR #25, opened 2018-06-11
Allow to limit state to specified interfaces

tyll:
This makes it easier to create a partial state to change.

### DIFF
--- a/nmstatectl/nmstatectl.py
+++ b/nmstatectl/nmstatectl.py
@@ -17,6 +17,7 @@
 from __future__ import absolute_import
 
 import argparse
+import fnmatch
 import json
 import logging
 import os
@@ -48,6 +49,10 @@ def setup_subcommand_show(subparsers):
     parser_show.set_defaults(func=show)
     parser_show.add_argument('--yaml', help='Output as yaml', default=False,
                              action='store_true')
+    parser_show.add_argument(
+        '--only', default='*',
+        help='Show only specified interfaces (comma-separated)'
+    )
 
 
 def setup_subcommand_set(subparsers):
@@ -66,8 +71,27 @@ def print_state(state, use_yaml=False):
                          separators=(',', ': ')))
 
 
+def _filter_interfaces(state, patterns):
+    """
+    return the states for all interfaces from `state` that match at least one
+    of the provided patterns.
+    """
+    showinterfaces = []
+
+    for interface in state['interfaces']:
+        for pattern in patterns:
+            if fnmatch.fnmatch(interface['name'], pattern):
+                showinterfaces.append(interface)
+                break
+    return showinterfaces
+
+
 def show(args):
     state = netinfo.show()
+    if args.only != '*':
+        patterns = [p for p in args.only.split(',')]
+        state['interfaces'] = _filter_interfaces(state, patterns)
+
     print_state(state, use_yaml=args.yaml)
 
 


nmstatectl/nmstatectl.py
```diff
@@ -44,6 +45,10 @@ def setup_subcommand_show(subparsers):
     parser_show.set_defaults(func=show)
     parser_show.add_argument('--yaml', help='Output as yaml', default=False,
                              action='store_true')
+    parser_show.add_argument(
+        '--only', default='*',
```
EdDev: We will probably have other entities (beyond interfaces), so how about just naming it "interfaces" or "filter-interfaces"?
"--interfaces" looks nice to me.

And we can use "all" instead of "*", sometimes "*" is reserved for other things in bash.

nmstatectl/nmstatectl.py
```diff
@@ -20,6 +20,7 @@
 from __future__ import absolute_import
```
EdDev: Commit message:
Can you please use "filter" instead of "limit"?
It will be also useful if you give an example in the commit message on options to use it.

nmstatectl/nmstatectl.py
```diff
@@ -55,6 +60,17 @@ def setup_subcommand_set(subparsers):
 
 def show(args):
     state = netinfo.show()
+    if args.only != '*':
+        patterns = [p for p in args.only.split(',')]
+        showinterfaces = []
+
+        for interface in state['interfaces']:
```
EdDev: Please extract it to a function, so it will look something nice like this:

for pattern in patterns:
     filtered_interfaces += _filter_interfaces(state['interfaces'], pattern)
state['interfaces'] = filtered_interfaces

nmstatectl/nmstatectl.py
tyll: I guess we could just use the same filter for the other entities as well. I would address this when there are other entities and the use-cases are clear. Since "*" is the default, it is not needed to specify `*` on the command-line, therefore I would keep it this way.

nmstatectl/nmstatectl.py
tyll: done

nmstatectl/nmstatectl.py
tyll: I do not see how this could work, because if `_filter_interfaces()` returns `None`, the None object would be added to the state. And the resulting function would just be
```
def _filter_interfaces(state, pattern):
    if fnmatch.fnmatch(state['name'], pattern):
        return state
```

I am no conviced that this is enough code for a funtion.

tyll:
How do we proceed here?

nmstatectl/nmstatectl.py
EdDev: Fair enough regarding the asterisk.

Regarding the name, this is a command line api, changing it later is problematic.
How would the user use --only to filter other things?
(Interesting enough, it the last few days several "other" things came up frequently)

tyll: I assume the other entities would have names as well and --only would only the entities for which the pattern matches, since interfaces are the only entities currently, it can only match interfaces.

nmstatectl/nmstatectl.py
EdDev: _filter_interfaces returns a list of interfaces states, if none pass the filter, then it returns an empty list.
It is much more readable that way (IMO).

nmstatectl/nmstatectl.py
EdDev: Yes, they will probably have a name or some other id, but also a "context" (interfaces).
Can we use --only with another argument that gives the context as well? Maybe something crazy like: --only interfaces:a,b,c,d
Sorry for nagging on this, but it feels too generic and we will have another tree branch pretty soon.

EdDev:
Please leave only the filter patch without the tests in this PR.
We will take care of the tests later on, there are several issues with them.

tyll:
I removed the tests from this PR.

PR #28, opened 2018-06-13
Support reading IP state

tyll:


tyll:
This is just reading the IP state from #19

libnmstate/netinfo.py
```diff
@@ -91,6 +96,30 @@ def _devinfo_bond(dev, devinfo):
             }
     return {}
 
+def _devinfo_ip(dev, devinfo):
+    ip4info = {}
+    connection = dev.get_active_connection()
```
EdDev: Please move all NM related details to nm.* , here we should only be concern with the higher format,
Something like this:
```
def _devinfo_ipv4(dev):
    act_con = nm.connection.get_device_active_connection(dev)
    return {'ipv4': nm.ipv4.get_state(act_con)}
```

libnmstate/netinfo.py
```diff
@@ -91,6 +96,30 @@ def _devinfo_bond(dev, devinfo):
...
+    if connection:
```
EdDev: We can reduce indentation and make it easier to reach by exiting early:

if not act_connection:
    return {'enabled': False}
...

libnmstate/netinfo.py
```diff
@@ -91,6 +96,30 @@ def _devinfo_bond(dev, devinfo):
...
+        ip4config = connection.get_ip4_config()
+        if ip4config:
+            ip4info['enabled'] = True
+            addresslist = []
+
+            addresses = ip4config.get_addresses()
+            for address in addresses:
```
EdDev: This can be converted into a list comprehension.

libnmstate/nmclient.py
```diff
@@ -39,3 +39,16 @@ def client(refresh=False):
     if _nmclient is None or refresh:
         _nmclient = NM.Client.new(None)
     return _nmclient
+
+
+def connection_ensure_setting(connection, setting_type):
```
EdDev: This is used in the next patch?

tyll: Sorry, this was left from before and can go.

tyll:
This was modified in #34

PR #29, opened 2018-06-13
Support setting IPv4 addresses

tyll:
It works for the on-disk config but not for the runtime config but the error is in some other code, it works when the state is applied twice.


tyll:
Changes were done in #35

PR #31, opened 2018-06-18
Decouple nm provider

EdDev:
The NM package should include all the logic on how to access NM.
The translator needs to include all the translation logic between the formats while the individual NM modules (which represents the NM API structure) should provide access to NM API.

### DIFF
--- a/libnmstate/nm/translator.py
+++ b/libnmstate/nm/translator.py
@@ -18,6 +18,8 @@
 # Refer to the README and COPYING files for full details of the license
 #
 
+import copy
+
 import six
 
 from libnmstate import nmclient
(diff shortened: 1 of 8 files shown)


libnmstate/nm/translator.py
```diff
@@ -59,6 +59,35 @@ def get_bond_options(bond_conf):
...
+
+    @staticmethod
+    def get_bond_info(bondinfo):
+        bond_options = bondinfo.get('options')
+        if not bond_options:
+            return {}
+        bond_slaves = bondinfo['slaves']
+
+        bond_mode = bond_options['mode']
+        del bond_options['mode']
```
phoracek: doing this you change original object passed as argument, it would be safer to create a copy.

PR #44, opened 2018-06-28
Use address instead of addresses to match the spec

tyll:


EdDev:
Yes, we should use this one.
Also, we need to fix the top level 'interfaces', as it should look like this:
```
interfaces:
  interface:
    - name: eth0
      type: ethernet
    - name: eth1
      type: ethernet
```

And if we want to be consistent, 'slaves' should be 'slave' for bonds.

tyll:
Do we still want to do this or keep it as it is?

EdDev:
Yes, I want to be consistent with https://tools.ietf.org/html/rfc8344.
Could you please rebase? The schema needs to be updated.

tyll:
I  rebased it, the schema already uses address afaics:
https://github.com/nmstate/nmstate/blob/master/libnmstate/schemas/operational-state.yaml#L250

PR #45, opened 2018-07-01
libnmstate: Support ovs-bridge

EdDev:
OVS bridges are now reported with their supported properties and
connected ports/intefaces.

Note: OVS bonds and internal interfaces are not supported yet.

### DIFF
--- a/libnmstate/netinfo.py
+++ b/libnmstate/netinfo.py
@@ -28,9 +28,13 @@ def show():
 
 def interfaces():
     info = []
-    for dev in nm.device.list_devices():
-        devinfo = nm.device.get_device_common_info(dev)
+
+    devices_info = [(dev, nm.device.get_device_common_info(dev))
+                    for dev in nm.device.list_devices()]
+
+    for dev, devinfo in devices_info:
         type_id = devinfo['type_id']
+
         iface_info = nm.translator.Nm2Api.get_common_device_info(devinfo)
 
         act_con = nm.connection.get_device_active_connection(dev)
@@ -39,6 +43,10 @@ def interfaces():
         if nm.bond.is_bond_type_id(type_id):
             bondinfo = nm.bond.get_bond_info(dev)
             iface_info.update(_ifaceinfo_bond(bondinfo))
+        elif nm.ovs.is_ovs_bridge_type_id(type_id):
+            iface_info['bridge'] = nm.ovs.get_ovs_info(dev, devices_info)
+        elif nm.ovs.is_ovs_port_type_id(type_id):
+            continue
 
         info.append(iface_info)
 
--- a/libnmstate/nm/ovs.py
+++ b/libnmstate/nm/ovs.py
@@ -1,0 +1,102 @@
+#
+# Copyright 2018 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+from libnmstate import nmclient
+
+from . import connection
+from . import device
+
+
+def is_ovs_bridge_type_id(type_id):
+    return type_id == nmclient.NM.DeviceType.OVS_BRIDGE
+
+
+def is_ovs_port_type_id(type_id):
+    return type_id == nmclient.NM.DeviceType.OVS_PORT
+
+
+def is_ovs_interface_type_id(type_id):
+    return type_id == nmclient.NM.DeviceType.OVS_INTERFACE
+
+
+def get_ovs_info(bridge_device, devices_info):
+    port_profiles = _get_slave_profiles(bridge_device, devices_info)
+
+    return {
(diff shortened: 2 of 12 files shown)


libnmstate/nm/ovs.py
```diff
@@ -0,0 +1,102 @@
...
+        bridge_options['stp'] = bridge_setting.props.stp_enable
+        bridge_options['rstp'] = bridge_setting.props.rstp_enable
+        bridge_options['fail-mode'] = bridge_setting.props.fail_mode or ''
+        bridge_options['mcast-snooping-enable'] = (
+            bridge_setting.props.mcast_snooping_enable)
+
+    return bridge_options
+
+
+def _get_slave_profiles(master_device, devices_info):
```
tyll: I was wondering if it makes sense to avoid `master`/`slave` and use `leader`/`member` instead.

libnmstate/netinfo.py
```diff
@@ -28,9 +28,13 @@ def show():
 
 def interfaces():
     info = []
-    for dev in nm.device.list_devices():
-        devinfo = nm.device.get_device_common_info(dev)
+
+    devices_info = [(dev, nm.device.get_device_common_info(dev))
+                    for dev in nm.device.list_devices()]
+
+    for dev, devinfo in devices_info:
```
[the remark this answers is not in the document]
EdDev: I needed to pass on the list of devices (devices_info) in order to detect the bridge slaves and the port slaves.
Hopefully, libnm can provide a "get_slaves" functionality like it has for the bond, then we will be able to avoid this.

libnmstate/nm/ovs.py
EdDev: master/slave is a naming convention I am familiar with, not only from NM but in general.
No sure from where leader/member originates, could you share why you think it is more appropriate?
One can also suggest parent/child naming, or layer-up/layer-down, but somehow when it relates to network stuff, I always end up with master/slave.

tyll: Wikipedia explains this:
https://en.wikipedia.org/wiki/Master/slave_(technology)#Appropriateness_of_usage

EdDev:
Known problems:
- Removing a bridge is possible, but it will not delete the port profiles and will detach the current interface profiles.
- The port properties, like 'vlan-tag' have not been tested at all.

tyll:
The schema does not match, not sure how I got here:
```
nmstatectl show
  File "/root/.local/bin/nmstatectl", line 10, in <module>
    sys.exit(main())
  File "/root/.local/lib/python2.7/site-packages/nmstatectl/nmstatectl.py", line 44, in main
    return args.func(args)
  File "/root/.local/lib/python2.7/site-packages/nmstatectl/nmstatectl.py", line 90, in show
    state = netinfo.show()
  File "/root/.local/lib/python2.7/site-packages/libnmstate/netinfo.py", line 25, in show
    validator.verify(report)
  File "/root/.local/lib/python2.7/site-packages/libnmstate/validator.py", line 33, in verify
    js.validate(data, validation_schema)
  File "/root/.local/lib/python2.7/site-packages/jsonschema/validators.py", line 541, in validate
    cls(schema, *args, **kwargs).validate(instance)
  File "/root/.local/lib/python2.7/site-packages/jsonschema/validators.py", line 130, in validate
    raise error
jsonschema.exceptions.ValidationError: {'state': 'down', 'type': 'ovs-interface', 'name': 'testbvr', 'ipv4': {'enabled': False}} is not valid under any of the given schemas

Failed validating u'oneOf' in schema[u'properties'][u'interfaces'][u'items']:
    {u'oneOf': [{u'allOf': [{u'properties': {u'type': {u'enum': [u'unknown'],
                                                       u'type': u'string'}}},
                            {u'$ref': u'#/definitions/interface-rw'},
                            {u'$ref': u'#/definitions/interface-ro'}]},
                {u'allOf': [{u'$ref': u'#/definitions-ethernet/interface'},
                            {u'$ref': u'#/definitions/interface-rw'},
                            {u'$ref': u'#/definitions/interface-ro'}]},
                {u'allOf': [{u'$ref': u'#/definitions-bond/interface'},
                            {u'$ref': u'#/definitions/interface-rw'},
                            {u'$ref': u'#/definitions/interface-ro'}]},
                {u'allOf': [{u'$ref': u'#/definitions-ovs-bridge/interface'},
                            {u'$ref': u'#/definitions/interface-rw'},
                            {u'$ref': u'#/definitions/interface-ro'}]},
                {u'allOf': [{u'$ref': u'#/definitions-dummy/interface'},
                            {u'$ref': u'#/definitions/interface-rw'},
                            {u'$ref': u'#/definitions/interface-ro'}]}],
     u'required': [u'name'],
     u'title': u'Interface',
     u'type': u'object'}

On instance[u'interfaces'][4]:
    {'ipv4': {'enabled': False},
     'name': 'testbvr',
     'state': 'down',
     'type': 'ovs-interface'}

```

tyll:
The commit message of the last commit starts with `bnmstate` instead of `libnmstate`.

tyll:
I created the testbvr with `ovs-vsctl add-br testbvr`.

EdDev:
You used an interface type which is not supported yet: "ovs-interface". It is used for OVS internal interface and other OVS created devices. For the moment, only system interfaces are supported.
For bridges, you need to use "ovs-bridge".

tyll:
It happens with `nmsate show`, so it is reporting something that it does
not support in the schema.

Edward Haas <notifications@github.com> schrieb am Do., 5. Juli 2018, 13:29:

EdDev:
Ok, I see it now. It can be reported but cannot be defined and it is not appearing in the schema.
I will add it to the schema and raise if it is detected on set.
Thanks.

EdDev:
@tyll , I pushed the changes. Could you try it now?

tyll:
@EdDev looks better now, thank you!

PR #47, opened 2018-07-18
NM mainloop

EdDev:
The NM provider is executing libnm async funtions.
For proper operation and in order to serialize the actions, GLib
main-loop use has been embedded through a mainloop object.

The async funtions against libnm are registered to the mainloop object
and then executed one after the other from that queue.
The callback functions for each async command are responsible to execute
the next action in the mainloop object queue.

### DIFF
--- a/libnmstate/netapplier.py
+++ b/libnmstate/netapplier.py
@@ -15,6 +15,8 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
+from contextlib import contextmanager
+
 import collections
 import copy
 import six
@@ -54,8 +56,18 @@ def _apply_ifaces_state(interfaces_desired_state, interfaces_current_state):
 
     generate_ifaces_metadata(ifaces_desired_state, ifaces_current_state)
 
-    _add_interfaces(ifaces_desired_state, ifaces_current_state)
-    _edit_interfaces(ifaces_desired_state, ifaces_current_state)
+    with _setup_providers():
+        _add_interfaces(ifaces_desired_state, ifaces_current_state)
+        _edit_interfaces(ifaces_desired_state, ifaces_current_state)
+
+
+@contextmanager
+def _setup_providers():
+    mainloop = nmclient.mainloop()
+    yield
+    if mainloop.actions_exists():
+        mainloop.execute_next_action()
+        mainloop.run(timeout=20)
 
 
 def generate_ifaces_metadata(ifaces_desired_state, ifaces_current_state):
--- a/libnmstate/netapplier.py
+++ b/libnmstate/netapplier.py
@@ -27,6 +27,10 @@ from libnmstate import nmclient
 from libnmstate import validator
 
 
+class ApplyError(Exception):
+    pass
+
+
 def apply(desired_state):
     validator.verify(desired_state)
     validator.verify_capabilities(desired_state, netinfo.capabilities())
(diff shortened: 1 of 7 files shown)


EdDev:
Agreed, I will cover tests and docs about it in the coming days. Thanks!

thom311:
> The NM provider is executing libnm async funtions.

This makes it sound like there could be multiple providers for nmstate in the future. In my opinion, that is a misconception about what nmstate should do at its core.

---

The branch adds the `_Mainloop` class, which is an improvement. It's better than before, on that grounds the branch should be merged.

However, I am still not convinced that the `_Mainloop` makes anything simpler than just running the mainloop directly. Now, you have this `mainloop` instance. You can queue a list of actions (`push_action()`) and run them all together via the contextmanager `_setup_providers()`. But essentially, it iterates the mainloop one by one (`execute_next_action()`). In fact, even internal callbacks like `_delete_connection_callback()` are responsible to chain up the mainloop by calling `execute_next_action()`.

That is not what glib applications usually do (AFAIK). Usually, they would do something like
```C
main() {
    /* some initalization */
    g_main_loop_run();
    /* shutdown */
    return success;
}
```
There is only one place where the mainloop is started (with `run`, not `iterate`).

Note, that in the (very) rare case that you need it, you can create multiple GLib mainloops, and run them nested. NetworkManager uses that nowhere at all, though our tests do it, and glib internally does it.

On the other hand, if you design an API (a library), the library offers an async API to initiate requests. It does not (itself) iterate the mainloop. Instead, the library user invokes async requests (by calling a function of the library), the function registers callbacks to the mainloop, and expects the caller to run the mainloop. In particular, the async function can internally invoke a series of asynchronous operations, before invoking the final callback. Also, the pending async request must be cancellable (for example, by using glib's `GCancellable`), which instructs the library to abort the call in a timely manner and invoke the callback. The callback must be invoked always exactly once (also when cancelled).

The reason for `_Mainloop` is synchronous API like `def delete(dev)`. I think building a synchronous library on top of something asynchronous (D-Bus communication) is suboptimal. Provided you want part of nmstate to be a public library, or that the code is structured in a way, where there is an (internal) library of helper functions. Clearly, at some point, you need to wait for the result of the async API.  But that should outside the scope of this "library", and being moved to the highest layers of the code -- the one that also calls `g_main_loop_run()`.

That's how I think I'd do it. The network system-linux-role is a counter example for that, because it does not provide a library. The entire role is said "highest layer", building on libnm's asynchronous API. For anything sufficiently complex, that does not seem the best solution to me.


But sure. Whatever works for you :)

EdDev:
I will try to answer what I can, but this may require a broader discussion.
The patch is trying to provide a basic framework to run async actions in a serialized manner. The serialization is accomplished by adding to the callback tail a call to the next action. Potentially, an action callback may call multiple actions.
We call the first action before running the mainloop to start the sequence, I guess we could do so internally in the mainloop.run() if you think it is better/nicer.

nmstate is aimed to apply a declarative state and/or present it. I am not clear how you see it being used by a user in an async manner. The library is the programmatic frontend, at least this is the use case I see at the moment.

NM is the core provider/backend of nmstate, it is in our best interest to keep it the only one used whenever possible. However, we already know that there are vendor specific devices that need to be supported, especially for OpenStack. There are also edge cases where we may need a quick temporary solution to support some options which are not available at a specific version. Having this flexibility is important IMO for making it interested and useful for multiple potential consumers/users.

PR #50, opened 2018-07-22
Transaction support

EdDev:
Use NM checkpoint feature to implement transactions.

Based on #47

### DIFF
--- a/libnmstate/nm/checkpoint.py
+++ b/libnmstate/nm/checkpoint.py
@@ -1,0 +1,113 @@
+#
+# Copyright 2018 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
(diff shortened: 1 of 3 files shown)


libnmstate/nm/checkpoint.py
```diff
@@ -0,0 +1,106 @@
...
+    dbus = None
+
+
+DBUS_STD_PROPERTIES_IFNAME = 'org.freedesktop.DBus.Properties'
+
+_nmdbus_manager = None
+
+
+def has_checkpoint_capability():
+    return dbus is not None
```
phoracek: It's not so much of a checkpoint capability, rather dbus capability. Should not be this module and other dbus/NM modules under the same hood? Sharing dbus connection.

libnmstate/nm/checkpoint.py
```diff
@@ -0,0 +1,106 @@
...
+            # this command should fail.
+            self.destroy()
+        else:
+            self.rollback()
+            raise
+
+    def create(self):
+        devs = []
+        timeout = self._timeout
+        cp_flags = 0x07
```
phoracek: please use a constant

libnmstate/nm/checkpoint.py
EdDev: Currently the checkpoint implementation is implemented using dbus directly and not libnm. This will probably change after Centos/RHEL 7.6 is released.
Therefore, checkpoint is available if the system has python-dbus there.
dbus is not a capability, that is an implementation detail of checkpoint.

libnmstate/nm/checkpoint.py
EdDev: Done

PR #57, opened 2018-08-03
Report and set: mtu,speed, auto-negotiation and duplex

tyll:


libnmstate/nm/wired.py
```diff
@@ -0,0 +1,37 @@
...
+
+from libnmstate import nmclient
+
+
+def get_info(active_connection):
+    """
+    Provides the current active values for an active connection.
+    """
+    info = {}
+    # FIXME: Maybe use
```
EdDev: Looks like we fetch the "desired" config, is it assured to be the actual? I was wandering the same for the other info we collect.

libnmstate/nm/wired.py
```diff
@@ -0,0 +1,37 @@
...
+    # active_connection.get_devices()[0].get_applied_connection(0) to get the
+    # real runtime connection...
+    try:
+        remote_connection = active_connection.get_connection()
```
EdDev: Seems we just use it as a "connection" and not as a "remote" one, but it is a nit.

libnmstate/nm/wired.py
```diff
@@ -0,0 +1,37 @@
...
+        wired_config = remote_connection.get_setting(nmclient.NM.SettingWired)
```
EdDev: Why not use get_setting_wired() ?

libnmstate/nm/wired.py
```diff
@@ -0,0 +1,37 @@
...
+        mtu = wired_config.get_property(nmclient.NM.SETTING_WIRED_MTU)
```
EdDev: Using wired_config.props.mtu is a bit nicer IMO.

libnmstate/nm/wired.py
```diff
@@ -0,0 +1,37 @@
...
+        info['mtu'] = mtu
+    except AttributeError:
```
EdDev: Will this pattern work well with future properties here?
There seem to be two levels to this: The existence of the wired config and the existence of the specific property.

EdDev:
I feel bad to ask it (as I have not done well with this myself), but any chance you can add a test for the wired module?

libnmstate/nm/wired.py
tyll: This might be something that we need to address in general. Currently we are using `dev.get_active_connection()` but from a discussion with @thom311 I understood that the active connection is only a pointer and might show to a configuration that is not yet, applied. But Using `dev.get_applied_connection(0)` would return the live configuration.

In general I think I would wish that there was a clear settings interface as a class that settings such as for the wired parameters could inherit. Then the parent class could provide an interface to the connection that would need to be changed or inspected to return the actual info. Maybe just provide the connection to the constructor and have two methods apply and show, where apply gets the nmstate to apply and show returns the partial state with the settings for the respective NM setting.

libnmstate/nm/wired.py
tyll: In NM it is a RemoteConnection (as opposed to a SimpleConnection that `dev.get_applied_connection(0)` returns.

libnmstate/nm/wired.py
tyll: I hope to reuse the code and make it more generic when adding more settings. Then only the argument to `get_setting()` is generic, which is a bit nicer that to do something like `getattr(remote_connection, 'get'setting_wired')()`.

libnmstate/nm/wired.py
tyll: I did not know `wired_config.props`... Maybe @thom311 can explain what is better to be used. :-)

libnmstate/nm/wired.py
tyll: We will need an extra try/except block for each setting to make sure that only individual settings are omitted.

tyll:
I would like to add a test but I tried to figure out how to properly do the setting, first. Will need to address this in the week with August 13 because I will be at the Fedora Flock conference in the mean time.

libnmstate/nm/wired.py
EdDev: I do not see how you will make it more generic this way, I only see here the need to depend on nmclient.NM.* (and require a separate mock for it), especially when for each of these you need to check specific attributes.
But I guess we will see it when you add another property.

libnmstate/nm/wired.py
EdDev: Under the hood it should use the same calls.

tyll:
So I rebased it to report the MTU from the device property and set it via a wired setting. However, strictly speaking getting the MTU information does now not belong in the wired module since it is not reading the wired configuration.

libnmstate/nm/applier.py
```diff
@@ -194,4 +200,5 @@ def _build_connection_profile(iface_desired_state, base_con_profile=None):
         ovs_port_options = iface_desired_state.get('options')
         settings.append(ovs.create_port_setting(ovs_port_options))
 
-    return connection.create_profile(settings)
+    new_profile = connection.create_profile(settings)
```
EdDev: Why is this change needed?

libnmstate/nm/applier.py
```diff
@@ -68,6 +69,7 @@ def prepare_edited_ifaces_configuration(ifaces_desired_state):
             cur_con_profile = connection.get_device_connection(nmdev)
         new_con_profile = _build_connection_profile(
             iface_desired_state, base_con_profile=cur_con_profile)
+
```
EdDev: nit: not related change

libnmstate/nm/wired.py
```diff
@@ -0,0 +1,47 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+from libnmstate import nmclient
+
+DEFAULT_MTU = 1500
+
+
+def create_setting(iface_state, base_con_profile):
+    mtu = iface_state.get("mtu")
```
EdDev: Single quote please.

libnmstate/nm/wired.py
```diff
@@ -0,0 +1,47 @@
...
+    if not mtu:
+        return None
+
+    if base_con_profile:
+        wired_setting = base_con_profile.get_setting_wired()
```
EdDev: This is not how it works with the connection-settings, see: duplicate_settings()
I remember it was a recommendation from thaller.
Could you please re-check this?

libnmstate/nm/wired.py
```diff
@@ -0,0 +1,47 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+from libnmstate import nmclient
+
+DEFAULT_MTU = 1500
```
EdDev: Who is using this one (or: why do we need it?)?

libnmstate/nm/wired.py
tyll: done

libnmstate/nm/wired.py
tyll: removed it, it was leftover from the previous version

libnmstate/nm/applier.py
tyll: it makes debugging easier because it allows to add `import pdb; pdb.set_trace()` after the assignment before the function returns.

libnmstate/nm/applier.py
tyll: done

libnmstate/nm/wired.py
tyll: @thom311 does this need to be changed somehow?

libnmstate/nm/applier.py
EdDev: If you'll need to embed such lines in the code, then you can split it when you actually debug.
Anyway, with this reasoning we will have to double or triple the lines of code in any program, I prefer to be driven by readability and cleanness and not by a rare need of using a debugger. A temporary variable is nice only if it gives the reader more information (and sometime if it helps in avoiding breaking a line into two).

tyll: I prefer temporary variables in general because they allow to digest the code in smaller pieces. I can also remove this change if you prefer it.

EdDev: It is a nit, I will not fight over it.

libnmstate/nm/wired.py
```diff
@@ -0,0 +1,61 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+from libnmstate import nmclient
+
+
+def create_setting(iface_state, base_con_profile):
```
EdDev: I would prefer to see this function do one thing and not two.
"create" should mean it is creating a new setting.
"update" or "set" can be a function for editing an existing setting.

So we would have something like this:
if base_con_profile:
    wired_setting = wired.duplicate_setting(base_con_profile)
else:
     wired_setting = wired.create_setting(...) # No need for args if there are no "defaults".
if wired_setting:
    wired.set_setting(wired_setting, speed=speed, mtu=mtu,...)

tyll: AFAIU the `else` is wrong because the `base_con_profile` might not contain a `wired_setting` (at least in theory, maybe it will always have one). IMHO the decision whether or not to create a setting would be better in the module to make it more clear for later refactoring in an interface to know what to do.

EdDev:
I do not see the unit tests.

tyll:
Fixed the commit, the unit tests should be there now.

EdDev:
@tyll , the PR failed CI on flake8, could you please have a look? I also do not see the new wired module covered in the report, although the tests have been running.

tyll:
I fixed the flake issues (I hope).

libnmstate/nm/wired.py
```diff
@@ -0,0 +1,147 @@
...
+    duplex = ethernet.get('duplex')
+    auto_negotiation = ethernet.get('auto-negotiation')
+
+    if not (mtu or speed or duplex or auto_negotiation):
+        return None
+
+    if auto_negotiation == YES:
+        auto_negotiation = True
+
+    if auto_negotiation == NO:
```
EdDev: elif fits here

libnmstate/nm/wired.py
```diff
@@ -0,0 +1,147 @@
...
+        wired_setting.set_property(nmclient.NM.SETTING_WIRED_AUTO_NEGOTIATE,
+                                   auto_negotiation)
+
+    return wired_setting
+
+
+# based on:
+# https://github.com/rlisagor/pynetlinux/blob/master/pynetlinux/ifconfig.py
+# https://elixir.bootlin.com/linux/v4.19-rc1/source/include/uapi/linux/ethtool.h
+def minimal_ethtool(interface):
```
EdDev: Could you create a dedicated module for this? ethtool.py under libnmstate would be nice.
Also, please review VDSM ethtool module, it includes socket protection (making sure it is closed):
https://github.com/oVirt/vdsm/blob/master/lib/vdsm/network/ethtool.py

libnmstate/nm/wired.py
```diff
@@ -0,0 +1,147 @@
...
+
+    if auto_negotiation is not None:
+        wired_setting.set_property(nmclient.NM.SETTING_WIRED_AUTO_NEGOTIATE,
+                                   auto_negotiation)
+
+    return wired_setting
+
+
+# based on:
+# https://github.com/rlisagor/pynetlinux/blob/master/pynetlinux/ifconfig.py
```
EdDev: I prefer you mention this in the commit message, comments in code are bad (they rot) but if we have them, they better explain the non trivial flow.

libnmstate/nm/wired.py
```diff
@@ -0,0 +1,147 @@
...
+def get_info(device):
+    """
+    Provides the current active values for a device
+    """
+    info = {}
+
+    iface = device.get_iface()
+    try:
+        info['mtu'] = device.get_mtu()
+    # maybe all device actually have a get_mtu() method
```
EdDev: Please remove the comment.

libnmstate/nm/wired.py
```diff
@@ -0,0 +1,147 @@
...
+    except AttributeError:
+        pass
+
+    ethernet = {}
+    try:
+        ethernet['speed'] = device.get_speed()
+    # DeviceWifi, DeviceGeneric, DeviceTun do not have a get_speed() method
```
EdDev: Same, please remove the comment. You can mention it in the doctext and/or commit message.

libnmstate/nm/wired.py
```diff
@@ -0,0 +1,147 @@
...
+    except AttributeError:
+        pass
+
+    # Duplex state is not available via Network Manager:
```
EdDev: Same, please use the doctext or commit message.
You could also add the BZ as reference.

tests/lib/nm/wired_test.py
```diff
@@ -0,0 +1,75 @@
...
+import pytest
+
+from lib.compat import mock
+
+from libnmstate import nm
+
+
+@pytest.fixture
+def NM_mock():
+    with mock.patch.object(nm.connection.nmclient, 'NM') as m:
```
EdDev: nm.wired.nmclient should be mocked.

libnmstate/nm/wired.py
tyll: Do I Need `IFREQ_FORMAT = '16sPi'` - it works with `16sP` currently without the buffer length. I do not see any indication that there should be a buffer length. I added a finally statement to close the socket.

libnmstate/nm/wired.py
tyll: The complex flow is deducted from the references, therefore it make sense to me to add this to the docstring.

libnmstate/nm/wired.py
tyll: I added the BZ but I do not see how adding this to the commit message or doctext would improve this since the context to the line of code would be lost.

tests/lib/nm/wired_test.py
tyll: fixed

libnmstate/nm/wired.py
EdDev: How about moving all this to a function, there you can use doctext.

libnmstate/nm/wired.py
```diff
@@ -0,0 +1,96 @@
...
+    iface = device.get_iface()
+    try:
+        info['mtu'] = int(device.get_mtu())
+    except AttributeError:
+        pass
+
+    ethernet = {}
+    try:
+        ethernet['speed'] = int(device.get_speed())
+    # DeviceWifi, DeviceGeneric, DeviceTun do not have a get_speed() method
```
EdDev: Somehow the comment here is gone.
I think it is clear without this comment.

tyll: As a sign of good faith I removed it. :-)

PR #60, opened 2018-08-23
Convert the schema format to YAML

EdDev:


tyll:
I did not see any errors related to the schema when testing with bond interfaces, however there seem to be other errors present, see #61 and #62. Since nmstatectl sometimes work and then AFAIU the schema is checked and it fails when I change something obviously broken, this can be merged IMHO.

PR #65, opened 2018-09-02
Set admin state once when master/slave interfaces are involved

EdDev:
When there are bonds or OVS bridges, the current admin-state handling may execute an activation/deactivation multiple times on a single iface (on slaves in particular).

We should avoid this and perform the action once per device.

### DIFF
--- a/libnmstate/nm/applier.py
+++ b/libnmstate/nm/applier.py
@@ -15,6 +15,8 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
+import six
+
 from . import bond
 from . import connection
 from . import device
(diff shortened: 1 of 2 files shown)


libnmstate/nm/applier.py
```diff
@@ -79,28 +81,33 @@ def prepare_edited_ifaces_configuration(ifaces_desired_state):
...
+        action(dev)
+
 
-def _set_dev_state(iface_state, func_action):
+def _get_devs_from_iface_state(iface_state):
     nmdev = device.get_device_by_name(iface_state['name'])
+    devs = []
     if nmdev:
-        devs = [nmdev]
+        devs += [nmdev]
```
tyll: Why do you use `+=` here? AFAICS `devs` will always be empty so it can just be assigned to the new list.

libnmstate/nm/applier.py
```diff
@@ -79,28 +81,33 @@ def prepare_edited_ifaces_configuration(ifaces_desired_state):
...
+            devs_actions.update({dev: device.delete for dev in devs})
         else:
             raise UnsupportedIfaceStateError(iface_desired_state)
 
+    for dev, action in six.viewitems(devs_actions):
+        action(dev)
+
 
-def _set_dev_state(iface_state, func_action):
+def _get_devs_from_iface_state(iface_state):
```
tyll: How about naming this `_get_affected_devices(interface_state)`. This specifies which devices are meant, uses full words and does not repeat itself.

libnmstate/nm/applier.py
```diff
@@ -79,28 +81,33 @@ def prepare_edited_ifaces_configuration(ifaces_desired_state):
 
 
 def set_ifaces_admin_state(ifaces_desired_state):
+    devs_actions = {}
     for iface_desired_state in ifaces_desired_state:
+        devs = _get_devs_from_iface_state(iface_desired_state)
         if iface_desired_state['state'] == 'up':
-            _set_dev_state(iface_desired_state, device.activate)
+            devs_actions.update({dev: device.activate for dev in devs})
         elif iface_desired_state['state'] == 'down':
```
tyll: Maybe check `iface_desired_state['state'] in ('down', 'absent')` to show that the same action happens for both cases.

libnmstate/nm/applier.py
EdDev: Just to express it is already defined. After it was defined at the top, one may assume it will not get overwritten afterwards. Feels safer for future changes.

libnmstate/nm/applier.py
EdDev: Done.
I left the 'iface' short cut as we use it across the code base and it is a known keyword.

libnmstate/nm/applier.py

PR #67, opened 2018-09-04
Support setting and getting a decription

tyll:
This contains commits from #57

### DIFF
--- a/libnmstate/nm/user.py
+++ b/libnmstate/nm/user.py
@@ -1,0 +1,75 @@
+#
+# Copyright 2018 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+""" Use a Network Manager User Setting to store and retrieve information that
+does not fit somewhere else such as an interface's description.
+
+https://lazka.github.io/pgi-docs/#NM-1.0/classes/SettingUser.html
+"""
+
+from libnmstate.nm import connection as nm_connection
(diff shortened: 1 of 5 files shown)


tests/lib/nm/user_test.py
```diff
@@ -0,0 +1,51 @@
...
+from libnmstate import nm
+
+
+@pytest.fixture
+def NM_mock():
+    with mock.patch.object(nm.user.nmclient, 'NM') as m:
+        yield m
+
+
+def test_create_setting_None(NM_mock):
```
EdDev: Naming of tests suggestion:
- test_create_setting_with_no_values
- test_create_setting_given_existing_config
- test_create_setting_given_non_existing_config

EdDev:
Two other comments:
- The commit subject: Can you please add 'interface'?  Description may appear in other places.
- Question: The test coverage decreased although you added tests, any idea why?

libnmstate/nm/user.py
```diff
@@ -0,0 +1,72 @@
...
+#
+""" Use a Network Manager User Setting to store and retrieve a device
+description:
+
+https://lazka.github.io/pgi-docs/#NM-1.0/classes/SettingUser.html
+"""
+
+from libnmstate.nm import nmclient
+
+NMSTATE_DESCRIPTION = "nmstate.description"
```
EdDev: This is interface specific, we may need other descriptions in the future.

tests/lib/nm/user_test.py
tyll: It tests the case where no settings needs to be created, how about `test_create_no_setting`?

libnmstate/nm/user.py
tyll: changed to `nmstate.interface.desciption`

tests/lib/nm/user_test.py
EdDev: sounds good, thanks.

tyll:
Change the commit message. The tests do not cover everything and the coverage is less than the coverage of the other code I guess. A problem I see with the tests is that they need so much mocking that I wonder if we actually test it properly instead of just how we assume that NM works. Ideally we would mock NM in a way that it assumes a certain network configuration and the API works as designed instead of mocking the API hoping that the resembles the API properly.

EdDev:
I agree, the over mocking needed makes these tests hard to write and even harder to maintain.
We could decide not to test these low level driver modules, reasoning that they do not contain any significant business logic and the burden of maintaining them is too high.
I think we can take this decision when we will cover them using the integration/functional tests.

PR #69, opened 2018-09-06
Fix cleanup metadata before verify + adding sleep

EdDev:


libnmstate/netapplier.py
```diff
@@ -118,6 +119,16 @@ def generate_ifaces_metadata(ifaces_desired_state, ifaces_current_state):
     )
 
 
+def clean_ifaces_metadata(ifaces_state):
```
tyll: How about calling this `remove_ifaces_metadata` since it does not clean the metadata but cleans the state from the metadata by removing the metatada.

EdDev: ok

PR #71, opened 2018-09-06
Do not merge duplex and speed for auto-negotiation

tyll:
When auto-negotiation is enabled, duplex and speed ethernet settings
should not be merged from the current state to be able to distinguish
whether a user specified the settings or not.

### DIFF
--- a/libnmstate/netapplier.py
+++ b/libnmstate/netapplier.py
@@ -47,6 +47,7 @@ def _apply_ifaces_state(interfaces_desired_state, verify_change):
     ifaces_desired_state = _index_by_name(interfaces_desired_state)
     ifaces_current_state = _index_by_name(netinfo.interfaces())
 
+    ifaces_desired_state = sanitize_ethernet_state(ifaces_desired_state)
     generate_ifaces_metadata(ifaces_desired_state, ifaces_current_state)
 
     with _transaction():
@@ -117,6 +118,22 @@ def generate_ifaces_metadata(ifaces_desired_state, ifaces_current_state):
         get_slaves_func=_get_ovs_slaves_from_state,
         set_metadata_func=_set_ovs_bridge_ports_metadata
     )
+
+
+def sanitize_ethernet_state(ifaces_desired_state):
+    """
+    If auto-negotiation, speed and duplex settings are not provided, they need
+    to be set to None to not override them with the values from the current
+    settings since the current settings are read from the device state and not
+    from the actual configuration.  This makes it possible to distiguish
+    whether a user specified these values in the later configuration step.
+    """
+    for iface_state in six.viewvalues(ifaces_desired_state):
+        ethernet = iface_state.setdefault("ethernet", {})
+        ethernet.setdefault("auto-negotiation", None)
+        ethernet.setdefault("speed", None)
+        ethernet.setdefault("duplex", None)
+    return ifaces_desired_state
 
 
 def remove_ifaces_metadata(ifaces_state):
--- a/libnmstate/nm/wired.py
+++ b/libnmstate/nm/wired.py
@@ -44,14 +44,24 @@ def create_setting(iface_state, base_con_profile):
 
     if auto_negotiation:
         wired_setting.props.auto_negotiate = True
-        if not speed:
+        if not speed and not duplex:
             wired_setting.props.speed = 0
-
-        if not duplex:
             wired_setting.props.duplex = None
+
+        elif not speed:
+            ethtool_results = minimal_ethtool(str(iface_state['name']))
+            speed = ethtool_results['speed']
+        elif not duplex:
+            ethtool_results = minimal_ethtool(str(iface_state['name']))
+            duplex = ethtool_results['duplex']
 
     elif auto_negotiation is False:
         wired_setting.props.auto_negotiate = False
+        ethtool_results = minimal_ethtool(str(iface_state['name']))
+        if not speed:
+            speed = ethtool_results['speed']
+        if not duplex:
+            duplex = ethtool_results['duplex']
 
     if speed:
         wired_setting.props.speed = speed
--- a/tests/lib/nm/wired_test.py
+++ b/tests/lib/nm/wired_test.py
@@ -48,11 +48,18 @@ def test_create_setting_mtu(NM_mock):
     assert setting.props.mtu == 1500
 
 
-def test_create_setting_auto_negotiation_False(NM_mock):
+@mock.patch.object(nm.wired, 'minimal_ethtool',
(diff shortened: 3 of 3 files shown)


libnmstate/netapplier.py
```diff
@@ -119,6 +120,22 @@ def generate_ifaces_metadata(ifaces_desired_state, ifaces_current_state):
...
 
 
+def sanitize_ethernet_state(ifaces_desired_state):
+    """
+    If auto-negotiation is enabled and speed and duplex settings are not
+    provided, they need to be set to None to not override them with the values
+    from the current settings. This makes it possible to distiguish whether a
+    user specified these values in the later configuration step.
+    """
+    for interface in ifaces_desired_state:
```
EdDev: for iface_state in six.viewvalues(ifaces_desired_state)

libnmstate/nm/wired.py
```diff
@@ -44,12 +44,17 @@ def create_setting(iface_state, base_con_profile):
 
     if auto_negotiation:
         wired_setting.props.auto_negotiate = True
-        if not speed:
+        if not speed and not duplex:
```
EdDev: I think we should not set here the wired setting, only update the speed per what is needed.
Then, at the stage below, we apply everything in once location.
The function is becoming too long, please also consider splitting it into smaller functions.

I'm also not clear why you merged speed & duplex cases.

libnmstate/netapplier.py
tyll: Why is `viewvalues()` needed here?

libnmstate/nm/wired.py
tyll: If speed or duplex are provided by the user, both need to be set for Network Manager. The missing value needs to be taken from the current device state.

Setting the speed later could work, for duplex we would need a special value that makes it possible to distinguish it from not being set by the user since it is set to None.

EdDev: When we get here, all 3 values exist, either as None, as a user input (+ "sanitized" the input) or as a merge result.
My question is, why can't we handle them individually:
if speed is None:
   speed = ...
if duplex is None:
   duplex = ...

Then, apply explicitly.

And now that I think about it, why do we need to read the speed/duplex? The merge from the netinfo should do that, no?

libnmstate/netapplier.py
tyll: Thank you, good idea. I will change this.

libnmstate/nm/wired.py
tyll: There was a bug in the previous code, therefore I had to change it a little bit.

auto-negotiation, speed and duplex are set to None if they are not specified by the user and therefore are not merged in netinfo (this is what the sanitization step does) and we need to get them from the device if the values are needed.

When only auto-negotiation was specified and set to True, then nmstate needs to reset the speed and duplex settings to their default values (`0` and `None`). When we first set `speed` to `None` in this case and later want to apply this value to `props.duplex`, we cannot distinguish "duplex was not specified by the user" (in which case it would be `None`) from "duplex needs to be set to None  for auto-negotiation". Therefore I set the props directly when it is clear that it needs to be set to None.

For `speed` it could work if I change the conditional to `if speed is not None` since `0` is false.

EdDev: Maybe I'm missing something, but why do we care at the "apply" stage the original source of why `duplex` is `None`?
After the desired state has been processed: normalized/sanitized/canonicalized/metadatalized , we no longer should care why it is like it is, the lower "drivers" should just apply it.

What am I missing here?

tyll: Uh, somehow my comment got lost. :-/

I guess the confusion comes from where the nmstate state is converted into network manager config. NM allows more than nmstate, for example NM allows to define a "use the default speed setting" but nmstate only supports actual values. At the same time, nmstate allows to only set partial aspects of the state (for example disabling auto-negotiation without setting speed or duplex).

Somewhere in nmstate, we need to be able to decide what to do in which case. Currently this is happening in the NM create setting code for the wired setting and speed/duplex are passed as None, if the user did not specify them. However, when auto-negotiation is disabled and the user does not specify speed/duplex, then IMHO nmstate should use the current runtime speed/duplex settings. Therefore the NM create setting code needs to fetch them, since they are not in the function variables (they are set to None).

EdDev: > when auto-negotiation is disabled and the user does not specify speed/duplex, then IMHO nmstate should use the current runtime speed/duplex settings.

It makes sense to me too, I would just argue that this logic fits the "merging" step and not in the final step that applies the state. When looking at the function now, it is clearly doing more than one thing (at its abstract level), which could be avoided if we move this logic away.

I would have preferred smaller functions as well, IMO it is a bit messy at the moment.

Lets try to refactor this later on when we'll have functional/integration tests in place.

tyll:
CI should work now

tests/lib/nm/wired_test.py
```diff
@@ -49,10 +49,21 @@ def test_create_setting_mtu(NM_mock):
 
 
 def test_create_setting_auto_negotiation_False(NM_mock):
-    setting = nm.wired.create_setting(
-        {'ethernet': {'auto-negotiation': False}}, None)
+    def mock_ethtool_result(device):
+        assert device == "nmstate_test"
```
EdDev: - This is already checked in line 66.
- Please use single quotes.
- I think you can use return_value and avoid the function + place the mock as a decorator: nicer structure.

tyll: Handled all three items and squashed changes together in a single commit.

tests/lib/nm/wired_test.py
```diff
@@ -49,10 +49,19 @@ def test_create_setting_mtu(NM_mock):
 
 
 def test_create_setting_auto_negotiation_False(NM_mock):
-    setting = nm.wired.create_setting(
-        {'ethernet': {'auto-negotiation': False}}, None)
+    ethtool_result = {'speed': 1337, 'duplex': 'mocked', 'auto-negotiation':
+                      'mocked'}
+
+    with mock.patch('libnmstate.nm.wired.minimal_ethtool',
```
EdDev: nit: You could have used a decorator for this:
```
@mock.patch.object(nm.wired, 'minimal_ethtool', return_value={'speed': 1337, 'duplex': 'mocked', 'auto-negotiation': 'mocked'})
def test_create_setting_auto_negotiation_False(NM_mock, mock_ethtool):
```
(I may be wrong about the order of the args in the test function)

I'm ok with the current version as well, let me know if I should wait or accept as is.

tyll: if made it a decorator

PR #73, opened 2018-09-10
nmstatectl: Add edit command to edit state

tyll:
Use EDITOR variable to open an editor with the current state and apply
it after the editor is closed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,12 @@ Change to desired state:
 nmstatectl set desired-state.json
 ```
 
+Edit current state of eth3 in a text editor:
+```shell
+nmstatectl edit --only eth3
+```
+
+
 `nmstatectl` will also read from stdin when no file is specified:
 
 
--- a/nmstatectl/nmstatectl.py
+++ b/nmstatectl/nmstatectl.py
@@ -21,8 +21,11 @@ import fnmatch
 import json
 import logging
 import os
+import subprocess
 import sys
+import tempfile
 
+from six.moves import input
 import yaml
 
 from libnmstate import netapplier
@@ -37,11 +40,28 @@ def main():
     parser = argparse.ArgumentParser()
 
     subparsers = parser.add_subparsers()
+    setup_subcommand_edit(subparsers)
     setup_subcommand_show(subparsers)
     setup_subcommand_set(subparsers)
 
     args = parser.parse_args()
     return args.func(args)
+
+
+def setup_subcommand_edit(subparsers):
+    parser_edit = subparsers.add_parser('edit',
+                                        help='Edit network state in EDITOR')
+    parser_edit.set_defaults(func=edit)
+    parser_edit.add_argument('--json', help='Edit as JSON', default=True,
+                             action='store_false', dest='yaml')
+    parser_edit.add_argument(
+        'only', default='*', nargs='?', metavar='interfaces',
+        help='Edit only specified interfaces (comma-separated)'
+    )
+    parser_edit.add_argument(
+        '--no-verify', action='store_false', dest='verify', default=True,
+        help='Verify that the desired state is now part of the current state.'
+    )
 
 
 def setup_subcommand_show(subparsers):
@@ -68,34 +88,28 @@ def setup_subcommand_set(subparsers):
     parser_set.set_defaults(func=apply)
 
 
-def print_state(state, use_yaml=False):
-    if use_yaml:
-        print(yaml.dump(state, default_flow_style=False))
+def edit(args):
+    state = _filter_state(netinfo.show(), args.only)
+
+    if not state['interfaces']:
+        sys.stderr.write('ERROR: No such interface\n')
+        return os.EX_USAGE
+
+    if args.yaml:
+        suffix = '.yaml'
+        txtstate = _make_pretty_yaml(state)
     else:
-        print(json.dumps(state, indent=4, sort_keys=True,
-                         separators=(',', ': ')))
+        suffix = '.json'
+        txtstate = _make_pretty_json(state)
 
+    new_state = _get_edited_state(txtstate, suffix, args.yaml)
+    if not new_state:
+        return os.EX_DATAERR
 
-def _filter_interfaces(state, patterns):
-    """
-    return the states for all interfaces from `state` that match at least one
-    of the provided patterns.
-    """
-    showinterfaces = []
+    print('Applying the following state: ')
+    print_state(new_state, use_yaml=args.yaml)
 
-    for interface in state['interfaces']:
-        for pattern in patterns:
-            if fnmatch.fnmatch(interface['name'], pattern):
-                showinterfaces.append(interface)
-                break
-    return showinterfaces
-
-
-def _filter_state(state, whitelist):
-    if whitelist != '*':
-        patterns = [p for p in whitelist.split(',')]
-        state['interfaces'] = _filter_interfaces(state, patterns)
-    return state
+    netapplier.apply(new_state, verify_change=args.verify)
 
 
 def show(args):
@@ -124,3 +138,113 @@ def apply(args):
     netapplier.apply(state, verify_change=args.verify)
     print('Desired state applied: ')
     print_state(state, use_yaml=use_yaml)
+
+
+def _filter_state(state, whitelist):
+    if whitelist != '*':
+        patterns = [p for p in whitelist.split(',')]
+        state['interfaces'] = _filter_interfaces(state, patterns)
+    return state
+
+
+def _filter_interfaces(state, patterns):
+    """
+    return the states for all interfaces from `state` that match at least one
+    of the provided patterns.
+    """
+    showinterfaces = []
+
+    for interface in state['interfaces']:
+        for pattern in patterns:
+            if fnmatch.fnmatch(interface['name'], pattern):
+                showinterfaces.append(interface)
+                break
+    return showinterfaces
+
+
+def _make_pretty_yaml(state):
+    return yaml.dump(state, default_flow_style=False)
+
+
+def _make_pretty_json(state):
+    return json.dumps(state, indent=4, sort_keys=True, separators=(',', ': '))
+
+
+def _get_edited_state(txtstate, suffix, use_yaml):
+    while True:
+        txtstate = _run_editor(txtstate, suffix)
+
+        if txtstate is None:
+            return None
+
+        new_state, error = _parse_state(txtstate, use_yaml)
+
+        if error:
+            if not _try_edit_again(error):
+                return None
+        else:
+            return new_state
+
+
+def _run_editor(txtstate, suffix):
+    editor = os.environ.get('EDITOR', 'vi')
+    with tempfile.NamedTemporaryFile(suffix=suffix,
+                                     prefix='nmstate-') as statefile:
+        statefile.write(txtstate)
+        statefile.flush()
+
+        try:
+            subprocess.check_call([editor, statefile.name])
+            statefile.seek(0)
+            return statefile.read()
+
+        except subprocess.CalledProcessError:
+            sys.stderr.write('Error running editor, aborting...\n')
+            return None
+
+
+def _parse_state(txtstate, parse_yaml):
+    error = ''
+    state = {}
+    if parse_yaml:
+        try:
+            state = yaml.load(txtstate)
+        except yaml.parser.ParserError as e:
+            error = 'Invalid YAML syntax: %s\n' % e
+        except yaml.parser.ScannerError as e:
+            error = 'Invalid YAML syntax: %s\n' % e
+    else:
+        try:
+            state = json.loads(txtstate)
+        except ValueError as e:
+            error = 'Invalid JSON syntax: %s\n' % e
+
+    if not error and 'interfaces' not in state:
+        error = 'Invalid state: should contain "interfaces" entry.\n'
+
+    return state, error
+
+
+def _try_edit_again(error):
+    """
+    Print error and ask for user feedback. Return True, if the state should be
+    edited again and False otherwise.
+    """
+
+    sys.stderr.write('ERROR: ' + error)
+    response = ''
+    while response not in ('y', 'n'):
+        response = input('Try again? [y,n]:\n'
+                         'y - yes, start editor again\n'
+                         'n - no, throw away my changes\n'
+                         '> ').lower()
+        if response == 'n':
+            return False
+    return True
+
+
+def print_state(state, use_yaml=False):
+    if use_yaml:
+        print(_make_pretty_yaml(state))
+    else:
+        print(_make_pretty_json(state))
--- a/tests/integration/nmstatectl_edit_test.py
+++ b/tests/integration/nmstatectl_edit_test.py
@@ -1,0 +1,41 @@
+#
+# Copyright 2018 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+import os
+
+
+from .testlib import cmd as libcmd
+
+
+RC_SUCCESS = 0
+
+
+def test_edit_abort():
+    runenv = dict(os.environ)
+    env = {'EDITOR': 'false'}
+
+    runenv.update(env)
+
+    cmds = ['nmstatectl', 'edit', 'lo']
+    ret = libcmd.exec_cmd(cmds, env=runenv)
+    rc, out, err = ret
+
+    assert_rc(rc, os.EX_DATAERR, ret)
+
+
+def assert_rc(actual, expected, return_tuple):
+    assert actual == expected, 'rc={}, out={}, err={}'.format(*return_tuple)


EdDev:
Based on the implementation and the required documentation, won't it be more suited to add to the 'set' command a '--editor' flag?

tyll:
It might work as well but it would make the command-line arguments a little bit uglier, since there would be a new `--yaml` flag that would only make sense with `--edit` (I would prefer edit instead of editor) and the file argument would not make sense together with `--edit`.

tyll:
Regarding ´--yaml` IMHO it should be ignored when `--edit` is not present. Otherwise we would need to remove the auto-detection but I think the auto-detection is useful.

What would be the semantics for `nmstatectl set --edit foo.json` Should it read the state from foo.json and then open the editor? Should it write the current state to foo.json and then open the editor? Should it read the state, write to a temp file and when the editor exits without an error, write the result in foo.json? Should it just overwrite foo.json if it exists or does it need a force flag to overwrite it?

`nmstate set edit` could also mean to read the state from the file `edit`.

To me it seems that `nmstate edit` has the cleanest UI even though it requires some duplicate code to specify similar flags.

EdDev:
> Regarding ´--yamlIMHO it should be ignored when--edit` is not present. Otherwise we would need to remove the auto-detection but I think the auto-detection is useful.

I'm ok with giving an error if it is not used with ```--edit```.

> What would be the semantics for nmstatectl set --edit foo.json
I think it fits editing current state, so it should create a new file. If an existing one exists, it should give an error or ask if it is ok to overwrite (I prefer the error as it fits scripting).
Editing an existing file also sounds appealing, but I do not know how to merge the two options is a simple to understand manner. Maybe a --baseconfig=current/filename is appropriate.

> To me it seems that nmstate edit has the cleanest UI even though it requires some duplicate code to specify similar flags.
The problem is future duplication, not current duplication. When we add flags, we will need to consider both and we may well forget about it. It is safer IMO to find a way to reuse it, especially when both relate to changing the configuration.

tyll:
Actually it will always process flags both from set (currently `--verify`) and show (`--yaml, --only`). Also even if we add other flags to one of these commands in the future, we still need to add special behavior for the edit case to make sure it makes sense or add error messages in case the flag only makes sense with show and edit. Also if we forget extra flags in the future, we or users will notice the need to add them to the edit command and we can add them in a controlled manner instead of now figuring out what to do with the options that are not meant for edit or set.

EdDev:
If we merge somehow ```set``` and ```edit``` we will need to take care of improper flags and I consider that a good thing.
Perhaps we are should separate and pipeline functionality instead of merging everything together. For example, we could say that the input for ```edit``` is a file/stdout like we have for set, then edit output is a file or stdout that will get pipelined into ```set```.
What do you think?
```
nmstatectl show --only eth0 | nmstatectl edit | nmstatectl set --verify
```
Now that we removed the input, edit and set look more in common, so we could merge them.

tyll:
Having the edit command is about convenience and making it easy to understand/show how to change a state - at least this was my motiviation. Using the pipeline makes it complicated again. `nmstatectl edit` only special feature would be to automatically set the correct filetype for syntax highlighting, there is already a tool that does everything else:

`nmstatectl show  --only eth3 | vipe | nmstatectl set --verify`

vipe for vi pipe (it is part of moreutils).

I do not follow how this would make it easier to merge the commands by the way. Maybe we could change the internal infrastructure to make `edit` use both the set and show parser and then call the respective handlers but with some magic to use a python buffer instead of stdout/stdin but this all seems to be a lot of effort for just ensuring that we will add the right new commandline flags to edit/set/show when we only adjust one of them.

tyll:
IMHO we are failing for the "perfect is the enemy of good" trap here when we try to make this command now perfect for all the possible future changes. The actual duplicate code is minimal and by requiring the user to create a pipe we would be pushing the complexity to the user.

Try to explain what to a user how to edit a state, it is a lot simplier to say: run nmstatectl edit then to say you need to first show the current state with nmstatectl show and then edit it with nmstatectl edit command. And also when you try to understand what is happening, nmstatectl edit is possible to be read faster and with less thinking.

What kind of other tools do you know that implement editing a current state using shell pipe? Ancient tools like crontab support setting, deleting, editing and showing the current crontab with different command-line flags instead of using `crontab -l | crontab -e` you can just use `crontab -e`. Also newer tools such as systemd try to make it easiert for the user. Instead of running `systemctl enable foo; systemstl start foo` you can just use `systemctl enable --now foo`.

If `edit` should be implemented using `show` and `set`, this could happen internally in nmstate but it is also something that could be done when it is required instead of blocking this change now IMHO.

EdDev:
> IMHO we are failing for the "perfect is the enemy of good" trap here when we try to make this command now perfect for all the possible future changes.

From my side, it is small & decoupled vs big & coupled. With just `edit`, you do two actions implicitly and limit it to current state editing, when you could just as well edit a file.
The fact that flags and their function are already part of other commands is a smell IMO, repeating them is possible but not nice if there is another way to tackle it.

I am more concern about the cmd api over the internal implementation, as you said, the internal can be changed easily, but the external part needs to be stable.

I suggest to bring another opinion to the table. Can you invite someone from your team who can guide us to a conclusion?

tyll:
> From my side, it is small & decoupled vs big & coupled. With just edit, you do two actions implicitly and limit it to current state editing, when you could just as well edit a file.

If there is a use case for editing a file and applying it, this feature can still be added but this would also need discussion about how it would be useful. Also adding `--edit` to `set` and maybe even adding `--base-state` makes `set` big and coupled from the command-line perspective.

> The fact that flags and their function are already part of other commands is a smell IMO, repeating them is possible but not nice if there is another way to tackle it.

The code to handle these flags is already shared as much as possible. I changed the filtering for show to use a shared function now and show is now only two lines.

> I am more concern about the cmd api over the internal implementation, as you said, the internal can be changed easily, but the external part needs to be stable.

This is another argument for using edit as a command, since it makes a simple API that can be extended where it makes sense instead of inheriting the options and behavior from other commands that might not make sense for it.

> I suggest to bring another opinion to the table. Can you invite someone from your team who can guide us to a conclusion?

ok.

thom311:
I find most important, that the command line tool allows convenient reading/writing to stdout/stdin, so that complex operations can be build by chaining them together. I didn't know `vipe`, but that is awesome.

Anyway, an `edit` command for convenience seems good to me. Maybe two top-level commands:
 - `edit`, more the high-level convenience function, possibly with few options but a specific use-case.
 - `set`, more low-level and accepts JSON/YAML from various sources and does something specific with it.

`set` probably supports various options where to get the input from: file, stdin, get-current-state-from-nm. Then it has various modes what do with with the input: verify-only, maybe-print-to-stdout, maybe-modify, maybe-commit-in-nm.
The `--edit` option seems like a step that is done somewhere in the output-chain (probably very early). Ok, that makes sense too.

I really don't know. But makes sense :) That's all from me.

The implementation is actually rather simple (nice). So, maybe this `--edit` step could be rather easily added in both modes.

EdDev:
@thom311 , you left me confused. Which approach you favour here?
`nmstatectl edit`
`nmstatectl set --edit`

Or something else?

thom311:
I think I like both. No preference, sorry.

EdDev:
@tyll , anyone else you can bring aboard for this?

@phoracek , do you have an opinion?

tyll:
We could also implement both `set --edit` and just `edit`. Then we start with edit, think of some good semantics for when a file is specified with set and then add it as well. However, I would prefer to not yet discuss the semantics for when a file is specified.

tyll:
I have some more ideas. For `nmstate edit` instead of using `--only`, the interfaces to fetch the states from should probably just be arguments, i.e. `nmstate edit eth0` instead of `nmstatectl edit --only eth0`. Someone could also use`nmstate edit eth0` to change the interface in the state file from eth0 to eth1 to add a new state for eth1 based on eth0. Therefore `--only` might be misleading.

In case `nmstatectl set --edit` wins, if no file is specified and stdin is a tty, it would make sense to just use the current state. However, there should also be a flag to specify which interface state to edit. But adding `--only` to set could also mean for `nmstatectl set --only eth0 state.yml` to get only the state for `eth0` from `state.yml`. When using this together with `--edit` it becomes unclear whether or not it would also use states for other interfaces from `state.yml` after the edit operation.

danwinship:
`edit` as its own command is consistent with `kubectl` in kubernetes. (The usual workflow there is `kubectl create -f FILE` to create something from a yaml or json input; `kubectl get TYPE OBJECT` or `kubectl describe TYPE OBJECT` to view an object; `kubectl edit TYPE OBJECT`, `kubectl patch TYPE OBJECT -p JSON-PATCH`, or `kubectl replace -f FILE` to modify an object; and `kubectl delete TYPE OBJECT` or `kubectl delete -f FILE` to delete.)

n0-1:
This thing reminds me of 'crontab' tool: There I could do:
> $ crontab -l >/tmp/foo
> $ vim /tmp/foo
> $ crontab /tmp/foo
And also combine the above into a single:
> $ crontab -e
Hence IMO making `nmstatectl edit` (or `nmstatectl set --edit`, at your preference) present the user with prefilled content is well-known and convenient behaviour and therefore worth the effort.

EdDev:
You are all raising good points, but I think the core disagreement here relates to the existence of `set` and `edit` commands, meaning different things. I do not mind changing `set` to `edit`, I mind having them both, each having its own semantics.

I prefer one single command that covers `set`/`edit` in a consistent manner.
Note that there is no `create` involved, as the current state is always there, there is no need to create it.
I would prefer to have a single command that:
- Edits the existing state based on the existing state. (as the current edit proposal does)
- Edits the existing state based on an already existing saved state (stdin or a file).
- Edit specific components/interfaces as suggested by @tyll , that is a cool idea as well.

tyll:
So, if I change `nmstatectl set` to behave like `nmstatectl edit` when no file is specified and stdin is a tty, would this be ok for you? It would also need `--only`, which should IMHO only limit the state loaded into the editor but allow the nmsatectl to accept the full state from the editor. The other behaviour (editing a file) could be added later.

EdDev:
I think your other option of doing `nmstatectl edit eth0` is much nicer. You could use `"eth0,eth1,bond*"` to extend it. But if you prefer `--only` I'm fine with it as well.

The point is to have only `edit`, not have both.
So something like this seems nice:
- Edit current state in the editor: `nmstatectl edit`  (you could add `-i` to specify interactive: open an editor, but it can be implicit)
- Edit just eth0 in the editor: `nmstatectl edit eth0` (`-i` can be implicit here as well)
- Edit given state from stdin: `cat state.yaml | nmstatectl edit`  (Use `-i` to control if it should open an editor or not)
- Edit given a file: `nmstatectl edit --file state.yaml` (If you prefer not to implement now, I'm ok with using stdin to do the same, the `-i` is relevant here as well)

tyll:
Ok, so you would like to rename set to edit?

Edward Haas <notifications@github.com> schrieb am Mo., 17. Sep. 2018, 15:57:

EdDev:
On one side it is renaming, on the other side it uses an interactive editing when appropriate (implicitly) and allows interactive editing in other scenarios when explicitly requested. The best of all worlds :)

If you prefer the implicit option only for now, I'm ok with that as well.

tyll:
Using `nmstatectl edit --file state.yaml` to set the state specified in `state.yaml` does not look as good as using `nmstatectl set state.yaml`. IMHO ` nmstatectl edit --file state.yaml` is no easily comprehensible as not editing the file `state.yml` but actually using the file to edit the system state. We could do some heuristics based on the arguments given to `edit` to deice whether it is a file or an interface but I am not sure that it would be the right thing to do.

EdDev:
Yes, I agree, `--file` can be interpreted as both input and output.
So we can just rename it to something clearer like: `--state-source` or `--state-to-apply` (or whatever makes it clearer). This also opens up the ability to save the resulted state from an interactive editing in the future.

tyll:
The more I think about this, the less I am convinced that having only set or edit is better than both. It makes  naming the options/arguments a lot harder to design and therefore I would assume it makes it harder to comprehend as well.I guess the issue is that for edit it would be natural to specify the interfaces as arguments but for set to use the input file.

`nmstatectl edit eth0`
`nmstatectl edit --cc eth0.json eth0`
`nmstatectl edit --template all.json eth1`

`nmstatectl set eth0.json`
`nmstatectl set --interactive eth0.json`
`nmstatectl set all.json eth0`

Probably it would also make sense for show to behave like this:

`nmstatectl show eth0`

EdDev:
@tyll , can you please open an issue for an enhancement so we can discuss it there?
Please define the requirements without the implementation details and then suggest the implementation.

tyll:
@cathay4t please take a look here

nmstatectl/nmstatectl.py
```diff
@@ -90,12 +116,49 @@ def _filter_interfaces(state, patterns):
...
+
+        try:
+            subprocess.check_call([editor, statefile.name])
 
+            statefile.seek(0)
+
+            if args.yaml:
+                new_state = yaml.load(statefile)
+            else:
+                new_state = json.load(statefile)
```
cathay4t: Please handle possible exceptions of `json.load()` and `yaml.load()`.

nmstatectl/nmstatectl.py
```diff
@@ -90,12 +116,49 @@ def _filter_interfaces(state, patterns):
...
-        patterns = [p for p in args.only.split(',')]
+def _filter_state(state, whitelist):
+    if whitelist != '*':
+        patterns = [p for p in whitelist.split(',')]
         state['interfaces'] = _filter_interfaces(state, patterns)
+    return state
+
+
+def edit(args):
+    state = _filter_state(netinfo.show(), args.only)
```
cathay4t: Please check whether specified interface exists:
```python
    if not state['interfaces']:
        sys.stderr.write('ERROR: No such interface\n')
        return 1
```

nmstatectl/nmstatectl.py
```diff
@@ -90,12 +116,49 @@ def _filter_interfaces(state, patterns):
...
+
+            if args.yaml:
+                new_state = yaml.load(statefile)
+            else:
+                new_state = json.load(statefile)
+
+            print('Applying the following state: ')
+            print_state(new_state, use_yaml=args.yaml)
+
+            netapplier.apply(new_state, verify_change=args.verify)
```
cathay4t: Please handle exception from `netapplier.apply()` by print error message instead of traceback.

nmstatectl/nmstatectl.py
```diff
@@ -90,12 +116,49 @@ def _filter_interfaces(state, patterns):
...
+        txtstate = pretty_json(state)
+    with tempfile.NamedTemporaryFile(suffix=suffix,
+                                     prefix='nmstate-') as statefile:
+        statefile.write(txtstate)
+        statefile.flush()
+
+        try:
+            subprocess.check_call([editor, statefile.name])
 
+            statefile.seek(0)
```
cathay4t: If possible, please allow user to regret the `edit` action by delete all lines in the tempfile. That's what git did.

nmstatectl/nmstatectl.py
EdDev: An empty state is not an error.

nmstatectl/nmstatectl.py
EdDev: Handle what exactly? Exploding when something is wrong sounds a good solution to me.

nmstatectl/nmstatectl.py
EdDev: What is wrong with a traceback? It is more informative.

nmstatectl/nmstatectl.py
cathay4t: For example, invalid json/yaml syntax.

nmstatectl/nmstatectl.py
cathay4t: A traceback will trigger abrt and traceback text itself might also scare non-tech people.

nmstatectl/nmstatectl.py
EdDev: Ok, so how would you recommend handling it?

nmstatectl/nmstatectl.py
EdDev: We need the traceback to determine the problem.
If you would like to build an error library that translates exceptions into nice messages and somehow provides the needed info for debugging, please add an issue for it. I think it has low priority at the moment.
Most python tools I know transparently leave the exceptions as is and do not try to wrap them for users. Perhaps it makes sense to wrap general errors we expect users to encounter.

nmstatectl/nmstatectl.py
cathay4t:
```diff
diff --git a/nmstatectl/nmstatectl.py b/nmstatectl/nmstatectl.py
index 7241635..fffbd37 100644
--- a/nmstatectl/nmstatectl.py
+++ b/nmstatectl/nmstatectl.py
@@ -143,10 +143,23 @@ def edit(args):
             statefile.seek(0)
-            if args.yaml:
-                new_state = yaml.load(statefile)
-            else:
-                new_state = json.load(statefile)
+            try:
+                if args.yaml:
+                    new_state = yaml.load(statefile)
+                else:
+                    new_state = json.load(statefile)
+            except ValueError as e:
+                sys.stderr.write("ERROR: Invalid JSON syntax: %s\n" % e)
+                return 1
+            except yaml.parser.ParserError:
+                sys.stderr.write("ERROR: Invalid YAML syntax: %s\n" % e)
+                return 1
+
+            if 'interfaces' not in new_state:
+                sys.stderr.write("ERROR: Invalid JSON/YAML syntax: "
+                                 "should contain 'interfaces' entry.\n")
+                return 1
+
             print('Applying the following state: ')
             print_state(new_state, use_yaml=args.yaml)
```

nmstatectl/nmstatectl.py
tyll: We could add a `--verbose` or `--debug` flag that would enable tracebacks and not show them otherwise. Dominik also noticed the traceback in the `--verify` demo and I agree that it does not look so nice.

nmstatectl/nmstatectl.py
tyll: How about making this a warning at least? I agree with @cathay4t that it makes sense to report this so that nmstate does not silently nothing when the user mistyped the interface name.

nmstatectl/nmstatectl.py
tyll: Should there be an extra case to catch empty files? You mentioned it in another comment IIRC.

tyll:
@cathay4t the biggest dispute is about how to design the command line. Is there a command-line for editing the state that feels most natural for you?

cathay4t:
@tyll I think the `virsh edit` command did quit well. I will take look into its design choices.

nmstatectl/nmstatectl.py
cathay4t: Oh. I was waiting a decision before coding anything out.

cathay4t:
@tyll Just some notes on how `virsh edit` handle things:

 * Report error on non-exist domain(vm guest).
```
error: failed to get domain 'ahah'
error: Domain not found: no domain with matching name 'ahah'
```
 * Provide error on syntax issue and ask whether user want to try it again or abort:
```
error: (domain_definition):105: Extra content at the end of the document
dfa
^
Failed. Try again? [y,n,i,f,?]:
y - yes, start editor again
n - no, throw away my changes
i - turn off validation and try to redefine again
f - force, try to redefine again
? - print this help
Failed. Try again? [y,n,i,f,?]:
```
 * Handle empty file as try again:
```
error: (domain_definition):1: Document is empty
(null)
^
Failed. Try again? [y,n,i,f,?]:
y - yes, start editor again
n - no, throw away my changes
i - turn off validation and try to redefine again
f - force, try to redefine again
? - print this help
Failed. Try again? [y,n,i,f,?]:
```

A good thing to have `retry or quit` asking: user's modification will not silently dropped. We should not ask user to input everything again just because they forgot to add some space or comma(that's pretty common mistake in JSON).

EdDev:
@cathay4t , you are taking the discussion to a new direction.
We are not focused to create a dedicated editor for nmstate, nor have the resources to invest in making it well featured.
@tyll has introduced it to allow easy editing of the current state, as this has become a common exercise when we demo nmstate new capabilities.

Please focus the discussion on how the command line should look like: How and when the edit window should be opened? Do we need a special command for it? Etc...
We should continue the discussion in our next meeting.

nmstatectl/nmstatectl.py
tyll: I added your code as a separate commit.

nmstatectl/nmstatectl.py
tyll: Since this should happen in a similar way for `nmstatectl set`, let's do this in a separate PR.

nmstatectl/nmstatectl.py
tyll: This is now handled.

tyll:
@cathay4t I added the prompt as well, I hit the problem with commas and JSON too often :-)
IMHO it is now good enough to be merged.

EdDev:
I have a bit of a problem with the review on this PR: In this project we do not squash and merge the PR, we expect commits in the PR to remain as is and get rebased.
Therefore, splitting commits inside the PR for changing things that have been created in the PR makes less sense and is confusing.
There are also changes that are not in the context of this PR, or at least affects the other subcommands.

Could you please split this PR and possible squash commits so it will be simpler for me to review?
BTW: I am reviewing on a per commit, or at least trying to.

tyll:
I squashed and split it, so fire up your comments!

tyll:
@EdDev regarding signed-off: the sqashed commit now contains code that @cathay4t provided in a comment to my code - does he now have to add his signed-off-by line to the commit message somehow?

EdDev:
@tyll , I do not think it is a significant code that is worth the trouble. I would consider it as a comment given and applied.
@cathay4t , are you ok with this?

cathay4t:
@tyll There is no need to include my sign-off here.

README.md
```diff
@@ -60,6 +60,12 @@ Change to desired state:
 nmstatectl set desired-state.json
 ```
 
+Edit state of eth3 in editor and verify it afterwards:
```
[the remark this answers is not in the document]
tyll: done

nmstatectl/nmstatectl.py
```diff
@@ -37,13 +40,30 @@ def main():
...
 
     args = parser.parse_args()
     return args.func(args)
 
 
+def setup_subcommand_edit(subparsers):
+    parser_edit = subparsers.add_parser('edit',
+                                        help='Edit network state in EDITOR')
+    parser_edit.set_defaults(func=edit)
+    parser_edit.add_argument('--yaml', help='Edit as YAML', default=False,
```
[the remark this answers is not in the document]
tyll: done

nmstatectl/nmstatectl.py
```diff
@@ -68,12 +88,19 @@ def setup_subcommand_set(subparsers):
...
+
+
+def pretty_json(state):
+    return json.dumps(state, indent=4, sort_keys=True, separators=(',', ': '))
+
+
 def print_state(state, use_yaml=False):
     if use_yaml:
-        print(yaml.dump(state, default_flow_style=False))
+        print(pretty_yaml(state))
```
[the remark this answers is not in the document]
tyll: The one line appears to be harder to read to me.

nmstatectl/nmstatectl.py
```diff
@@ -98,6 +125,77 @@ def _filter_state(state, whitelist):
...
+def edit(args):
+    state = _filter_state(netinfo.show(), args.only)
+
+    if args.yaml:
+        suffix = '.yaml'
+        txtstate = pretty_yaml(state)
+    else:
+        suffix = '.json'
+        txtstate = pretty_json(state)
+    with tempfile.NamedTemporaryFile(suffix=suffix,
```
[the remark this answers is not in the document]
tyll: The suffix is needed for vim file type detection.

nmstatectl/nmstatectl.py
```diff
@@ -98,6 +125,77 @@ def _filter_state(state, whitelist):
...
+                                     prefix='nmstate-') as statefile:
+        statefile.write(txtstate)
+        statefile.flush()
+
+        while True:
+            txtstate = run_editor(txtstate, suffix)
+            error = ""
```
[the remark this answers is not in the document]
tyll: done

nmstatectl/nmstatectl.py
```diff
@@ -68,12 +88,19 @@ def setup_subcommand_set(subparsers):
     parser_set.set_defaults(func=apply)
 
 
+def pretty_yaml(state):
```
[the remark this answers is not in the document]
EdDev: The you can use `get` prefix. You can leave it as is if you do not feel comfortable with any other verb.

nmstatectl/nmstatectl.py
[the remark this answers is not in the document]
EdDev: It is a common pattern in python (cleaner IMO). Up to you.

nmstatectl/nmstatectl.py
```diff
@@ -98,6 +125,77 @@ def _filter_state(state, whitelist):
...
+            except ValueError as e:
+                error = "Invalid JSON syntax: %s\n" % e
+            except yaml.parser.ParserError as e:
+                error = "Invalid YAML syntax: %s\n" % e
+
+            if not error and (not new_state or 'interfaces' not in new_state):
+                error = "Invalid state: should contain 'interfaces' entry.\n"
+
+            if error:
+                sys.stderr.write("ERROR: ")
```
EdDev: Won't a simple concatenation ('+') work?

nmstatectl/nmstatectl.py
```diff
@@ -98,6 +125,77 @@ def _filter_state(state, whitelist):
...
+                sys.stderr.write(error)
+                response = ""
+                while response not in ("y", "n"):
+                    response = input("Try again? [y,n]:\n"
```
EdDev: It will be nice to accept both upper and lower cases input. (normalize to lower-case).

nmstatectl/nmstatectl.py
```diff
@@ -98,6 +125,77 @@ def _filter_state(state, whitelist):
...
+                                     "y - yes, start editor again\n"
+                                     "n - no, throw away my changes\n"
+                                     "> ")
+                if response == "n":
+                    return 1
+            else:
```
EdDev: Won't it be nicer to use the `error` as the while condition (instead of `True`)

nmstatectl/nmstatectl.py
```diff
@@ -98,6 +125,77 @@ def _filter_state(state, whitelist):
...
+
+        print('Applying the following state: ')
+        print_state(new_state, use_yaml=args.yaml)
+
+        netapplier.apply(new_state, verify_change=args.verify)
+
+
+def run_editor(txtstate, suffix):
+    editor = os.environ.get('EDITOR', 'vi')
+    with tempfile.NamedTemporaryFile(suffix=suffix,
```
EdDev: This code snippet appeared already above.
How about creating a contextmanager to share it?
`with temp_config(txtstate) as f:`

nmstatectl/nmstatectl.py
```diff
@@ -98,6 +125,77 @@ def _filter_state(state, whitelist):
...
+        statefile.write(txtstate)
+        statefile.flush()
+
+        try:
+            subprocess.check_call([editor, statefile.name])
+            statefile.seek(0)
+            return statefile.read()
+
+        except subprocess.CalledProcessError:
+            sys.stderr.write('Error running editor, aborting...\n')
```
EdDev: Don't you want to report the exception info?

tyll: The idea is to allow users to quit vim with `:cq` without changing the state. The exception would no be interesting then. Not sure if there are other error cases where it would be.

nmstatectl/nmstatectl.py
tyll: yes, it would work, too. changed i.

nmstatectl/nmstatectl.py
tyll: changed

nmstatectl/nmstatectl.py
tyll: while True with break is AFAIK the python pattern for do..until. Using error as the while condition means that it needs to be set to True before entering the loop which seems worse to me.

nmstatectl/nmstatectl.py
tyll: good catch, the first occurrence was a leftover from refactoring, I just removed it. :-)

nmstatectl/nmstatectl.py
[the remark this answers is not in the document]
tyll: I can live with a make prefix. If I had a free wish, the state would be an object with respective methods. :-)

nmstatectl/nmstatectl.py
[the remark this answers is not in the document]
tyll: Not sure if it is common when there are function calls involved ;-)

`print(_make_pretty_yaml(state) if use_yaml else _make_pretty_json(state))` might be a possibility but with the `_make` prefix it is too long for a single line anyhow. Therefore I prefer to keep it

nmstatectl/nmstatectl.py
tyll: added the error

nmstatectl/nmstatectl.py
```diff
@@ -98,6 +125,77 @@ def _filter_state(state, whitelist):
...
+                                     prefix='nmstate-') as statefile:
+        statefile.write(txtstate)
+        statefile.flush()
+
+        while True:
+            txtstate = run_editor(txtstate, suffix)
+            error = ""
+
+            if txtstate is None:
+                return 1
```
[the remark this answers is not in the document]
tyll: You said the ctl is primarily for us, do we really need to distinguish these errors? Seems to me overkill at the moment since there is an API that could be used to get proper exceptions instead of calling nmstatectl and checking the returncode.

nmstatectl/nmstatectl.py
```diff
@@ -98,6 +125,77 @@ def _filter_state(state, whitelist):
...
+
+            try:
+                if args.yaml:
+                    new_state = yaml.load(txtstate)
+                else:
+                    new_state = json.loads(txtstate)
+            except ValueError as e:
```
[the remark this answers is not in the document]
tyll: changed

nmstatectl/nmstatectl.py
```diff
@@ -98,6 +125,77 @@ def _filter_state(state, whitelist):
...
+                error = "Invalid JSON syntax: %s\n" % e
+            except yaml.parser.ParserError as e:
+                error = "Invalid YAML syntax: %s\n" % e
+
+            if not error and (not new_state or 'interfaces' not in new_state):
```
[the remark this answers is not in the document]
tyll: I introduce `interfaces_undefined` which feels more natural to me in this case.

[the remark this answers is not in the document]
tyll: Actually using an extra variable does not work because `new_state` is not defined when `error` is True. I reverted the change.

nmstatectl/nmstatectl.py
```diff
@@ -98,6 +125,77 @@ def _filter_state(state, whitelist):
...
+                error = "Invalid state: should contain 'interfaces' entry.\n"
+
+            if error:
```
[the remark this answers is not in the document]
tyll: changed this

nmstatectl/nmstatectl.py
```diff
@@ -98,6 +125,77 @@ def _filter_state(state, whitelist):
...
+    else:
+        suffix = '.json'
+        txtstate = pretty_json(state)
+    with tempfile.NamedTemporaryFile(suffix=suffix,
+                                     prefix='nmstate-') as statefile:
+        statefile.write(txtstate)
+        statefile.flush()
+
+        while True:
+            txtstate = run_editor(txtstate, suffix)
```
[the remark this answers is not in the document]
tyll: fixed

tyll:
Everything should be addressed now. :-)

nmstatectl/nmstatectl.py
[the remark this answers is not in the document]
tyll: changed

tyll:
ci test please

tyll:
ci test please

PR #74, opened 2018-09-10
README: Update state example

tyll:
Add the recently added Ethernet properties.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,16 +72,23 @@ Desired/Current state example:
 {
     "interfaces": [
         {
+            "description": "Production Network",
+            "ethernet": {
+                "auto-negotiation": true,
+                "duplex": "full",
+                "speed": 1000
+            },
             "ipv4": {
                 "address": [
                     {
-                        "ip": "192.168.122.1",
+                        "ip": "192.0.2.142",
                         "prefix-length": 24
                     }
                 ],
                 "enabled": true
             },
-            "name": "eth0",
+            "mtu": 1500,
+            "name": "eth3",
             "state": "up",
             "type": "ethernet"
         }


README.md
```diff
@@ -72,16 +72,23 @@ Desired/Current state example:
...
+            "ethernet": {
+                "auto-negotiation": true,
+                "duplex": "full",
+                "speed": 1000
+            },
             "ipv4": {
                 "address": [
                     {
-                        "ip": "192.168.122.1",
+                        "ip": "192.0.2.142",
```
EdDev: No need to change this.

tyll: The new IP address makes it easier to actually test and reproduce this state since it is reserved for documentatio and examples in [RFC5737](https://tools.ietf.org/html/rfc5737). The other IP conflicts with the default virtual bridge IPs and I do not see any value in having to manually update the IP address when we support more properties in the state.

tyll:
Updated the commit message

README.md
EdDev: The rfc specify these addresses for documentation only and recommends/warns that implemantations may block them locally and publicly, so it will not be useful to use them using copy/paste.
Said that, I’m not against this, but the reasoning is not fitting here. I would also prefer the change in its own commit.

tyll: I use them without any problems on Linux for a while to make sure that tests do not conflict with other assigned IP addresses. It is useful for us to use with copy and paste to make it easy to update the example state here. Doing this in a separate commit seems to be overkill for me since it is just a dump from one of our test systems to show the currently supported properties. Any manual fiddling with the state only bears the risk of accidentally publishing something that is not correct. Even if you would be using a different IP address (from RFC5737) for the next example, it is not a problem for me.

EdDev: The fact that is works today does not mean it will work in the future and as you reference the RFC, it does warns and recommends not to use these in a running env. If you ask me, the original reasoning for even having these addresses is to make sure they cannot be copy/paste, protecting against duplication.
And the reasoning to have focused commits is for us to go fast, especially in the review process.

Anyway, I can live with this as is, taking it.

PR #75, opened 2018-09-11
Show interface name first

tyll:
When showing or editing a state, show the interface name first. This makes it easier to find a specific interface state or to edit the state in yaml, since the interface name now appears after the dash instead of other options that one might want to remove.

### DIFF
--- a/nmstatectl/nmstatectl.py
+++ b/nmstatectl/nmstatectl.py
@@ -30,6 +30,7 @@ import yaml
 
 from libnmstate import netapplier
 from libnmstate import netinfo
+from nmstatectl.prettystate import PrettyState
 
 
 def main():
@@ -95,12 +96,14 @@ def edit(args):
         sys.stderr.write('ERROR: No such interface\n')
         return os.EX_USAGE
 
+    pretty_state = PrettyState(state)
+
     if args.yaml:
         suffix = '.yaml'
-        txtstate = _make_pretty_yaml(state)
+        txtstate = pretty_state.yaml
     else:
         suffix = '.json'
-        txtstate = _make_pretty_json(state)
+        txtstate = pretty_state.json
 
     new_state = _get_edited_state(txtstate, suffix, args.yaml)
     if not new_state:
@@ -160,14 +163,6 @@ def _filter_interfaces(state, patterns):
                 showinterfaces.append(interface)
                 break
     return showinterfaces
-
-
-def _make_pretty_yaml(state):
-    return yaml.dump(state, default_flow_style=False)
-
-
-def _make_pretty_json(state):
-    return json.dumps(state, indent=4, sort_keys=True, separators=(',', ': '))
 
 
 def _get_edited_state(txtstate, suffix, use_yaml):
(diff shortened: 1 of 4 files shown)


nmstatectl/nmstatectl.py
```diff
@@ -93,10 +115,45 @@ def pretty_yaml(state):
...
+        ordered_state[key] = value
+
+    return ordered_state
+
+
+def order_state(state):
+    try:
+        iface_states = state.pop('interfaces')
+        state['interfaces'] = []
+        for iface_state in sorted(iface_states, key=itemgetter("name")):
```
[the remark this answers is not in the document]
tyll: changed

nmstatectl/nmstatectl.py
```diff
@@ -93,10 +115,45 @@ def pretty_yaml(state):
...
+            iface_state = order_iface_state(iface_state)
+            state['interfaces'].append(iface_state)
+    except KeyError:
```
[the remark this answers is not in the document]
tyll: changed. Do not remember what it was needed for.

tyll:
Everything should be addressed now. Not sure about travis - flake8 complains about the long URL but since it is a URL, breaking it into two lines would be bad, too.

tyll:
I added type and state to the top as well

nmstatectl/prettystate.py
```diff
@@ -0,0 +1,91 @@
...
+    def json(self):
+        return json.dumps(self.state, indent=4, separators=(',', ': '))
+
+
+def represent_ordereddict(dumper, data):
+    """
+    Represent OrderedDict as regular dictionary
+
+    Source: https://stackoverflow.com/questions/16782112/can-pyyaml-dump-dict-items-in-non-alphabetical-order
+    """  # noqa: E501
```
EdDev: I prefer you break the line over breaking format rules.

nmstatectl/prettystate.py
```diff
@@ -0,0 +1,91 @@
...
+    return state
+
+
+def order_iface_state(iface_state):
+    ordered_state = OrderedDict()
+
+    for setting in ('name', 'type', 'state'):
+        try:
+            ordered_state[setting] = iface_state.pop(setting)
+        except KeyError:
```
EdDev: I guess you need this for the pop action.
How about this? (less noisy IMO):
```
key = iface_state.pop(setting, None)
if key is not None:
    ordered_state[setting] = key
```

nmstatectl/prettystate.py
EdDev: 60 votes is not that convincing, but if you prefer it like this lets go with the exception.

nmstatectl/prettystate.py
EdDev: The taste changes with context (and age :) )
Your argument is valid, leave it.

nmstatectl/prettystate.py
tyll: All but one answer are for keeping the URL this way.

tyll:
ci test please

PR #78, opened 2018-09-20
Containarized integration tests

EdDev:
Run integration tests in a Centos7 container (using Docker).


automation/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM centos:7
+ENV container docker
+RUN (cd /lib/systemd/system/sysinit.target.wants/; for i in *; do [ $i == \
```
tyll: Why is this needed? Please add some explanation as a comment.

automation/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+     systemd-tmpfiles-setup.service ] || rm -f $i; done); \
+    rm -f /lib/systemd/system/multi-user.target.wants/*;\
+    rm -f /etc/systemd/system/*.wants/*;\
+    rm -f /lib/systemd/system/local-fs.target.wants/*; \
+    rm -f /lib/systemd/system/sockets.target.wants/*udev*; \
+    rm -f /lib/systemd/system/sockets.target.wants/*initctl*; \
+    rm -f /lib/systemd/system/basic.target.wants/*;\
+    rm -f /lib/systemd/system/anaconda.target.wants/*;
+
+RUN yum -y install NetworkManager epel-release git cairo*; \
```
tyll: AFAIK it is best practice to run "yum clean all" after a yum installation step to avoid stale yum caches in cached docker layers. Also it might be better to use `&&` instead of `;` to chain commands (also for the systemd cleanup) but I am not sure if this makes docker abort if something fails (which it should IMHO).

EdDev: Change all except the systemd part, which I prefer to keep as the original test from the official Centos page.

automation/Dockerfile
EdDev: Done

PR #85, opened 2018-10-12
nmstatectl: Verify state by default

tyll:
This is another split from #73

### DIFF
--- a/nmstatectl/nmstatectl.py
+++ b/nmstatectl/nmstatectl.py
@@ -61,8 +61,9 @@ def setup_subcommand_set(subparsers):
                             'stdin is used when no file is specified.',
                             nargs='?')
     parser_set.add_argument(
-        '--verify', action='store_true',
-        help='Verify that the desired state is now part of the current state.'
+        '--no-verify', action='store_false', dest='verify', default=True,
+        help='Do not verify that the state was completely set and disable '
+        'rollback to previous state.'
     )
     parser_set.set_defaults(func=apply)
 


EdDev:
I think we need to update the documentation as well.

tyll:
The documentation does not seem to mention `--verify` - also checked the other repo with examples. Only the demo will not be accurate anymore.

tyll:
To test the network system role I created veth devices to have some kind of dummy devices. Maybe this can be done here as well?

EdDev:
In the docker env, eth0 is actually a veth and we should be able to create several of them when running the container.
I just thought to create an iface by defining it.. a dummy is a good candidate as we are supposed to support dummy type, unfortunately it does not work well on first try (probably we need the wait-for-connection-active mechanism).

nmstatectl/nmstatectl.py
```diff
@@ -61,7 +61,7 @@ def setup_subcommand_set(subparsers):
                             'stdin is used when no file is specified.',
                             nargs='?')
     parser_set.add_argument(
-        '--verify', action='store_true',
+        '--no-verify', action='store_false', dest='verify', default=True,
         help='Verify that the desired state is now part of the current state.'
```
cathay4t: The help text seems does not explain `no-verify`. Maybe something like: `Disable the verify of desired state is now part of the current state.`

EdDev: I would suggest: `Do not verify that the desired state is part of the current state post execution.`

tyll: Good catch. I changed it to:

`Do not verify that the state was completely set and disable rollback to previous state.`

How do you like it?

PR #91, opened 2018-10-17
nm: Extend debug logging information for connections

EdDev:
The connection profile lifecycle is now better logged, assisting with
debugging of edge scenarios.


libnmstate/nm/connection.py
```diff
@@ -68,9 +68,9 @@ def update_profile(base_profile, new_profile):
     base_profile.replace_settings_from_connection(new_profile)
 
 
-def commit_profile(connection_profile, save_to_disk=True):
+def commit_profile(connection_profile, save_to_disk=True, metadata=None):
```
tyll: AFAICS, metadata should be always called nmdev since the callback assumes that it will be nmdev and not arbitrary metadata.

libnmstate/nm/applier.py
```diff
@@ -56,7 +56,7 @@ def edit_existing_ifaces(con_profiles):
         if nmdev:
             cur_con_profile = connection.get_device_connection(nmdev)
         if cur_con_profile:
-            connection.commit_profile(connection_profile)
+            connection.commit_profile(connection_profile, metadata=nmdev)
```
tyll: This needs `s/metadata/nmdev`, too.

PR #92, opened 2018-10-17
libnmstate: Interface desire state sanitize adjustment

EdDev:
nmstate is processing the desired state of an interface and adds base
information for the 'ethernet' properties (auto-negotiation, speed and
duplex).
The 'ethernet' properties are relevant only for ethernet type interfaces
and not for virtual ones. In addition, the base values used in the
sanitize step may cause the post apply verification to fail.

The interface ethernet state sanitize is now limited to ethernet type
interfaces and special handling has been added in the verification step
to allow proper comparison between the desired and the current state.

### DIFF
--- a/libnmstate/netapplier.py
+++ b/libnmstate/netapplier.py
@@ -310,6 +310,19 @@ def assert_ifaces_state(ifaces_desired_state, ifaces_current_state):
         iface_cstate = ifaces_current_state[ifname]
         iface_dstate = _canonicalize_desired_state(
             ifaces_desired_state[ifname], iface_cstate)
+
+        _cleanup_iface_ethernet_state_sanitize(iface_dstate, iface_cstate)
+
         if iface_dstate != iface_cstate:
             raise DesiredStateIsNotCurrentError(
                 'desired: {}\ncurrent: {}'.format(iface_dstate, iface_cstate))
+
+
+def _cleanup_iface_ethernet_state_sanitize(desired_state, current_state):
+    ethernet_desired_state = desired_state.get('ethernet')
+    if ethernet_desired_state:
+        ethernet_current_state = current_state.get('ethernet', {})
+        for key in ('auto-negotiation', 'speed', 'duplex'):
+            if ethernet_desired_state.get(key, None) is None:
+                ethernet_desired_state.pop(key, None)
+                ethernet_current_state.pop(key, None)
--- a/libnmstate/netapplier.py
+++ b/libnmstate/netapplier.py
@@ -47,7 +47,8 @@ def _apply_ifaces_state(interfaces_desired_state, verify_change):
     ifaces_desired_state = _index_by_name(interfaces_desired_state)
     ifaces_current_state = _index_by_name(netinfo.interfaces())
 
-    ifaces_desired_state = sanitize_ethernet_state(ifaces_desired_state)
+    ifaces_desired_state = sanitize_ethernet_state(ifaces_desired_state,
+                                                   ifaces_current_state)
     generate_ifaces_metadata(ifaces_desired_state, ifaces_current_state)
 
     with _transaction():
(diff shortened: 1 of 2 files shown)


libnmstate/netapplier.py
```diff
@@ -58,7 +59,7 @@ def _apply_ifaces_state(interfaces_desired_state, verify_change):
             _edit_interfaces(ifaces_desired_state, ifaces_current_state)
         if verify_change:
             # FIXME: Remove this sleep after adding wait for state mechanism.
-            time.sleep(2)
+            time.sleep(5)
```
tyll: Is this intentional here? Maybe this should be its own commit with an explanation.

EdDev: No, a leftover. Thanks.

libnmstate/netapplier.py
```diff
@@ -120,19 +121,22 @@ def generate_ifaces_metadata(ifaces_desired_state, ifaces_current_state):
...
     """
-    for iface_state in six.viewvalues(ifaces_desired_state):
-        ethernet = iface_state.setdefault("ethernet", {})
-        ethernet.setdefault("auto-negotiation", None)
-        ethernet.setdefault("speed", None)
-        ethernet.setdefault("duplex", None)
+    for ifname, iface_state in six.viewitems(ifaces_desired_state):
+        if ifname in ifaces_current_state:
+            iface_current_state = ifaces_current_state[ifname]
+            if iface_current_state['type'] == 'ethernet':
```
tyll: IMHO it would be better to check if the current state contains an ethernet subtree here. The method to create the current state needs to be changed to only add the ethernet subtree if the interface type is ethernet.

Also wired.create_setting() should only be called if the interface type is ethernet.

libnmstate/netapplier.py
```diff
@@ -120,19 +121,22 @@ def generate_ifaces_metadata(ifaces_desired_state, ifaces_current_state):
     )
 
 
-def sanitize_ethernet_state(ifaces_desired_state):
+def sanitize_ethernet_state(ifaces_desired_state, ifaces_current_state):
     """
     If auto-negotiation, speed and duplex settings are not provided, they need
```
tyll:
```suggestion
    If auto-negotiation, speed and duplex settings are not provided, but exist in the current state, they need
```

libnmstate/netapplier.py
```diff
@@ -120,19 +121,22 @@ def generate_ifaces_metadata(ifaces_desired_state, ifaces_current_state):
...
     from the actual configuration.  This makes it possible to distiguish
     whether a user specified these values in the later configuration step.
     """
-    for iface_state in six.viewvalues(ifaces_desired_state):
-        ethernet = iface_state.setdefault("ethernet", {})
-        ethernet.setdefault("auto-negotiation", None)
-        ethernet.setdefault("speed", None)
-        ethernet.setdefault("duplex", None)
+    for ifname, iface_state in six.viewitems(ifaces_desired_state):
+        if ifname in ifaces_current_state:
```
tyll: AFAICS this could be simplified to something like

```
if 'ethernet' in ifaces_current_state.get(ifname, {})
    ethernet = ...
```

libnmstate/netapplier.py
```diff
@@ -306,6 +310,15 @@ def assert_ifaces_state(ifaces_desired_state, ifaces_current_state):
         iface_cstate = ifaces_current_state[ifname]
         iface_dstate = _canonicalize_desired_state(
             ifaces_desired_state[ifname], iface_cstate)
+
+        iface_ethernet_dstate = iface_dstate.get('ethernet')
```
tyll: IMHO it should be explained what is going on here to be able to know this still in two weeks. I needed the comment in the sanitize method as well to be able to understand the issue. :-)
Maybe it would be also better to put in its own method to be able to use variables such as current_state and desired_state since `...cstate/dstate` are hard to distinguish.

libnmstate/netapplier.py

libnmstate/netapplier.py
EdDev: I massaged it a bit.
I think we should check against the type and not against a subsequent property.

libnmstate/netapplier.py
EdDev: I will add the changes not to report the `ethernet` subtree, but see my previous response regarding the verification here.
Thanks for the two points.

libnmstate/netapplier.py
EdDev: Extracting to smaller methods is always good, thanks for pointing this out.

libnmstate/netapplier.py
```diff
@@ -120,19 +121,22 @@ def generate_ifaces_metadata(ifaces_desired_state, ifaces_current_state):
...
 
 
-def sanitize_ethernet_state(ifaces_desired_state):
+def sanitize_ethernet_state(ifaces_desired_state, ifaces_current_state):
     """
-    If auto-negotiation, speed and duplex settings are not provided, they need
-    to be set to None to not override them with the values from the current
-    settings since the current settings are read from the device state and not
+    If auto-negotiation, speed and duplex settings are not provided,
+    but exist in the current statet, they need to be set to None
```
tyll: There is a typo: `statet`

tyll:
There is a typo in the commit message (desire vs desired) and I think it should start with a verb after the colon and include the word `ethernet`. Maybe it should even be separate commits

- ignore unspecified ethernet settings for verification
- sanitize ethernet settings only for type ethernet
- report ethernet settings only for type ethernet (this change is currently missing)
- change wired settings only for type ethernet (this change is currently misssing)

libnmstate/netapplier.py

EdDev:
> There is a typo in the commit message (desire vs desired) and I think it should start with a verb after the colon and include the word ethernet. Maybe it should even be separate commits
> 
Ok, fair enough.
>     ignore unspecified ethernet settings for verification
>     sanitize ethernet settings only for type ethernet
>     report ethernet settings only for type ethernet (this change is currently missing)
Adding
>     change wired settings only for type ethernet (this change is currently misssing)
I answered on this but github hides the comment after changing the line.
"For the wired setting, it will not work as there are properties that are used there for virtual devices (like mtu)."

EdDev:
ci test please

PR #87, opened 2018-10-13
automation: Connect 2 additional networks to docker

EdDev:
Adding 2 additional links to the running docker (3 in total) to allow
complex test configurations to be supported.

PR #93, opened 2018-10-17
tests: Add an integ lib test for adding static IPv4

EdDev:
Configuring a static IPv4 address on an interface.

The assertion is embedded in the transaction verification,
which is on by default.

### DIFF
--- a/tests/integration/static_ip_address_test.py
+++ b/tests/integration/static_ip_address_test.py
@@ -1,0 +1,59 @@
+#
+# Copyright 2018 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+import copy
+
+from libnmstate import netapplier
+
+from .testlib import assertlib
+from .testlib import statelib
+from .testlib.statelib import INTERFACES
+
+
+def test_add_static_ipv4_with_full_state():
+    desired_state = statelib.show_only(('eth1',))
+    eth1_desired_state = desired_state[INTERFACES][0]
+
+    eth1_desired_state['state'] = 'up'
+    eth1_desired_state['ipv4']['enabled'] = True
+    eth1_desired_state['ipv4']['address'] = [
+        {'ip': '192.168.122.250', 'prefix-length': 24}
+    ]
+    netapplier.apply(copy.deepcopy(desired_state))
+
+    assertlib.assert_state(desired_state)
+
+
+def test_add_static_ipv4_with_min_state():
+    desired_state = {
+        INTERFACES: [
+            {
+                'name': 'eth2',
+                'type': 'ethernet',
+                'state': 'up',
+                'ipv4': {
+                    'enabled': True,
+                    'address': [
+                        {'ip': '192.168.122.251', 'prefix-length': 24}
+                    ]
+                }
+            }
+        ]
+    }
+    netapplier.apply(copy.deepcopy(desired_state))
+
+    assertlib.assert_state(desired_state)
--- a/tests/integration/testlib/assertlib.py
+++ b/tests/integration/testlib/assertlib.py
@@ -1,0 +1,76 @@
+#
+# Copyright 2018 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+import collections
+import copy
+import six
+
+from . import statelib
+from .statelib import INTERFACES
+
+
+def assert_state(desired_state):
+    """Given a state, assert it against the current state."""
+
+    filtered_current_state = statelib.filter_current_state(desired_state)
+    normalized_desired_state = _normalize_desired_state(desired_state,
+                                                        filtered_current_state)
+
+    assert normalized_desired_state == filtered_current_state
+
+
+def _normalize_desired_state(desired_state, current_state):
+    """
+    Given the desired and current state, complement the desired state with all
+    missing properties such that it will be a "full" state description.
+    Note: It is common for the desired state to include only partial config,
+    expecting nmstate to complement the missing parts from the current state.
+    """
+    desired_state = copy.deepcopy(desired_state)
+    desired_interfaces_state = desired_state[INTERFACES]
+
+    for current_iface_state in current_state[INTERFACES]:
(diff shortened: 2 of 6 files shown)


tyll:
This needs #87 and #92 to be merged

tests/integration/apply_ip_interface_test.py
```diff
@@ -0,0 +1,48 @@
...
+        ]
+    }
+
+
+def test_add_static_ipv4(eth1_state):
+    eth1_state['interfaces'][0]['ipv4']['enabled'] = True
+    eth1_state['interfaces'][0]['ipv4']['address'] = [
+        {'ip': '192.168.122.250', 'prefix-length': 24}
+    ]
+    netapplier.apply(eth1_state)
```
tyll: IMHO some kind of assertion should be added after applying the state.

EdDev: @tyll , I agree. I am waiting for the dependencies to get merged to make it easier for me to work on this.

EdDev:
ci test please

EdDev:
ci test please

EdDev:
Depends on #100 and #101 
It also requires another fix for IPv6 setting and reporting which seems to be faulty based on this test results.

EdDev:
Also depends now on #102

tests/integration/testlib/statelib.py
```diff
@@ -0,0 +1,52 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+from libnmstate import netinfo
+
+
+INTERFACES = 'interfaces'
```
tyll: Shouldn't this constant be part of libnmstate?

tests/integration/static_ip_address_test.py
```diff
@@ -0,0 +1,37 @@
...
+
+from libnmstate import netapplier
+
+from .testlib import assertlib
+from .testlib import statelib
+
+
+def test_add_static_ipv4():
+    desired_state = statelib.show_only(('eth1',))
+    eth1_desired_state = desired_state['interfaces'][0]
```
tyll: This is not using `INTERFACES`... is this intended?

tests/integration/testlib/statelib.py
EdDev: I guess it should be promoted there. A PR on that will be nice.
(where to locate it is less obvious I guess)

tests/integration/static_ip_address_test.py
EdDev: Thanks for noticing, changing.

tests/integration/testlib/statelib.py
```diff
@@ -0,0 +1,52 @@
...
+        INTERFACES: [{'name': ifname} for ifname in ifnames]
+    }
+    return filtered_current_state(base_filter_state)
+
+
+def filtered_current_state(desired_state):
+    """
+    Given the desired state, return a filtered version of the current state
+    which includes only entities that have been mentioned in the desired state.
+    In case there are no interfaces for filtering, all are reported
```
tyll: The doctext mentions entities without explaining what they are and used interfaces in the next sentence, maybe use `entities such as interfaces` and then `entities` in the second sentence.

tests/integration/testlib/assertlib.py
```diff
@@ -0,0 +1,75 @@
...
+def assert_state(desired_state):
+    """Given a state, assert it against the current state."""
+    filtered_current_state = statelib.filtered_current_state(desired_state)
+    normalized_desired_state = _normalize_desired_state(desired_state,
+                                                        filtered_current_state)
+
+    assert normalized_desired_state == filtered_current_state
+
+
+def _normalize_desired_state(desired_state, current_state):
```
tyll: Isn't something like this part of libnmstate?

tests/integration/testlib/statelib.py
EdDev: Done

tyll: Also IMHO filtered is to unpecific for the function. Since it works on entities, this should be part of the name.

tests/integration/testlib/assertlib.py
EdDev: We want the test to assert the outcome even though the production code is already doing that (you pointed that out well in the last review).
This is the test version of such an assert which require normalization of the data.
Hopefully, we will not repeat bugs in production here, or at least have different bugs.

I also hope we will have less logic here, as we do not need the inner metadata and other special handling, we expect nmstate to return back something that is simpler.

tests/integration/static_ip_address_test.py
```diff
@@ -0,0 +1,38 @@
...
+
+from libnmstate import netapplier
+
+from .testlib import assertlib
+from .testlib import statelib
+from .testlib.statelib import INTERFACES
+
+
+def test_add_static_ipv4():
+    desired_state = statelib.show_only(('eth1',))
```
tyll: IMHO the test should work without getting the current state of eth1 but specify the desired state indepentently and then check if it is properly applied. Otherwise the test condition depends on the previous state of the test setup which seems to be wrong.

tests/integration/testlib/statelib.py
EdDev: Could you please explain what you mean? It should probably be a verb: `filter_current_state`. And it works on a state, I think it is obvious the state has entities.

tests/integration/static_ip_address_test.py
EdDev: I think this is safer as it does not assume anything about what exists now.
My main concern is when we add other properties which may be a must: I do not want to revisit the test. In this solution, we get the template from what is supported now by nmstate and edit it.

We may want to add tests that check specific completion of data (when we provide only minimum data and expect nmstate to merge it with current), but I think that is not the same test... at least, it is a more messy one.

I agree we need not to drag problems or states between tests. For that I plan another change to be executed on every executed test, I will post it after we finish this initial one.

tests/integration/testlib/assertlib.py
```diff
@@ -0,0 +1,75 @@
...
+import collections
+import copy
+import six
+
+from . import statelib
+from .statelib import INTERFACES
+
+
+def assert_state(desired_state):
+    """Given a state, assert it against the current state."""
```
cathay4t: Maybe a new line after `"""`?

EdDev: Done

tests/integration/testlib/statelib.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    Report the current state, filtering based on the given interface names.
+    """
+    base_filter_state = {
+        INTERFACES: [{'name': ifname} for ifname in ifnames]
+    }
+    return filter_current_state(base_filter_state)
+
+
+def filter_current_state(desired_state):
```
tyll: Maybe use a name like `get_corresponding_current_state` - it is strange that something that filters the current state gets a desired state as argumentr. Or maybe make the current state an argument as well and use a matching name. Since it get's the current state in the function I would even argue that this should be a class method for a class that represents the current state - but since it is not a class, it is a strange method.

tests/integration/static_ip_address_test.py
EdDev: I just added a second test, hope this is now ok.

tests/integration/testlib/statelib.py
EdDev: See #110

EdDev:
ci test please

tests/integration/testlib/statelib.py
tyll: Thank you

tests/integration/testlib/assertlib.py
tyll: IMHO normalization does not describe this. For me it means that the state is changed to an equivalent state.

tyll: Btw. is it expected that this will only work in the docker environment but not with real NICs? I guess then the ethernet subtree will be added and might break the expectations.

EdDev: I have renamed things in #111 , lets discuss it there.
I am not sure we have a problem with `ethernet`, from "outside" I am not sure the special handling is relevant (there are no None/s and stuff). But lets see how it works out through an added test later on.

PR #94, opened 2018-10-17
tests: Add an nmstatectl integration test for 'set'

EdDev:
Configuring a static IPv4 address on an interface.

### DIFF
--- a/tests/integration/nmstatectl_test.py
+++ b/tests/integration/nmstatectl_test.py
@@ -22,6 +22,7 @@ from libnmstate.schema import Constants
 from .testlib import cmd as libcmd
 
 
+SET_CMD = ['nmstatectl', 'set']
 SHOW_CMD = ['nmstatectl', 'show']
 
 RC_SUCCESS = 0
@@ -48,6 +49,20 @@ LOOPBACK_YAML_CONFIG = b"""- name: lo
   ipv6:
     enabled: false
   mtu: 65536"""
+
+ETH1_YAML_CONFIG = b"""interfaces:
+- name: eth1
+  state: up
+  type: ethernet
+  mtu: 1500
+  ipv4:
+    address:
+    - ip: 192.0.2.250
+      prefix-length: 24
+    enabled: true
+  ipv6:
+    enabled: false
+"""
 
 
 def test_missing_operation():
@@ -97,6 +112,13 @@ def test_show_command_only_non_existing():
 
     state = json.loads(out)
     assert len(state[Constants.INTERFACES]) == 0
+
+
+def test_set_command_with_yaml_format():
+    ret = libcmd.exec_cmd(SET_CMD, stdin=ETH1_YAML_CONFIG)
+    rc, out, err = ret
+
+    assert_rc(rc, RC_SUCCESS, ret)
 
 
 def assert_rc(actual, expected, return_tuple):
--- a/tests/integration/testlib/cmd.py
+++ b/tests/integration/testlib/cmd.py
@@ -20,7 +20,7 @@ import re
 import subprocess
 
 
-def exec_cmd(cmd, env=None):
+def exec_cmd(cmd, env=None, stdin=None):
     """
     Execute cmd in an external process, collect its output and returncode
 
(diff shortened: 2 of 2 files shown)


tests/integration/testlib/cmd.py
```diff
@@ -20,7 +20,7 @@
 import subprocess
 
 
-def exec_cmd(cmd, env=None):
+def exec_cmd(cmd, env=None, stdin=None):
```
tyll: How about moving this into a separate commit? It is a prerequisite for the other change but not really related.

tests/integration/nmstatectl_test.py
```diff
@@ -73,5 +88,12 @@ def test_show_command_with_yaml_format():
...
 
 
+def test_set_command_with_yaml_format():
+    ret = libcmd.exec_cmd(SET_CMD, stdin=ETH1_YAML_CONFIG)
+    rc, out, err = ret
+
+    assert_rc(rc, RC_SUCCESS, ret)
+
+
 def assert_rc(actual, expected, return_tuple):
```
tyll: IMHO it is strange that the same value is passed twice to the method, once as actual and once in return_tuple.

tyll:
The new test fails in the standard-ci.

tests/integration/nmstatectl_test.py
EdDev: This function does not know that the `expected` is in the `return_tuple`. I do not see anything nicer here, expect perhaps to use a separate function that always creates the string on input at the caller.

Anyway, it is not part of this PR.

tests/integration/testlib/cmd.py
EdDev: We introduce it with the usage and it is not that complex. I find it simpler to review this way.
But if you insist, I will comply.

tests/integration/nmstatectl_test.py
tyll: You are right, it is not part of this PR. However, I do not follow what you are writing. The method is called `assert_rc` and it assumes that the return code is the first value of `return_tuple`. At the same time `actual` implies that it is the return code from `return_tuple`. I guess the method needs a doc string to explain when `actual =! return_tuple[0]` because it does not make sense to me to call it this way.

tests/integration/nmstatectl_test.py
```diff
@@ -49,6 +50,20 @@
...
   mtu: 65536"""
 
+ETH1_YAML_CONFIG = b"""interfaces:
+- name: eth1
+  state: up
+  type: ethernet
+  mtu: 1500
+  ipv4:
+    address:
+    - ip: 192.168.122.250
```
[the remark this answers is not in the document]
EdDev: Done

tests/integration/nmstatectl_test.py
EdDev: I will try to name and structure it better in a following patch.

tests/integration/testlib/cmd.py
EdDev: In production code, we can "show" how a new unit is used using a test. In tests, when introducing a new "service" option, it is convenient to "show" how to use it or assure it really works by using it (in the same patch).

There are however cases where the  size of the new "service" or test that uses it is too big to make sense, there separation is preferred. Also, when introducing the new "service", only one needs to use it in the same commit, the other (if there are any), better be changed in a separate commit/s.

Again, because it is a grey area, I am fine with either ways, as long as they are merged together (with gerrit, it is harder to assure they get merged together).

tyll:
required for #227

tests/integration/nmstatectl_test.py
tyll: Thank you!

tests/integration/nmstatectl_test.py
```diff
@@ -99,5 +114,12 @@ def test_show_command_only_non_existing():
     assert len(state[Constants.INTERFACES]) == 0
 
 
+def test_set_command_with_yaml_format():
+    ret = libcmd.exec_cmd(SET_CMD, stdin=ETH1_YAML_CONFIG)
+    rc, out, err = ret
```
tyll: just a minor nitpick, IMHO it is not obvious what `err` means. It seems to be `stderr` but I would not assume this just based on the variable name. Using `stdout` and `stderr` here would be more obvious since the input to `exec_cmd` is also `stdin`.

EdDev: IMO `stderr` and `stdout` may be used as either reserved keywords or just as package names, it is "too" common to use it like this.
On the other hand, it is used as an argument name by other existing function (including `exec_cmd`).

I don't really know, it feel uncomfortable for me to use std* names, perhaps due to my c/c++ history. We will then argue if this is a stream or a string and never end this discussion.
Leaving the coder to choose a reasonable name is a good middle ground IMO.

PR #95, opened 2018-10-17
tox: coverall fails to run due to urllib3 new version

EdDev:
Use older version of urllib3: 1.23

The error seen on CI:

File
"/home/travis/build/nmstate/nmstate/.tox/coveralls/local/lib/python2.7/
site-packages/requests/compat.py",
line 48, in <module>
    from urllib3.packages.ordered_dict import OrderedDict

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -106,6 +106,7 @@ commands =
 [testenv:coveralls]
 passenv = TRAVIS TRAVIS_*
 deps =
+    urllib3==1.23
     coveralls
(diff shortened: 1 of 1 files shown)


tyll:
LGTM. How do we track that we might need to revisit this in the future to ensure that urllib3 will not become too outdated?

EdDev:
I do not have a good answer to that, but as this is a tooling for the CI, it is not that important.
We can just revisit this when the next thing fails :)

PR #100, opened 2018-10-18
Docker container stabilization

EdDev:
Fix the docker container and its run to stabilize network availability and NM loading.

PR #102, opened 2018-10-19
Drop tox from the integ tests

EdDev:
automation: Drop tox usage

Instead of using tox, install all the dependencies
in the container image.

This approach should reduce the instability due to dns connectivity
problems from inside the container to the outside world.

Depends on #100

### DIFF
--- a/automation/run-integration-tests.sh
+++ b/automation/run-integration-tests.sh
@@ -37,4 +37,14 @@ docker exec $USE_TTY -i $CONTAINER_ID /bin/bash -c 'ip addr flush eth1 && ip add
 docker exec $USE_TTY -i $CONTAINER_ID /bin/bash -c 'nmcli dev; nmcli con; ip addr; ip route; cat /etc/resolv.conf; ping -c 3 github.com || true'
 
 pyclean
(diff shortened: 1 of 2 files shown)


EdDev:
@cathay4t I took you suggestion and drop tox. Hopefully this will solve CI instability regarding DNS.

automation/run-integration-tests.sh
```diff
@@ -37,4 +37,14 @@ docker exec $USE_TTY -i $CONTAINER_ID /bin/bash -c 'ip addr flush eth1 && ip add
 docker exec $USE_TTY -i $CONTAINER_ID /bin/bash -c 'nmcli dev; nmcli con; ip addr; ip route; cat /etc/resolv.conf; ping -c 3 github.com || true'
 
 pyclean
-docker exec $USE_TTY -i $CONTAINER_ID /bin/bash -c 'cd /workspace/nmstate && tox -e check-integ-py27'
+docker exec $USE_TTY -i $CONTAINER_ID /bin/bash -c '
+  cd /workspace/nmstate &&
+  pip install . &&
```
tyll: Why does this work? Why does `/root/.local/bin` not need to be added to PATH?

PR #106, opened 2018-10-22
netapplier: Show diff if verification fails

tyll:
This makes netapplier show a nice diff:

```
libnmstate.netapplier.DesiredStateIsNotCurrentError: 
desired:
========
name: voidbr
type: unknown
state: up
ipv4:
  address:
  - ip: 192.168.23.1
    prefix-length: 24
  enabled: true
ipv6:
  address:
  - ip: fe80::3fd:5f3f:db52:72a2
    prefix-length: 64
  enabled: true
mtu: 1500000

current:
========
name: voidbr
type: unknown
state: up
ipv4:
  address:
  - ip: 192.168.23.1
    prefix-length: 24
  enabled: true
ipv6:
  address:
  - ip: fe80::3fd:5f3f:db52:72a2
    prefix-length: 64
  enabled: true
mtu: 1500

difference:
===========
--- desired
+++ current
@@ -14 +14 @@
-mtu: 1500000
+mtu: 1500
```


libnmstate/netapplier.py
```diff
@@ -315,7 +318,35 @@ def assert_ifaces_state(ifaces_desired_state, ifaces_current_state):
 
         if iface_dstate != iface_cstate:
             raise DesiredStateIsNotCurrentError(
-                'desired: {}\ncurrent: {}'.format(iface_dstate, iface_cstate))
+                _format_desired_current_state_diff(iface_dstate,
+                                                   iface_cstate)
+            )
+
+
+def _format_desired_current_state_diff(desired_state, current_state):
```
EdDev: How about moving this to prettystate? Calling it using prettystate.diff_state(...) fits nicely.

libnmstate/netapplier.py
```diff
@@ -315,7 +318,35 @@ def assert_ifaces_state(ifaces_desired_state, ifaces_current_state):
...
+    pretty_desired_state = PrettyState(desired_state).yaml
+    pretty_current_state = PrettyState(current_state).yaml
+
+    diff = "".join(difflib.unified_diff(
```
EdDev: Single quotes please (unlit line 334)

libnmstate/netapplier.py
```diff
@@ -315,7 +318,35 @@ def assert_ifaces_state(ifaces_desired_state, ifaces_current_state):
...
+        pretty_desired_state.splitlines(True),
+        pretty_current_state.splitlines(True),
+        fromfile="desired",
+        tofile="current", n=0))
+    return (
+        '\n'
```
EdDev: How about using '\n'.join(['', 'desired', '=====', '{pretty_desired_state}', ...]).format(pretty_desired_state, ...)
Not a must, but it may look nicer without the multi '\n'.

libnmstate/prettystate.py
```diff
@@ -27,7 +27,10 @@ class PrettyState(object):
     def __init__(self, state):
 
         yaml.add_representer(OrderedDict, represent_ordereddict)
-        self.state = order_state(state)
+        if 'interfaces' in state:
```
EdDev: This is not that nice, it is depending on the caller: every new caller scenario will require changes here.
How about require the caller to add all it needs to add to make it here unified (add the `interfaces` key)?
Another option is to input in the constructor the order function, leaving the responsibility to the caller to choose the correct one.

tyll: Initially I made it the responsibility of the caller as well, however `order_state()` requires the `interface` key to be present, therefore we can just check for it directly:

```
iface_states = state.pop('interfaces', [])
```

Therefore I would prefer to make it simplier for the caller for now.

libnmstate/netapplier.py
tyll: This will require to make the labels "desired", "current" for the state part of the parameters. How about we change this when there is another user for the differ that makes it clearer what parameters we need.

libnmstate/netapplier.py
tyll: fixed... How about we switch to double quotes? Then we can use the black format style as well, add it to tox and do not have to change this by hand. :-) https://github.com/ambv/black

Or is there some other automatic code formatter that we could use that defaults to single ticks?

libnmstate/netapplier.py
tyll: I like the multiple newline characters more... Another option would be to define a template at the beginning of the file

```
DIFF_TEMPLATE = """
desired
======
{desired]
...
```

Since we do not want to spend too much time on these features, maybe just keep it the way it is? :-)

libnmstate/prettystate.py
EdDev: It is one `if` now but has the potential of increasing in the future (it will happen, `interfaces` are not the only root subtree we will have for sure). There is nothing common in something that in the entry point has a switch (multi if) statement.
Why not let the caller just add it? {'interfaces': ifaces_states}

If you have strong feeling on this point, another options is to just extract the `if` part out of the constructor to a factory function that will do exactly the same.

libnmstate/netapplier.py
tyll: changed this

libnmstate/prettystate.py
```diff
@@ -56,12 +82,16 @@ def represent_ordereddict(dumper, data):
...
 def order_state(state):
-    iface_states = state.pop('interfaces', [])
-    state['interfaces'] = [
-        order_iface_state(iface_state) for iface_state in sorted(
-            iface_states, key=itemgetter('name')
-        )
-    ]
+    iface_states = state.pop('interfaces', None)
+
+    state = order_iface_state(state)
```
EdDev: I think this needs to be in the `else` block of the condition below.
If the input is "full", we either have an empty dict here or in the future something other than an interface state...
I would locate it below and add a comment about the assumption (assume that it is an iface state).

libnmstate/prettystate.py
```diff
@@ -15,6 +15,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
+import difflib
```
EdDev: To keep an alphabetic order, it should be below the collections.

libnmstate/prettystate.py
tyll: When we add other things than interfaces, then the full input can contain both AFAIU. Therefore 

```
if interfaces:
...
elif routes:
...
else:
order_iface_state
```

will not work, because we need to order both interface and route states. Also he order_iface_state will also order all other dictionaries properly, therefore IMHO the current approach is good.

libnmstate/prettystate.py
tyll: yes, fixed.

PR #110, opened 2018-10-23
Schema const interfaces

EdDev:


EdDev:
Depends on #93

libnmstate/schema.py
```diff
@@ -26,3 +26,7 @@ def load(schema_name):
 
 
 ifaces_schema = load('operational-state')
+
+
+class Const(object):
```
tyll: I would prefer non-abbreviated names where possible, for example `constants` and it seems to be more natural to have them in a module than in a class on first thought. Btw. do you have plans to add all identifiers as constants?

EdDev: It fits an Enum IMO, so having it under a Class feels correct to me. The subject also fits the schema module as well.
Changing the class/enum name.

EdDev: I guess it should happen, not sure if that is an important task though.

PR #111, opened 2018-10-24
Refactor testlib

EdDev:
Based on #93

### DIFF
--- a/tests/integration/testlib/assertlib.py
+++ b/tests/integration/testlib/assertlib.py
@@ -15,62 +15,19 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
-import collections
-import copy
-import six
+from libnmstate import netinfo
 
 from . import statelib
-from .statelib import INTERFACES
 
 
-def assert_state(desired_state):
+def assert_state(desired_state_data):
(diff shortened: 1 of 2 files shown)


EdDev:
@tyll unless you have any objection, I will merge this one.

tyll:
The description in the commit message should be in the imperative mood IMHO: https://en.wikipedia.org/wiki/Imperative_mood instead of using third person passive speech (see the example from @cathay4t)  - also it is `state-related functions` AFAIK (with a hyphen) because it is an adjective. But these are minor issues, so you can also merge if you prefer not to correct this. Please keep it in mind for the future, though. :-)

EdDev:
How about now?

tyll:
Correct IMHO would be 
```
Move the state-related logic from `testlib.assertlib` to
the new class `State` in `testlib.statelib`.
```
* Move instead of Moving (imperative mood is how you would command someone to do something)
* state-related instead of state related
* it is `the new class State` and not `a new class State` - `a new class named State` would also be possible (I also added backticks around State, since it is actually code AFAICS.

EdDev:
Done

EdDev:
ci test please

PR #116, opened 2018-10-28
Nm disable no auto default

EdDev:
This PR is all about removing the special NetworkManager.conf setting which makes sure to avoid creating a connection profile automatically on interfaces.

nmstate should be able to handle this without special config.

NOTE: Tests show (with #98 ) that it is not working as expected yet. There are from time to time timeouts on waiting for the activation callback. Further investigation is needed.
Nevertheless, the changes merit a proper review.
NOTE2: The last commit needs to be removed before merging, in parallel to the creation of a new image.


EdDev:
Dependent on: #118 , #119 , #120

EdDev:
> - Maybe add a test to check that profiles that are going to be activated are not reported

This is probably  related to #120 , but I do not know how to check it in a stable manner: I can start a profile, but nothing assures it will still be there when we check it. Perhaps a dynamic IPv4 profile will do the job.
>  - On first thought it sounds better to me to wait for the profile to be fully activated and report it then. Otherwise it seems to be a bad race condition.IMHO it would be better to track this problem somewhere.

Some connections we should wait for and some we should not (like the auto ones which nmstate did not asked for). The ones we do control, we need to wait for them using a callback: Opened https://nmstate.atlassian.net/browse/NMSTATE-67 to track it.

EdDev:
ci test please

PR #117, opened 2018-10-28
automation: Collect artifacts for the CI to save

EdDev:
Collect the journalctl output and cores in case they exists.
The oVirt STDCI will save these for a limited period under the job
artifacts.


EdDev:
Confirmed to work: https://jenkins.ovirt.org/job/nmstate_nmstate_standard-check-pr/221/artifact/check-patch.el7.x86_64/journal.log

PR #120, opened 2018-10-28
nm: Expose device connection profile only if active

EdDev:
Fetching a device connection profile based on the device should only
consider activated connection profiles and not in transient ones.

The caller code is usually not expecting the returned connection profile
to be in between states. In a transient state, the connection profile
may miss several expected properties, like the interfaces it is bound
to.

### DIFF
--- a/libnmstate/nm/connection.py
+++ b/libnmstate/nm/connection.py
@@ -134,7 +134,9 @@ def set_master_setting(con_setting, master, slave_type):
 
 def get_device_connection(nm_device):
     act_connection = get_device_active_connection(nm_device)
-    if act_connection:
+    if (act_connection and
+            act_connection.props.state ==
+            nmclient.NM.ActiveConnectionState.ACTIVATED):
         return act_connection.props.connection
     return None
 
--- a/tests/lib/nm/connection_test.py
+++ b/tests/lib/nm/connection_test.py
@@ -125,8 +125,10 @@ def test_set_master_setting():
     assert con_setting_mock.props.slave_type == 'slave-type'
 
 
-def test_get_device_connection():
+def test_get_device_connection(NM_mock):
     dev_mock = mock.MagicMock()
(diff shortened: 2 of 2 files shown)


libnmstate/nm/connection.py
```diff
@@ -134,7 +134,9 @@ def set_master_setting(con_setting, master, slave_type):
 
 def get_device_connection(nm_device):
     act_connection = get_device_active_connection(nm_device)
-    if act_connection:
+    if (act_connection and
+            act_connection.props.state ==
+            nmclient.NM.ActiveConnectionState.ACTIVATED):
```
tyll: https://github.com/nmstate/nmstate/pull/116#issuecomment-433741650
> > * Maybe add a test to check that profiles that are going to be activated are not reported
> 
> This is probably related to #120 , but I do not know how to check it in a stable manner: I can start a profile, but nothing assures it will still be there when we check it. Perhaps a dynamic IPv4 profile will do the job.
> 
> > * On first thought it sounds better to me to wait for the profile to be fully activated and report it then. Otherwise it seems to be a bad race condition.IMHO it would be better to track this problem somewhere.
> 
> Some connections we should wait for and some we should not (like the auto ones which nmstate did not asked for). The ones we do control, we need to wait for them using a callback: Opened https://nmstate.atlassian.net/browse/NMSTATE-67 to track it.

What does it mean not to wait for the automatically generated profiles? Wouldn't those be the profiles that we would want to change with nmstate? @cathay4t what is your opinion here?

cathay4t:
Root cause(I guess):
 * When any NIC/interface does not have NetworkManager(short to NM) profile bounded to it, NM will create a default new profile with empty `connection.interface-name` property while `802-3-ethernet.mac-address` is used to match the NIC.

Still working on a stable reproducer.

EdDev:
Closing this one for the time being.
We will need eventually to decide how to treat devices in transient state when attempting to configure them.

PR #121, opened 2018-10-28
tests integ: Adding ethernet ipv4 tests

SvetaHBar:
Two new tests were added to the ethernet integration tests:
- remove static ipv4
- edit static ipv4 address and prefix

### DIFF
--- a/tests/integration/static_ip_address_test.py
+++ b/tests/integration/static_ip_address_test.py
@@ -17,11 +17,37 @@
 
 import copy
 
+import pytest
+
 from libnmstate import netapplier
 
 from .testlib import assertlib
 from .testlib import statelib
 from .testlib.statelib import INTERFACES
+
+
+IPV4_ADDRESS1 = '192.0.2.251'
+IPV4_ADDRESS2 = '192.0.2.252'
+
+
+@pytest.fixture
+def setup_eth1_ipv4():
+    desired_state = {
+        INTERFACES: [
+            {
+                'name': 'eth1',
+                'type': 'ethernet',
+                'state': 'up',
+                'ipv4': {
+                    'enabled': True,
+                    'address': [
+                        {'ip': IPV4_ADDRESS1, 'prefix-length': 24}
+                    ]
+                }
+            }
+        ]
+    }
+    netapplier.apply(desired_state)
 
 
 def test_add_static_ipv4_with_full_state():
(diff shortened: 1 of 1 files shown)


tests/integration/static_ip_address_test.py
```diff
@@ -16,13 +16,32 @@
...
+    desired_state = {
+        INTERFACES: [
+            {
+                'name': 'eth1',
+                'type': 'ethernet',
+                'state': 'up',
+                'ipv4': {
+                    'enabled': True,
+                    'address': [
+                        {'ip': '192.168.122.251', 'prefix-length': 24}
```
tyll: Please use a designated test address, such as 192.0.2.251, see 
https://tools.ietf.org/html/rfc5737#section-3 - this will reduce problems with address conflicts.

tests/integration/static_ip_address_test.py
```diff
@@ -57,3 +76,46 @@ def test_add_static_ipv4_with_min_state():
...
+    desired_state = {
+        INTERFACES: [
+            {
+                'name': 'eth1',
+                'type': 'ethernet',
+                'state': 'up',
+                'ipv4': {
+                    'enabled': True,
+                    'address': [
+                        {'ip': '192.168.122.64', 'prefix-length': 23}
```
tyll: Please use a designated test address, such as 192.0.2.251, see 
https://tools.ietf.org/html/rfc5737#section-3 - this will reduce problems with address conflicts.

tests/integration/static_ip_address_test.py
```diff
@@ -57,3 +76,46 @@ def test_add_static_ipv4_with_min_state():
...
+                }
+            }
+        ]
+    }
+
+    print desired_state
+    netapplier.apply(copy.deepcopy(desired_state))
+
+    assertlib.assert_state(desired_state)
+
```
tyll: Please use only two newlines between method definitions here.

tests/integration/static_ip_address_test.py
```diff
@@ -57,3 +76,46 @@ def test_add_static_ipv4_with_min_state():
...
+                'name': 'eth1',
+                'type': 'ethernet',
+                'ipv4': {
+                    'enabled': False
+                }
+            }
+        ]
+    }
+
+    print desired_state
```
tyll: This will not work with python3 and does not belong here.

tests/integration/static_ip_address_test.py
```diff
@@ -16,6 +16,7 @@
 #
 
 import copy
+import pytest
```
EdDev: Please separate from the above with a newline

tests/integration/static_ip_address_test.py
SvetaHBar: Done

tests/integration/static_ip_address_test.py
SvetaHBar: Done

tyll:
Thank you very much!

PR #123, opened 2018-10-29
Execute all actions in mainloop

EdDev:
Depends on #122

### DIFF
--- a/libnmstate/nm/nmclient.py
+++ b/libnmstate/nm/nmclient.py
@@ -96,7 +96,11 @@ class _MainLoop(object):
         if not isinstance(timeout, six.integer_types):
             raise TypeError('timeout is expected to be an integer')
 
+        if not self.actions_exists():
+            return _MainLoop.SUCCESS
+
         with self._idle_timeout(timeout):
+            self._register_first_action()
             self._mainloop.run()
 
         if self._error == _MainLoop.RUN_TIMEOUT_ERROR:
@@ -151,3 +155,10 @@ class _MainLoop(object):
         logging.warning('NM main-loop timed out.')
         mainloop.quit()
         return _MainLoop.FAIL
+
+    def _register_first_action(self):
+        GLib.timeout_add(1, self._execute_action_once, None)
+
+    def _execute_action_once(self, _):
+        self.execute_next_action()
+        return False
--- a/tests/integration/iface_admin_state_test.py
+++ b/tests/integration/iface_admin_state_test.py
@@ -1,0 +1,40 @@
+#
+# Copyright 2018 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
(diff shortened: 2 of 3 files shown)


libnmstate/nm/nmclient.py
```diff
@@ -92,8 +92,7 @@ def actions_exists(self):
 
     def run(self, timeout):
         if timeout is None:
-            self._mainloop.run()
-            return _MainLoop.SUCCESS
+            timeout = 3600
```
tyll: This does not seem to be a sane default. Nobody will want to wait 60 minutes IMHO. How about 20 seconds as used in netapplier.

tests/integration/iface_admin_state_test.py
```diff
@@ -0,0 +1,38 @@
...
+    desired_state = {
+        INTERFACES: [
+            {
+                'name': 'eth1',
+                'type': 'ethernet',
+                'state': 'down',
+            }
+        ]
+    }
+    netapplier.apply(copy.deepcopy(desired_state))
```
tyll: How about applying the state twice to ensure that the interface is really down when it is set down again.

libnmstate/nm/nmclient.py
EdDev: it is not a default, it is the alternative of "no timeout". Previously it was indefinitely, now it is limited to "long time".
It should not be used unless in a debug mode or something.

tests/integration/iface_admin_state_test.py
EdDev: The fixture assures that: that is the base state.
But I can add an assert here to make sure we do start from the correct state in the test.

libnmstate/nm/nmclient.py
tyll: indefinitely is wrong, too, I agree. If there needs to be an insanely high timeout, then it could be set to a higher value in a debug build. But it should also not block for 60 minutes by accident. Also if it should not be used, then it can be a sane value like 20 seconds anyhow, so there is no reason to set it that high. :-)

tests/integration/iface_admin_state_test.py
tyll: The assert would be great to make this obvious.

libnmstate/nm/nmclient.py
EdDev: It was originally copied from the anisble roles, so this is a translation of that indefinitely.
I would prefer no one will set it to None, till change to raise on it.

EdDev: Ahh and this should have been discussed on #122

tests/integration/iface_admin_state_test.py
EdDev: See #122

libnmstate/nm/nmclient.py
EdDev: Should be discussed on #122

tests/integration/iface_admin_state_test.py
EdDev: I added it

PR #124, opened 2018-10-29
automation: Wait for the dbus service to come up in the container

EdDev:
This should fix the random CI failures due to the dbus service too slow to come up.

### DIFF
--- a/automation/run-integration-tests.sh
+++ b/automation/run-integration-tests.sh
@@ -92,7 +92,7 @@ mkdir -p $EXPORT_DIR
 
 CONTAINER_ID="$(docker run --privileged -d -v /sys/fs/cgroup:/sys/fs/cgroup:ro -v $PROJECT_PATH:/workspace/nmstate -v $EXPORT_DIR:$CONT_EXPORT_DIR $DOCKER_IMAGE)"
 trap run_exit EXIT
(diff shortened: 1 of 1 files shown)


automation/run-integration-tests.sh
```diff
@@ -92,7 +92,7 @@ mkdir -p $EXPORT_DIR
 
 CONTAINER_ID="$(docker run --privileged -d -v /sys/fs/cgroup:/sys/fs/cgroup:ro -v $PROJECT_PATH:/workspace/nmstate -v $EXPORT_DIR:$CONT_EXPORT_DIR $DOCKER_IMAGE)"
 trap run_exit EXIT
-docker_exec 'systemctl start dbus.socket'
+docker_exec 'while ! systemctl status dbus &> /dev/null; do sleep 1; done'
```
tyll: I guess systemctl is-active would be the better command here. Not sure how long this should take but the sleep could be reduced to 100 ms or so if it is usually less than a second.

EdDev: It takes 1sec on my machine, so it is much costly to retry 10 times on this.
The best solution seems to wait a second.
I left the output so we can see it on CI.
Used the `is-active`, thanks.

PR #126, opened 2018-10-29
prettystate: Do not change the input state

tyll:
Without this, editing does not work.


EdDev:
Please try to say a bit more in the commit message and mention that a test was added.

PR #127, opened 2018-10-29
readme: Add basic coding-guidelines and commit template

EdDev:
For consistency I'm adding some basic guidelines.
These are taken from oVirt VDSM project.

### DIFF
--- a/README.developer.md
+++ b/README.developer.md
@@ -1,0 +1,28 @@
+# NMState Coding Guidelines
+
+- NMState is written primarily in Python, and its coding style should follow
+  the best practices of Python coding unless otherwise declared.
+- PEP8 is holy.
+- Tests are holy.
+  Production code must be covered by unit tests and/or basic integration tests.
+  When too many mocks are required, it is often a smell that the tested code
+  is not well structured or in some cases a candidate for integration tests.
+- Commit message should use the provided commit-template.txt as template:
+  You can set the template for nmstate by configuring git:
+  `git config commit.template commit-template.txt` 
+- Packages, modules, functions, methods and variables should use
+  underscore_separated_names.
+- Class names are in CamelCase.
+- Imports should be grouped in the following order:
+-- Standard library imports
+-- Related third party imports
+-- Local application-specific or library-specific imports.
+- All indentation is made of the space characters.
+  Tabs are evil. In makefiles, however, tabs are obligatory.
+  White space between code stanzas are welcome. They help to create breathing
+  while reading long code.
+  However, splitting stanzas into helper functions could be even better.
+- Prefer single quotes (') whenever possible.
+
+Note: This is a modified version of VDSM conding guidelines:
+https://www.ovirt.org/develop/developer-guide/vdsm/coding-guidelines/
--- a/commit-template.txt
+++ b/commit-template.txt
@@ -1,0 +1,15 @@
+topic: Short summary under 50 chars (if possible)
+
+The topic should be a module name, a subsystem name, or a feature name,
+helping to understand what is this patch about.
+
+Please follow the seven rules of a great Git commit message:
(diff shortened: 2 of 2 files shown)


commit-template.txt
```diff
@@ -0,0 +1,11 @@
+topic: Short summary under 50 chars (if possible)
+
```
tyll: How about quoting the seven rules for a good commit message from https://chris.beams.io/posts/git-commit/

IMHO at least using the imperative mood is missing here.

README.developer.md
```diff
@@ -0,0 +1,18 @@
+# NMState Coding Guidelines
+
+- NMState is written primarily in Python, and its coding style should follow the best practices of Python coding unless otherwise declared.
+- PEP8 is holy.
+- Tests are holy. Production code must be covered by unit tests and basic integration tests.
```
tyll: We do not follow this completely, therefore "must" is a little bit too strong.

README.developer.md
```diff
@@ -0,0 +1,18 @@
+# NMState Coding Guidelines
+
+- NMState is written primarily in Python, and its coding style should follow the best practices of Python coding unless otherwise declared.
+- PEP8 is holy.
```
tyll: How about mentioning that commits should be checked with tox here, it includes a pep8 check afaik.

README.developer.md
```diff
@@ -0,0 +1,18 @@
...
+  You can set the template for nmstate by configuring git: `git config commit.template commit-template.txt` 
+- Packages, modules, functions, methods and variables should use underscore_separated_names.
+- Class names are in CamelCase.
+- Imports should be grouped in the following order:
+-- Standard library imports
+-- Related third party imports
+-- Local application-specific or library-specific imports.
+- All indentation is made of the space characters. Tabs are evil. In makefiles, however, tabs are obligatory.
+  White space between code stanzas are welcome. They help to create breathing while reading long code. However, splitting stanzas into helper functions could be even better.
+
```
tyll: The preference of single ticks should be mentioned here and ideally also some tooling to help to conform to this (or we use double quotes for which tooling exists ;-)).

README.developer.md
cathay4t: I am OK with `must`. But it could be better we could say more about how to create unit test and integration test like:
```
Please refer to commit <commit_id_with_unit_test_integration_test> as an example.
```

EdDev: 'must': We should follow it, and I hope you will nack my patches if it does not. But I should rephrase it to: "unit tests and/or basic integration tests". There are tests which are better covered using the integration tests and not unit tests (the ones who need too much mocking).

About the 'how': Well, I do hope the one who attempts to post a change knows where too look. I usually go to tests to learn how to change something or how it currently works.

README.developer.md
EdDev: I got used to single quotes and it feels more minimalistic.. the python way. As we are already there, and I like it so much, I will indeed add it :)
I do not have tooling, but from experience after you get used to it, it sticks.

commit-template.txt
EdDev: Ok, will try to embed it here.

tyll:
@EdDev your caller to callee preference is also missing but this should also have tooling support.

EdDev:
@tyll I have a bunch of other preferences in the queue, lets push this basic version and then continue with the others in a following patch.

I also like tooling, but I think it cannot be a condition for guidance and preferences. Some even cannot have tooling as exceptions are acceptable. With the caller-callee preference, it sometimes does not work with python due to the interpreter processing or because of an contradicting preference (one I can think of is: pytest prefixes).

EdDev:
I opened https://nmstate.atlassian.net/browse/NMSTATE-70 to commit for doing this in this sprint.
I hope this is enough.

PR #128, opened 2018-10-30
[INVESTIGATION] Tests with conn setting debug info

EdDev:


EdDev:
ci test please

EdDev:
@tyll , @cathay4t , it seems like the problems we have in the current nmstate flow is with how we explicitly activate the connections and the `autoconnect` option enabled on the new connection.
One is causing the connection to be activated automatically and the other is attempting to cancel it and issue a new (explicit) activation.

Any suggestions?

thom311:
It you add/modify a profile with autoconnect=true, then it may auto-activate right away. With `Update2` D-Bus method, this can be blocked for modification. Blocking for add is not supported (yet). I guess, one could add it with `connection.autoconnect=no`, and then issue a `Update2()` to enable autoconnect on the profile and blocking it at the same time. See also https://github.com/linux-system-roles/network/issues/35#issue-303190135

thom311:
Update2() is a super-set of all other similar D-bus methods. Meaning, in libnm, `commit_changes()` actually calls `Update2()`: https://cgit.freedesktop.org/NetworkManager/NetworkManager/tree/libnm/nm-remote-connection.c?id=3648c58bc5f0e5195ad946b8a3e5efe65c89e4c0#n267

Only caveat is, that Update2() is relatively new API. You may or may not have it available (check with d-feet).

PR #129, opened 2018-10-30
integration tests: Normalize IPv4 address usage

tyll:


tests/integration/static_ip_address_test.py
```diff
@@ -113,7 +115,7 @@ def test_edit_static_ipv4_address_and_prefix(setup_eth1_ipv4):
                 'ipv4': {
                     'enabled': True,
                     'address': [
-                        {'ip': IPV4_ADDRESS2, 'prefix-length': 23}
+                        {'ip': IPV4_ADDRESS2, 'prefix-length': 24}
```
EdDev: @tyll this is wrong, the test is checking if the address *and* prefix changed. Please revert back this part.

tests/integration/static_ip_address_test.py
```diff
@@ -25,9 +25,11 @@
 from .testlib import statelib
 from .testlib.statelib import INTERFACES
 
-
+# TEST-NET addresses: https://tools.ietf.org/html/rfc5737#section-3
 IPV4_ADDRESS1 = '192.0.2.251'
 IPV4_ADDRESS2 = '192.0.2.252'
+IPV4_ADDRESS3 = '198.51.100.249'
```
EdDev: I do not understand why additional addresses were needed, using the existing ones could be enough.
But it is not harming anything, just raising a note for the future.

tests/integration/static_ip_address_test.py
tyll: Good catch, I changed the prefix to `30`, `23` would be outside the range of the test network.

tests/integration/static_ip_address_test.py
tyll: Maybe only one extra address was needed. My motivation was to avoid problems with setting the same address to different interfaces (eth1 and eth2). However, this should probably also be a test case - I wonder how nmstate will handle this.

EdDev: Yes, this is a good test idea. @SvetaHBar could you please check this?

PR #132, opened 2018-10-30
Temporary disable the connection setting autoconnect

EdDev:


tyll:
this breaks travis

EdDev:
I did mention it in the commit message.
I think going one step backward in order to go forward is better than block new integration tests. CI also fails randomly because of it.

I will fix it in the weekend or early next week. Will open a Jira issue on it.

EdDev:
Opened https://nmstate.atlassian.net/browse/NMSTATE-71 to track re-enabling autoconnect.

PR #133, opened 2018-10-31
Docker: Allow to start shell with variable

tyll:
Make it possible to enter a shell in the integration test docker image
before it would exit. This makes it easier to debug/develop integration
tests.

### DIFF
--- a/automation/run-integration-tests.sh
+++ b/automation/run-integration-tests.sh
@@ -85,13 +85,22 @@ function run_exit {
     remove_container
 }
 
+function open_shell {
+    res=$?
+    [ "$res" -ne 0 ] && echo "*** ERROR: $res"
+    set +o errexit
+    docker_exec 'echo "pytest tests/integration --pdb" >> ~/.bash_history'
+    docker_exec 'cd /workspace/nmstate; exec /bin/bash'
+    run_exit
+}
+
 cd $EXEC_PATH 
 docker --version && cat /etc/resolv.conf && ping -c 1 github.com
 
 mkdir -p $EXPORT_DIR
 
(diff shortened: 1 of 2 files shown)


automation/run-integration-tests.sh
```diff
@@ -85,13 +85,22 @@ function run_exit {
...
+}
+
 cd $EXEC_PATH 
 docker --version && cat /etc/resolv.conf && ping -c 1 github.com
 
 mkdir -p $EXPORT_DIR
 
 CONTAINER_ID="$(docker run --privileged -d -v /sys/fs/cgroup:/sys/fs/cgroup:ro -v $PROJECT_PATH:/workspace/nmstate -v $EXPORT_DIR:$CONT_EXPORT_DIR $DOCKER_IMAGE)"
-trap run_exit EXIT
+[ -n "$debug_exit_shell" ] && trap open_shell EXIT || trap remove_container EXIT
```
EdDev: Why did you changed run_exit?

If I understand correctly, the shell will be opened after the run_tests is executed or if an error causes the script to exit. Right?

automation/run-integration-tests.sh
```diff
@@ -85,13 +85,22 @@ function run_exit {
     remove_container
 }
 
+function open_shell {
+    res=$?
+    [ "$res" -ne 0 ] && echo "*** ERROR: $res"
+    set +o errexit
+    docker_exec 'echo "pytest tests/integration --pdb" >> ~/.bash_history'
+    docker_exec 'cd /workspace/nmstate; exec /bin/bash'
+    remove_container
```
EdDev: It will be safer to leave run_exit (in case it will change in the future).
Or do you see a problem with this?

tyll: No, this was just because I assumed that the artifacts collection is not needed for manual runs. But it does not hurt to keep it.

automation/run-integration-tests.sh
tyll: yes, the change was unintentional.

tyll: You understand it correctly. I updated the branch.

PR #135, opened 2018-11-04
nm: Fix connection debug logging formatting

EdDev:


EdDev:
No idea how this slipped in, but we need master fixed.
(detected while debugging bond creation)

EdDev:
Fixing my own miserable mistake. If there is something you would like me to change, please let me know. I'm merging this as is for now to allow other fixes in the queue to run on CI.

PR #137, opened 2018-11-05
Ignore absent ifaces from dstate

EdDev:
depends on #136

### DIFF
--- a/tests/integration/iface_admin_state_test.py
+++ b/tests/integration/iface_admin_state_test.py
@@ -17,6 +17,8 @@
 
 import copy
 
+import pytest
+
 from libnmstate import netapplier
 
 from .testlib import assertlib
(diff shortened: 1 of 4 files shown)


tests/integration/iface_admin_state_test.py
```diff
@@ -38,3 +40,20 @@ def test_set_a_down_iface_down():
     netapplier.apply(copy.deepcopy(desired_state))
 
     assertlib.assert_state(desired_state)
+
+
+@pytest.mark.xfail(reason='Some ifaces cannot be removed', strict=True)
```
tyll: Should there also be an xfail for virtual interfaces that cannot be set to down, because down equals absent for them?

EdDev: Yes, but note that it exposes a larger scope problem: The way we handle `down` is completely wrong.
Do we have a Jira issue on this?

EdDev:
> Looks good to me, wonder if state down needs to be normalized to absent for virtual interfaces.

One  problem at a time :)
We should try it with bond and see what happens (not sure if we need to reject `down` or normalize it)

tests/integration/iface_admin_state_test.py
tyll: I do not think that there is a Jira issue.

PR #138, opened 2018-11-05
Bond base integration tests (creation and removal)

EdDev:
Depends on #136 #137

### DIFF
--- a/tests/integration/bond_test.py
+++ b/tests/integration/bond_test.py
@@ -1,0 +1,53 @@
+#
+# Copyright 2018 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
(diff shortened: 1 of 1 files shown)


tests/integration/bond_test.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+@pytest.fixture
+def bond99_base():
+    state = yaml.load(BOND99_YAML_BASE)
+    netapplier.apply(copy.deepcopy(state))
+
+    yield copy.deepcopy(state)
+
+    state[INTERFACES][0]['state'] = 'absent'
```
tyll: IMHO removing the bond should be tested explicitly in a test and not only in a fixture.

EdDev: Good point, thanks.

tests/integration/bond_test.py
```diff
@@ -0,0 +1,53 @@
...
+    mode: balance-rr
+    slaves:
+    - eth1
+    - eth2
+"""
+
+
+def test_add_and_remove_bond_with_two_slaves():
+    state = yaml.load(BOND99_YAML_BASE)
+    netapplier.apply(copy.deepcopy(state))
```
tyll: Is the deepcopy necessary here? If it is, wouldn't it be better to do it in netapplier.apply?

EdDev: Yes, because `apply` may change the dict we later want to compare against.
Unfortunately, we do not have a way to make sure no one changes our data.

EdDev: Ohh missed the second part. Yes, we need to move it into apply().

PR #140, opened 2018-11-06
readme: Add Clean Code guidelines for developers

EdDev:


README.developer.md
```diff
@@ -14,15 +14,38 @@
...
+- Name classes using a noun.
+- Name functions/methods using a verb.
+- Make them as small as possible.
+- Should do one thing only and do it well.
+  One thing means one level of abstraction.
+  The names and code should reflect that.
+- Methods/functions should be organized per level of abstraction,
+  where callee sits below their caller.
+- Avoid output-arguments (arguments to output data out of a function/method).
+- Don’t use boolean arguments, use 2 functions/methods instead.
```
tyll: This seems to be a little bit hypocritical since our public API `netinfo.show()` and `netapplier.apply()` both use boolean arguments. I would maybe agree with `Try to avoid boolean arguments, consider using two functions/methods instead.`

README.developer.md
```diff
@@ -14,15 +14,38 @@
...
+- Comply to CQS, either:
```
tyll: Please avoid acronyms without introducing them. Also this is something that is controversial and we do not follow this in the current code and the python stdlib does not as well. I would remove this.

tyll:
Also I think @cathay4t should also review this before we merge it.

EdDev:
ci test please

EdDev:
@cathay4t , could you please review this one.

PR #143, opened 2018-11-07
Ipv6 fixes

EdDev:


tests/integration/ethernet_mtu_test.py
```diff
@@ -34,7 +34,6 @@ def test_increase_iface_mtu():
     eth1_desired_state = desired_state[INTERFACES][0]
     eth1_desired_state['state'] = 'up'
     eth1_desired_state['mtu'] = 1900
-    eth1_desired_state['ipv6']['enabled'] = True
```
tyll: Thre is a FIXME about IPv6 in this file that needs to be removed as well when these lines are removed.

EdDev:
ci test please

tests/integration/ethernet_mtu_test.py
EdDev: thanks, I missed that. Fixed.

tests/integration/ethernet_mtu_test.py
```diff
@@ -26,15 +26,12 @@
...
 
-# FIXME: Once IPv6 disabling is supported, below IPv6 codes should be removed.
-
 
 def test_increase_iface_mtu():
     desired_state = statelib.show_only(('eth1',))
     eth1_desired_state = desired_state[INTERFACES][0]
     eth1_desired_state['state'] = 'up'
     eth1_desired_state['mtu'] = 1900
-    eth1_desired_state['ipv6']['enabled'] = True
```
tyll: IMHO the test is flawed. Instead of taking the full current state of eth1 and changing some things it should only take the necessary items (name, type) and add the items it would like to change. At least this is what we discussed is needed for users to do for ethernet and IPv6 settings IIRC.

EdDev: Sorry, I do not understand. It is reasonable for the user to get the current state and modify the parts he wants, sending it back. What is wrong about it?
The `edit` command does exactly that.

EdDev:
ci test please

EdDev:
@tyll , this should now work on both the container and VM (passed on mine).

libnmstate/sysctl.py
```diff
@@ -0,0 +1,44 @@
...
+import errno
+
+
+def enable_ipv6(dev):
+    disable_ipv6(dev, disable=False)
+
+
+def disable_ipv6(dev, disable=True):
+    try:
+        with open('/proc/sys/net/ipv6/conf/%s/disable_ipv6' % dev, 'w') as f:
```
tyll: This should use .format() IMHO, since we use it everywhere else, too.

libnmstate/sysctl.py
```diff
@@ -0,0 +1,44 @@
...
+    except IOError as e:
+        if e.errno == errno.ENOENT and disable:
+            # IPv6 stack is (already) not available on this device
+            return
+        raise
+
+
+def is_disabled_ipv6(dev='default'):
+    try:
+        with open('/proc/sys/net/ipv6/conf/%s/disable_ipv6' % dev) as f:
```
tyll: Please use .format()

libnmstate/sysctl.py
```diff
@@ -0,0 +1,44 @@
...
+#
+
+import errno
+
+
+def enable_ipv6(dev):
+    disable_ipv6(dev, disable=False)
+
+
+def disable_ipv6(dev, disable=True):
```
tyll: Here is a boolean argument that you do not like. Since the method is called disable and has a disable parameter, it should maybe better be called `configure_ipv6()` or `set_ipv6()`. Or make the argument the number and then have two functions for enabling/disabling that call this function with the right number.

libnmstate/sysctl.py
```diff
@@ -0,0 +1,44 @@
...
+        with open('/proc/sys/net/ipv6/conf/%s/disable_ipv6' % dev, 'w') as f:
+            f.write('1' if disable else '0')
+    except IOError as e:
+        if e.errno == errno.ENOENT and disable:
+            # IPv6 stack is (already) not available on this device
+            return
+        raise
+
+
+def is_disabled_ipv6(dev='default'):
```
tyll: It would be more natural to call this `is_ipv6_disabled()`. Also it should probably return a boolean instead of the number.

tests/integration/conftest.py
```diff
@@ -32,6 +34,25 @@ def logging_setup():
...
+    NM is learning the initial IPv6 stack state for an interface when the
+    interfaces is first managed. Therefore, this function makes sure the
+    learned state is IPv6 disabled so NM will return the interfaces to this
+    state when marked with IPv6 method=ignore.
+
+    FIXME: This special handling should be dropped when either NM fixes its
+    behaviour or an explicit disable_ipv6=1 is executed by nmstate.
+    """
+    for ifname in ('eth1', 'eth2'):
+        cmd.exec_cmd(('nmcli', 'device', 'set', ifname, 'managed', 'no'))
```
tyll: This code should be part of libnmstate and not part of the test.

libnmstate/sysctl.py
EdDev: Sure.
I just copied this from vdsm.

tests/integration/ethernet_mtu_test.py
tyll: A real user will inspect the full state and see what needs to be changed and setting the state to up with ipv6 disabled would not be allowed with the edit command as well.

libnmstate/sysctl.py
EdDev: Many like using the old stile for a *single* variable scenario, it is shorter and cleaner.
If you insist on this, I can change it (let me know).

libnmstate/sysctl.py
EdDev: Done

tests/integration/conftest.py
EdDev: I tried embedding it in libnmstate but it is not possible.
This code makes sure the test env is ready for use by nmstate with the ipv6 limitation in mind.

tests/integration/ethernet_mtu_test.py
EdDev: It is allowed if we follow the workaround based on the latest findings.

tyll: But it is still not allowed for the regular user since the workaround only happens in the test and not when the user actually used edit. IMHO this make the tests meaningless. Also since it is breaking the mtu tests, it shows that the mtu tests are too broad because they should not test the IPv6 behavior.

EdDev: The "regular" user may have ipv6 disabled in the first place, you do not have control over it one way or the other. We can just say in what conditions it works now.

It breaks other tests because this is an integration test and not unit test, integration tests do break unrelated things. But this last point has nothing to do with this PR anyway.

EdDev:
@tyll this version is following your suggestion, raising test iface up.

tests/integration/ethernet_mtu_test.py
```diff
@@ -80,40 +72,33 @@ def test_decrease_to_zero_iface_mtu():
...
-def test_decrease_to_min_ethernet_frame_size_iface_mtu():
+def test_decrease_to_ipv6_min_ethernet_frame_size_iface_mtu():
     desired_state = statelib.show_only(('eth1',))
-    origin_desired_state = copy.deepcopy(desired_state)
     eth1_desired_state = desired_state[INTERFACES][0]
-    eth1_desired_state['state'] = 'up'
-    # the min is 64 - 18 = 46
-    eth1_desired_state['mtu'] = 40
-    eth1_desired_state['ipv6']['enabled'] = True
+    eth1_desired_state['mtu'] = 1280
```
tyll: Why no test for values below 1280? 1279 would also be a good candidate.

libnmstate/nm/ipv6.py
```diff
@@ -48,10 +48,8 @@ def get_info(active_connection):
 
 def create_setting(config):
```
tyll: IMHO it would make sense to explain why the IP6_CONFIG methods are chosen like this to ensure that we understand the code when we look at it again in the future.

EdDev:
ci test please

tests/integration/ethernet_mtu_test.py
SvetaHBar: syre

PR #150, opened 2018-11-11
IPv6 tests: added new IPv6 static tests

SvetaHBar:
the tests are added:
- Remove IPv6 static test - failed
- Add static ipv6 with min state
- Edit static ipv6 and prefix
- Add ifaces with same static ipv6 in one transaction
- Add ifaces with same static ipv6 to existing

Signed-off-by: Sveta Haas <sveta.haas@gmail.com>

### DIFF
--- a/tests/integration/static_ip_address_test.py
+++ b/tests/integration/static_ip_address_test.py
@@ -58,6 +58,28 @@ def setup_eth1_ipv4():
     netapplier.apply(desired_state)
 
 
+@pytest.fixture
+def setup_eth1_ipv6():
+    desired_state = {
+        INTERFACES: [
+            {
+                'name': 'eth1',
+                'type': 'ethernet',
+                'state': 'up',
+                'ipv6': {
+                    'enabled': True,
+                    'address': [
+                        {'ip': IPV6_ADDRESS1, 'prefix-length': 64}
+                    ]
+                }
+            }
+        ]
+    }
+    netapplier.apply(desired_state)
+
+    return desired_state
+
+
 def test_add_static_ipv4_with_full_state():
     desired_state = statelib.show_only(('eth1',))
     eth1_desired_state = desired_state[INTERFACES][0]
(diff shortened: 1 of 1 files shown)


tests/integration/static_ip_address_test.py
```diff
@@ -257,3 +277,119 @@ def test_add_static_ipv6_with_no_address():
...
+                        {'ip': IPV6_ADDRESS2, 'prefix-length': 30}
+                    ]
+                }
+            }
+        ]
+    }
+
+    netapplier.apply(copy.deepcopy(desired_state))
+
+    assertlib.assert_state(desired_state)
```
EdDev: This does not confirm that the previous address is not in the current state.

tests/integration/static_ip_address_test.py
```diff
@@ -257,3 +277,119 @@ def test_add_static_ipv6_with_no_address():
...
+
+
+def test_edit_static_ipv6_address_and_prefix(setup_eth1_ipv6):
+    desired_state = {
+        INTERFACES: [
+            {
+                'name': 'eth1',
+                'type': 'ethernet',
+                'state': 'up',
```
EdDev: - `state` is not required (you depend on the fixture which already handled it).
- ipv6.enabled may also be dropped (not sure) as it is already enabled by the fixture.

tests/integration/static_ip_address_test.py
```diff
@@ -58,6 +58,26 @@ def setup_eth1_ipv4():
     netapplier.apply(desired_state)
 
 
+@pytest.fixture
+def setup_eth1_ipv6():
```
EdDev: Multiple tests are using the fixture, therefore, it is better add a teardown to remove the address.
On the other hand, we have a general fixture (in conftest) which takes eth1/2 down and up again before each test.

Ahhh that `cleanup_test_interfaces` needs to be changed, it is a bad too wide fixture.

tests/integration/static_ip_address_test.py
```diff
@@ -257,3 +277,119 @@ def test_add_static_ipv6_with_no_address():
...
+                }
+            }
+        ]
+    }
+    netapplier.apply(copy.deepcopy(desired_state))
+
+    assertlib.assert_state(desired_state)
+
+
+@pytest.mark.xfail
```
EdDev: Better give a reason and make it strict so we will not miss it when we fix the problem.

tests/integration/static_ip_address_test.py
```diff
@@ -257,3 +277,119 @@ def test_add_static_ipv6_with_no_address():
...
+def test_remove_static_ipv6(setup_eth1_ipv6):
```
EdDev: It is `disable` not `remove`.

tests/integration/static_ip_address_test.py
SvetaHBar: Done

tests/integration/static_ip_address_test.py
SvetaHBar: Done

tests/integration/static_ip_address_test.py
SvetaHBar: Done

tests/integration/static_ip_address_test.py
SvetaHBar: Fixed

SvetaHBar:
ci test please

tests/integration/static_ip_address_test.py
```diff
@@ -257,3 +279,108 @@ def test_add_static_ipv6_with_no_address():
...
+
+
+def test_edit_static_ipv6_address_and_prefix(setup_eth1_ipv6):
+    setup_eth1_ipv6[INTERFACES][0]['ipv6']['address'] = [
+         {'ip': IPV6_ADDRESS2, 'prefix-length': 64}
+     ]
+
+    netapplier.apply(copy.deepcopy(setup_eth1_ipv6))
+
+    assert setup_eth1_ipv6[INTERFACES][0]['ipv6']['address'] == [
```
EdDev: This is checking the desired state and not the current one.
The current needs to be read using `show`.

SvetaHBar: Fixed

tests/integration/static_ip_address_test.py
```diff
@@ -255,3 +277,130 @@ def test_add_static_ipv6_with_no_address():
...
+                        {'ip': IPV6_ADDRESS2, 'prefix-length': 24}
+                    ]
+                }
+            }
+        ]
+    }
+
+    netapplier.apply(copy.deepcopy(desired_state))
+    eth1_desired_state = desired_state[INTERFACES][0]
+    print eth1_desired_state
```
EdDev: Leftovers

SvetaHBar: Fixed

SvetaHBar:
ci test please

SvetaHBar:
ci test please

PR #152, opened 2018-11-12
Vlan support

EdDev:


tests/integration/nm/vlan_test.py
```diff
@@ -0,0 +1,83 @@
...
+
+def _delete_vlan(devname):
+    nmdev = nm.device.get_device_by_name(devname)
+    with _mainloop():
+        nm.device.deactivate(nmdev)
+        nm.device.delete(nmdev)
+
+
+@contextmanager
+def _mainloop():
```
tyll: Isn't there some other test that calls the mainloop as well? How about having some generic shared code for all the things that are not specific for this test?

tests/integration/nm/vlan_test.py
```diff
@@ -0,0 +1,83 @@
...
+    mloop = nm.nmclient.mainloop()
+    yield
+    success = mloop.run(timeout=5)
+    assert success, 'Mainloop error: ' + mloop.error
+
+
+def _get_vlan_ifname(state):
```
tyll: Does the vlan interface name always consiste of the base interface and the VLAN ID? Why do we need two settings for the same data then?

PR #153, opened 2018-11-12
netapplier: Retry desired vs current states verification

EdDev:
Perform the desired vs current state verification several times before
declaring failure.

This is a temporary measure taken which should be dropped when the
waiting mechanism is put in place.


libnmstate/netapplier.py
```diff
@@ -63,13 +64,32 @@ def _apply_ifaces_state(interfaces_desired_state, verify_change):
...
+            verified_successfully = False
+            for retry in range(4):
+                try:
+                    _verify_change(ifaces_desired_state)
+                except DesiredStateIsNotCurrentError:
+                    time.sleep(1)
+                else:
+                    verified_successfully = True
+                    break
+            if not verified_successfully:
```
tyll: I think this variable is not require and it should work with an `else` for the for loop. But the current way might be easier to read if one is not familiar with the "for... else..." syntax.

tyll: Another option to implenent do..until would be to use:

while True... break. Then there would be only one call of `_verify_change` instead of two. This would make it less likely that only one of them is adjusted in the future.

EdDev: Lets use `else`, it is the python way. Thanks.

PR #154, opened 2018-11-14
packaging: Create packaging Dockerfile and README

EdDev:
Create a Dockerfile for distributing nmstate through containers.
Provide documentation on how to package nmstate and distribute to PyPI
and docker-hub.

### DIFF
--- a/packaging/Dockerfile
+++ b/packaging/Dockerfile
@@ -1,0 +1,39 @@
+# This Dockerfile is based on the recommendations provided in the
+# Centos official repository (https://hub.docker.com/_/centos/).
+# It enables systemd to be operational.
+FROM centos:7.5.1804
+ENV container docker
+RUN (cd /lib/systemd/system/sysinit.target.wants/; for i in *; do [ $i == \
+     systemd-tmpfiles-setup.service ] || rm -f $i; done); \
+    rm -f /lib/systemd/system/multi-user.target.wants/*;\
+    rm -f /etc/systemd/system/*.wants/*;\
+    rm -f /lib/systemd/system/local-fs.target.wants/*; \
+    rm -f /lib/systemd/system/sockets.target.wants/*udev*; \
+    rm -f /lib/systemd/system/sockets.target.wants/*initctl*; \
+    rm -f /lib/systemd/system/basic.target.wants/*;\
+    rm -f /lib/systemd/system/anaconda.target.wants/*;
+
+RUN yum -y install iproute NetworkManager epel-release cairo* && \
+    yum -y install python2-pip gcc python-devel gobject-introspection-devel && \
+    yum clean all && \
+    pip install -U \
+      pip \
+      \
+      pluggy==0.6 \
+      PyGObject && \
+    \
+    install -o root -g root -d /etc/sysconfig/network-scripts && \
+    echo -e "[logging]\nlevel=TRACE\ndomains=ALL\n" > /etc/NetworkManager/conf.d/97-docker-build.conf && \
+    echo -e "[device]\nmatch-device=*\nmanaged=0\n" >> /etc/NetworkManager/conf.d/97-docker-build.conf && \
+    sed -i 's/#RateLimitInterval=30s/RateLimitInterval=0/ ; s/#RateLimitBurst=1000/RateLimitBurst=0/' /etc/systemd/journald.conf
+
+# Patch with NetworkManager fix for BZ#1642625
+RUN yum-config-manager --add-repo https://copr.fedorainfracloud.org/coprs/till/nmstate-NetworkManager/repo/epel-7/till-nmstate-NetworkManager-epel-7.repo && \
+    yum-config-manager --enable till-nmstate-NetworkManager && \
+    yum -y update NetworkManager
+
+RUN pip install nmstate
+
+VOLUME [ "/sys/fs/cgroup" ]
+
+CMD ["/usr/sbin/init"]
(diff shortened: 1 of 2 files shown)


packaging/README.md
```diff
@@ -0,0 +1,33 @@
+# Distribute and Install nmstate
+The packaging and distribution of nmstate is currently implemented through
```
tyll: IMHO the sentence is more complicated that necessary, it could just be:

`NMState is distributed on PyPI and Docker Hub.`

Also it would be great to agree on a capitalization for NMState. The main web page calls it NMState and not nmstate.

packaging/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+First prepare and upload to PyPI and then build and upload the docker image.
+ 
+
+## Package and Upload to PyPI
+The following procedures can be reviewed in detail [here](https://packaging.python.org/tutorials/packaging-projects/#uploading-the-distribution-archives).
+```
+cd <project-path>
+python setup.py sdist bdist_wheel
+twine upload dist/*
```
tyll: What is twine? Maybe explain it and where to get it. Is it the twine package on Fedora?

packaging/README.md
```diff
@@ -0,0 +1,33 @@
...
+```
+
+## Build a new container image and push to the docker hub
+
+- CentOS 7:
+```
+cd <project-path/packaging>
+sudo docker build --rm -t local/centos7-nmstate .
+docker tag local/centos7-nmstate nmstate/centos7-nmstate:<ver>
```
tyll: What is the value of `<ver>` supposed to be? Also IMHO it would be better to write a shell or python script for creating a release instead of documenting the steps in a README.

packaging/Dockerfile
```diff
@@ -0,0 +1,42 @@
+# This Dockerfile is based on the recommendations provided in the
```
tyll: Can we create a common docker image for preparing CentOS7 for distribution and the CI and then just do the extra changes for CI/distribuition in a separate Dockerfile? Or have one Dockerfile create multiple images?

packaging/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+    yum -y install python2-pip gcc python-devel gobject-introspection-devel && \
+    yum clean all && \
+    pip install -U \
+      pip \
+      setuptools \
+      pbr \
+      tox \
+      \
+      pluggy==0.6 \
+      PyGObject && \
```
tyll: Do we need to install this via pip for CentOS? Is there nothing packaged to be used? Also I suppose we do not need pip, setuptools, pbr and tox in the distributed image but only for CI, do we?

EdDev: Right, need to remove it.
Not sure about PyGObject, perhaps pygobject2 is enough. I think I had problems with this.
Will check.

packaging/README.md
EdDev: Done

packaging/README.md
EdDev: Line 9 points the reader to a link which describe this in more detail, I just wanted to paster here the commands used to build and upload the package... just for convenience.

packaging/README.md
EdDev: `<ver>` is the version number.
A script will be nice, but I do not think it is a must at the moment. I am not that focused on the release process at the moment, it will happen eventually as it is the right thing to do.

packaging/Dockerfile
EdDev: So I think we can use this image as the base, the CI can be based on it and add its adjustments.

packaging/Dockerfile
EdDev: I was unable to replace these, leaving it to the rpm packaging effort.

packaging/README.md
tyll: But how do we choose the version number? If I had to follow these instructions, I would not know which number would be correct. The shell script can be just all commands in a single file with `set -e`, so it would not be much extra work but it makes it easier for us to test this change.

tyll: How about we just use lastest for the docker version for now?

tyll: There seems to be a 0.1 tag already, I suppose this should be removed: https://hub.docker.com/r/nmstate/centos7-nmstate/tags/

EdDev: `latest` and `0.1` are the same. This is a way to track what `latest` means.
With NMState we use 0.1bXX where XX is the build num (it could have been a commit hash like we do with other projects).

EdDev: 0.1 means it is the first one that works :)
0.1b1 is the next one.

We need to user of this image to know which `latest` he is using. Latest is just a pointer to the last version.

I expect to release it every time we update the one on PyPI.
Again, I think it belongs to a different PR as it discusses a release flow/procedure. And we do not have a release process yet.

packaging/Dockerfile
cathay4t: The PR https://github.com/nmstate/nmstate/pull/162 contains removed all the `pip` command.

cathay4t:
LGTM. Thanks.

packaging/Dockerfile
```diff
@@ -0,0 +1,39 @@
+# This Dockerfile is based on the recommendations provided in the
+# Centos official repository (https://hub.docker.com/_/centos/).
+# It enables systemd to be operational.
+FROM centos:7
```
tyll: This should now also use the version that is used in integration tests

EdDev: Done

packaging/Dockerfile
EdDev: @cathay4t , lets take it and then start sharing these two images. The dev one should depend on this one (assuming we change this to rpm/s as well).

EdDev:
Got resolved by #265

PR #155, opened 2018-11-14
integ test: Decrease ipv6 iface mtu to lower then min

SvetaHBar:
New test aimed to check that mtu value cannot be decreased lower than
1280

### DIFF
--- a/tests/integration/ethernet_mtu_test.py
+++ b/tests/integration/ethernet_mtu_test.py
@@ -102,3 +102,17 @@ def test_decrease_to_ipv6_min_ethernet_frame_size_iface_mtu():
     netapplier.apply(copy.deepcopy(desired_state))
 
     assertlib.assert_state(desired_state)
+
+
+def test_decrease_to_lower_than_min_ipv6_iface_mtu():
+    original_state = statelib.show_only(('eth1',))
+    desired_state = copy.deepcopy(original_state)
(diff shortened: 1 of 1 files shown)


tests/integration/ethernet_mtu_test.py
```diff
@@ -102,3 +102,17 @@ def test_decrease_to_ipv6_min_ethernet_frame_size_iface_mtu():
     netapplier.apply(copy.deepcopy(desired_state))
 
     assertlib.assert_state(desired_state)
+
+
+def test_decrease_to_lower_than_min_ipv6_iface_mtu():
+    desired_state = statelib.show_only(('eth1',))
+    origin_desired_state = copy.deepcopy(desired_state)
```
tyll: IMHO `original` makes more sense than `origin` here. Also if you are touching this, it would seem to be more logical to me to first assign to `original_state` and then create a `desired_state` with a deep copy from it.

tests/integration/ethernet_mtu_test.py
```diff
@@ -102,3 +102,17 @@ def test_decrease_to_ipv6_min_ethernet_frame_size_iface_mtu():
...
+    eth1_desired_state = desired_state[INTERFACES][0]
+    eth1_desired_state['mtu'] = 68
```
tyll: The MTU value should be motivated. 1279 would be enough to be lower than the IPv6 minimum mtu.

PR #156, opened 2018-11-18
nm: Activate new virtual interfaces

EdDev:
NMState supports several virtual interfaces, all having a master-slave
relationship (bond, ovs-bridge). The master interface has been activated
so far by its slaves (which automatically perform such an activation
when they are activated).
With the introduction of VLAN interfaces, it is now required for the
virtual interface to be explicitly activated.

Activating a new interface is now supported by executing an activation
command on the (remote) connection profile. Existing interfaces are
still activated using their device object.

### DIFF
--- a/libnmstate/nm/applier.py
+++ b/libnmstate/nm/applier.py
@@ -85,21 +85,68 @@ def prepare_edited_ifaces_configuration(ifaces_desired_state):
     return con_profiles
 
 
-def set_ifaces_admin_state(ifaces_desired_state):
+def set_ifaces_admin_state(ifaces_desired_state, con_profiles=()):
+    """
+    Control interface admin state by activating, deactivating and deleting
+    devices connection profiles.
+
+    The `absent` state results in deactivating the device and deleting
+    the connection profile.
+    FIXME: The `down` state is currently handled in the same way.
+
+    For new virtual devices, the `up` state is handled by activating the
+    new connection profile. For existing devices, the device is activated,
+    leaving it to choose the correct profile.
+
+    In order to activate correctly the interfaces, the order is significant:
+    - New interfaces (virtual interfaces).
+    - Master interfaces.
+    - All the rest.
+    """
+    new_ifaces = _get_new_ifaces(con_profiles)
+    new_ifaces_to_activate = set()
+    master_ifaces_to_activate = set()
     devs_actions = {}
+
     for iface_desired_state in ifaces_desired_state:
-        devs = _get_affected_devices(iface_desired_state)
-        if iface_desired_state['state'] == 'up':
-            devs_actions.update({dev: (device.activate,) for dev in devs})
-        elif iface_desired_state['state'] in ('down', 'absent'):
-            devs_actions.update(
-                {dev: (device.deactivate, device.delete) for dev in devs})
+        nmdev = device.get_device_by_name(iface_desired_state['name'])
+        if not nmdev:
+            ifname = iface_desired_state['name']
+            if ifname in new_ifaces and iface_desired_state['state'] == 'up':
+                    new_ifaces_to_activate.add(ifname)
         else:
-            raise UnsupportedIfaceStateError(iface_desired_state)
+            if iface_desired_state['state'] == 'up':
+                master_iface_types = ovs.BRIDGE_TYPE, bond.BOND_TYPE
+                if iface_desired_state['type'] in master_iface_types:
+                    master_ifaces_to_activate.add(nmdev)
+                else:
+                    devs_actions[nmdev] = (device.activate,)
+            elif iface_desired_state['state'] in ('down', 'absent'):
+                nmdevs = _get_affected_devices(iface_desired_state)
+                for nmdev in nmdevs:
+                    devs_actions[nmdev] = (device.deactivate, device.delete)
+            else:
+                raise UnsupportedIfaceStateError(iface_desired_state)
+
+    for ifname in new_ifaces_to_activate:
+        device.activate(dev=None, connection_id=ifname)
+
+    for dev in master_ifaces_to_activate:
+        device.activate(dev)
 
     for dev, actions in six.viewitems(devs_actions):
         for action in actions:
             action(dev)
+
+
+def _get_new_ifaces(con_profiles):
+    ifaces_without_device = set()
+    for con_profile in con_profiles:
+        ifname = con_profile.get_interface_name()
+        nmdev = device.get_device_by_name(ifname)
+        if not nmdev:
+            ifaces_without_device.add(ifname)
+    return ifaces_without_device
 
 
 def _get_affected_devices(iface_state):
--- a/libnmstate/nm/device.py
+++ b/libnmstate/nm/device.py
@@ -20,13 +20,22 @@ import logging
 from . import nmclient
 
 
-def activate(dev, connection=None):
+def activate(dev=None, connection_id=None):
+    """Activate the given device or remote connection profile."""
+    mainloop = nmclient.mainloop()
+    mainloop.push_action(
+        _safe_activate_async, dev, connection_id)
+
+
+def _safe_activate_async(dev, connection_id):
     client = nmclient.client()
     mainloop = nmclient.mainloop()
+    connection = None
+    if connection_id:
+        connection = client.get_connection_by_id(connection_id)
     specific_object = None
     user_data = mainloop, dev
-    mainloop.push_action(
-        client.activate_connection_async,
+    client.activate_connection_async(
         connection,
         dev,
         specific_object,
--- a/tests/integration/nm/vlan_test.py
+++ b/tests/integration/nm/vlan_test.py
@@ -67,9 +67,7 @@ def _create_vlan(vlan_desired_state):
         con_profile = nm.connection.create_profile(
             (con_setting, vlan_setting, ipv4_setting, ipv6_setting))
         nm.connection.add_profile(con_profile, save_to_disk=False)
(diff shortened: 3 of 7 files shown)


libnmstate/nm/applier.py
```diff
@@ -85,23 +86,73 @@ def prepare_edited_ifaces_configuration(ifaces_desired_state):
...
+    """
+    Control interface admin state by activating, deactivating and deleting
+    devices connection profiles.
+
+    The `down` and `absent` states are handled in the same manner (temporary)
+    by deactivating the device and deleting the connection profile.
+
+    For new virtual devices, the `up` state is handled by activating the
+    new connection profile. For existing devices, the device is activated,
+    leaving it to choose the correct profile.
```
tyll: What is the advantage of activating the device instead of always activating the profile that we want?

libnmstate/nm/applier.py
```diff
@@ -85,23 +86,73 @@ def prepare_edited_ifaces_configuration(ifaces_desired_state):
     return con_profiles
 
 
-def set_ifaces_admin_state(ifaces_desired_state):
+def set_ifaces_admin_state(ifaces_desired_state, con_profiles=()):
+    """
+    Control interface admin state by activating, deactivating and deleting
+    devices connection profiles.
+
+    The `down` and `absent` states are handled in the same manner (temporary)
```
tyll: How about:
```
The `absent` state results in deactivating the device and deleting the connection profile.
FIXME: The `down` state is currently handled in the same way.
```

This makes it more clear that the down state may change in the future.

libnmstate/nm/applier.py
```diff
@@ -85,23 +86,73 @@ def prepare_edited_ifaces_configuration(ifaces_desired_state):
...
+    by deactivating the device and deleting the connection profile.
+
+    For new virtual devices, the `up` state is handled by activating the
+    new connection profile. For existing devices, the device is activated,
+    leaving it to choose the correct profile.
+
+    In order to activate correctly the interfaces, the order is significant:
+    - New interfaces (virtual interfaces).
```
tyll: Why does it need to be this order? What happens when we want to use a VLAN on top of a bond interface, doesn't the bond interface need to be activated first?

libnmstate/nm/device.py
```diff
@@ -20,13 +20,28 @@
 from . import nmclient
 
 
-def activate(dev, connection=None):
+def activate(dev, fetch_remote_connection_func=None):
+    """
+    Activating the given device or remote connection profile.
```
tyll: minor detail: Better start the description in imperative mood, i.e. `Activate the given device... `

libnmstate/nm/device.py
```diff
@@ -20,13 +20,28 @@
 from . import nmclient
 
 
-def activate(dev, connection=None):
+def activate(dev, fetch_remote_connection_func=None):
```
tyll: Shouldn't the dev parameter be optional?

tests/integration/nm/vlan_test.py
```diff
@@ -35,7 +35,7 @@ def test_create_and_remove_vlan():
         assert vlan_desired_state == vlan_current_state
 
     # FIXME: Remove this sleep after adding wait for state mechanism.
-    time.sleep(2)
+    time.sleep(5)
```
tyll: This should probably be a separate PR/commit, since it is unrelated/fixes problems for other PRs as well.

tests/integration/nm/vlan_test.py
```diff
@@ -67,9 +67,10 @@ def _create_vlan(vlan_desired_state):
     with mainloop():
         con_profile = nm.connection.create_profile((con_setting, vlan_setting))
         nm.connection.add_profile(con_profile, save_to_disk=False)
-    with mainloop():
-        remote_profile = nm.nmclient.client().get_connection_by_id(ifname)
-        nm.device.activate(dev=None, connection=remote_profile)
+        nm.device.activate(
+            dev=None,
+            fetch_remote_connection_func=(
+                lambda: nm.nmclient.client().get_connection_by_id(ifname)))
```
tyll: It seems that `activate()` could be easily be simplified by just accepting a connection id instead of the fetch function.

tests/lib/nm/device_test.py
```diff
@@ -41,10 +41,11 @@ def mainloop_mock():
 
 def test_activate(client_mock, mainloop_mock):
     dev = 'foodev'
-    nm.device.activate(dev)
+    mainloop_mock.push_action = lambda func, dev, fetch: func(dev, fetch)
 
-    mainloop_mock.push_action.assert_called_once_with(
-        client_mock.activate_connection_async,
+    nm.device.activate(dev, fetch_remote_connection_func=None)
```
tyll: The fetch function is by default None, why specify it explicitly here? It seems that there is a test missing in case it is not None.

libnmstate/nm/applier.py
```diff
@@ -85,23 +86,73 @@ def prepare_edited_ifaces_configuration(ifaces_desired_state):
...
+                for nmdev in nmdevs:
+                    devs_actions[nmdev] = (device.deactivate, device.delete)
+            else:
+                raise UnsupportedIfaceStateError(iface_desired_state)
+
+    for ifname in new_ifaces_to_activate:
+        device.activate(
+            dev=None,
+            fetch_remote_connection_func=(
+                lambda: nmclient.client().get_connection_by_id(ifname))
```
tyll: It seems that activate() could be easily be simplified by just accepting a connection id instead of the fetch function.

libnmstate/nm/applier.py
EdDev: Good point, we need new VLAN/s last in the 'new' interfaces section.
It's actually even more complex with bridges and friends.

The order matters only for masters before their slaves. Tests failed when activating the slave before the master, as the master explicit activation collided with the slaves triggering the same implicitly.
This may become a problem with the introduction of other device types and more scenarios.

libnmstate/nm/applier.py
EdDev: The function is triggered only later by the mainloop, there is no connection ID at this time, it has not been created yet (the connection is created in the mainloop as well).

libnmstate/nm/device.py
EdDev: It can be. I'll change it.

libnmstate/nm/device.py
EdDev: Done

tests/integration/nm/vlan_test.py
EdDev: Right

tests/integration/nm/vlan_test.py
EdDev: see previous answer to this

tests/lib/nm/device_test.py
EdDev: Done

EdDev: Regarding the missing test: It is covered by the nm.vlan test in the integration suite.
I will start taking out these unit nm tests in future patches.

EdDev:
ci test please

EdDev:
ci test please

EdDev:
ci test please

libnmstate/nm/applier.py
EdDev: > his might true for now, but there is no document of NM confirmed this.

It is document in our code:
https://github.com/nmstate/nmstate/blob/master/libnmstate/nm/connection.py#L105

> How about use nmclient.activate_connection_async where you can define device without connection/profile ID?

I think nmstate is supposed to be specific when controlling the interfaces. I am not clear how we could do that without a unique identifier (we have today the device and the profile ID).
But there may be a need to add other identifiers in the future, if and when that  happens we could add it like we added this one.

libnmstate/nm/applier.py
tyll: It does not need to happen in this PR, but could we instead just check the errors in case of collisions and ignore errors if it just says that we want to activate an already activated connection?

libnmstate/nm/applier.py
tyll: I can agree with the first point. The second point seems to be wrong because the best profile will always be the one that nmstate created - otherwise we will not reach the defined state in nmstate. There is no clear argument for the third one since it seems to be more of a gut feeling.

libnmstate/nm/applier.py
EdDev: Ahh, you are absolutely right. I was blind.
Fixed! Thank you!

libnmstate/nm/device.py
```diff
@@ -20,13 +20,28 @@
 from . import nmclient
 
 
-def activate(dev, connection=None):
+def activate(dev=None, connection_id=None):
+    """
+    Activate the given device or remote connection profile.
+    For the remote connection profile, a fetcher function is provided which
```
tyll: The docstring needs to be adjusted for the change from fetch function to connection_id

EdDev: Done

EdDev:
ci test please

libnmstate/nm/device.py
```diff
@@ -20,13 +20,25 @@
 from . import nmclient
 
 
-def activate(dev, connection=None):
+def activate(dev=None, connection_id=None):
+    """
+    Activate the given device or remote connection profile.
+    For the remote connection profile, the connection ID is provided.
```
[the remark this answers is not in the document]
EdDev: Fair enough, removing.

PR #160, opened 2018-11-19
Vlan: Support the VLAN interface type in libnmstate

EdDev:
Depends on #156

PR #165, opened 2018-11-21
Nm wait for con activation

EdDev:
Based on #160 , #156

### DIFF
--- a/libnmstate/nm/device.py
+++ b/libnmstate/nm/device.py
@@ -20,6 +20,94 @@ import logging
 from . import nmclient
 
 
+class ActivationError(Exception):
+    pass
+
+
+class AlternativeACState(object):
+    UNKNOWN = 0
+    ACTIVE = 1
+    ACTIVATING = 2
+    FAIL = 3
+
+
+class ActiveConnection(object):
+
+    def __init__(self, active_connection):
+        self._act_con = active_connection
+        nmdevs = active_connection.get_devices()
+        self._nmdev = nmdevs[0] if nmdevs else None
+
+        self.handlers = set()
+
+        self.refresh_state()
+
+    def refresh_state(self):
+        self._state = self._act_con.get_state()
+        self._state_reason = self._act_con.get_state_reason()
+        self._alternative_state = AlternativeACState.UNKNOWN
+
+        nm_acs = nmclient.NM.ActiveConnectionState
+        if self._state == nm_acs.DEACTIVATED:
+            unable_to_activate = (
+                    not self._nmdev or
+                    (
+                            self._state_reason is not None and
+                            self._state_reason != nm_acs.DEVICE_DISCONNECTED
+                    ) or
+                    self._nmdev.get_active_connection() is not self._act_con
+            )
+            if unable_to_activate:
+                self._alternative_state = AlternativeACState.FAIL
+            # Use the device-state as an alternative to determine if active.
+            elif (self._nmdev <= nmclient.NM.DeviceState.DISCONNECTED or
+                    self._nmdev > nmclient.NM.DeviceState.DEACTIVATING):
+                self._alternative_state = AlternativeACState.FAIL
+
+    @property
+    def is_active(self):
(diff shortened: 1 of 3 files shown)


EdDev:
ci test please

EdDev:
ci test please

libnmstate/nm/device.py
```diff
@@ -20,6 +20,96 @@
 from . import nmclient
 
 
+class ActivationError(Exception):
+    pass
+
+
+class AlternativeACState(object):
```
tyll: Why is it called Alternative?

libnmstate/nm/device.py
```diff
@@ -20,6 +20,96 @@
...
+        nmdevs = active_connection.get_devices()
+        self._nmdev = nmdevs[0] if nmdevs else None
+
+        self.handlers = set()
+
+    def refresh_state(self):
+        self._state = self._act_con.get_state()
+        self._state_reason = self._act_con.get_state_reason()
+        self._alternative_state = (
+            nmclient.NM.ActiveConnectionStateReason.UNKNOWN)
```
tyll: Should this be AlternativeACState.UNKNOWN?

libnmstate/nm/device.py
```diff
@@ -20,6 +20,96 @@
...
+            )
+            if unable_to_activate:
+                self._alternative_state = AlternativeACState.FAIL
+            # Use the device-state as an alternative to determine if active.
+            elif (self._nmdev <= nmclient.NM.DeviceState.DISCONNECTED or
+                    self._nmdev > nmclient.NM.DeviceState.DEACTIVATING):
+                self._alternative_state = AlternativeACState.FAIL
+
+    @property
+    def is_active(self):
```
tyll: The `is_` prefix does not make sense for properties IMHO. Being a property implies that it is somthing that the class is.

libnmstate/nm/device.py
```diff
@@ -20,6 +20,96 @@
...
+                return (
+                    _is_device_master_type(self._nmdev) and
+                    nmclient.NM.DeviceState.IP_CONFIG <= nmdev_state <=
+                    nmclient.NM.DeviceState.ACTIVATED
+                )
+
+        return False
+
+    @property
+    def is_activating(self):
```
tyll: See the comment about the `is_` prefix above.

libnmstate/nm/device.py
```diff
@@ -60,14 +150,47 @@ def _active_connection_callback(src_object, result, user_data):
...
         mainloop.quit('Connection activation failed on %s: error=unknown' %
                       nmdev.get_iface())
     else:
-        devname = act_con.props.connection.get_interface_name()
+        devname = nm_act_con.props.connection.get_interface_name()
+        logging.debug('Connection activation initiated: dev=%s, con-state=%s',
+                      devname, nm_act_con.props.state)
+
+        ac = ActiveConnection(nm_act_con)
+        ac.refresh_state()
```
tyll: IMHO it makes sense to refresh the state in the constructor or initialize with the correct state from the beginning.

EdDev: Yes it can be refresh inside the constructor.
But as the object represents a state of several underlying objects (of NM), and those states are only true at the time of "refresh", the object state is never "exact". I think it is clearer for such objects to get constructed with an init state and only at later stages (before use) to be refreshed explicitly.

If you have strong feeling on this, I can change it. Please let me know.

libnmstate/nm/device.py
tyll: My suggestion is to rename it to `active`.

libnmstate/nm/device.py
tyll: Why does it need to be manually refreshed at all? How about refreshing the state whenever one of the properties is accessed?

EdDev: I just tried to change it. Looks like I need to call refresh from 4 places now: `is_active`, `is_activating`, `reason` and `state`. (for the last I could use a direct call to self._act_con.get_state(), but not for the other 3)

PR #166, opened 2018-11-22
tests integ: Fix NM VLAN test with IPv4/6 disabled

EdDev:


EdDev:
I agree, that is better.I will be able to change it only later (3h).

EdDev:
@tyll , come to think of it, I am not sure this solves the problem we see on CI.
It solves the problem of waiting for the connection to reach active state, but that is probably only relevant for #165 patch.

EdDev:
Hope the delay time increase will solve it now.

PR #169, opened 2018-11-23
NM: Refresh NM client before querying devices

tyll:
This create a segmentation fault locally but might be a good thing to do to fix problems with the vlan test, since the vlan device seems to be cached by the client even when it is already gone and cannot be retrieved in a different nmclient instance. I guess ideally the mainloop would run for longer to take care of this.

### DIFF
--- a/libnmstate/nm/device.py
+++ b/libnmstate/nm/device.py
@@ -163,7 +163,7 @@ def _delete_connection_callback(src_object, result, user_data):
 
 
 def get_device_by_name(devname):
-    client = nmclient.client()
+    client = nmclient.client(refresh=True)
     return client.get_device_by_iface(devname)
 
 


EdDev:
I just realized this myself with @thaller help this morning/noon.
But the refresh should not be done here... as this is in many times in the mainloop (we may call it from apply).
See #170 .

tyll:
@thom311 this test get's libnm(?) to crash. Not sure if you like to debug it.

thom311:
@tyll Yes, interested :) Can you help getting more information? Backtrace would be best. Or maybe run it under valgrind?

Als `G_DEBUG=fatal-warnings` is helpful. Actually, I recommend that you set that environment variable in all your tests, so that libnm/glib dumps core on all warnings. These kinds of warnings always indicate a bug somewhere.

tyll:
Sent logs and core created with `G_DEBUG=fatal-warnings,gc-friendly G_SLICE=always-malloc valgrind --num-callers=100 --log-file=/tmp/log pytest -vkk vlan` to @thom311 

OPnened a new PR to set `G_DEBUG=fatal-warnings` in integration tests (#181).

PR #170, opened 2018-11-23
tests integ: Fix nm VLAN tests by refreshing nm client

EdDev:
Remove the sleep introduced in previous patched and add an NM client
refresh.
While outside the mainloop, the client needs to be re-created in order
to retrieve updated information form NM.

### DIFF
--- a/tests/integration/nm/vlan_test.py
+++ b/tests/integration/nm/vlan_test.py
@@ -16,7 +16,6 @@
 #
 
 from contextlib import contextmanager
-import time
 
 from libnmstate import nm
 
@@ -34,9 +33,6 @@ def test_create_and_remove_vlan():
         vlan_current_state = _get_vlan_current_state(vlan_desired_state)
         assert vlan_desired_state == vlan_current_state
 
-    # FIXME: Remove this sleep after adding wait for state mechanism.
-    time.sleep(5)
-
     assert not _get_vlan_current_state(vlan_desired_state)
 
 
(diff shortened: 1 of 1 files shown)


tests/integration/nm/vlan_test.py
```diff
@@ -50,6 +46,7 @@ def _vlan_interface(state):
 
 
 def _get_vlan_current_state(vlan_desired_state):
+    nm.nmclient.client(refresh=True)
```
tyll: IMHO it would be better to do this in the get_device_by_name() function but since it segfaults there, this is an acceptable workaround. Please add a FIXME, though, so we remove this eventually.

EdDev: It cannot be done in the production NM package code because there we are already operating inside the mainloop (for apply at least). I guess that is the cause of the segfaults.

Regarding the FIXME, I do not think we can "fix" it. In the NM integration tests we run the low level NM api, and create the mainloop ourself in the test. This last assert checks if the interface has been removed (those action have been executed in the mainloop), and it does not need a mainloop, just a client refresh.
Bottom line, there is no fix here unless you can think of something else we can do to make this nicer.

EdDev:
Now it fails on something else :)

EdDev:
ci test please

tests/integration/nm/vlan_test.py
tyll: Keeping the mainloop running would ensure that the client will be refreshed AFAIU, so it would be a fix... But I see that it would be difficult to do this in the test code as well.

PR #174, opened 2018-11-28
integ tests: Added "eth1_up" argument to the vlan test

SvetaHBar:


EdDev:
> integ tests: Added "eth1_up" argument to the vlan test

How about: Add eth1_up fixture as a requirement to the vlan test

PR #175, opened 2018-11-28
linux-bridge type interface

EdDev:


libnmstate/nm/bridge.py
```diff
@@ -0,0 +1,131 @@
...
+    bridge_setting = None
+    if base_con_profile:
+        bridge_setting = base_con_profile.get_setting_bridge()
+        if bridge_setting:
+            bridge_setting = bridge_setting.duplicate()
+
+    if not bridge_setting:
+        bridge_setting = nmclient.NM.SettingBridge.new()
+
+    bridge_setting.props.ageing_time = bridge['mac-ageing-time']
```
tyll: Does it make sense to use constants here for these values?

tests/integration/nm/linux_bridge_test.py
```diff
@@ -0,0 +1,133 @@
...
+
+from .testlib import mainloop
+
+
+BRIDGE0 = 'brtest0'
+ETH1 = 'eth1'
+
+
+@pytest.fixture
+def bridge_default_config():
```
tyll: It would be nice to also have a simple test with only the minimal state necessary for a bridge.

EdDev: Sure, will add one.

tests/integration/nm/linux_bridge_test.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+BRIDGE0 = 'brtest0'
+ETH1 = 'eth1'
+
+
+@pytest.fixture
+def bridge_minimum_config():
+    return {
+        LB.CONFIG_SUBTREE: {
```
tyll: Is it good to use these constants here in the test? The advantage is that it ensures that the code also uses the values that are defined in the constant, the disadvantage is that the test does not test whether the values in the constants are correct.

libnmstate/nm/bridge.py
```diff
@@ -0,0 +1,143 @@
...
+        bridge_setting = nmclient.NM.SettingBridge.new()
+
+    for key, val in six.viewitems(bridge):
+        if key == LB.MAC_AGEING_TIME:
+            bridge_setting.props.ageing_time = val
+        elif key == LB.GROUP_FORWARD_MASK:
+            bridge_setting.props.group_forward_mask = val
+        elif key == LB.MULTICAST_SNOOPING:
+            bridge_setting.props.multicast_snooping = val
+        elif key == LB.STP_SUBTREE:
```
tyll: It seems that this subtree handling could belong in its own helper method.

EdDev: Thank you for reminding my own rule: Make the functions as small as possible.

PR #177, opened 2018-11-29
integ tests: Add and remove two vlans on same inface

SvetaHBar:


tests/integration/vlan_test.py
```diff
@@ -35,6 +36,18 @@ def test_add_and_remove_vlan(eth1_up):
     assert not current_state[INTERFACES]
 
 
+def test_add_and_remove_two_vlans_on_inface(eth1_up):
+    with two_vlans_on_same_interface() as desired_state:
+        print(desired_state)
```
tyll: The print calls should be removed

tests/integration/vlan_test.py
```diff
@@ -35,6 +36,18 @@ def test_add_and_remove_vlan(eth1_up):
...
+        assertlib.assert_state(desired_state)
+
+    vlan_interfaces = [i['name'] for i in desired_state[INTERFACES]]
+    print(vlan_interfaces)
```
tyll: this, too

tests/integration/vlan_test.py
```diff
@@ -35,6 +36,18 @@ def test_add_and_remove_vlan(eth1_up):
...
+    current_state = statelib.show_only(vlan_interfaces)
+    print(current_state)
```
tyll: and this

tests/integration/vlan_test.py
```diff
@@ -25,6 +25,7 @@
 
 
 VLAN_IFNAME = 'eth1.101'
+VLAN1_IFNAME = 'eth1.102'
```
tyll: IMHO it would be more natural to call the second vlan interface VLAN2_IFNAME.

tests/integration/vlan_test.py
```diff
@@ -64,3 +77,50 @@ def vlan_interface(ifname, vlan_id):
                 ]
             }
         )
+
+
+@contextmanager
+def two_vlans_on_same_interface():
```
tyll: Since is is about two vlans on eth1, it makes more sense to call it `two_vlans_on_eth1`.

PR #180, opened 2018-11-30
Integration tests: Test downing a vlan interface

tyll:

### DIFF
--- a/tests/integration/testlib/assertlib.py
+++ b/tests/integration/testlib/assertlib.py
@@ -18,6 +18,7 @@
 from libnmstate import netinfo
 
 from . import statelib
+from .statelib import INTERFACES
 
 
 def assert_state(desired_state_data):
@@ -34,3 +35,10 @@ def assert_state(desired_state_data):
     full_desired_state.normalize()
 
     assert full_desired_state.state == current_state.state
+
+
+def assert_absent(*ifnames):
+    """ Assert that a interface is not present in the current state """
+
+    current_state = statelib.show_only(ifnames)
+    assert not current_state[INTERFACES]
--- a/tests/integration/vlan_test.py
+++ b/tests/integration/vlan_test.py
@@ -16,6 +16,9 @@
 #
 
 from contextlib import contextmanager
+import os.path
+
+import yaml
 
 from libnmstate import netapplier
 
@@ -26,6 +29,8 @@ from .testlib.statelib import INTERFACES
 
 VLAN_IFNAME = 'eth1.101'
 VLAN2_IFNAME = 'eth1.102'
+
+PATH_MAX = 4096
 
 
 def test_add_and_remove_vlan(eth1_up):
@@ -60,6 +65,59 @@ def test_set_vlan_iface_down(eth1_up):
 
         current_state = statelib.show_only((VLAN_IFNAME,))
         assert not current_state[INTERFACES]
+
+
+def test_add_down_remove_vlan(eth1_up):
+    with yaml_state('vlan101_eth1_up.yml',
+                    cleanup='vlan101_eth1_absent.yml') as desired_state:
+        assertlib.assert_state(desired_state)
+        with yaml_state('vlan101_eth1_down.yml') as desired_state:
+            assertlib.assert_absent(VLAN_IFNAME)
+
+    assertlib.assert_absent(VLAN_IFNAME)
+
+
+@contextmanager
+def yaml_state(initial, cleanup=None):
+    desired_state = load_example(initial)
+
+    netapplier.apply(desired_state)
+    try:
+        yield desired_state
+    finally:
+        if cleanup:
+            netapplier.apply(load_example(cleanup))
+
+
+def load_example(name):
+    examples = find_examples_dir()
+
+    with open(os.path.join(examples, name)) as yamlfile:
+        state = yaml.load(yamlfile)
+
+    return state
+
+
+def find_examples_dir():
+    path = ''
+    parent = '../'
+    rootdir = '/'
+    examples = None
+    for _ in range(PATH_MAX / len('x/')):
+        maybe_examples = os.path.abspath(os.path.join(path, 'examples'))
+        if os.path.isdir(maybe_examples):
+            examples = maybe_examples
+            break
+
+        if os.path.abspath(path) == rootdir:
+            break
+
+        path = parent + path
+
+    if examples:
+        return examples
+    else:
+        raise RuntimeError('Cannot find examples directory')
 
 
 @contextmanager
--- a/examples/vlan101_eth1_absent.yml
+++ b/examples/vlan101_eth1_absent.yml
@@ -1,0 +1,8 @@
+---
+interfaces:
+  - name: eth1.101
+    type: vlan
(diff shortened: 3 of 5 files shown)


tyll:
@EdDev here you can see the error when downing an interface

tests/integration/vlan_test.py
```diff
@@ -35,6 +37,47 @@ def test_add_and_remove_vlan(eth1_up):
     assert not current_state[INTERFACES]
 
 
+def test_add_down_remove_vlan(eth1_up):
+    up = yaml.load("""
```
EdDev: Can you please move these `""...""` to be constants? (not the conversion to a dict)

tests/integration/vlan_test.py
```diff
@@ -35,6 +37,47 @@ def test_add_and_remove_vlan(eth1_up):
...
+interfaces:
+- name: eth1.101
+  type: vlan
+  state: absent
+  vlan:
+    base-iface: eth1
+    id: 101
+""")
+
+    netapplier.apply(up)
```
EdDev: Please use a contextmanager in order to properly cleanup even on failure (of the apply or assert)

EdDev:
@tyll , I would be interested in understanding what is restored by the checkpoint when the `down` apply fails. It is supposed to restore the vlan device and profile. Is it indeed the case?

tests/integration/vlan_test.py
tyll: changed

tests/integration/vlan_test.py
tyll: changed it, this will hopefully show that absent fails to clean up ;-)

tyll:
> @tyll , I would be interested in understanding what is restored by the checkpoint when the `down` apply fails. It is supposed to restore the vlan device and profile. Is it indeed the case?

I believe it fails to restore the relationship between the device and the profile, i.e. there is no record anymore that the profile created the device because after down failed, absent will only remove the profile but not the device. Need to show  this to @thom311 for confirmation.

tyll:
@EdDev seems like your changes fixed this test case

tests/integration/testlib/assertlib.py
```diff
@@ -34,3 +35,10 @@ def assert_state(desired_state_data):
     full_desired_state.normalize()
 
     assert full_desired_state.state == current_state.state
+
+
+def assert_absent(ifname):
```
EdDev: Can you use a list of ifaces as input? `*ifnames` could work nicely here.

tyll:
> Aren't the tests duplicating test_set_vlan_iface_down now?

yes, it seems so. Nevertheless, I find the yaml approach useful since it allows to easily copy the test data for manual testing. Should I remove the `test_set_vlan_iface_down` test?

tests/integration/testlib/assertlib.py
tyll: fixed this

tyll:
I also made the cleanup argument in `yaml_state()` optional to simplify the test.

tyll:
The examples are now separated

tests/integration/vlan_test.py
```diff
@@ -62,6 +65,53 @@ def test_set_vlan_iface_down(eth1_up):
         assert not current_state[INTERFACES]
 
 
+def test_add_down_remove_vlan(eth1_up):
```
EdDev: How about `test_set_vlan_state_down` ?

tests/integration/vlan_test.py
```diff
@@ -62,6 +65,53 @@ def test_set_vlan_iface_down(eth1_up):
...
+                    cleanup='vlan101_eth1_absent.yml') as desired_state:
+        assertlib.assert_state(desired_state)
+        with yaml_state('vlan101_eth1_down.yml') as desired_state:
+            assertlib.assert_absent(VLAN_IFNAME)
+
+    assertlib.assert_absent(VLAN_IFNAME)
+
+
+@contextmanager
+def yaml_state(initial, cleanup=None):
```
EdDev: I would prefer only the `load_example` to know about the input format (yaml), all the rest can be used with regular dicts.

tests/integration/vlan_test.py
```diff
@@ -62,6 +65,53 @@ def test_set_vlan_iface_down(eth1_up):
...
+    with open(os.path.join(examples, name)) as yamlfile:
+        state = yaml.load(yamlfile)
+
+    return state
+
+
+def find_examples_dir():
+    path = ''
+    examples = None
+    for _ in range(1466):
```
EdDev: Can you use a constant and name it so I'll understand what is this number all about?

tyll: It is 4096/3 + 1 (4096 is PATH_MAX here) and 3 is the length of `../`. It is an upper bound to avoid `while True`. Do you have a suggestion for the name of the constant?

tests/integration/vlan_test.py
tyll: The original problem was that it was not possible to remove the vlan after trying to down it, so the whole cycle was important. Since there is no checkpoint rollback anymore, this is not tested and also it does not fail reliable anymore when testing it locally.

tests/integration/vlan_test.py
tyll: We can split out a generic method when we need it but I would like to also have a yaml one to avoid having to load all the examples explicitly in the tests.

tests/integration/vlan_test.py
EdDev: I lost you here, sorry. I though we are checking that the `down` is working.
You want to check that `absent` is working after `down`?

tests/integration/vlan_test.py
EdDev: IMO the logic of the contextmanager should not be mixed with the format the the config resides in.
You can use fixtures for the config reading if you do not want to explicitly read the config from within the test. But I see no problem in explicitly telling in the test from where you get the config/state.

tests/integration/vlan_test.py
EdDev: `PATH_MAX/len('../') + 1` should do the job.

tests/integration/vlan_test.py
tyll: Not sure if we are discussing the same thing.
In the tests I am fine with specifying the file name of the state files but I am not fine with loading the files there as well, since this is something that the context manager can do for me.

About mixing the format: If the context manager is for dictionaries, then the logic also mixes the requirement of getting a dictionary instead of getting a filename with the remaining logic.

If you need this context manager also with pure dictionary support, we can add it and build one for yaml on top of it. But I would prefer to do this in a seperate step when we need it.

tests/integration/vlan_test.py
tyll: Maybe it is also both... It has been too long ago since I developed the test but at that time there was also an issue with not being able to remove a vlan after down failed. I did not yet manage to replicate this yesterday (at least not reliable). Nevertheless, the test is for the full cycle of adding, downing and removing the vlan, therefore the name suits IMHO.

tyll:
ci test please

Issue #188, opened 2018-12-06
Failed to disable ipv4 on eth0


phoracek:
I was trying to drop ipv4 from eth0 and attach it to OVS bridge, but it failed with `ipv4.gateway: gateway cannot be set if there are no addresses configured`.

config.yaml:

```yaml
interfaces:
- name: eth0
  type: ethernet
  state: up
  ipv4:
    enabled: false
```

`nmstatectl set config.yaml` output:

```
2018-12-06 21:57:38,531 root         DEBUG    Checkpoint /org/freedesktop/NetworkManager/Checkpoint/23 created for all devices: 60
2018-12-06 21:57:38,565 root         WARNING  IPv6 link local address fe80::5055:ff:fed1:5501/64 is ignored when applying desired state
2018-12-06 21:57:38,565 root         DEBUG    Connection settings for duplicate_settings:
id: eth0
iface: eth0
uuid: 4513218f-4a9e-4843-9c27-18138bcdc8f1
type: 802-3-ethernet
autoconnect: False
autoconnect_slaves: <enum NM_SETTING_CONNECTION_AUTOCONNECT_SLAVES_DEFAULT of type NM.SettingConnectionAutoconnectSlaves>
2018-12-06 21:57:38,567 root         DEBUG    Executing NM action: func=commit_changes_async, args=(True, <Gio.Cancellable object at 0x7f784af35730 (GCancellable at 0x7f783c0623a0)>, <function _commit_changes_callback at 0x7f784cb85ed8>, (<libnmstate.nm.nmclient._MainLoop object at 0x7f784af1c810>, <NM.DeviceEthernet object at 0x7f784af351e0 (NMDeviceEthernet at 0x1638490)>))
2018-12-06 21:57:38,572 root         ERROR    NM main-loop aborted: Connection update failed: error=nm-connection-error-quark: ipv4.gateway: gateway cannot be set if there are no addresses configured (7), dev=eth0/<enum NM_DEVICE_STATE_ACTIVATED of type NM.DeviceState>
[...]

PR #190, opened 2018-12-06
disable ipv4.gateway if enabled=False

phoracek:
Should solve https://github.com/nmstate/nmstate/issues/188


cathay4t:
@phoracek IMHO we should clear all configurations when `enabled=False` in stead of duplicating existing configure and remove gateway value.

cathay4t:
Thanks for the PR, I have create another PR to fix this issue.

tyll:
The other PR is #197

PR #193, opened 2018-12-17
netapplier: Ignore `down` state (for virt iface) in verification step

EdDev:
Currently, when in the desired state a virtual interface is set as
`down`, it is removed (treated as `absent`).
When the verification step occurs, checking that the desired state is
included in the current state, these virtual interfaces should be
treated like the `absent` ones.

### DIFF
--- a/libnmstate/netapplier.py
+++ b/libnmstate/netapplier.py
@@ -69,14 +69,14 @@ def _apply_ifaces_state(interfaces_desired_state, verify_change):
 
 
 def _verify_change(ifaces_desired_state):
-    # FIXME: Remove this sleep after adding wait for state mechanism.
+    ifaces_current_state = _index_by_name(netinfo.interfaces())
+    ifaces_desired_state = _remove_absent_iface_entries(
+        ifaces_desired_state)
+    ifaces_desired_state = _remove_down_virt_iface_entries(
+        ifaces_desired_state)
+    ifaces_desired_state = remove_ifaces_metadata(ifaces_desired_state)
 
-        ifaces_current_state = _index_by_name(netinfo.interfaces())
-        ifaces_desired_state = _remove_absent_iface_entries(
-            ifaces_desired_state)
-        ifaces_desired_state = remove_ifaces_metadata(ifaces_desired_state)
-
-        assert_ifaces_state(ifaces_desired_state, ifaces_current_state)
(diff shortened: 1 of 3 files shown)


libnmstate/netapplier.py
```diff
@@ -360,9 +358,21 @@ def _sort_lag_slaves(desired_state, current_state):
...
-    return {ifname: ifstate
-            for ifname, ifstate in six.viewitems(ifaces_desired_state)
-            if ifstate.get('state') != 'absent'}
+    ifaces = {}
+    for ifname, ifstate in six.viewitems(ifaces_desired_state):
+        iface_admin_state = ifstate.get('state')
+        is_absent = iface_admin_state == 'absent'
+        is_virt_down = (
+            iface_admin_state == 'down' and
+            ifstate.get('type') in ('bond',
```
tyll: IMHO the list of virtual interfaces should be a constant or this check should be its own method to ensure that we do not have multiple lists of virtual types.

libnmstate/netapplier.py
```diff
@@ -360,9 +358,21 @@ def _sort_lag_slaves(desired_state, current_state):
 
 
 def _remove_absent_iface_entries(ifaces_desired_state):
```
tyll: With the change, the method's name does not match it's purpose correctly anymore. Seems to me that it would be better to have two methods, one for absent interfaces and one for down virtual interfaces.

EdDev: I considered it an extension to the `absent`, because we handle both the same. When we will not handle them the same, we will not need to remove them as we do here.

libnmstate/netapplier.py
EdDev: Makes sense. I will move it to the schema module.

libnmstate/netapplier.py
tyll: if you like to keep them in one method, then the method should be renamed to reflect that it also handles down interfaces. i.e. `_remove_down_virtual_and_absent_inface_entries()` which seems to be more ugly to me. But the current name is misleading now.

tyll:
this needs to be rebased

libnmstate/netapplier.py

tyll:
ci test please

tyll:
ci test please

PR #201, opened 2018-12-18
NM/device: Handle activation failures for connections

### LINKED ISSUE #198: _active_connection_callback() wrongfully assumes that only devices are activated
https://github.com/nmstate/nmstate/blob/39784d2e4593fc685f9d7752d3a92cc2aeeb307f/libnmstate/nm/device.py#L146

assumes that user_data will always contain a device. However, if a connection instead of a device is activated, there is no device and this logging will create an execption:
```
2018-12-14 13:40:46,169 root         DEBUG    Executing NM action: func=_safe_activate_async, args=(None, 'ovs-br0')
  File "/usr/lib/python2.7/site-packages/libnmstate/nm/device.py", line 148, in _active_connection_callback
    nmdev.get_iface(), e))
AttributeError: 'NoneType' object has no attribute 'get_iface'
```

The bug is present on multiple lines in the method.

tyll:
When a connection is activated, the user data for the active connection
callback does not contain a device object. Therefore report the
connection ID instead of the device instead.

fixes #198

### DIFF
--- a/libnmstate/nm/device.py
+++ b/libnmstate/nm/device.py
@@ -127,7 +127,7 @@ def _safe_activate_async(dev, connection_id):
     if connection_id:
         connection = client.get_connection_by_id(connection_id)
     specific_object = None
-    user_data = mainloop, dev, cancellable
+    user_data = mainloop, dev, connection_id, cancellable
     client.activate_connection_async(
         connection,
         dev,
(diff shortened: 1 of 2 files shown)


libnmstate/nm/device.py
```diff
@@ -135,24 +135,33 @@ def _safe_activate_async(dev, connection_id):
 
 
 def _active_connection_callback(src_object, result, user_data):
-    mainloop, nmdev, cancellable = user_data
+    mainloop, nmdev, connection_id, cancellable = user_data
     mainloop.drop_cancellable(cancellable)
     try:
         nm_act_con = src_object.activate_connection_finish(result)
     except Exception as e:
+        if nmdev:
```
EdDev: Could you please move this new code to a helper function that returns the two values?

libnmstate/nm/device.py
```diff
@@ -135,24 +135,33 @@ def _safe_activate_async(dev, connection_id):
...
-                    nmdev.get_iface(), e))
+                'Connection activation failed on {} {}: error={}'.format(
+                    activation_type, activation_object, e))
         return
 
     if nm_act_con is None:
-        mainloop.quit('Connection activation failed on %s: error=unknown' %
-                      nmdev.get_iface())
+        mainloop.quit('Connection activation failed on %s %s: error=unknown' %
+                      activation_type, activation_object)
```
EdDev: It cannot be done here, as these variables have not been defined yet.
With the helper func, you can just call it again.

libnmstate/nm/device.py
tyll: done

libnmstate/nm/device.py
tyll: I ensured that the variables are defined now

EdDev: You just do it even if there is no need to do it (as this is relevant for errors only).

libnmstate/nm/device.py
```diff
@@ -173,6 +177,19 @@ def _active_connection_callback(src_object, result, user_data):
                     ac.devname, ac.reason))
 
 
+def _get_activation_metadata(nmdev, connection_id):
+    if nmdev:
+        activation_type = "device"
```
EdDev: Single quotes please.

libnmstate/nm/device.py
tyll: changed

tyll:
ci test please

PR #203, opened 2018-12-20
integ tests: Test full rollback for vlans

SvetaHBar:


tyll:
Great test! Thank you!

tests/integration/vlan_test.py
```diff
@@ -22,7 +22,7 @@
 from .testlib import assertlib
 from .testlib import statelib
 from .testlib.statelib import INTERFACES
-
+import pytest
```
tyll: The import needs to be moved between the contextlib and the libnmstate imports.

tests/integration/vlan_test.py
```diff
@@ -45,6 +45,40 @@ def test_add_and_remove_two_vlans_on_same_iface(eth1_up):
     assert not current_state[INTERFACES]
 
 
+def test_rollback_for_vlans(eth1_up):
+    current_state = statelib.show_only([])
+    desired_state = {
```
tyll: Since this is the same state as in ` two_vlans_on_eth1()` it makes sense to move them to a constant and use it in both methods IMHO.

tests/integration/vlan_test.py
```diff
@@ -45,6 +45,40 @@ def test_add_and_remove_two_vlans_on_same_iface(eth1_up):
     assert not current_state[INTERFACES]
 
 
+def test_rollback_for_vlans(eth1_up):
+    current_state = statelib.show_only([])
```
tyll: Is there no method to show the full state? Passing an empty list here looks strange. This is also true for the other instance of this.

tests/integration/vlan_test.py
```diff
@@ -45,6 +45,40 @@ def test_add_and_remove_two_vlans_on_same_iface(eth1_up):
...
+                        }
+            }
+        ]
+    }
+
+    desired_state[INTERFACES][1]['invalid_key'] = 'foo'
+    with pytest.raises(netapplier.DesiredStateIsNotCurrentError):
+        netapplier.apply(desired_state)
+
+    current_state_after_apply = statelib.show_only([])
```
tyll: this is the other instance

PR #206, opened 2018-12-20
Test examples separately and add OVS example

tyll:


tyll:
@EdDev the CI shows the failure with OVS that we chatted about earlier

EdDev:
> the CI shows the failure with OVS that we chatted about earlier

Yes, I see. I will debug this, thanks!

EdDev:
Rebase and fixed the yaml examples:
- The `port` subtree position.
- Minimized the `absent` example. We seem to have a bug when removing a bridge and at the same time specifying its slave as `up`.

PR #208, opened 2018-12-27
nm: Do not raise explcitly within the checkpoint context

EdDev:
In case an exception occurs in a contextmanager, python will re-raise
the exception automatically after the __exit__ method is processed.
Raising from within __exit__ is not required in this case.

### DIFF
--- a/libnmstate/nm/checkpoint.py
+++ b/libnmstate/nm/checkpoint.py
@@ -88,7 +88,6 @@ class CheckPoint(object):
             self.destroy()
         else:
             self.rollback()
-            raise
 
     def create(self):
         devs = []


libnmstate/nm/checkpoint.py
```diff
@@ -88,7 +88,6 @@ def __exit__(self, exc_type, exc_val, exc_tb):
             self.destroy()
         else:
             self.rollback()
-            raise
```
dankenigsberg: https://docs.python.org/3/reference/datamodel.html#object.__exit__

> Note that __exit__() methods should not reraise the passed-in exception; this is the caller’s responsibility.

EdDev:
ci test please

PR #209, opened 2018-12-27
nm: Reduce noise when logging actions execution

EdDev:
The logging of the action function arguments is not helpful, makes it
difficult to follow the logs, therefore, they are removed.


dankenigsberg:
> The logging of the action function arguments is not helpful, makes it difficult to follow the logs, therefore, they are removed.

I have never looked at your logs myself, so I cannot really judge. code is fine.

PR #212, opened 2018-12-30
Linux bridge

EdDev:
Support Linux bridge in nmstate

### DIFF
--- a/libnmstate/appliers/linux_bridge.py
+++ b/libnmstate/appliers/linux_bridge.py
@@ -1,0 +1,41 @@
+#
+# Copyright 2018 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+import six
+
+
+def get_slaves_from_state(state, default=()):
+    ports = state.get('bridge', {}).get('port')
+    if ports is None:
+        return default
+    return [p['name'] for p in ports]
+
+
+def set_bridge_ports_metadata(master_state, slave_state):
+    _set_common_slaves_metadata(master_state, slave_state)
+
+    ports = master_state.get('bridge', {}).get('port', [])
+    port = next(
+        six.moves.filter(lambda n: n['name'] == slave_state['name'], ports),
+        {}
+    )
+    slave_state['_brport_options'] = port
+
+
+def _set_common_slaves_metadata(master_state, slave_state):
+    slave_state['_master'] = master_state['name']
+    slave_state['_master_type'] = master_state['type']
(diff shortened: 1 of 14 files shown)


libnmstate/appliers/linux_bridge.py
```diff
@@ -0,0 +1,41 @@
...
+
+    ports = master_state.get('bridge', {}).get('port', [])
+    port = next(
+        six.moves.filter(lambda n: n['name'] == slave_state['name'], ports),
+        {}
+    )
+    slave_state['_brport_options'] = port
+
+
+def _set_common_slaves_metadata(master_state, slave_state):
```
tyll: This seems to be a generic method that is not specific to linux bridges, can it maybe be moved to a different place?

libnmstate/appliers/linux_bridge.py
```diff
@@ -0,0 +1,41 @@
...
+    slave_state['_master'] = master_state['name']
+    slave_state['_master_type'] = master_state['type']
```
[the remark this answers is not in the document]
EdDev: - The `nm` package is indented for NetworkManager specific business logic, this code is API format processing, independent on any specific provider.
- The `netapplier` module needs to be split in a re-factoring effort, it is already too big, includes many unrelated things and therefore hard to read an maintain. It should end up as a glue of the general nmstate setup flow, with the idea that all individual steps will register into the flow.

For now I considered placing in the `appliers` package modules that handle specific focused subjects (for now, per iface type I guess, but it can evolve) that are not provider specific.
If you see a better split, you are welcome to suggest it.

libnmstate/appliers/linux_bridge.py
```diff
@@ -0,0 +1,41 @@
...
+
+def set_bridge_ports_metadata(master_state, slave_state):
+    _set_common_slaves_metadata(master_state, slave_state)
+
+    ports = master_state.get('bridge', {}).get('port', [])
+    port = next(
+        six.moves.filter(lambda n: n['name'] == slave_state['name'], ports),
+        {}
+    )
+    slave_state['_brport_options'] = port
```
tyll: IMHO it would be nicer to call this `_bridge_port_options`. I do not see an advantage in abbreviating this.

libnmstate/appliers/linux_bridge.py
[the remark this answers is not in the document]
tyll: Instead of using a directory it could also just be a file like `libnmstate/linux_bridge_applier.py` - subdirectories with only single files are IMHO more confusing. But if there are going to be more files in there, soon, it is probably OK.

libnmstate/appliers/linux_bridge.py
EdDev: I would need to change it for OVS as well and I prefer not to touch it for now.
When we will refactor netapplier, we should address this point.

To the point itself, I think that for keys, a short name fits nicer, especially when we need to debug it. But I have no strong feelings on this, it can be changed if you prefer the longer version.

libnmstate/appliers/linux_bridge.py
EdDev: Yes, I think you are right. I left/duplicated it here because I have no better place to place it in.
It should get out of here when we split the metadata handlers out of the netapplier.

libnmstate/appliers/linux_bridge.py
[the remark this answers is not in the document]
EdDev: I need to place there at least these for the metadata: `bond`, `ovs`, `team`.
Then, I want to place there anything that requires specific iface type logic but is not NM specific, not sure how much of this is there beyond the metadata stuff.
So yes, we will have there more soon.

PR #213, opened 2018-12-31
integ tests: Test rollback for bond

SvetaHBar:


tests/integration/bond_test.py
```diff
@@ -132,3 +134,38 @@ def test_add_bond_with_slaves_and_ipv4(eth1_up, eth2_up, setup_remove_bond99):
...
+                    }
+                ]
+            }
+
+    desired_state[INTERFACES][0]['invalid_key'] = 'foo'
+
+    with pytest.raises(netapplier.DesiredStateIsNotCurrentError):
+        netapplier.apply(desired_state)
+
+    time.sleep(2)
```
tyll: Why is the sleep needed here?

SvetaHBar: Sleep needed for let to rollback fully completed, before assert .

tyll: @EdDev this sounds to me like there is a bug in the rollback code and it should properly wait until the rollback is completed.

EdDev:
> this sounds to me like there is a bug in the rollback code and it should properly wait until the rollback is completed.

The rollback is performed in an async manner and as it occurs at the end, I do not think we should block on it. The only reason we block on the others is because we want to control the order of some things. We actually need to move to non blocking commands soon, serializing everything is not that good and causes us some problems (we need to do things in groups of commands).

Anyway, for this case, I think it is fine not to block on the rollback, it's not that we have anything else to do later on.

tyll:
Sure, the internal nmstate commands do not need to block/can run in parallel. However, this is the public method and IMHO it should block until either the state is fully applied or reverted.

EdDev:
But we block `apply` in order to verify at the end a stable state (block reasoning). For me, blocking on the rollback is a "special" scenario, I currently do not see the client requirement for this. This is mainly because there is no action suggested if it does not succeed or fail.

Lets try and agree on something like this: Open an RFE and suggest to optionally block on the rollback. If we find this needed, we can go forward and implement it (we may need NM support for this).

tyll:
Here is the RFE: https://nmstate.atlassian.net/browse/NMSTATE-100

PR #215, opened 2019-01-03
RPM packaging: Sync back changes from Fedora packaging

tyll:

@cathay4t I noticed that the Fedora spec uses Suggests instead of Recommends for NetworkManager. IMHO Recommends is better since by default nmstate needs to run on the same system as NetworkManager and Suggests are currently ignored by DNF.


tyll:
@cathay4t please review and approve

cathay4t:
Merged. Thanks for the patch.

PR #227, opened 2019-01-13
nmstatectl test assert_rc refactoring

EdDev:
Depends on #94

### DIFF
--- a/tests/integration/nmstatectl_test.py
+++ b/tests/integration/nmstatectl_test.py
@@ -70,7 +70,7 @@ def test_missing_operation():
     ret = libcmd.exec_cmd(cmds)
     rc, out, err = ret
 
-    assert_rc(rc, RC_FAIL2, ret)
+    assert rc == RC_FAIL2, format_exec_cmd_result(ret)
     assert "nmstatectl: error: invalid choice: 'no-such-oper'" in err
 
 
@@ -78,7 +78,7 @@ def test_show_command_with_no_flags():
     ret = libcmd.exec_cmd(SHOW_CMD)
     rc, out, err = ret
 
-    assert_rc(rc, RC_SUCCESS, ret)
+    assert rc == RC_SUCCESS, format_exec_cmd_result(ret)
     assert LOOPBACK_JSON_CONFIG in out
 
     state = json.loads(out)
@@ -89,7 +89,7 @@ def test_show_command_with_yaml_format():
     ret = libcmd.exec_cmd(SHOW_CMD + ['--yaml'])
     rc, out, err = ret
 
-    assert_rc(rc, RC_SUCCESS, ret)
+    assert rc == RC_SUCCESS, format_exec_cmd_result(ret)
     assert LOOPBACK_YAML_CONFIG in out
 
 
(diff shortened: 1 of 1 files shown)


tests/integration/nmstatectl_test.py
```diff
@@ -118,8 +118,12 @@ def test_set_command_with_yaml_format():
     ret = libcmd.exec_cmd(SET_CMD, stdin=ETH1_YAML_CONFIG)
     rc, out, err = ret
 
-    assert_rc(rc, RC_SUCCESS, ret)
+    assert_rc(rc, RC_SUCCESS, exec_cmd_report(ret))
 
 
-def assert_rc(actual, expected, return_tuple):
-    assert actual == expected, 'rc={}, out={}, err={}'.format(*return_tuple)
+def assert_rc(actual, expected, assert_report):
```
tyll: IMHO it would be better to make `assert_rc` less universal and more specific for the current use case, i.e. only pass `expected` and `return_tuple` into it. Currently it is just harder to use and does not follow the KISS principle.

tests/integration/nmstatectl_test.py
```diff
@@ -108,7 +108,7 @@ def test_show_command_only_non_existing():
     ret = libcmd.exec_cmd(SHOW_CMD + ['non_existing_interface'])
     rc, out, err = ret
 
-    assert_rc(rc, RC_SUCCESS, ret)
+    assert_rc(rc, RC_SUCCESS, exec_cmd_report(ret))
```
tyll: What is the benefit of calling a method instead of just writing

`assert rc == RC_SUCCESS, exec_cmd_report(ret)`

directly?.

tests/integration/nmstatectl_test.py
```diff
@@ -118,8 +118,12 @@ def test_set_command_with_yaml_format():
...
 
 
-def assert_rc(actual, expected, return_tuple):
-    assert actual == expected, 'rc={}, out={}, err={}'.format(*return_tuple)
+def assert_rc(actual, expected, assert_report):
+    assert actual == expected, assert_report
+
+
+def exec_cmd_report(exec_cmd_output):
```
tyll: I believe you once said that methods should be verbs and I think that report does not really match here. Also `exec_cmd_output` could maybe be better `exec_cmd_result`. With a verb it could be:

`format_exec_cmd_result(result)`

tyll: I think I would even rename `exec_cmd` to `exec` and make it return an object with a `__str__()` method that does what `exex_cmd_report` does.

tests/integration/nmstatectl_test.py
EdDev: You are right. In the process of changing it I misses the outer picture.

tests/integration/nmstatectl_test.py
EdDev: Dropping it now make sense, as you suggested in the previous comment.

tests/integration/nmstatectl_test.py
EdDev: I think it is an overkill to objectize this. We would need to create an object and then execute it (execution as part of the constructor is not a preferred approach).

Anyway, generating a report from the output seems reasonable to me.

Regarding the verb, I considered the `exec` prefix as such, but I'm ok with your suggestion.

EdDev:
ci test please

tests/integration/nmstatectl_test.py
tyll: I would also be fine with just having a result object but it looks good enough for me now as well.

PR #239, opened 2019-01-21
Enable autoconnect

EdDev:
Enable autoconnect on profile creation.

### DIFF
--- a/libnmstate/netapplier.py
+++ b/libnmstate/netapplier.py
@@ -1,5 +1,5 @@
 #
-# Copyright 2018 Red Hat, Inc.
+# Copyright 2018-2019 Red Hat, Inc.
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -305,7 +305,8 @@ def _edit_interfaces(ifaces_desired_state, ifaces_current_state):
         iface2prepare + proxy_ifaces)
     nm.applier.edit_existing_ifaces(ifaces_configs)
 
-    nm.applier.set_ifaces_admin_state(ifaces2edit)
+    nm.applier.set_ifaces_admin_state(ifaces2edit + proxy_ifaces,
+                                      con_profiles=ifaces_configs)
 
 
 def _canonicalize_desired_state(iface_desired_state, iface_current_state):
--- a/libnmstate/nm/connection.py
+++ b/libnmstate/nm/connection.py
@@ -142,7 +142,18 @@ def get_device_connection(nm_device):
 
 
 def get_device_active_connection(nm_device):
-    return nm_device.get_active_connection()
+    active_conn = None
+    if nm_device:
+        active_conn = nm_device.get_active_connection()
+    return active_conn
+
+
+def get_connection_by_id(connection_id):
+    client = nmclient.client()
+    conn = None
+    if connection_id:
+        conn = client.get_connection_by_id(connection_id)
+    return conn
 
 
 def _logging_connection_info(con_setting, source):
--- a/libnmstate/nm/connection.py
+++ b/libnmstate/nm/connection.py
@@ -1,5 +1,5 @@
 #
-# Copyright 2018 Red Hat, Inc.
+# Copyright 2018-2019 Red Hat, Inc.
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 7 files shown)


EdDev:
There seem to be a problem enabling autoconnect: https://bugzilla.redhat.com/show_bug.cgi?id=1667874

I will try to conceive a workaround for this issue (retry or sleep).

EdDev:
ci test please

tyll:
another possible workaround might be to activate it without autoconnect and enable autoconnect when it is already active in the configuration.

libnmstate/nm/device.py
```diff
@@ -138,6 +147,20 @@ def _safe_activate_async(dev, connection_id):
     )
 
 
+def _get_connection(client, connection_id):
+    conn = None
+    if connection_id:
+        conn = client.get_connection_by_id(connection_id)
```
tyll: This looks like we should have our own client class that just has a get_connection method that works like this.

libnmstate/nm/device.py
```diff
@@ -123,13 +121,24 @@ def _safe_activate_async(dev, connection_id):
     client = nmclient.client()
     mainloop = nmclient.mainloop()
     cancellable = mainloop.new_cancellable()
-    connection = None
-    if connection_id:
-        connection = client.get_connection_by_id(connection_id)
+
+    conn = _get_connection(client, connection_id)
```
tyll: IMHO connection is a better variable name than conn.

libnmstate/nm/device.py
```diff
@@ -126,6 +126,9 @@ def _safe_activate_async(dev, connection_id):
     connection = None
     if connection_id:
         connection = client.get_connection_by_id(connection_id)
+    if connection:
```
tyll: Should there be a check that ensures that it is only called with dev or connection_id? Is it going to be called with both but the connection_id invalid?

EdDev:
The autoconnect is supposed to resolve two issues: State persistence over boot and auto activation of the devices (like in the case of bond with dhcp).

I think that the last proposal will probably not solve the second problem.

libnmstate/nm/connection.py
```diff
@@ -108,9 +108,9 @@ def create_setting(con_name, iface_name, iface_type):
     con_setting.props.interface_name = iface_name
     con_setting.props.uuid = str(uuid.uuid4())
     con_setting.props.type = iface_type
-    con_setting.props.autoconnect = False
```
tyll: The copyright header needs to be updated.

libnmstate/nm/device.py
```diff
@@ -17,6 +17,7 @@
 
 import logging
 
+from . import connection
```
tyll: The copyright header needs to be updated.

libnmstate/nm/device.py
tyll: Eventiually it should be a  class that inherits from the NM client class. Moving it to a module is not an improvement IMHO. It does not need to be fixed now, though.

libnmstate/netapplier.py
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright 2018 Red Hat, Inc.
```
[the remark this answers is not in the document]

libnmstate/nm/nmclient.py
```diff
@@ -68,17 +68,24 @@ def __init__(self):
...
             func, args, kwargs = action
             logging.debug('Executing NM action: func=%s', func.__name__)
             func(*args, **kwargs)
         else:
             logging.debug('NM action queue exhausted, quiting mainloop')
             self._mainloop.quit()
 
+    def execute_last_action(self):
+        func, args, kwargs = self._last_action
+        logging.debug('Executing NM (last) action: func=%s', func.__name__)
```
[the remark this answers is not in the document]
EdDev: Done

libnmstate/nm/device.py
```diff
@@ -161,6 +161,10 @@ def _active_connection_callback(src_object, result, user_data):
             logging.debug(
                 'Connection activation canceled on %s %s: error=%s',
                 activation_type, activation_object, e)
+        elif is_connection_unavailable(e):
+            logging.warning('Connection unavailable on %s %s, retrying',
+                            activation_type, activation_object)
+            mainloop.execute_last_action()
```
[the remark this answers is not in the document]
EdDev: We have the mainloop `timeout` to protect us from such a scenario.

PR #240, opened 2019-01-21
RPMs: Require python-setuptools

tyll:
nmstatectl needs python-setuptools to import `pkg_resources`:
```
Traceback (most recent call last):
  File "/usr/bin/nmstatectl", line 5, in <module>
    from pkg_resources import load_entry_point
ImportError: No module named pkg_resources
```

Signed-off-by: Till Maas <opensource@till.name>

PR #241, opened 2019-01-21
PoC for docker packaging and using a base image

tyll:
We need to figure out an idea for which image names and tags to use, we have at least the following:

- docker base image for CentOS 7, Fedora 28, 29, Rawhide
- docker images with the stable releases built from RPM (do we also want them built from pip?)
    - ideally there would also be a simple test to check that the dependencies work, the current integration tests clobber missing dependencies (i.e. for python-setuptools as shown in #240).
- docker images with the development RPMs (to be able to test and correct problems with the docker images quicker) and to allow users to test unreleased bug fixes

TODO:
[ ] Make the integration tests use the base image(s)
[ ] Add base image for Fedora

This is intendend to replace #154 eventually

### DIFF
--- a/packaging/Dockerfile.centos_7_git_rpm
+++ b/packaging/Dockerfile.centos_7_git_rpm
@@ -1,0 +1,20 @@
+# This Dockerfile is based on the recommendations provided in the
+# Centos official repository (https://hub.docker.com/_/centos/).
+# It enables systemd to be operational.
+FROM centos:7.5.1804
+ENV container docker
+COPY docker_enable_systemd.sh docker_sys_config.sh ./
+RUN bash ./docker_enable_systemd.sh && rm ./docker_enable_systemd.sh -f
+
+# libibverbs is a dependency for openvswitch that is missing as a dependency in the RPM
+RUN yum -y upgrade && \
+    yum -y install NetworkManager NetworkManager-ovs openvswitch libibverbs && \
+    yum -y install epel-release yum-utils && \
+    yum-config-manager --add-repo=https://copr.fedorainfracloud.org/coprs/nmstate/nmstate-git-el7/repo/epel-7/nmstate-nmstate-git-el7-epel-7.repo && \
+    yum -y install nmstate && \
+    yum clean all && \
+    bash ./docker_sys_config.sh && rm ./docker_sys_config -f
+
+VOLUME [ "/sys/fs/cgroup" ]
+
+CMD ["/usr/sbin/init"]
--- a/packaging/Dockerfile.centos_7_git_rpm
+++ b/packaging/Dockerfile.centos_7_git_rpm
@@ -1,19 +1,9 @@
-# This Dockerfile is based on the recommendations provided in the
-# Centos official repository (https://hub.docker.com/_/centos/).
-# It enables systemd to be operational.
-FROM centos:7.5.1804
-ENV container docker
-COPY docker_enable_systemd.sh docker_sys_config.sh ./
-RUN bash ./docker_enable_systemd.sh && rm ./docker_enable_systemd.sh -f
+FROM nmstate/centos_7_base
 
-# libibverbs is a dependency for openvswitch that is missing as a dependency in the RPM
-RUN yum -y upgrade && \
-    yum -y install NetworkManager NetworkManager-ovs openvswitch libibverbs && \
-    yum -y install epel-release yum-utils && \
+RUN yum -y install yum-utils && \
     yum-config-manager --add-repo=https://copr.fedorainfracloud.org/coprs/nmstate/nmstate-git-el7/repo/epel-7/nmstate-nmstate-git-el7-epel-7.repo && \
     yum -y install nmstate && \
-    yum clean all && \
-    bash ./docker_sys_config.sh && rm ./docker_sys_config -f
+    yum clean all
 
 VOLUME [ "/sys/fs/cgroup" ]
 
--- a/packaging/Dockerfile.centos_7_stable_rpm
+++ b/packaging/Dockerfile.centos_7_stable_rpm
@@ -1,0 +1,19 @@
+# This Dockerfile is based on the recommendations provided in the
+# Centos official repository (https://hub.docker.com/_/centos/).
+# It enables systemd to be operational.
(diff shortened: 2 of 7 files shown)


tyll:
@cathay4t @EdDev it would be great if you could comment on this approach and tag/image/file naming. If we are in agreement I will finalize this.

packaging/Dockerfile.centos_7_git_rpm
```diff
@@ -0,0 +1,20 @@
+# This Dockerfile is based on the recommendations provided in the
+# Centos official repository (https://hub.docker.com/_/centos/).
+# It enables systemd to be operational.
+FROM centos:7.5.1804
```
EdDev: This one should be based on the previous image.
`FROM centos7-nmstate-base`
And add only the additions here.

packaging/Dockerfile.centos_7_git_rpm
```diff
@@ -0,0 +1,20 @@
+# This Dockerfile is based on the recommendations provided in the
```
EdDev: The name `git_rpm` is a bit strange. If I understand correctly, this image contains master. So you could use `master` or `master_copr`.

packaging/Dockerfile.centos_7_stable_rpm
```diff
@@ -0,0 +1,19 @@
+# This Dockerfile is based on the recommendations provided in the
```
EdDev: How about `latest-epel-testing`

packaging/Dockerfile.fedora_29_git_rpm
```diff
@@ -0,0 +1,20 @@
+# This Dockerfile is based on the recommendations provided in the
+# Centos official repository (https://hub.docker.com/_/centos/).
+# It enables systemd to be operational.
+FROM fedora:29
+ENV container docker
+COPY docker_enable_systemd.sh docker_sys_config.sh ./
+RUN bash ./docker_enable_systemd.sh && rm ./docker_enable_systemd.sh -f
+
+# libibverbs is a dependency for openvswitch that is missing as a dependency in the RPM
+RUN yum -y upgrade && \
```
EdDev: Don't we need to use `dnf`?

packaging/Dockerfile.centos_7_git_rpm
```diff
@@ -1,19 +1,9 @@
-# This Dockerfile is based on the recommendations provided in the
-# Centos official repository (https://hub.docker.com/_/centos/).
-# It enables systemd to be operational.
-FROM centos:7.5.1804
-ENV container docker
-COPY docker_enable_systemd.sh docker_sys_config.sh ./
-RUN bash ./docker_enable_systemd.sh && rm ./docker_enable_systemd.sh -f
+FROM nmstate/centos_7_base
```
EdDev: We are currently using nmstate/centos7-nmstate-* format, lets keep the convention.

packaging/Dockerfile.centos_7_git_rpm
```diff
@@ -1,19 +1,9 @@
...
 
-# libibverbs is a dependency for openvswitch that is missing as a dependency in the RPM
-RUN yum -y upgrade && \
-    yum -y install NetworkManager NetworkManager-ovs openvswitch libibverbs && \
-    yum -y install epel-release yum-utils && \
+RUN yum -y install yum-utils && \
```
EdDev: Isn't this suppose to be part of base?

tyll: No, the CI system does not need to install the COPR repository.

packaging/Dockerfile.centos_7_git_rpm
tyll: IMHO Copr is an implementation detail that is not relevant for the image name. `git` is the name that we use in the Copr repo, i.e. `nmstate-git-el7` - so should we rename that copr to `nmstate-git-master-el7`?

packaging/Dockerfile.centos_7_stable_rpm
tyll: `epel-testing` is IMHO implementation detail. The idea is to provide the stable/tagged nmstate releases via docker and epel-testing is current repo that is used for this.

packaging/Dockerfile.fedora_29_git_rpm
tyll: It is the same.

packaging/Dockerfile.centos_7_stable_rpm
EdDev: `rpm` is also an implementation detail.
`latest-release`, just `latest` or just leave the stable then?

packaging/Dockerfile.fedora_29_git_rpm
EdDev: Fedora abandon `dnf` in favour of `yum` as an alias? I remember we needed to install a special rpm to use `yum` on Fedora.

tyll:
> `rpm` and `git` are also implementation details, I would prefer not to see them in the name.

So do we want the Docker images only for the RPM packages and not for the pip installation? Then we can drop the rpm name. If not, then we need something to distinguish those two kinds.

git is currently based on the name of the COPR. IMHO latest(-release) does not convey the notion that it is an unreleased version.

packaging/Dockerfile.centos_7_stable_rpm
tyll: Please see my answer in the conversaion.

EdDev:
There are a lot of points to cover here, I think we need a meeting on this one.

But from the version of images point of view, I think that you can use tags only for incremental changes, which usually match the need to update nmstate version or some dependencies, not the OS.
We could decide to just provide the Dockerfile/s, and not the image.

Regarding pypi, I think we can use it on an Ubuntu image. There is no reason to use it for CentOS/Fedora/RHEL if we have rpm/s available. The dev image provides the 'install from source'.

tyll:
should we maybe then just skip this for now? Initially I hoped to get this ready for the conferences. It does not seem to be urgent anymore.

tyll:
I will create a new PR for Fedora once #233 is merged.

PR #245, opened 2019-01-24
Bump version to 0.0.5

tyll:


tyll:
it would be great to merge this next to avoid to have too many RPMs on COPR with the wrong version number

PR #252, opened 2019-01-30
Create and remove linux  bridge with min desired state

SvetaHBar:


tests/integration/linux_bridge_test.py
```diff
@@ -0,0 +1,94 @@
...
+
+import yaml
+
+from libnmstate import netapplier
+
+from .testlib import statelib
+from .testlib.statelib import INTERFACES
+
+
+IPV4_ADDRESS1 = '192.0.2.251'
```
tyll: These addresses do not seem to be used.

tests/integration/linux_bridge_test.py
```diff
@@ -0,0 +1,94 @@
...
+        mac-ageing-time: 300
+        multicast-snooping: true
+        stp:
+          enabled: true
+          forward-delay: 15
+          hello-time: 2
+          max-age: 20
+          priority: 32768
+      port:
+        - name: eth1
```
tyll: It would be geat to also test more than one port, adding/removing ports to an existing bridge, setting up a VLAN on top of a bridge etc. This does not need to happen in this PR but it makes me wonder if we should start writing a high-level test plan (maybe with user stories) and then see if we cover all the scenarios.

EdDev: I agree. I think we can add such a plan in the jira issues in advance and then implement them. New features should have these in advance, as part of the feature story.
Anyway, this discussion is part of our retrospection/planning.

tests/integration/linux_bridge_test.py
SvetaHBar: You right , fixed.

PR #253, opened 2019-01-30
Integration tests: Show live logs and timestamp

tyll:


automation/run-integration-tests.sh
```diff
@@ -69,6 +69,11 @@ function run_tests {
       pytest \
         --verbose --verbose \
         --log-level=DEBUG \
+        --log-date-format='%Y-%m-%d %H:%M:%S' \
```
EdDev: I was hopping the `logging_setup` fixture will set the basics. Any chance we can do all this there or in the code in general?

I guess the other option is to set a config logger file.

There are currently 3 places we setup the logger: pytest, nmstatectl (main) and integration tests conftest module.
We should converge them all somehow.

If we cannot find any better solution, then we can take this one as a temporally solution.

cathay4t:
ci test please.

cathay4t:
ci test please.

PR #255, opened 2019-02-05
Provide requirements for ubuntu (on travis)

EdDev:
With this addition, `gi.require_version('NM', '1.0')` should not fail.
However, it is questionable if this is needed at all, because for unit tests, we do not need it on the system.

In addition, we need to be able and fail with a meaningful error so the user can take actions (installing the relevant packages).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ addons:
   apt:
     packages:
     - libgirepository1.0-dev
+    - gir1.2-networkmanager-1.0
 
 install:
   - pip install tox


tyll:
The change looks good to me but it seems to make the tests on Ubuntu now fail. The idea to use a Fedora container to run the unit tests was to not have to care about what is needed on Ubuntu. :-)

EdDev:
> The idea to use a Fedora container to run the unit tests was to not have to care about what is needed on Ubuntu. :-)

You are right, but the idea is for the unit tests to run on any platform that the developers runs on.  After creating this PR I did questioned its relevance. I think I'll close it.

PR #256, opened 2019-02-05
nm: Exit with a human readable error when missing libnm introspection

EdDev:


libnmstate/nm/nmclient.py
```diff
@@ -44,7 +43,15 @@ def client(refresh=False):
...
-        _nmclient = NM.Client.new(None)
+        if NM:
+            _nmclient = NM.Client.new(None)
+        else:
+            logging.error(
+                'Missing introspection data for libnm'
+                'please make sure to install it prior to running nmstate.\n'
+                'Check the documentation for more information.'
+            )
+            exit(1)
```
tyll: IMHO it is wrong to exit here. This would for example kill an ansible proceess instead of allowing it to properly report back an error message.

libnmstate/nm/nmclient.py
```diff
@@ -44,7 +43,15 @@ def client(refresh=False):
...
     # running
     if _nmclient is None or refresh:
-        _nmclient = NM.Client.new(None)
+        if NM:
+            _nmclient = NM.Client.new(None)
+        else:
+            logging.error(
+                'Missing introspection data for libnm'
+                'please make sure to install it prior to running nmstate.\n'
+                'Check the documentation for more information.'
```
tyll: IMHO we can add the path of the missing file here (`/usr/lib64/girepository-1.0/NM-1.0.typelib`) to make it easier for users to find the appropriate package instead of having to find the documentation first.

libnmstate/nm/nmclient.py
```diff
@@ -27,7 +27,6 @@
     gi.require_version('NM', '1.0')  # NOQA: F402
     from gi.repository import NM     # pylint: disable=no-name-in-module
 except ValueError:
-    # Allow the error to pass in case we are running in a unit test context
```
tyll: The comment is still valid, is it not? Otherwise the detailed error message should happen here.

EdDev: Not exactly. It is not for the unit test, it is for the user who attempts to run it without the proper dependencies installed.
Running stuff from the root on import is nasty, we use this to avoid the noise here and manage it in a more controlled manner.

libnmstate/nm/nmclient.py
EdDev: I'm afraid each distribution has it in a different path. This is the path on Ubuntu: `/usr/lib/x86_64-linux-gnu/girepository-1.0`

libnmstate/nm/nmclient.py
EdDev: I had no better idea and focused on a cli client. Any suggestion? You prefer an exception?

EdDev: I'll use an exception.

EdDev:
ci test please

libnmstate/nm/nmclient.py
tyll: Do we actually have the information in the documentation?

EdDev: We are missing a lot of documentation.
It should have appeared here: https://nmstate.github.io/#install

I will add it.

EdDev: @tyll , I have pushed #306 in the README.

PR #258, opened 2019-02-06
integ tests:Slave IP addr learned via DHCP added as static IP to linux bridge

SvetaHBar:


tests/integration/dhcp_test.py
```diff
@@ -219,3 +219,79 @@ def _clean_up():
         os.unlink(DNSMASQ_CONF_PATH)
     except FileNotFoundError:
         pass
+
+
+def test_slave_ipaddr_learned_via_dhcp_added_as_static_to_linux_bridge(dhcp_env
```
EdDev: If you use `ip` instead of `ipaddr`, will it fit one line?

SvetaHBar: yes, fixed

tests/integration/dhcp_test.py
```diff
@@ -219,3 +219,79 @@ def _clean_up():
...
+                'name': 'linux-br0',
+                'type': 'linux-bridge',
+                'state': 'up',
+                'ipv4': {
+                    'enabled': True,
+                    'dhcp': False,
+                    'address': dhcpcli_ip
+                },
+                'bridge': {
+                    'options': {
```
tyll: IMHO it would be nicer to remove all these options to see if it works with the default settings.

tests/integration/dhcp_test.py
```diff
@@ -219,3 +219,79 @@ def _clean_up():
...
+                             'forward-delay': 15,
+                             'hello-time': 2,
+                             'max-age': 20,
+                             'priority': 32768
+                             }
+                     },
+                    'port': [
+                        {
+                         'name': 'dhcpcli',
+                         'stp-hairpin-mode': False,
```
tyll: IMHO it would be great to keep the name only here.

SvetaHBar: Current change is causing the bug - port verification is failed in assert_ifaces_state func

tests/integration/dhcp_test.py
SvetaHBar: Done

PR #260, opened 2019-02-06
nmstatectl: Allow to set multiple states

tyll:

### DIFF
--- a/nmstatectl/nmstatectl.py
+++ b/nmstatectl/nmstatectl.py
@@ -86,7 +86,7 @@ def setup_subcommand_set(subparsers):
     parser_set = subparsers.add_parser('set', help='Set network state')
     parser_set.add_argument('file', help='File containing desired state. '
                             'stdin is used when no file is specified.',
-                            nargs='?')
+                            nargs='*')
     parser_set.add_argument(
         '--no-verify', action='store_false', dest='verify', default=True,
         help='Do not verify that the state was completely set and disable '
@@ -127,16 +127,24 @@ def show(args):
 
 
 def apply(args):
-    # read from file when it is not '-' unless '-' actually exists
-    if args.file and (args.file != '-' or os.path.isfile(args.file)):
-        with open(args.file) as statefile:
-            statedata = statefile.read()
-    elif not sys.stdin.isatty() or args.file == '-':
+    if args.file:
+        for statefile in args.file:
+            if statefile == '-' and not os.path.isfile(statefile):
+                statedata = sys.stdin.read()
+            else:
+                with open(statefile) as statefile:
+                    statedata = statefile.read()
+
+            apply_state(statedata, args.verify)
+    elif not sys.stdin.isatty():
         statedata = sys.stdin.read()
+        apply_state(statedata, args.verify)
     else:
         sys.stderr.write('ERROR: No state specified\n')
         return 1
 
+
+def apply_state(statedata, verify_change):
     use_yaml = False
     # JSON dictionaries start with a curly brace
     if statedata[0] == '{':
@@ -144,7 +152,7 @@ def apply(args):
     else:
         state = yaml.load(statedata)
         use_yaml = True
-    netapplier.apply(state, verify_change=args.verify)
+    netapplier.apply(state, verify_change)
     print('Desired state applied: ')
     print_state(state, use_yaml=use_yaml)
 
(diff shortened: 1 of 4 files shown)


tests/integration/testlib/__init__.py
```diff
@@ -15,6 +15,38 @@
...
+
 import pytest
 
+PATH_MAX = 4096
+
+
 pytest.register_assert_rewrite(__name__ + '.assertlib')
+
+
+def find_examples_dir():
```
EdDev: I would prefer you move it to examplelib.py

nmstatectl/nmstatectl.py
```diff
@@ -86,7 +86,7 @@ def setup_subcommand_set(subparsers):
     parser_set = subparsers.add_parser('set', help='Set network state')
     parser_set.add_argument('file', help='File containing desired state. '
```
EdDev: Now it is `File(s)`

tests/integration/testlib/__init__.py
tyll: Why? It seems to be a little bit too much to create a separate file for just one method.

nmstatectl/nmstatectl.py
tyll: The help output is fine IMHO, since it shows that file occurs more than once:
```
./ncl set --help
usage: ncl set [-h] [--no-verify] [file [file ...]]

positional arguments:
  file         File containing desired state. stdin is used when no file is
               specified.

optional arguments:
  -h, --help   show this help message and exit
  --no-verify  Do not verify that the state was completely set and disable
               rollback to previous state.
```

tests/integration/testlib/__init__.py
EdDev: because it is a bad practice to define non package common helpers in __init__ files. It may cause a dependency loop in the future.
I prefer to have there nothing, package constants or something that is needed to serve the whole package. Yiu may end up having many unrelated small helpers in there, which is a problem by itself.

tyll: Those are all theoretical problems that we do not have right now...

EdDev: If there is no super good reason, I prefer not to add code to the init files.

tyll: moved it

nmstatectl/nmstatectl.py
```diff
@@ -127,24 +127,35 @@ def show(args):
...
-    # read from file when it is not '-' unless '-' actually exists
-    if args.file and (args.file != '-' or os.path.isfile(args.file)):
-        with open(args.file) as statefile:
-            statedata = statefile.read()
-    elif not sys.stdin.isatty() or args.file == '-':
+    if args.file:
+        for statefile in args.file:
+            if statefile == '-' and not os.path.isfile(statefile):
+                statedata = sys.stdin.read()
+
```
EdDev: nit: No need for this newline

tests/integration/testlib/exampleslib.py
```diff
@@ -0,0 +1,47 @@
+#
```
EdDev: Shouldn't the name be singular rather than plural? (example vs examples)?

tests/integration/nmstatectl_test.py
```diff
@@ -16,10 +16,12 @@
 #
 
 import json
+import os.path
```
EdDev: `import os` should be enough.

tests/integration/testlib/exampleslib.py
tyll: I changed it but honestly I think having a single file for just one method is wrong.

tests/integration/nmstatectl_test.py
tyll: Actually `import os` is doing more, so `import os.path` is enough because the other os content is not needed. I do not see why it would be better to import everything, therefore I will keep it for now. I am fine with changing this if you convince me why `import os` is better.

nmstatectl/nmstatectl.py
tyll: IMHO it matches the other newlines in this method.

tyll:
ci test please

tyll:
ci test please

nmstatectl/nmstatectl.py
EdDev: A newline after ‘else’ is very strange.

tyll: I only saw newlines before the else. I removed them now, too. I really wish a tool would just take care of this formatting so I do not need to waste brain cycles on this. ;-)

PR #268, opened 2019-02-17
Integ tests nm: OVS tests (bridge and single system port)

EdDev:


tests/integration/nm/ovs_test.py
```diff
@@ -0,0 +1,185 @@
...
+
+def _attach_port_to_bridge(port_state):
+    port_profile_name = nm.ovs.PORT_PROFILE_PREFIX + port_state['name']
+
+    _create_proxy_port(port_profile_name, port_state)
+    _enslave_interface(port_profile_name, port_state)
+
+
+def _enslave_interface(port_profile_name, port_state):
+    slave_nmdev = nm.device.get_device_by_name(port_state['name'])
```
EdDev: Use `OB.PORT_NAME`

tests/integration/nm/ovs_test.py
```diff
@@ -0,0 +1,185 @@
...
+
+    with mainloop():
+        _create_bridge_iface(iface_bridge_settings)
+        ports_state = bridge_desired_state[OB.CONFIG_SUBTREE][OB.PORT_SUBTREE]
+        for port_state in ports_state:
+            _attach_port_to_bridge(port_state)
+
+
+def _attach_port_to_bridge(port_state):
+    port_profile_name = nm.ovs.PORT_PROFILE_PREFIX + port_state['name']
```
EdDev: Use `OB.PORT_NAME`

tests/integration/nm/ovs_test.py
```diff
@@ -0,0 +1,185 @@
...
+
+    with _bridge_interface(bridge_desired_state):
+        bridge_current_state = _get_bridge_current_state()
+        assert bridge_current_state == bridge_default_config
+        assert not bridge_current_state[OB.CONFIG_SUBTREE][OB.PORT_SUBTREE]
+
+    assert not _get_bridge_current_state()
+
+
+def test_create_and_remove_bridge(eth1_up, bridge_default_config):
```
EdDev: Change name: Add `with_system_port`

libnmstate/schema.py
```diff
@@ -67,3 +67,20 @@ class LinuxBridge(object):
     PORT_STP_PRIORITY = 'stp-priority'
     PORT_STP_HAIRPIN_MODE = 'stp-hairpin-mode'
     PORT_STP_PATH_COST = 'stp-path-cost'
+
+
+class OvsBridge(object):
```
tyll: Seems it is either OVS or OvS but not Ovs.

tests/integration/nm/ovs_test.py
```diff
@@ -0,0 +1,185 @@
...
+
+BRIDGE0 = 'brtest0'
+ETH1 = 'eth1'
+
+
+@pytest.fixture
+def bridge_minimum_config():
+    return {
+        OB.CONFIG_SUBTREE: {
+            OB.PORT_SUBTREE: []
```
tyll: Is an empty port subtree required for the minimum config?

tests/integration/nm/ovs_test.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+def test_create_and_remove_minimum_config_bridge(bridge_minimum_config,
+                                                 bridge_default_config):
+    bridge_desired_state = bridge_minimum_config
+
+    with _bridge_interface(bridge_desired_state):
+        bridge_current_state = _get_bridge_current_state()
+        assert bridge_current_state == bridge_default_config
+        assert not bridge_current_state[OB.CONFIG_SUBTREE][OB.PORT_SUBTREE]
```
tyll: IMHO the assert should be that it is the empty list

tests/integration/nm/ovs_test.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+def _attach_port_to_bridge(port_state):
+    port_profile_name = nm.ovs.PORT_PROFILE_PREFIX + port_state['name']
+
+    _create_proxy_port(port_profile_name, port_state)
+    _enslave_interface(port_profile_name, port_state)
+
+
+def _enslave_interface(port_profile_name, port_state):
```
tyll: I would prefer names that avoid using slave as a name. How about `_plug_interface()`?

tests/integration/nm/ovs_test.py
tyll: `port_nmdev`...

libnmstate/nm/ovs.py
```diff
@@ -170,6 +176,10 @@ def _get_bridge_port_info(port_profile, devices_info):
 def _get_bridge_options(bridge_device):
     bridge_options = {}
     con = connection.get_device_connection(bridge_device)
+    if not con:
+        logging.debug('No active connection detected for dev %s',
+                      bridge_device)
+        con = bridge_device.get_available_connections()[0]
```
tyll: What is the idea here?

EdDev: This is indeed strange, but for an ovs bridge device, it may exist as a profile without a backing active connection.
It is enough to pass the tests now, but I expect this to need some tweaking to pass some other edge scenarios.

I hope to be guided by the tests/scenarios, anything else is just guessing. Intuitive it is not.

libnmstate/schema.py
EdDev: It can be any of the 3, I see no problem here. Even libnm uses the same.
But if you commented on this, I guess it is important to you... changing.

tests/integration/nm/ovs_test.py
EdDev: This is testing the nm ovs provider module, not nmstate api. For this test, I think we are ok, it just helps handling the ports in the tests uniformly.

The minimum here is in terms of the properties.

tests/integration/nm/ovs_test.py
EdDev: Done

tests/integration/nm/ovs_test.py
EdDev: Done

tests/integration/nm/ovs_test.py
EdDev: Done

tests/integration/nm/ovs_test.py
EdDev: Ok, lets try 'connect'.

tests/integration/nm/ovs_test.py
EdDev: In this context (and detail), `port` does not fit. We have both an interface and a port, ovs/nm differentiating between them.

Changed it to `iface_nmdev`.

tests/integration/nm/ovs_test.py
```diff
@@ -0,0 +1,185 @@
...
+        _create_bridge(state)
+        yield
+    finally:
+        _delete_iface(BRIDGE0)
+        for p in state[OB.CONFIG_SUBTREE][OB.PORT_SUBTREE]:
+            _delete_iface(nm.ovs.PORT_PROFILE_PREFIX + p[OB.PORT_NAME])
+            _delete_iface(p[OB.PORT_NAME])
+
+
+def _get_bridge_current_state():
```
tyll: For me the feelings that the test cases should be classes becomes stronger. Given that we have many methods that operate on the constants instead of (only) their arguments lets me think that it would be nice to have a BridgeTestcase class with properties such as `bridge_name` and methods like `_get_current_state(self)`. It is somehow like a class implemented as a module currently.

EdDev: Currently it is just forking the Linux bridge nm tests.
Refactoring this is a good idea and it should probably fit the Linux bridge one as well.

Could you please add it as an issue and assign it to the current sprint?

libnmstate/nm/ovs.py
tyll: This should somehow be documented then.

tyll: IMHO the logic which connection to select (in this case the first available connection) should be part of the `nm.applier` code instead of the different device modules to ensure that they all take the same choice.

tests/integration/nm/ovs_test.py
```diff
@@ -98,7 +98,6 @@ def _create_bridge_iface(iface_bridge_settings):
 def _delete_iface(devname):
     nmdev = nm.device.get_device_by_name(devname)
     with mainloop():
-        nm.device.deactivate(nmdev)
```
tyll: There is a typo in the commit message `ceanup`.

EdDev: Done

libnmstate/nm/ovs.py
EdDev: This is an interesting issue.
- Currently, the applier is dependent and calling the individual modules. The specific modules cannot depend on it as it will cause a dependency loop. We could reverse the dependency as a refactoring step, making the individual modules a plugin to the nm applier.
- I'm not convinced there is a shared logic regarding which connection to use at the moment. In most cases, when a profile is present and relevant, it is activated and appears in 'activated' state. With OVS it is not following that behaviour, so an adjustment was needed.
- Regarding documentation, I was hoping that the debug message is enough. It states why we are in this condition and then the `bridge_device.get_available_connections()[0]` is self explanatory. I guess I can wrap this in a special function and document it, but I am not sure how to name it or what to exactly say.
We could also comment this in the commit message, will that satisfy your request here?

EdDev:
ci test please

libnmstate/nm/ovs.py
tyll: Not sure, why there needs to be a dependency loop.My idea is that `get_ovs_info()` is called with a profile as its argument instead of the device name. But now that I think of this, `_get_bridge_options()` should not access a non-active profile to get the options since currently Nmstate only reports the runtime state and not the configuration state. The information from `bridge_device.get_available_connections()[0]` is not guaranteed to match the actual runtime state. I guess the current code is a result of the problem that NM sometimes forgets that it manages a OVS bridge. Therefore the device is there but there is no active connection.

tyll: About the docs: The justification, why accessing the first available profile is correct, is missing.

EdDev: I do not understand how that is related to the 'applier' then, the `get_ovs_info` is called from netinfo, not from the applier.
Not accessing a non active profile is a correct general note, but NM is not creating a bridge if no interface is connected to it. I can reflect that using an `xfail` test instead. Is this what you prefer?

Doc related: The justification or explanation is in the commit message, telling long stories in the code is not a recommended practice (it rots).

EdDev:
ci test please

tests/integration/nm/ovs_test.py
```diff
@@ -0,0 +1,185 @@
...
+            _delete_iface(nm.ovs.PORT_PROFILE_PREFIX + p[OB.PORT_NAME])
+            _delete_iface(p[OB.PORT_NAME])
+
+
+def _get_bridge_current_state():
+    nm.nmclient.client(refresh=True)
+    nmdev = nm.device.get_device_by_name(BRIDGE0)
+    devices_info = [(dev, nm.device.get_device_common_info(dev))
+                    for dev in nm.device.list_devices()]
+    return nm.ovs.get_bridge_info(nmdev, devices_info) if nmdev else {}
```
tyll: IMHO it is hard to read that the complex lines 103-104 are ignored when nmdev evaluates to False. Would be easiert to read to have a `if nmdev: result = ... else: result = {}; return result`, i.e. the condition first instead at the end.

EdDev: For me, it is easier to read it as is. I have two parameters, and we return one of two options.
If we add the conditions, we will have more code lines, which normally is not a good thing.

Using `x if a else y` is pretty common in python, especially if it lands on a single line.

libnmstate/nm/ovs.py
```diff
@@ -123,6 +123,10 @@ def is_ovs_interface_type_id(type_id):
     return type_id == nmclient.NM.DeviceType.OVS_INTERFACE
 
 
+def get_bridge_info(bridge_device, devices_info):
+    return {'bridge': get_ovs_info(bridge_device, devices_info)}
```
tyll: This seems like it should use `CONFIG_SUBTREE` instead of `'bridge'`.

EdDev: Done

tests/integration/nm/ovs_test.py
tyll: `x if a else y` by itself is not always a problem, however this code is actually:

`nm.ovs.get_bridge_info(nmdev, [(dev, nm.device.get_device_common_info(dev))
                    for dev in nm.device.list_devices()]) if nmdev else {}` because `devices_info` is only used if `nmdev` is considered to be true. Given that `nm.ovs.get_bridge_info()` is just a one-line function to wrap this in a dictionary it seems to be very inconsistent to me to try to put too much complexity in as little lines as possible here. Also I do not agree that more lines of code are a bad thing if the same code is just compressed into less lines.
I guess this code smells again because the absence of a class which seems to require to generate and pass the complex `devices_info` through a lot of methods. Looking more at it, AFAICS `devices_info` does not access any variable in the current scope which makes it even more strange that it actually needs to be generated at this location. `nm.ovs.get_bridge_info()` could just generate the same information by itself.

Nevertheless, I am not going to block the review on this but it does not seemt to be the code quality that you strive for.

PR #269, opened 2019-02-17
automation: Disable pytest live-log

EdDev:
The live-log option creates a lot of noise while running the
integration tests with no special added value.

Only when a failure occurs, the logs need to be examined.


tyll:
This change does not do anything to enable the logs on a failure it just disables them.  I do not really follow what the problem is. Why does the log hurt if there is no failure? I do not look into the logs in this case and then I am not troubled. Is it only hurting you, when you run it locally? How about to enable these flags then only in the CI?

EdDev:
I do not see where they are useful. If you have the scenario, please share it.
Anything that is logged to stdout/stderr is shown if a specific test fails.

It hurts locally and if I need to check the CI logs. The output if full of relevant information, making it hard to locate the things that are important.

EdDev:
ci test please

PR #284, opened 2019-02-27
integ tests: Added bond context manager

SvetaHBar:


tests/integration/bond_test.py
```diff
@@ -55,6 +56,38 @@ def setup_remove_bond99():
     netapplier.apply(remove_bond)
 
 
+@contextmanager
+def bond_interface(bond_name, slaves_name):
```
tyll: IMHO the first argument could be just `name` and the second one should be just `slaves`. Since the slaves are a list, the name suffix does not make sense there.

tyll: And the slaves could be a keyword-argument that defaults to None and is set to an empty list if it is None - or maybe an empty tuple works as well instead of None.

SvetaHBar: Why slaves could be a keyword argument with default value? I think in our case better to leave it as required value, w/o default option

tests/integration/bond_test.py
```diff
@@ -171,84 +204,44 @@ def test_rollback_for_bond(eth1_up, eth2_up):
...
-        }
-
-    netapplier.apply(bond_state)
+def test_add_slave_to_bond_without_slaves(eth1_up):
 
-    bond_state[INTERFACES][0]['link-aggregation']['slaves'] = ['eth1']
+    with bond_interface("bond99", []):
 
-    netapplier.apply(bond_state)
+        bond_current = statelib.show_only(('bond99',))
```
EdDev: You can just edit the dict that the contextmanager returns:
```
with bond_interface(name='bond99', slaves=[]) as bond_state:
    bond_state[INTERFACES][0]['link-aggregation']['slaves'] = ['eth1']
    ...
```

tests/integration/bond_test.py
```diff
@@ -171,84 +204,44 @@ def test_rollback_for_bond(eth1_up, eth2_up):
...
-                }
-
-            ]
-        }
-
-    netapplier.apply(bond_state)
+def test_add_slave_to_bond_without_slaves(eth1_up):
 
-    bond_state[INTERFACES][0]['link-aggregation']['slaves'] = ['eth1']
+    with bond_interface("bond99", []):
```
EdDev: Please use single quotes.

tests/integration/bond_test.py
tyll: The advantage of a keyword argument is that it gives an example for the value. Also it is valid to not specify any slaves by specifying a list, therefore it makes IMHO sense to make it optional.

SvetaHBar: I think we should not allow a bond creation with slaves default = None( except one specific test), almost in all cases bond created with slaves .

tests/integration/bond_test.py
```diff
@@ -55,6 +56,38 @@ def setup_remove_bond99():
     netapplier.apply(remove_bond)
 
 
+@contextmanager
+def bond_interface(name, slaves=[]):
```
EdDev: 1. I was under the impression that you do not want to have defaults for slaves.
2. It is not allowed to have mutable defaults, you can only use `None` or a tuple for that.

I am in favour of having no default.

tests/integration/bond_test.py
```diff
@@ -171,112 +204,54 @@ def test_rollback_for_bond(eth1_up, eth2_up):
...
-                        'slaves': []
-                    },
-                }
-
-            ]
-        }
+def test_add_slave_to_bond_without_slaves(eth1_up):
 
-    netapplier.apply(bond_state)
+    with bond_interface('bond99', []) as bond_state:
```
EdDev: If I'm not mistaken, Till asked to use kwargs when calling the contextmanager:
`with bond_interface(name='bond99', slaves=[]) as bond_state:`

tests/integration/bond_test.py
SvetaHBar: Test without cleanup now using context manager as well.
Test which using 'setup_remove_bond99' will be fixed in other patch ( bond dict includes ip and options)

SvetaHBar: Default was added by mistake .

EdDev:
ci test please

EdDev:
ci test please

PR #286, opened 2019-03-01
Nm connection refactoring

EdDev:


libnmstate/nm/connection.py
```diff
@@ -21,26 +21,67 @@
...
+
+    def import_by_id(self, con_id):
+        self._con_profile = None
+        if con_id:
+            self._con_profile = self._nmclient.get_connection_by_id(con_id)
+
+    def update(self, profile):
+        self._con_profile.replace_settings_from_connection(profile)
+
+    def add(self, save_to_disk=True):
```
tyll: IMHO in the end this should be a method of an NM class (or maybe NMClient), since it is about adding the profile object to NM.

libnmstate/nm/connection.py
```diff
@@ -21,26 +21,67 @@
...
+        self._mainloop.push_action(
+            self._nmclient.add_connection_async,
+            self._con_profile,
+            save_to_disk,
+            self._mainloop.cancellable,
+            _add_connection_callback,
+            user_data,
+        )
+
+    def commit(self, save_to_disk=True, nmdev=None):
```
tyll: This belongs into an NM class as well IMHO.

libnmstate/nm/connection.py
```diff
@@ -21,26 +21,67 @@
...
+        ac = get_device_active_connection(nmdev)
+        if ac:
+            self._con_profile = ac.props.connection
+
+    def import_by_id(self, con_id):
+        self._con_profile = None
+        if con_id:
+            self._con_profile = self._nmclient.get_connection_by_id(con_id)
+
+    def update(self, profile):
```
tyll: This should take a ConnectionProfile object. Otherwise the abstraction is broken.

libnmstate/nm/connection.py
```diff
@@ -21,26 +21,67 @@
 from . import nmclient
 
 
-def create_profile(settings):
-    con_profile = nmclient.NM.SimpleConnection.new()
-    for setting in settings:
-        con_profile.add_setting(setting)
+class ConnectionProfile(object):
+
+    def __init__(self, profile=None):
```
tyll: IMHO the options to create a ConnectionProfile should be expressed in the constructor and not in extra create or import methods. When creating a ConnectionProfile with profile=None (the default), then commit() and update() and add() will most likely not work or in other words, the object is not properly initialized.

EdDev: The object can be in a state that some of its actions are not valid.
I could convert this constructor argument to an import_by_profile if you prefer.

Expressing how the object should be created through the arguments is ugly, especially when the combination of the input may or may not be valid due to collisions. I think that explicitly describing the steps through methods is much better and clearer. it also allows the object to expand in a nicer manner (through additional actions on it).

libnmstate/nm/connection.py
EdDev: I'm not convinced there is an abstraction problem here.
I am ok with your suggestion, but it depends if it is easy to use at the caller.
I mean, if we allow to input in the constructor the profile, then it is easy, otherwise it will be a lot of steps for the caller.

libnmstate/nm/connection.py
EdDev: I think it is valid to say something like: A profile is adding itself to the system.
There is no one correct solution, both may be correct, but one is easier to use then the other.

libnmstate/nm/connection.py
```diff
@@ -21,85 +21,120 @@
...
-
-    return con_profile
+class ConnectionProfile(object):
+
+    def __init__(self, profile=None):
+        self._con_profile = profile
+        self._nmclient = nmclient.client()
+        self._mainloop = nmclient.mainloop()
+
+    def create(self, settings):
```
EdDev: `setting` input should be changed in following PR/s to a base Setting object.
(As the change is incremental, only after all setting type will use it, we can change it here)

libnmstate/nm/connection.py
EdDev: Usage concurs with you for now, changing this.

tyll:
This block #287

libnmstate/nm/connection.py
```diff
@@ -137,36 +137,83 @@ def commit_profile(connection_profile, save_to_disk=True, nmdev=None):
...
+        con_setting.props.id = con_name
+        con_setting.props.interface_name = iface_name
+        con_setting.props.uuid = str(uuid.uuid4())
+        con_setting.props.type = iface_type
+        con_setting.props.autoconnect = True
+        con_setting.props.autoconnect_slaves = (
+            nmclient.NM.SettingConnectionAutoconnectSlaves.YES)
+
+        self._setting = con_setting
+        self._logging_connection_info('create_setting')
```
tyll: the logging message needs to be adjusted

libnmstate/nm/connection.py
```diff
@@ -137,36 +137,83 @@ def commit_profile(connection_profile, save_to_disk=True, nmdev=None):
...
+        new = nmclient.NM.SettingConnection.new()
+        new.props.id = base.props.id
+        new.props.interface_name = base.props.interface_name
+        new.props.uuid = base.props.uuid
+        new.props.type = base.props.type
+        new.props.autoconnect = True
+        new.props.autoconnect_slaves = base.props.autoconnect_slaves
+
+        self._setting = new
+        self._logging_connection_info('duplicate_settings')
```
tyll: the logging message is outdated

libnmstate/nm/connection.py
```diff
@@ -137,36 +137,83 @@ def commit_profile(connection_profile, save_to_disk=True, nmdev=None):
...
+        con_setting.props.uuid = str(uuid.uuid4())
+        con_setting.props.type = iface_type
+        con_setting.props.autoconnect = True
+        con_setting.props.autoconnect_slaves = (
+            nmclient.NM.SettingConnectionAutoconnectSlaves.YES)
+
+        self._setting = con_setting
+        self._logging_connection_info('create_setting')
+
+    def import_by_profile(self, con_profile):
```
tyll: IMHO this method should take a ConnectionProfile object. Also I still think that the object should be fully initialized in the constructor.

libnmstate/nm/applier.py
```diff
@@ -54,16 +54,17 @@ def prepare_new_ifaces_configuration(ifaces_desired_state):
 
 def edit_existing_ifaces(con_profiles):
     for connection_profile in con_profiles:
-        devname = connection_profile.get_interface_name()
+        devname = connection_profile.profile.get_interface_name()
```
tyll: The `interface_name` should be a property of the ConnectionProfile object

libnmstate/nm/applier.py
```diff
@@ -146,7 +148,7 @@ def set_ifaces_admin_state(ifaces_desired_state, con_profiles=()):
 def _get_new_ifaces(con_profiles):
     ifaces_without_device = set()
     for con_profile in con_profiles:
-        ifname = con_profile.get_interface_name()
+        ifname = con_profile.profile.get_interface_name()
```
tyll: The `interface_name` should be a property of the ConnectionProfile object.

libnmstate/nm/bond.py
```diff
@@ -45,9 +45,10 @@ def get_bond_info(nm_device):
 
 
 def get_options(nm_device):
-    con = connection.get_device_connection(nm_device)
-    if con:
-        bond_settings = con.get_setting_bond()
+    con = connection.ConnectionProfile()
+    con.import_by_device(nm_device)
+    if con.profile:
+        bond_settings = con.profile.get_setting_bond()
```
tyll: In the future, the ConnectionProfile object should have a property with a BondSetting object that containsthe bond_settings and that should be used here.

libnmstate/nm/bridge.py
```diff
@@ -128,9 +128,10 @@ def get_slaves(nm_device):
 
 def _get_bridge_setting(nmdev):
     bridge_setting = None
-    bridge_profile = connection.get_device_connection(nmdev)
-    if bridge_profile:
-        bridge_setting = bridge_profile.get_setting_bridge()
+    bridge_con_profile = connection.ConnectionProfile()
+    bridge_con_profile.import_by_device(nmdev)
+    if bridge_con_profile.profile:
+        bridge_setting = bridge_con_profile.profile.get_setting_bridge()
```
tyll: In the future, the ConnectionProfile object should have a property with a BridgeSetting object that contains the bridge_settings and that should be used here.

libnmstate/nm/connection.py
```diff
@@ -137,36 +137,83 @@ def commit_profile(connection_profile, save_to_disk=True, nmdev=None):
     con_profile.commit(save_to_disk, nmdev)
 
 
+class ConnectionSetting(object):
+
+    def __init__(self, con_setting=None):
```
tyll:
```
if not con_setting:
    con_setting = nmclient.NM.SettingConnection.new()
    con_setting.props.uuid = str(uuid.uuid4())
    con_setting.props.autoconnect = True
    con_setting.props.autoconnect_slaves = (
￼	            nmclient.NM.SettingConnectionAutoconnectSlaves.YES)

# name, interface, type can be attributes to be set afterwards or we require either these values in init or con_setting

```

libnmstate/nm/connection.py
```diff
@@ -137,36 +137,83 @@ def commit_profile(connection_profile, save_to_disk=True, nmdev=None):
...
+        con_setting.props.autoconnect = True
+        con_setting.props.autoconnect_slaves = (
+            nmclient.NM.SettingConnectionAutoconnectSlaves.YES)
+
+        self._setting = con_setting
+        self._logging_connection_info('create_setting')
+
+    def import_by_profile(self, con_profile):
+        base = con_profile.get_setting_connection()
+        new = nmclient.NM.SettingConnection.new()
```
tyll: This could be removed if __init__ always creates a SettingConnection object

libnmstate/nm/connection.py
```diff
@@ -137,36 +137,83 @@ def commit_profile(connection_profile, save_to_disk=True, nmdev=None):
...
+        new.props.id = base.props.id
+        new.props.interface_name = base.props.interface_name
+        new.props.uuid = base.props.uuid
+        new.props.type = base.props.type
+        new.props.autoconnect = True
```
tyll: This can be removed if this is set in `__ini__` IMHO. It is also not nice to hardcode this in multiple places.

libnmstate/nm/applier.py

libnmstate/nm/bond.py
EdDev: Yes, it should have a list of Setting objects, with concrete instances (like BondSetting).

libnmstate/nm/connection.py
EdDev: This object represents a connection setting.
It may serve creating:
- A new one from scratch (`create`).
- Objectizing an existing setting, by getting the settings in the constructor. Consider this a copy constructor.
- Load/copy setting from the system/NM through a given profile.

Your suggestion supports partially options 1 (due to the missing args), fully option 2 and not option 3.
To support all, we will need to input it all, and then decide on an order or rules on how to read the input (to guess about what the caller intended).
If we are explicitly placing these options through different methods we make the api cleaner, and allow the methods/constructor focusing only on one thing, and not two.
It is also open to extensions (in case we will need to load the settings from other sources).

libnmstate/nm/connection.py
EdDev: You mean changing it to `create`?

libnmstate/nm/connection.py
EdDev: I agree with the input type of ConnectionProfile. Changing.

libnmstate/nm/connection.py
EdDev: If the constructor is doing everything, you are right. But as explained earlier, it is now split per explicit action.

libnmstate/nm/connection.py
EdDev: It is expressing what we do today: Hard coding it, overwriting anything we touch to have it enabled.
When the need will arise we will change it.

libnmstate/nm/connection.py
EdDev: I guess you mean the method name, right?

libnmstate/nm/connection.py
tyll: maybe `ConnectionSetting.create() done`

libnmstate/nm/connection.py
tyll: yes

libnmstate/nm/connection.py
tyll: The current way is uglier than having multiple arguments in the constructor and it is also misleading that a constructor does not create the actual object but that there is a separate create method. @cathay4t, what do you think?

IMHO the other constructors should be at least class methods that return a proper object. But actually create() is IMHO just the constructor and the import_by methods actually belong in other classes so that a Device object returns a ConnectionProfile or the client can be used to look up a ConnectionProfile based on the ID.

libnmstate/nm/connection.py
tyll: The current solution requires the class to know about the main loop which it would not if only the client runs the main loop if necessary.

libnmstate/nm/connection.py
```diff
@@ -21,157 +21,171 @@
...
+            user_data,
+        )
+
+    @property
+    def profile(self):
+        return self._con_profile
+
+    @property
+    def devname(self):
+        if self._con_profile:
```
tyll: This if is a result of the object not being properly initialized after it was created and could go if this was not the case.

libnmstate/nm/connection.py
```diff
@@ -21,157 +21,171 @@
...
+        self._mainloop.push_action(
+            self._con_profile.commit_changes_async,
+            save_to_disk,
+            self._mainloop.cancellable,
+            self._commit_changes_callback,
+            user_data,
+        )
+
+    @property
+    def profile(self):
```
tyll: Why make the profile first internal and then provide it to the outside nonetheless?

libnmstate/nm/connection.py
EdDev: It is a common practice to create "unloaded" objects and it allows better maintenance and extendibility options.
At the minimum, it is certainly not wrong.

A constructor with unused arguments is wrong.
Functions with unused arguments (in an OR relationship) are also wrong, unfortunately we do have them as well (although I consider constructor more important in this regard).

This PR is proposing one way of abstracting things. I am sure there are other ways to do it.
Up to now I was not convinced that this approach is not good or that it cannot be a base for future refactoring, but if you have something better, please suggest it through a PR so we can discuss it (can be a new PR or over this one).
(except of not agreeing on the approach, I also do not see how this current work can be adjusted to fit with your proposals)

libnmstate/nm/connection.py
EdDev: The same can be said for the function/method that runs it... It either needs to know about the client or about the object. One of them needs to know about the client.
Both approaches are valid IMO, and if one is better then the other, it can be changed with a proper explanation on why it is better.

libnmstate/nm/connection.py
EdDev: Because it is part of the object state and it is currently needed by others outside the object.
As an example, the profile is loaded as part of the object initialization (import or create).

libnmstate/nm/connection.py
EdDev: There are consequences to the approach chosen, that does not make something better or worse.
The disadvantage of the constructor you are suggestion is stronger from my point of view.

libnmstate/nm/connection.py
EdDev: OK

libnmstate/nm/connection.py
tyll: Having default values for arguments including constructors is best practice in Python and commonly used, see for example: https://docs.python.org/3.4/library/subprocess.html#subprocess.Popen

EdDev: These are not defaults, we are talking about multiple arguments which sometime collide between themself.
I do not think that having a constructor with tens of arguments and with potentially new introduced ones in the future is a good thing.

EdDev:
I'm summarizing the discussion on the PR regarding the ConnectionProfile and ConnectionSetting classes instantiation and initialization to organize it a bit.

The PR proposed to separate the initialization of the object from it instantiation, due to the many forms and sources of the initialization stage. It should be noted that these objects are a kind of proxy to the libnm objects.
It looks like this:
```
o = O()  # Instantiated only object
o = O(other_o.libnm_object)  # Instantiated and initialized based on an existing libnm object.
o.create(new_settings) # Init it based on new settings
o.import_by_id(id) # Init it based on an ID, loading the libnm object from the system based on the id.
o.import_by_dev(dev) # Init it based on a device, loading the libnm object from the system based on the device.
```

If I understand the main concern from @tyll about this approach, it is the intermediate state of the object between instantiation and initialization.
I do not think it is a problem, it is just a definition of how this object needs to be used.

The problem with putting everything in the constructor is: the duplication and mess required in the constructor, and its required adjustment whenever something new needs to be added (affecting the constructor signature and the logic in it, which in turn affects all users of it, instead of only the new parts).
There are other reasons of why to use later initialization, but they are less relevant now.

Unifying it all under the constructor will look something like this:
(This is actually fixable in a language that supports overloading the constructor, but unfortunately it cannot be done in python).
```
class O():
  def __init__(libnm_obj=None, new_settings=None, id=None, device=None):
    if libnm_obj is not None:
      ...
    elif new_settings is not None:
      ...
    elif id is not None:
      ...
    elif device is not None:
      ...
    else:
      raise Exception()
```
If two parameters are used by mistake requires more validation in the constructor.

Now, although I think it is perfectly fine to use the original approach, especially in our case here, I can suggest an option to use a builder to do this, not exposing the non initialized object but still preserve the cleanless:
```
class O():
  def __init__(libnm_obj=None):
    ...
  @staticmethod
  def create(new_settings):
    o = O()
    o._create(new_settings)
    return o
  ...
```

libnmstate/nm/connection.py
tyll: Actually I believe that the code to get the profile from a device or by UUID should be moved to a device or client class. Since having the current objects is an improvement over not having them, I am also fine with doing a less strict review and revisit this later.

libnmstate/nm/applier.py
```diff
@@ -48,16 +48,17 @@ def prepare_new_ifaces_configuration(ifaces_desired_state):
 
 def edit_existing_ifaces(con_profiles):
     for connection_profile in con_profiles:
-        devname = connection_profile.get_interface_name()
+        devname = connection_profile.devname
```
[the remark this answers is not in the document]
EdDev: > It is hard to understand why this code makes sense - if it does. AFAIU

We are talking about a syntactic change of objectizing the connection entity.
Nothing should have changed logically, your comment on this are just, but should be addressed through either a separate PR (refactoring) or through a ticket.

> this indicates that the connection profile object should return a device object IMHO

That creates a dependency which is not needed between the two (connection and device).
Anyway, that is just an addition which can be added later when we will have a device object wrapper.

> I do not really get the idea behind this or why this complicated flow is needed.

This is just an ugly way to check if the the profile in question is of a new device/profile or of an existing one.
It is a subsequence of using two stages (the first being `prepare_edited_ifaces_configuration`).

Hopefully this can be fixed and refactor in the near future, probably as apart of merging the two mainloop runs as we already discussed in the past.

libnmstate/nm/connection.py
EdDev: I'm pretty sure the objects proposed here and in the next PR (for the device) will continue evolving.
This PR and refactoring in general is a never ending task and many times only after reorganizing stuff you see that things still need to move around.

I consider usage to dictate eventually what is better suited. Usage is usually what makes us refactor things in the first place.

libnmstate/nm/applier.py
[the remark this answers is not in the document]
tyll: > This is just an ugly way to check if the the profile in question is of a new device/profile or of an existing one.
> It is a subsequence of using two stages (the first being `prepare_edited_ifaces_configuration`).

I see. This could also solved inside the object so that it knows whether or not it is new or changed and calling ConnectionProfile.commit() will just do the right thing without this complicated check.

tyll:
> I'm summarizing the discussion on the PR regarding the ConnectionProfile and ConnectionSetting classes instantiation and initialization to organize it a bit.
> 
> The PR proposed to separate the initialization of the object from it instantiation, due to the many forms and sources of the initialization stage. It should be noted that these objects are a kind of proxy to the libnm objects.
> It looks like this:
> 
> ```
> o = O()  # Instantiated only object
> o = O(other_o.libnm_object)  # Instantiated and initialized based on an existing libnm object.
> o.create(new_settings) # Init it based on new settings
> o.import_by_id(id) # Init it based on an ID, loading the libnm object from the system based on the id.
> o.import_by_dev(dev) # Init it based on a device, loading the libnm object from the system based on the device.
> ```
> If I understand the main concern from @tyll about this approach, it is the intermediate state of the object between instantiation and initialization.

IMHO this use case is unnatural, therefore I would prefer something like this:

```
class ConnectionProfile(object):
    def __init__(self, profile=None):
        if not profile:
            profile = nmclient.NM.SimpleConnection.new()

        self._con_profile = profile  #  IMHO it could also make sense to allow profile to be a UUID instead of the SimpleConnection object and then fetch the corresponding SimpleConnection from NM

    def add_settings(self, settings):
        for setting in settings:
            self._con_profile.add_setting(setting)

class NMClient(object):
...
    def get_profile(con_id):
        connection_profile = None
        nm_profile = nmclient.get_connection_by_id(con_id)
        if nm_profile:
            connection_profile = ConnectionProfile(nm_profile)


class NMDevice(object):
....
    def get_active_profile(self):
        connection_profile = None
        ac = get_device_active_connection(nmdev)
        if ac:
            profile = ac.props.connection
            connection_profile = ConnectionProfile(profile=profile)

        return connection_profile
```

EdDev:
> IMHO this use case is unnatural, therefore I would prefer something like this:

I have no idea to what `unnatural` case you refer to but I have provided in my examples the use cases for the usage itself. You can look for `ConnectionProfile` and see that all initialization options are used.

Please provide actual usage alternatives from the code that uses the new class.

The proposal for use an NMClient class is not something I think is a good thing at the moment, but you are welcome to propose it as a following refactor step and we can examine its usefulness.

> IMHO it could also make sense to allow profile to be a UUID instead of the SimpleConnection object and then fetch the corresponding SimpleConnection from NM

That will require checking the type of the input argument and will make even worse the conditions I previously explained that should not exist in the constractor.

It will be more useful to answer the specific usage cases presented in the summary given, which relate directly to how the existing code uses the connection profiles. And to propose a solution to the points I presented as problematic when doing it all in the constructor.

libnmstate/nm/connection.py
```diff
@@ -21,157 +21,171 @@
...
+    def set_master(self, master, slave_type):
+        if master is not None:
+            self._setting.props.master = master
+            self._setting.props.slave_type = slave_type
+
+    @property
+    def setting(self):
+        return self._setting
+
+    def _logging_connection_info(self, source):
```
tyll: This should be `_log_connection_info()`

tyll:
> > IMHO this use case is unnatural, therefore I would prefer something like this:
> 
> I have no idea to what `unnatural` case you refer to but I have provided in my examples the use cases for the usage itself. You can look for `ConnectionProfile` and see that all initialization options are used.

For someone not used to Nmstate, the object is special compared to other objects in Python IMHO.

 
> Please provide actual usage alternatives from the code that uses the new class.

Code from this PR:
```
    con = connection.ConnectionProfile()
    con.import_by_device(nm_device)
    if con.profile:
        bond_settings = con.profile.get_setting_bond()
```
Code with my class:

```
connection_profile = nm_device.get_active_profile()
if connection_profile:
    bond_settings = connection_profile.get_bond_settings()
```

Actually I realized that `get_active_profile()` should raise an exception, so the code would be:

```
try:
    connection_profile = nm_device.get_active_profile()
    bond_settings = connection_profile.get_bond_settings()
except KeyError:  # or some better matching exception:
    # do what needs to be done if there are no bond_settings
```

This will not initialize the object a little bit, then maybe intialize it more and does not need to check again if it is really initialized.

> 
> The proposal for use an NMClient class is not something I think is a good thing at the moment, but you are welcome to propose it as a following refactor step and we can examine its usefulness.
> 
> > IMHO it could also make sense to allow profile to be a UUID instead of the SimpleConnection object and then fetch the corresponding SimpleConnection from NM
> 
> That will require checking the type of the input argument and will make even worse the conditions I previously explained that should not exist in the constractor.

The arguments did not really convince me and I forgot them since this is AFAIK how to achieve overloading of methods in Python. However, it would also fine by me to make it a uuid parameter and have a second constructor as a classmethod that accepts a raw profile (IMHO this is a more internal way to create a wrapper object).

libnmstate/nm/connection.py
EdDev: Done

EdDev:
ci test please

PR #278, opened 2019-02-20
OVS internal interface: nmstate core

EdDev:
Depends on #277 , #271

PR #289, opened 2019-03-05
examples: Rename the yaml files

EdDev:
Fixing the Linux bridge name and changing the name format to be more
generic and dropping the interface names.

By removing the interface names, it provides the possibility to expand
the example.
In case a new example will require differentiation, the names can be
adjusted.


EdDev:
Note that the ovs bridge example has been changed in #278 (https://github.com/nmstate/nmstate/pull/278/commits/12ec6134cc410ead0be60ec0d99e7b775a90dd13).

tyll:
IMHO the interface names in the files are useful to easily see if they can be applied without a problem. If they state eth1 and one does not care about eth1, then they are good to go. Otherwise one needs to look into the yaml files to figure out what device they would change.

Also the words like setup, teardown make sense in the testing world but the examples are meant for users. for them up/down makes more sense - probably delete makes more sense than absent.

EdDev:
- `setup`/`teardown`: Ok, I will use `create` and `delete`.
- The reason `eth1` is not that good has been seen when adding another port to a bridge. I think, we can be fine with just starting giving general examples for the user to look into. If we want to give a complex example, then we can name it nicely with description of what it represents and not what it exactly contains.

tyll:
create and delete sound good, too. About eth1 ... if there should be multiple ports, we can also name all of them in the filename IMHO.

EdDev:
> if there should be multiple ports, we can also name all of them in the filename IMHO

I do not see the benefit of it. The filename needs to give the context or story subject, not these details. If we will have a vlan or a bond or whatever other complex scenario there may be, we cannot add all of them in the name.

EdDev:
ci test please

EdDev:
ci test please

tyll:
IMHO it would be more natural to write `create linuxbridge` than `linuxbridge create`. Also `up_create` etc as a name are strange.

EdDev:
> IMHO it would be more natural to write create linuxbridge than linuxbridge create. Also up_create etc as a name are strange.

I do not have strong feeling about this, but IMO it is more convenient to sort the examples per subject (device type) and not per action (create, delete). Looking for `linuxbridge` examples is more common than looking for `create` examples.

tyll:
> > IMHO it would be more natural to write create linuxbridge than linuxbridge create. Also up_create etc as a name are strange.
> 
> I do not have strong feeling about this, but IMO it is more convenient to sort the examples per subject (device type) and not per action (create, delete). Looking for `linuxbridge` examples is more common than looking for `create` examples.

This was another advantage of using up/down/present/absent because `linuxbrdige present` is the natural order AFAICS.

EdDev:
> This was another advantage of using up/down/present/absent because linuxbrdige present is the natural order AFAICS.

If this is the problem, we can use a noun: `linuxbrdige_creation`.
Using the names and values of the implementation is not explanatory, we would need to add the keys names in addition to the values. Summarizing the example using a subject is more appropriate. It is also more scalable (when multiple interfaces of different types with many options appear in the example).

tyll:
> > This was another advantage of using up/down/present/absent because linuxbrdige present is the natural order AFAICS.
> 
> If this is the problem, we can use a noun: `linuxbrdige_creation`.

Since Nmstate is descriptive, it is not really creation, it is about making it `up`. What was the original reason why up/down/absent are bad?

EdDev:
> Since Nmstate is descriptive, it is not really creation, it is about making it up. What was the original reason why up/down/absent are bad?

The fact that nmstate is descriptive has nothing to do with the example being all about the creation of an interface or some other operation.
The problem with the values of the state is about the inability to express in the name the setup parameters. This is especially true when we have 3 interfaces, some are ports, some may be slaves, some examples have IP/s, some have an MTU, etc.
I prefer not to see something like: `eth1_up_vlan101_bond0...`.

tyll:
> > Since Nmstate is descriptive, it is not really creation, it is about making it up. What was the original reason why up/down/absent are bad?
> 
> The fact that nmstate is descriptive has nothing to do with the example being all about the creation of an interface or some other operation.

Since it is declarative (this is what I meant), it is not about an operation but about a state. And the example will not create the linux bridge if it already exists - it ensures that it is up. Therefore the name of an action is misleading.

> The problem with the values of the state is about the inability to express in the name the setup parameters. This is especially true when we have 3 interfaces, some are ports, some may be slaves, some examples have IP/s, some have an MTU, etc.
> I prefer not to see something like: `eth1_up_vlan101_bond0...`.

So if it is not possible to express the names in a complex scenarion, we can not do it. But this does not require to make the simple examples harder to understand as well.

EdDev:
Are you claiming that “linuxbridge creation” is more complex than “linuxbridge_br0_eth1_eth2_up”?

Please suggest something that is not locked to details and still describes the content. Please keep it open to add other details and interfaces (including ports and slaves).

tyll:
> Are you claiming that “linuxbridge creation” is more complex than “linuxbridge_br0_eth1_eth2_up”?

No, I am saying that the examples should include as much details as possible. And if the example is to complex, I am also fine with names that do not contain all the details.

So for example

`vlan101_eth1.yml` is a perfect name for an example that creates VLAN 101 on eth1 IMHO.

> Please suggest something that is not locked to details and still describes the content. Please keep it open to add other details and interfaces (including ports and slaves).

I think we can remove the `_up` suffix from all the examples so we also do not need to invent another suffix like creation. If it is an example for a device I think everyone would assume that it is not about removing the device.

EdDev:
If we name the creation using `vlan101_eth1.yml`, then how do we name the `down` and `absent` states?
What will happen when we will want to add another vlan on the same interface? Or two vlans with the same ID on different interfaces?

I would prefer not to change the name frequently.
How about this(?):
- linuxbridge_{up,absent}
- ovsbridge_{up,absent}
- vlan_{up,down,absent}

tyll:
> If we name the creation using `vlan101_eth1.yml`, then how do we name the `down` and `absent` states?

Dow we need both down and absent examples? It seems to be a little bit excessive since down can be easily created from the "up" or "absent" example, depending on whether the device is already there.

The absent example could be called `vlan101_eth1_absent.yml`.

> What will happen when we will want to add another vlan on the same interface? Or two vlans with the same ID on different interfaces?

If we want to add more complex examples as well, then they can be added as new files with appropriate names. For this the name of the current file does not need to be changed.

> I would prefer not to change the name frequently.

IMHO it is valid to change the name if necessary.

> How about this(?):
> 
> * linuxbridge_{up,absent}
> * ovsbridge_{up,absent}
> * vlan_{up,down,absent}

linuxbridge{,_absent} etc. works for me (linuxbridge.yml and linuxbridge_absent.yml). Not sure about also adding a down example for VLANs. Also this will result in me using my own examples in case I need to do demos again because I would like to have names that remind me of their content.

Kind regards
Till

EdDev:
> linuxbridge{,_absent} etc. works for me (linuxbridge.yml and linuxbridge_absent.yml). 

I do not like the asymetric naming. We can use `present` if you prefer.

> Not sure about also adding a down example for VLANs.

I am ok to drop it.

> Also this will result in me using my own examples in case I need to do demos again because I would like to have names that remind me of their content.

The content you can be sure it contains is too general, these examples should evolve to include as much details as possible. If we will end up with 5 examples for each device type, it will be a mess.

Let me refresh the intent of this PR: The examples files should be flexible for change by adding content and they need to be consistent in the naming format.

tyll:
> > linuxbridge{,_absent} etc. works for me (linuxbridge.yml and linuxbridge_absent.yml).
> 
> I do not like the asymetric naming. We can use `present` if you prefer.

If you insist on `present`, I do not care that much. But it is not necessary IMHO. It would be strange to assume that an example about a linuxbridge shows how to remove it or to take it down, therefore adding `present` is redundant.

> > Also this will result in me using my own examples in case I need to do demos again because I would like to have names that remind me of their content.
> 
> The content you can be sure it contains is too general, these examples should evolve to include as much details as possible. If we will end up with 5 examples for each device type, it will be a mess.

I disagree. Smaller and simpler examples make it easier for users to take a look and try things out because they are not too overwhelming.

> Let me refresh the intent of this PR: The examples files should be flexible for change by adding content and they need to be consistent in the naming format.

IMHO these are requirements that only make things complicated and provide no clear gain. We as developers should be flexible and accept change when necessary. Optimizing now for unclear/not agreed upon future changes wastes time that we can better spent when the actual changes are there.

EdDev:
My attempt to have consistent names and use the existing examples to add to it content in the future failed.
Personally I think (and this discussion emphasised it for me) that examples belong to the documentation and tests, not as individual example files. 

I do agree that having simple examples is much better than having overloaded ones, something that makes the whole intent here irrelevant.

tyll:
@EdDev do I understand correctly that we can close this PR? I still think that examples files are good, though.

PR #293, opened 2019-03-10
Fix nm ovs tests by deleting dev

EdDev:
The nm/ integration tests have been leaving the OVS* devices as leftovers.
Fixing it by explicitly deleting these virt devices (NM bug) in the same way they are deleted in the production code.


EdDev:
ci test please

EdDev:
ci test please

libnmstate/nm/device.py
```diff
@@ -378,6 +375,10 @@ def _delete_device_callback(src_object, result, user_data):
         if mainloop.is_action_canceled(e):
             logging.debug('Device deletion aborted on %s: error=%s',
                           nmdev.get_iface(), e)
+        elif 'No such interface' in str(e):
```
cathay4t: How about:
```python
exept GLib.GError as e:
    if e.domain == 'g-dbus-error-quark' and e.code == 19:
...
```

EdDev: It makes sense, but we have a few problems popping up:
- pylint does not like the GLib.GError attributes (domain, code).
- It is unclear to me why the code is errno.ENODEV related. I could not find anything that specify that the error code has anything to do with the codes from errno.

The first point can be handled (hopefully) as we did in `is_action_canceled()`.

bengal: I think the correct way to test for an unregistered device is to check for the unknown-method error code of gdbus:

``
if e.matches(Gio.DBusError.quark(), Gio.DBusError.UNKNOWN_METHOD):
``

cathay4t: `errno.ENODEV` is by my guess. Clearly @bengal is right on the `Gio.DBusError.UNKNOWN_METHOD`.

EdDev: Done

EdDev:
ci test please

libnmstate/nm/device.py
```diff
@@ -374,6 +371,15 @@ def _delete_device_callback(src_object, result, user_data):
...
         success = src_object.delete_finish(result)
+    # pylint: disable=catching-non-exception
+    except nmclient.GLib.GError as e:
+        # pylint: enable=catching-non-exception
+        if e.matches(nmclient.Gio.DBusError.quark(),
+                     nmclient.Gio.DBusError.UNKNOWN_METHOD):
+            logging.debug('Device %s has been already deleted: error=%s',
+                          nmdev.get_iface(), e)
+            mainloop.execute_next_action()
+            return
```
cathay4t: Please do logging or exception for other `DBusError` here.

PR #298, opened 2019-03-13
automation: Container image for CentOS7 unit tests

EdDev:
The unit tests with the lint check can be executed locally
as follow: sudo ./automation/run-local-utests.sh

### DIFF
--- a/automation/run-local-utests.sh
+++ b/automation/run-local-utests.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+PROJECT=${PROJECT:-${PWD##*/}}
+TOXWORKDIR="$HOME/.cache/.tox-$PROJECT"
+CONTAINER_IMAGE="localhost/$PROJECT/centos7-$PROJECT-utest"
+
+mkdir -p "$TOXWORKDIR"
+
+podman run \
+       --rm \
+       -ti \
+       -v $PWD:/workspace/$PROJECT:Z \
+       -v $TOXWORKDIR:$TOXWORKDIR:Z \
+       $CONTAINER_IMAGE \
+       /bin/bash -c " \
+         cp -rf /workspace/$PROJECT /tmp/ \
+         && \
+         cd /tmp/$PROJECT \
+         && \
+         tox --workdir $TOXWORKDIR"
--- a/packaging/Dockerfile.centos7-nmstate-utest
+++ b/packaging/Dockerfile.centos7-nmstate-utest
@@ -1,0 +1,25 @@
+# This image definition is defining
+# a container image on which unit and lint tests
+# can be run locally through `tox`.
+
+FROM centos:7
+
+RUN yum -y upgrade \
+    && \
+    yum -y install \
+        NetworkManager \
+        NetworkManager-libnm \
+    && \
+    yum -y install epel-release \
+    && \
+    yum -y install \
+        python2-pip \
+        make \
+    && \
+    yum-builddep -y \
+        python-gobject \
+        dbus-python \
+    && \
+    yum clean all \
+    && \
+    pip install tox


EdDev:
I needed this to run locally the unit tests frequently and fast.

automation/run-local-utests.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+TOXWORKDIR="/tmp/.tox-nmstate"
```
tyll: This should be a patch in ~/.cache or a a securely created temporary directory.

automation/run-local-utests.sh
```diff
@@ -0,0 +1,19 @@
...
+
+mkdir -p "$TOXWORKDIR"
+
+docker run \
+       --rm \
+       -ti \
+       -v $PWD:/workspace/nmstate:Z \
+       -v $TOXWORKDIR:$TOXWORKDIR:Z \
+       nmstate/centos7-utest \
```
tyll: IMHO it would be better to not abbreviate unit test as utest.

EdDev: Well, it is not only unit tests, it also needs to cover lints (static analysis).

automation/run-local-utests.sh
EdDev: Done

automation/run-local-utests.sh
tyll: This does not make utests a better name :-P

EdDev: The maker has usually the privilege to name it if it is not completely wrong.

tyll: You said that unit test is wrong, so abbreviating it as utest is wrong as well....

tyll:
Why can the normal CI image not be used for this btw?

EdDev:
> Why can the normal CI image not be used for this btw?

Because I wanted the bare minimum to run these tests as an alternative to running it locally on my machine.
Static analysis and unit tests should run (almost) everywhere, having the minimum protects this.

automation/Dockerfile.centos7-utest
```diff
@@ -0,0 +1,24 @@
...
+    && \
+    yum -y install \
+        NetworkManager \
+        NetworkManager-libnm \
+    && \
+    yum -y install epel-release \
+    && \
+    yum -y install \
+        python2-pip \
+        dbus-devel \
```
EdDev: Using `yum-builddep -y dbus-python` should be better.

EdDev: Done

EdDev:
> Only script for CI or shared by developers should be in code base.

This change is about providing a developer (me) an option to run unit tests and static analysis frequently while developing.
How CI is relevant here I do not understand and regarding the developers, I think I represent some of them in this project.

Please explain what is wrong with this change that merits a nack.

cathay4t:
I simply run `tox` for unit test which covers py2 and py3 which is easier and quicker than docker way, IMHO.
If you think `tox` running on Fedora is not enough for unit test, the CI system should enable CentOS unit test also.

If you can convince @tyll to use this script and sign up for maintaining it, you will get my ack on this.

If this is for your own use only, please keep it in your personal folder.
I have express the same to tyll when he want to add his script using `vagrant` to setup VM for testing.

EdDev:
> I simply run tox for unit test which covers py2 and py3 which is easier and quicker than docker way, IMHO.

This change runs `tox`, but it runs it in a container because the dependencies required to run it are not available on RHEL7 (not I want to install them on my machine).

> If you think tox running on Fedora is not enough for unit test, the CI system should enable CentOS unit test also.

I do not mind running it on on a Fedora image that is limited to only the unit tests run dependencies.
Currently, the images that exist are prepared for integration tests with a lot of dependencies, one of which is systemd and NM services running there.

If you can convince @tyll to use this script and sign up for maintaining it, you will get my ack on this.

I am one of the maintainers in this project and I am the one who will maintain it.

> If this is for your own use only, please keep it in your personal folder.
> I have express the same to tyll when he want to add his script using vagrant to setup VM for testing.

As we are the main contributors to this project, our personal preference can be exposed in the project unless it is totally wrong or has a technical problem.
Any developer who will like to run the unit tests using this dedicated image can do so, it is a matter of choice.
If there is a better and faster solution to what I want to accomplish, I'm will be the first one to take it.

I would like to know if there is a technical reason this solution is wrong.

automation/run-local-utests.sh
```diff
@@ -0,0 +1,19 @@
...
+TOXWORKDIR="$HOME/.cache/.tox-nmstate"
+CONTAINER_IMAGE="nmstate/centos7-nmstate-utest"
+
+mkdir -p "$TOXWORKDIR"
+
+docker run \
+       --rm \
+       -ti \
+       -v $PWD:/workspace/nmstate:Z \
+       -v $TOXWORKDIR:$TOXWORKDIR:Z \
```
tyll: This will create root-owned files in `$HOME/.cache/.tox-nmstate`, will it not? Would you be open to use podman in unprivileged mode to run the unit tests?

Also I think it would be better to integrate the feature of copying the current directory into the other run-tests file and also running this container image.

EdDev: It can be run without sudo if you set the docker group. But using podman sounds good to me.

We have piggy-back these tests on the integration test image. I’m fine to run it instead of the current runs, but I did not want to depend on you guys to agree to that, as it introduces changing existing stuff. That could have been a different proposal.

cathay4t:
> I am one of the maintainers in this project and I am the one who will maintain it.

OK. Then you got me approve. We can merge it once rebased.

> 
> > If this is for your own use only, please keep it in your personal folder.
> > I have express the same to tyll when he want to add his script using vagrant to setup VM for testing.
> 
> As we are the main contributors to this project, our personal preference can be exposed in the project unless it is totally wrong or has a technical problem.
> Any developer who will like to run the unit tests using this dedicated image can do so, it is a matter of choice.
> If there is a better and faster solution to what I want to accomplish, I'm will be the first one to take it.
> 
> I would like to know if there is a technical reason this solution is wrong.

Technically, even running unit tests in docker save us from versioning of pytest/pylint/flake8, but that waste a lot time on waiting tox to installed the nmstate dependency. So personally, I use pylint, flake8
directly(but I never push that interest to this project) which save me a lot time. Anyway, if you sign up on
maintaining it, I am OK with it.

automation/run-local-utests.sh
```diff
@@ -0,0 +1,19 @@
...
+
+TOXWORKDIR="$HOME/.cache/.tox-nmstate"
+CONTAINER_IMAGE="nmstate/centos7-nmstate-utest"
+
+mkdir -p "$TOXWORKDIR"
+
+docker run \
+       --rm \
+       -ti \
+       -v $PWD:/workspace/nmstate:Z \
```
tyll: It would be nicer to make `nmstate` a variable as well (maybe `${project}` and it could default to `${PWD##*/}`), so it can also be used for other projects.

EdDev: Nice, thanks. Done.

EdDev:
@tyll , @cathay4t , please note that in order to maintain correctly the unit tests, running this or probably a Fedora image on the CI is preferred. We should protect against running the unit tests against a running service.

EdDev:
@tyll, @cathay4t, note that I changed it to use *podman* and it works great with a non-root user.

PR #300, opened 2019-03-17
yaml.load: Explicitly set yaml loader to SafeLoader

EdDev:
PyYaml 5.1 has deprecated the usage of yaml.load(input) with the default
loader. It has been accompanied by a message warning.

As the usage of yaml in nmstate is for external input, the loader has
been explicitly set to SafeLoader.

Ref: https://msg.pyyaml.org/load
https://github.com/yaml/pyyaml/commit/0cedb2a0697b2bc49e4f3841b8d4590b6b15657e

### DIFF
--- a/libnmstate/schema.py
+++ b/libnmstate/schema.py
@@ -22,7 +22,9 @@ import yaml
 
 def load(schema_name):
     return yaml.load(
-        pkgutil.get_data('libnmstate', 'schemas/' + schema_name + '.yaml'))
+        pkgutil.get_data('libnmstate', 'schemas/' + schema_name + '.yaml'),
+        Loader=yaml.SafeLoader
+    )
 
 
 ifaces_schema = load('operational-state')
--- a/nmstatectl/nmstatectl.py
+++ b/nmstatectl/nmstatectl.py
@@ -151,7 +151,7 @@ def apply_state(statedata, verify_change):
     if statedata[0] == '{':
         state = json.loads(statedata)
     else:
-        state = yaml.load(statedata)
+        state = yaml.load(statedata, Loader=yaml.SafeLoader)
         use_yaml = True
     netapplier.apply(state, verify_change)
     print('Desired state applied: ')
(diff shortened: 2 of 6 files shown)


EdDev:
ci test please

tyll:
IMHO it would be nicer to use `yaml.safe_load()` instead of specifying the loader.

EdDev:
IMHO it would be nicer to use yaml.safe_load() instead of specifying the loader.

Using such a wrapper can be done internally, but I do not think it is correct to use their wrapper.
Specifying the loader is explicit and opens the ability to replace it in the future or use one for mocking if needed.

I do not think it is worth using the wrapper they provide.

tyll:
> > IMHO it would be nicer to use yaml.safe_load() instead of specifying the loader.
> 
> Using such a wrapper can be done internally, but I do not think it is correct to use their wrapper.
> Specifying the loader is explicit and opens the ability to replace it in the future or use one for mocking if needed.
> 
> I do not think it is worth using the wrapper they provide.

Using `safe_load()` is documented in their documentation and is probably the more common way to use this loader for people using pyyaml. Also I do not see the value in making things complicated now in case it might be useful in the future (regarding mockig) and also as this PR shows, the method can be easily replaced regardless whether it is a loader that is specified or the name of a method.

EdDev:
> the method can be easily replaced regardless whether it is a loader that is specified or the name of a method.

We will lock ourself from creating a new loader with extensions or perhaps using something "safer".
I do not see the benefit of limiting ourself here.
I fail to see the disadvantage in using the loader directly.

tyll:
> > the method can be easily replaced regardless whether it is a loader that is specified or the name of a method.
> 
> We will lock ourself from creating a new loader with extensions or perhaps using something "safer".
> I do not see the benefit of limiting ourself here.

There is no lock out since we can always change it to use a custom loader when needed.

> I fail to see the disadvantage in using the loader directly.

It is a more complicated and uncommon way of doing this. The [PyYAML docs](https://pyyaml.org/wiki/PyYAMLDocumentation) mention the safe_load method as well. However, if you prefer the complicated approach, I do not care so much to keep discussing this.

PR #301, opened 2019-03-17
netapplier refactoring

EdDev:
The `netapplier` module has became too convoluted with too many responsibilities, making it hard to maintain and evolve.

This PR is all about splitting it apart into more focused modules, leaving it as the glue that arranges the flow/steps of an "apply" command.


EdDev:
ci test please

cathay4t:
@EdDev Can you give few comments about what this PR for?

libnmstate/state.py
```diff
@@ -0,0 +1,248 @@
...
+        other_state.sanitize_dynamic_ip()
+
+        self.canonicalize_interfaces(other_state)
+
+        self.normalize_for_verification()
+        other_state.normalize_for_verification()
+
+        self._assert_interfaces_equal(other_state)
+
+    def normalize_for_verification(self):
```
tyll: IMHO it would be nicer if the state property would just report the normalized/canonicalized state instead of requiring to call a special option to trigger the normalization.

libnmstate/state.py
```diff
@@ -0,0 +1,248 @@
...
+        self._clean_sanitize_ethernet()
+        self._sort_lag_slaves()
+        self._sort_bridge_ports()
+        self._canonicalize_ipv6()
+        self._remove_iface_ipv6_link_local_addr()
+        self._sort_ip_addresses()
+
+    def canonicalize_interfaces(self, other_state):
```
tyll: IMHO this should be called `merge()` since canonicalization also means normalization.

libnmstate/state.py
```diff
@@ -0,0 +1,248 @@
...
+                if ip.get('enabled') and (
+                        ip.get('dhcp') or ip.get('autoconf')):
+                    ip['address'] = []
+                else:
+                    for dhcp_option in ('auto-routes',
+                                        'auto-gateway',
+                                        'auto-dns'):
+                        ip.pop(dhcp_option, None)
+
+    def verify_interfaces(self, other_state):
```
tyll: Maybe there should be a verifier class that can then compare two states....

libnmstate/state.py
```diff
@@ -0,0 +1,248 @@
...
+            key=itemgetter('name')
+        )
+        return self._state
+
+    @property
+    def interfaces(self):
+        """ Indexed interfaces state """
+        return self._ifaces_state
+
+    def sanitize_ethernet(self, other_state):
```
tyll: I was wondering if it makes sense to actually store both the current full state and the desired (possibly partial) state and then let the NM drivers access both and merge the state as required. Then they can directly see what was explicitly stated by the user and decide what to do. The special handling can then be done in the actual driver.

EdDev: I think that the approach to push down the stack the merging has been attempted or discussed already (not sure).
If you go in this path, we will need a translator (between nmstate API and NM one) and will need to repeat this merging on each provider (assuming more will come in).
The question is, does it help differing the merge further in, removing the common code that does the merging.

I think that the sanitize step can be just removed by avoiding the problem of RW parameter side effects, like we did in the routing case. Whenever we have an issue that one RW parameter affects other RW parameters, we should split between configuration and running states.
`ethernet` and `ip-address` are good examples.

libnmstate/state.py
EdDev: it would result in having a getter changing the state, which is not nice at all (and surprising).
On the other hand, you could argue that on each call it will do it and not change the internal state, but then it is dependent on the usage (if we reuse the normalized object or not).

We can try it out I guess and see if it makes sense.

libnmstate/state.py
EdDev: Sounds reasonable to me, but you need to add `interfaces` at the moment (perhaps this will change).
Lets do it in a separate small PR, I prefer to keep this as the original name so it can be tracked.

libnmstate/state.py
tyll: So your idea is to have

```
ethernet:
    config: ...
    running: ...
```

and the same for IP addresses? This seems like an ugly solution. Will it be possible to change the running state in this example or only the config state?

Regarding the common code for merging, we can still provide common code to get the merged results if this is useful and provide access to the unmerged states.

Also I would only care about other provides when they are there instead of complicating things for events that might never happen.

PR #305, opened 2019-03-31
state: Rename canonicalize_interfaces to merge_interfaces

EdDev:


EdDev:
@tyll , per your request the canonicalize is changed to merge.

EdDev:
> IMHO it should be named merge_state so it does not need to be renamed when it gains support for routes or other entries.

Well, you are basically correct and we can wrap it with the general `merge` method. What I am not sure about at this stage is if it will indeed be called from the same place.
If it will, then this will be a correct move.
I do not think we have to do this now, just because I cannot answer the usage question, but if you prefer to I am fine with it.

@cathay4t , will it fit with the routes setting? Does it uses such a merge?

EdDev:
BTW, we may find that splitting this class per subjects is needed (interfaces, DNS, routes, etc..). Then we will need to use inheritance or composition, but it is hard to see how it will be or if indeed needed at this stage.

PR #306, opened 2019-04-03
readme: Update install instructions

EdDev:


README.md
```diff
@@ -39,15 +39,18 @@ IRC](https://freenode.net/kb/answer/chat).
 Nmstate uses `tox` to run unit tests and linters. Since Nmstate uses the binary
 module PyGObject it also requires the build dependencies for it.
 
-### RHEL 7.6
```
tyll: Why would you like to remove the instructions for RHEL? IMHO it makes sense to have them as well.

README.md
```diff
@@ -60,7 +63,35 @@ tox
 
 ## Runtime Environment
 
-Install (from sources) system-wide:
+### Install pre-requirements
+```shell
+sudo yum -y install epel-release
```
tyll: The dev installations do not use sudo but yum directly. Consistency would be nice.

README.md
```diff
@@ -60,7 +63,35 @@ tox
...
+    NetworkManager-libnm \
+    NetworkManager-ovs \
+    openvswitch \
+    python2-pyyaml \
+    python2-six \
+    python-gobject-base \
+    python-jsonschema \
+    python-setuptools \
+    python-ipaddress
+sudo echo -e "[device]\nmatch-device=*\nmanaged=0\n" >> \
```
tyll: How about recommending to run the script that we use for docker as well?

README.md
```diff
@@ -60,7 +63,35 @@ tox
...
+    /etc/NetworkManager/conf.d/97-docker-build.conf
+sudo echo -e "[main]\nno-auto-default=*\n" >> \
+    /etc/NetworkManager/conf.d/97-docker-build.conf
+```
```
tyll: openvswitch needs to be started and NetworkManager needs to be (re)started as well.

README.md
```diff
@@ -60,7 +63,35 @@ tox
...
+
+### Install nmstate from pip
+```shell
+sudo yum -y install python2-pip
+sudo pip install --upgrade nmstate
```
tyll: Is this really something that should be recommended? It does not seem to be good advice. IMHO it is better to recommend to install it for the user only.

README.md
```diff
@@ -60,7 +63,35 @@ tox
...
+    python-jsonschema \
+    python-setuptools \
+    python-ipaddress
+sudo echo -e "[device]\nmatch-device=*\nmanaged=0\n" >> \
+    /etc/NetworkManager/conf.d/97-docker-build.conf
+sudo echo -e "[main]\nno-auto-default=*\n" >> \
+    /etc/NetworkManager/conf.d/97-docker-build.conf
+```
+
+### Install nmstate from pip
```
tyll: This is actually PyPi not pip

README.md
```diff
@@ -60,7 +63,35 @@ tox
 
 ## Runtime Environment
 
-Install (from sources) system-wide:
+### Install pre-requirements
```
tyll: IMHO it would be nice to also explain the deps a little bit for #256, especially why the dependencies are needed.

tyll: Especially the word "introspection data" should be mentioned because this is what the error message in #256 says.

README.md
tyll: The `--upgrade` command also seems to be dangerous as it seems to make pip upgrade all the requirements as well. IMHO `pip uninstall nmstate; pip install nmstate` seems to be a better recommendation.

README.md
EdDev: I just thought that instructing RHEL users to use `epel` and `pip` is not a good move (or supported), but if you think it makes sense, then it is fine by me.

README.md
EdDev: Could you please recommend the text here?
Form my point of view, if someone is interested in understanding what each package is about, they can look it up.

README.md
EdDev: Ok, will drop the `sudo`.

README.md
EdDev: If we specify it here explicitly we do not need to elaborate on how to run that script and where it is. It's not that costly and simple enough, so I think it is good enough.

For the future, it will be nice if we can put all this requirements into the setup.py (or call a script from there), so all this will be done as part of the installation.

README.md
EdDev: Right, thanks.

README.md
EdDev: Done

README.md
EdDev: Ok, dropping the upgrade and replacing it with the uninstall-install.
nmstate in my view is a host level tool, no different from NM. I see no point installing it under a user.

README.md
tyll: Using EPEL and pip is the pragmatic way to be able to develop on RHEL and I would like to be be able to do it if necessary, so please keep it.

README.md
tyll: Probably it makes sense to also enable it, therefore

```
systemctl enable --now NetworkManager openvswitch
systemctl restart NetworkManager
```

Is better - also NetworkManager might need to be restarted after starting openvswitch (not 100% sure).

README.md
```diff
@@ -60,9 +63,39 @@ tox
...
+```shell
+yum -y install python2-pip
+pip uninstall nmstate; pip install nmstate
+```
+
+### Install nmstate from source
+Install system-wide:
 ```shell
-sudo pip install --upgrade .
+pip install --upgrade .
```
tyll: `pip uninstall nmstate; pip install .` would be better here, too. IMHO

README.md
```diff
@@ -60,9 +63,39 @@ tox
 
 ## Runtime Environment
 
-Install (from sources) system-wide:
+### Install pre-requirements
+```shell
+yum -y install epel-release
+yum -y install \
```
tyll: I guess in the future we should move this to their own shell scripts so that we can share them with the container specs to keep them in sync...

README.md
tyll: How about something like the following?

```
Despite the pure python dependencies (that can also be found in requirements.txt), Nmstate also needs NetworkManager to be running on the local system to be able to configure the local network state. To access NetworkManager Nmstate need libnm and the corresponding introspection data (`NM-1.0.typelib`, here provided by `python-gobject-base`). To manage OvS, Nmstate needs the packages `NetworkManager-ovs` and `openvswitch`. NetworkManager might need to be restarted after installing the OvS plugin and starting the openvswitch service.

Additionally, some bugs in NetworkManager require special configuration snippets.
```

README.md
```diff
@@ -60,9 +63,39 @@ tox
...
+    libibverbs \
+    NetworkManager \
+    NetworkManager-libnm \
+    NetworkManager-ovs \
+    openvswitch \
+    python2-pyyaml \
```
tyll: `dbus-python` should also be installed on CentOS/RHEL 7.

README.md
tyll: For host level installs it is IMHO better to (recommend to) install the RPM package instead of using pip.

README.md
```diff
@@ -60,9 +63,39 @@ tox
...
+    python-ipaddress
+echo -e "[device]\nmatch-device=*\nmanaged=0\n" >> \
+    /etc/NetworkManager/conf.d/97-docker-build.conf
+echo -e "[main]\nno-auto-default=*\n" >> \
+    /etc/NetworkManager/conf.d/97-docker-build.conf
+systemctl restart NetworkManager
+systemctl restart openvswitch
+```
+
+### Install nmstate from PyPi
```
tyll: Actually since we already suggest to enable epel, it would be even better to install it with

```yum --enable-repo epel-testing install nmstate``` - for this case a lot of the RPM dependencies do not need to be installed manually btw.

File: README.md (hunk 1 above)
Comment (EdDev) [reply README.md]: Ok, I will keep it.
But for the record, based on my experience, the pragmatic way is to use a container instead. We should probably add documentation for it as well.

File: README.md (hunk 7 above)
Comment (EdDev) [reply README.md]: Done

File: README.md (hunk 10 above)
Comment (EdDev) [reply README.md]: I think this is something you are adding through a different PR.
How come it works as is at the moment?

File: README.md (hunk 4 above)
Comment (EdDev) [reply README.md]: nmstate installation is not about deciding this for the user. I think NM is enabled by default when installed and regarding OVS, I do not know and prefer not to interfere.

File: README.md (hunk 11 above)
Comment (EdDev) [reply README.md]: I do not think it is nice to require `epel-testing` and I think the current steps are good enough.
This is also what we do in the containers, so these are "verified".

File: README.md (hunk 8 above)
Comment (EdDev) [reply README.md]: Done

File: README.md (hunk 10 above)
Comment (tyll) [reply README.md]: Currently it will not use the checkpoint feature if it is not installed but it would already use it if it is there. On RHEL it is typically installed for subscription manager etc. Not sure about CentOS.

File: README.md (hunk 11 above)
Comment (tyll) [reply README.md]: Not sure why using pypi is better than using epel testing... There is also the COPR repo. IMHO PyPi is more for people who cannot use RPMs.

File: README.md (hunk 4 above)
Comment (tyll) [reply README.md]: Documenting how it works best is not deciding this for the user.... You realize that these are steps that a user will do manually, so there is no inteference as long as the user does not do anything. But not telling them about the best options is strange.... Also making them add custom config files to NM is a lot worse that asking them to enable services.

File: README.md [hunk 12]
```diff
@@ -60,9 +73,52 @@ tox
...
+    NetworkManager-ovs \
+    openvswitch \
+    python2-pyyaml \
+    python2-six \
+    python-gobject-base \
+    python-jsonschema \
+    python-setuptools \
+    python-ipaddress
+echo -e "[device]\nmatch-device=*\nmanaged=0\n" >> \
+    /etc/NetworkManager/conf.d/97-nmstate.conf
```
tyll: Great that you changed the file names here! :+1:

README.md
EdDev: Done

README.md
EdDev: Done

README.md
```diff
@@ -60,9 +73,55 @@ tox
...
+    python-ipaddress \
+    python-jsonschema \
+    python-setuptools \
+    python2-pyyaml \
+    python2-six
+echo -e "[device]\nmatch-device=*\nmanaged=0\n" >> \
+    /etc/NetworkManager/conf.d/97-nmstate.conf
+echo -e "[main]\nno-auto-default=*\n" >> \
+    /etc/NetworkManager/conf.d/97-nmstate.conf
+systemctl restart NetworkManager
```
tyll: Is this the correct order?

EdDev: It does not matter, they are decoupled as fas as I know.
If this is not the case, it is a bug.

PR #307, opened 2019-04-03
nm/checkpoint: Do not destroy other checkpoints

tyll:
Not destroying other checkpoints will make checkpoint creation fail if
there is already a checkpoint for all devices. This prevents other users
to use another instance of Nmstate to change the state while a state
change is in progress.

### DIFF
--- a/libnmstate/nm/checkpoint.py
+++ b/libnmstate/nm/checkpoint.py
@@ -108,8 +108,8 @@ class CheckPoint(object):
             logging.debug('Checkpoint %s created for all devices: %s',
                           dbuspath, timeout)
             self._checkpoint = dbuspath
-        except dbus.exceptions.DBusException:
-            raise NMCheckPointCreationError()
+        except dbus.exceptions.DBusException as e:
+            raise NMCheckPointCreationError(str(e))
 
     def destroy(self):
         self._manager.interface.CheckpointDestroy(self._checkpoint)
--- a/libnmstate/nm/checkpoint.py
+++ b/libnmstate/nm/checkpoint.py
@@ -93,7 +93,6 @@ class CheckPoint(object):
         devs = []
         timeout = self._timeout
         cp_flags = (
-            CHECKPOINT_CREATE_FLAG_DESTROY_ALL |
             CHECKPOINT_CREATE_FLAG_DELETE_NEW_CONNECTIONS |
             CHECKPOINT_CREATE_FLAG_DISCONNECT_NEW_DEVICES
         )
--- a/libnmstate/nm/checkpoint.py
+++ b/libnmstate/nm/checkpoint.py
@@ -32,6 +32,12 @@ CHECKPOINT_CREATE_FLAG_DISCONNECT_NEW_DEVICES = 0x04
 _nmdbus_manager = None
 
 
+class NMCheckPointCreationError(Exception):
+    """ Error creating a Network Manager CheckPoint """
+
+    pass
+
+
 def has_checkpoint_capability():
     return dbus is not None
 
@@ -96,16 +102,22 @@ class CheckPoint(object):
             CHECKPOINT_CREATE_FLAG_DELETE_NEW_CONNECTIONS |
             CHECKPOINT_CREATE_FLAG_DISCONNECT_NEW_DEVICES
         )
-        cp = self._manager.interface.CheckpointCreate(devs, timeout, cp_flags)
-        logging.debug('Checkpoint %s created for all devices: %s', cp, timeout)
-        self._checkpoint = cp
+        try:
+            dbuspath = self._manager.interface.CheckpointCreate(devs, timeout,
+                                                                cp_flags)
+            logging.debug('Checkpoint %s created for all devices: %s',
+                          dbuspath, timeout)
+            self._checkpoint = dbuspath
+        except dbus.exceptions.DBusException:
+            raise NMCheckPointCreationError()
 
     def destroy(self):
         self._manager.interface.CheckpointDestroy(self._checkpoint)
         logging.debug('Checkpoint %s destroyed', self._checkpoint)
 
     def rollback(self):
-        result = self._manager.interface.CheckpointRollback(self._checkpoint)
-        logging.debug('Checkpoint %s rollback executed: %s',
-                      self._checkpoint, result)
+        result = self._manager.interface.CheckpointRollback(
+            self._checkpoint)
+        logging.debug('Checkpoint %s rollback executed: %s', self._checkpoint,
+                      result)
         return result
--- a/libnmstate/nm/checkpoint.py
+++ b/libnmstate/nm/checkpoint.py
@@ -17,11 +17,7 @@
 
 import logging
 
-try:
-    import dbus                         # pylint: disable=import-error
-except ImportError:
-    # dbus is set to None to indicate that checkpoint is not available
-    dbus = None
+import dbus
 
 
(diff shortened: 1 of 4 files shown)


libnmstate/nm/checkpoint.py
```diff
@@ -96,16 +99,24 @@ def create(self):
...
-        logging.debug('Checkpoint %s created for all devices: %s', cp, timeout)
-        self._checkpoint = cp
+        try:
+            dbuspath = self._manager.interface.CheckpointCreate(devs, timeout,
+                                                                cp_flags)
+            logging.debug('Checkpoint %s created for all devices: %s',
+                          dbuspath, timeout)
+            self._checkpoint = dbuspath
+        except dbus.exceptions.DBusException:
+            pass
```
EdDev: If `create` fails, it should explode.
If the caller is interested in doing something when it explodes, he should handle it.

If there is a need to ignore an exception, a very good reason for it is needed, otherwise it is a smell of something bad happening.

All the `if` checks below I consider a bad practice. If a caller uses them on an object that is not properly initialized, it should explode as well.

libnmstate/nm/checkpoint.py
```diff
@@ -93,7 +93,6 @@ def create(self):
         devs = []
         timeout = self._timeout
         cp_flags = (
-            CHECKPOINT_CREATE_FLAG_DESTROY_ALL |
```
EdDev: > Not destroying other checkpoints will make checkpoint creation fail if
> there is already a checkpoint for all devices. This prevents other users
> to use another instance of Nmstate to change the state while a state
> change is in progress.

Some other user of NM may have created such a checkpoint and not nmstate, so I am not clear how is this a good step here.
Previously we said something like: "If checkpoints existed before, we destroy them all and use ours."
Now we hope that there are no others which can kill us.

tyll: Sorry, I do not follow. You believe that it is a good idea to blow up other applications that create a checkpoint by unconditionally destroying it? This sounds like a bad idea to me. IMHO the right step is to notify the user and get them to rectify the situation.

Also there is the dream of a requirement to have a global lock and now you are worried that someone else might have a smaller lock that would block Nmstate... this does not make sense or I am not understanding what you mean with `kill us`.

libnmstate/nm/checkpoint.py
tyll: Can you please explain how to envision this with a context manager? Do you want something like this?

```
try:
    with nm.checkpoint.CheckPoint() as checkpoint:
        with _setup_providers():
            _add_interfaces(desired_state.interfaces, current_state.interfaces)
        with _setup_providers():
            current_state = state.State(
                {Constants.INTERFACES: netinfo.interfaces()}
            )
            _edit_interfaces(desired_state, current_state)
        if verify_change:
            _verify_change(desired_state)
except NMCheckPointError:
```

libnmstate/nm/checkpoint.py
EdDev: Having the checkpoint created by someone else is not related to the lock, as this 3rd party could have just left the checkpoint in place and is not currently changing the system.

Per our discussion, the scenario that convinces me that this makes sense is this one:
- User calls nmstate.apply with a rollback timeout.
- User calls nmstate.apply again, before sending the confirmation, with a rollback timeout. This should fail as the user should have either confirm the changes or explicitly cause a rollback.

Note: If the use calls apply again with no timeout mentioned, no checkpoint should be created and if an existing one is already there, it should be used (in case verify or an error occurs and requires a rollback).

I suggest you document all these scenarios in the acceptance test or some other documentation so we can monitor it.

libnmstate/nm/checkpoint.py
EdDev: Looks good to me.
Consider moving the try block into its own function so this will look a bit nicer.

tyll: Changed this ... did not move this in its own function because this was already a rebase hell and it is most of what the function already does. It might make sense to clean this up once this becomes a method of a state object so that there would not be a ton of variables to pass around.

libnmstate/nm/checkpoint.py
tyll: a) I do not plan to use the timeout parameter as you mentioned but instead have an explicit commit parameter that the caller needs to set to false if they want to commit the change themself
b) Re-usinga checkpoint is something that we can discuss in the future (to me it seems wrong) and we can add it afterwards
c) I have the opinion that a checkpoint should be always created to handle verification or Nmstate runtime errors

tyll:
@cathay4t it would be great if you could review this at your earliest convenience so I can continue with building on top of this.

tyll:
> Things I dislike:
> 
> * I don't think `NmstateConflictError` could be a child of permission error.

ok, should it inherit any other class?

>   And why not just use `NMCheckPointCreationError`?

`NMCHeckPointCreationError` is internal to the nm submodule. If we want to raise it in `netapplier()` it needs to be in the `nm.error` submodule and would make the nm submodule not independent. Therefore it seems to me that we need two exceptions.

> * The `NMCheckPointCreationError` does not hold any information other than it's name.
>   Please try to at least pass the error message to it instead of `raise NMCheckPointCreationError()`.

Good catch. I will change it.

>   And the only usage of it in `libnmstate` is convert it to `NmstateConflictError`, why the complex
>   when you can simple raise it as `NmstateConflictError` or `NMCheckPointCreationError`?
> * The only usage of checkpoint in `libnmstate` is `nm.checkpoint.CheckPoint()` in `netapplier.py`.
>   Other arguments like `autodestroy` and `dbuspath` is only used by test code only. I would like to
>   see them removed and only be added when they are needed by production code.
>   The same comments also applied to `nm.checkpoint.get_checkpoints()` which is also unused.

I would like to avoid rebasing both the frontend and backend code at the same time. If the code is good enough, I can also add the commits for the frontend code (nmstatectl).

cathay4t:
> > Things I dislike:
> > 
> > * I don't think `NmstateConflictError` could be a child of permission error.
> 
> ok, should it inherit any other class?

How about `RuntimeError`?

> 
> > And why not just use `NMCheckPointCreationError`?
> 
> `NMCHeckPointCreationError` is internal to the nm submodule. If we want to raise it in `netapplier()` it needs to be in the `nm.error` submodule and would make the nm submodule not independent. Therefore it seems to me that we need two exceptions.

Good point. You choice is better.
> 
> > And the only usage of it in `libnmstate` is convert it to `NmstateConflictError`, why the complex
> > when you can simple raise it as `NmstateConflictError` or `NMCheckPointCreationError`?
> > 
> > * The only usage of checkpoint in `libnmstate` is `nm.checkpoint.CheckPoint()` in `netapplier.py`.
> >   Other arguments like `autodestroy` and `dbuspath` is only used by test code only. I would like to
> >   see them removed and only be added when they are needed by production code.
> >   The same comments also applied to `nm.checkpoint.get_checkpoints()` which is also unused.
> 
> I would like to avoid rebasing both the frontend and backend code at the same time. If the code is good enough, I can also add the commits for the frontend code (nmstatectl).

Maybe you may create another PR containing:
 * Patch to introduce new function for `nm.checkpoint`.
 * Patch to use newly introduced function in `nmstatectl`

In this way, we can review the usage of newly added function.

tyll:
Updated the PR:
- Adjusted exceptions: Keep error and inherit RuntimeError
- removed commits that add backend support for later fronted changes

cathay4t:
@tyll We are ready to merge it after you rebase it.

tyll:
Thank you!

PR #308, opened 2019-04-04
Container dbus

tyll:
This is needed for #307 to ensure that python-dbus is avilable in the containers.


tyll:
This needs to get rebased after #315 is merged since they share some changes.

cathay4t:
@tyll Can you explain what did you clean up in commit `Fedora container spec: Cleanup package installation`?  Is that simple break packages into one-per-line?

tyll:
> @tyll Can you explain what did you clean up in commit `Fedora container spec: Cleanup package installation`? Is that simple break packages into one-per-line?

I broke it into one package per line, grouped them together and removed duplicates.

PR #309, opened 2019-04-04
mac-address: Support interface mac-address

EdDev:
Dependent on https://nmstate.atlassian.net/browse/NMSTATE-166

### DIFF
--- a/libnmstate/nm/wired.py
+++ b/libnmstate/nm/wired.py
@@ -17,17 +17,22 @@
 
 from libnmstate.ethtool import minimal_ethtool
 from libnmstate.nm import nmclient
+from libnmstate.schema import Interface
+
+
+ZEROED_MAC = '00:00:00:00:00:00'
 
 
 def create_setting(iface_state, base_con_profile):
     mtu = iface_state.get('mtu')
+    mac = iface_state.get(Interface.MAC)
 
     ethernet = iface_state.get('ethernet', {})
     speed = ethernet.get('speed')
     duplex = ethernet.get('duplex')
     auto_negotiation = ethernet.get('auto-negotiation')
 
-    if not (mtu or speed or duplex or (auto_negotiation is not None)):
+    if not (mac or mtu or speed or duplex or (auto_negotiation is not None)):
         return None
 
     wired_setting = None
@@ -38,6 +43,9 @@ def create_setting(iface_state, base_con_profile):
 
     if not wired_setting:
         wired_setting = nmclient.NM.SettingWired.new()
+
+    if mac:
+        wired_setting.props.cloned_mac_address = mac
 
     if mtu:
         wired_setting.props.mtu = mtu
@@ -83,6 +91,11 @@ def get_info(device):
         info['mtu'] = int(device.get_mtu())
     except AttributeError:
         pass
+
+    mac = device.get_hw_address()
+    # A device may not have a MAC or it may not yet be "realized" (zeroed mac).
+    if mac and mac != ZEROED_MAC:
+        info[Interface.MAC] = mac
 
     if device.get_device_type() == nmclient.NM.DeviceType.ETHERNET:
         ethernet = _get_ethernet_info(device, iface)
--- a/tests/integration/bond_test.py
+++ b/tests/integration/bond_test.py
@@ -23,11 +23,15 @@ import yaml
 from libnmstate import netapplier
 from libnmstate import netinfo
 from libnmstate.error import NmstateVerificationError
+from libnmstate.schema import Interface
 
 from .testlib import assertlib
 from .testlib import statelib
 from .testlib.statelib import INTERFACES
 
+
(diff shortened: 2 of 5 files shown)


tests/lib/nm/wired_test.py
```diff
@@ -42,6 +42,13 @@ def test_create_setting_duplicate(NM_mock):
         base_profile.get_setting_wired.return_value.duplicate.return_value
 
 
+def test_create_setting_mac(NM_mock):
+    setting = nm.wired.create_setting(
+        {'mac-address': '01:23:45:67:89:ab'}, None)
```
tyll: AFAIU, mac[0] & 0x02 should be true for local MAC addresses, i.e. use  `02:...` or `03:...` here.

tests/lib/nm/wired_test.py
```diff
@@ -96,11 +103,13 @@ def test_create_setting_speed_duplex(NM_mock):
 def test_get_info_with_invalid_duplex(ethtool_mock, NM_mock):
     dev_mock = mock.MagicMock()
     dev_mock.get_iface.return_value = 'nmstate_test'
+    dev_mock.get_hw_address.return_value = 'ab:cd:ef:01:23:45'
```
tyll: This might also not be a local MAC address IIUC.

libnmstate/nm/wired.py
```diff
@@ -39,6 +41,9 @@ def create_setting(iface_state, base_con_profile):
     if not wired_setting:
         wired_setting = nmclient.NM.SettingWired.new()
 
+    if mac:
+        wired_setting.props.cloned_mac_address = mac
```
tyll: should be check that it does not contains special NM values like random or stable?

tests/lib/nm/wired_test.py
EdDev: This is a unit test, who cares about these technicalities? It is also not wrong to use non local ones.
The usage or no usage of it matters for a production system.
Not even sure if it matters for an integration tests, although you may argue there that using a non-local one may collide with existing address.

libnmstate/nm/wired.py
EdDev: The schema accepts only a mac format.

tests/lib/nm/wired_test.py
tyll: If we can use any address, we can also use a correct one. It is not a blocker for me, just an observation.

tyll: Btw. it would be nice to also have integration tests, ideally for each interface type to ensure that it works across all types.

EdDev: The missing integration tests is a hard nack here.

EdDev:
I have added integration tests for the mac in the bond tests.
We should cover other interface types, but I think we should be safe as it is, because the mac is set explicitly in all the running tests anyway.

Added a ticket to make sure we add it later on: https://nmstate.atlassian.net/browse/NMSTATE-169

libnmstate/nm/wired.py
```diff
@@ -84,6 +89,11 @@ def get_info(device):
     except AttributeError:
         pass
 
+    mac = device.get_hw_address()
+    # A device may not have a MAC or it may not yet be "realized" (zeroed mac).
+    if mac and mac != '00:00:00:00:00:00':
```
tyll: not a blocker, should the zero mac be a constant?

tests/integration/linux_bridge_test.py
```diff
@@ -29,6 +29,8 @@
 from .testlib.statelib import INTERFACES
 
 
+MAC0 = '02:ff:ff:ff:ff:00'
```
tyll: This seems to be left accidentally.

tests/integration/bond_test.py
```diff
@@ -241,3 +245,20 @@ def test_remove_one_of_the_bond_slaves(eth1_up, eth2_up):
         bond99_cur_state = current_state[INTERFACES][0]
 
     assert bond99_cur_state['link-aggregation']['slaves'] == ['eth2']
+
+
+def test_set_bond_mac_address(eth1_up):
```
tyll: IMHO it would be nicer to see the MAC tests in their own file to make it easier to see if they cover all device types 
but it is not a blocker for me.

libnmstate/nm/wired.py
EdDev: Done

tests/integration/bond_test.py
EdDev: I think that we need to choose the main subject of the tests. Having each property as a subject does not fit well IMO.
But perhaps we can centralize all interface type agnostic tests into one test module.

tests/integration/linux_bridge_test.py
EdDev: Thanks, done

PR #312, opened 2019-04-04
Some small corrections for the installation README

tyll:


tyll:
Thank you! It is not about how to install Nmstate per-se but how to use pip.

PR #313, opened 2019-04-04
automation: Remove unecessary PYTHON_SITE_PATH_CMD

tyll:


tyll:
It seems this got lost from the review back then.

PR #314, opened 2019-04-04
setup.py: Workaround old setuptools in CentOS/RHEL 7

tyll:
setup.py breaks on CentOS7/RHEL 7, breaking the RPM and container build.
Workaround this by handling the one special case manually.


tyll:
Duplicate #315

PR #325, opened 2019-04-23
tests integ: Introduce the nm wired test module

EdDev:


tyll:
There is a whitespace error

tests/integration/nm/wired_test.py
```diff
@@ -0,0 +1,68 @@
...
+
+from libnmstate import nm
+from libnmstate import schema
+
+from .testlib import mainloop
+
+
+ETH1 = 'eth1'
+
+MTU0 = 1200
```
tyll: Why this MTU value?

tests/integration/nm/wired_test.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def test_interface_mtu_change(eth1_up):
```
tyll: Will this test restore the previous MTU once the test is done?

tests/integration/nm/wired_test.py
EdDev: It is *a* value, no specific reason.

tests/integration/nm/wired_test.py
EdDev: The eth1_up fixture hopefully does it by deleting the profile and redefining it.

EdDev: Confirmed: Any test which uses `eth1` will have to call `eth1_up` (or `eth1_down`) and that will reset the mtu value (like all the other values it may have).

EdDev:
ci test please

PR #326, opened 2019-04-23
Test against NetworkManager devel builds

tyll:


tyll:
The CentOS repo seems to be broken, therefore the copr tests fail for CentOS

tyll:
Actually it is not the repo that is broken but the problem that pip installs urllib3 into the system sitelib which then breaks installing the python-urllib3 package.  A good example why using pip as root without --user is a bad idea (and also not recommended by Red Hat documentation).

automation/Dockerfile
```diff
@@ -1,16 +1,21 @@
...
         radvd \
-    && \
-    pip install -U pip pytest==4.2.1 pytest-cov==2.6.1 python-coveralls && \
-    yum clean all
+        rpm-build \
+    && yum clean all
+
+RUN    pip install --upgrade pip pytest==4.2.1 pytest-cov==2.6.1
+
+RUN    pip3 install --user python-coveralls
```
cathay4t: Can you explain why old `pip install python-coveralls` does not works?

cathay4t:
> Actually it is not the repo that is broken but the problem that pip installs urllib3 into the system sitelib which then breaks installing the python-urllib3 package. A good example why using pip as root without --user is a bad idea (and also not recommended by Red Hat documentation).

Can we change every pip command to use `--user`?

automation/Dockerfile
tyll: I added the reasons for using Python3 to the commit message already. Is it enough?

tyll:
> > Actually it is not the repo that is broken but the problem that pip installs urllib3 into the system sitelib which then breaks installing the python-urllib3 package. A good example why using pip as root without --user is a bad idea (and also not recommended by Red Hat documentation).
> 
> Can we change every pip command to use `--user`?

For an unknown reason run-tests.sh fails to find pytest then and I did not want to spend time investigating this since it is good enough for now IMHO.

PR #332, opened 2019-04-24
Introduce nm.device.reapply

EdDev:
Reapply operation is introduced in the nm provider.
It includes a fallback to a profile activation in case the reapply fails.

Integration into nmstate core is planned in a separate PR.

Dependent on PR #325

### DIFF
--- a/libnmstate/nm/connection.py
+++ b/libnmstate/nm/connection.py
@@ -151,7 +151,7 @@ class ConnectionProfile(object):
                 logging.debug(
                     'Connection activation in progress: dev=%s, state=%s',
                     ac.devname, ac.state)
-                self._waitfor_active_connection_async(ac)
+                self.waitfor_active_connection_async(ac)
                 return
 
         specific_object = None
@@ -205,7 +205,7 @@ class ConnectionProfile(object):
             if ac.is_active:
                 self._mainloop.execute_next_action()
             elif ac.is_activating:
-                self._waitfor_active_connection_async(ac)
+                self.waitfor_active_connection_async(ac)
             else:
                 self._mainloop.quit(
                     'Connection activation failed on {}: reason={}'.format(
@@ -230,7 +230,7 @@ class ConnectionProfile(object):
 
         return activation_type, activation_object
 
-    def _waitfor_active_connection_async(self, ac):
+    def waitfor_active_connection_async(self, ac):
         ac.handlers.add(
             ac.nm_active_connection.connect(
                 'state-changed', self._waitfor_active_connection_callback, ac)
--- a/libnmstate/nm/device.py
+++ b/libnmstate/nm/device.py
@@ -49,6 +49,68 @@ def delete(dev):
         con_profile.delete()
 
 
+def reapply(dev, connection_profile):
+    mainloop = nmclient.mainloop()
+    mainloop.push_action(_safe_reapply_async, dev, connection_profile)
+
+
+def _safe_reapply_async(dev, connection_profile):
+    mainloop = nmclient.mainloop()
+    cancellable = mainloop.new_cancellable()
+
+    if _wait_for_active_connection_async(dev, connection_profile):
+        return
+
+    version_id = 0
+    flags = 0
+    user_data = mainloop, dev, cancellable
+    dev.reapply_async(
+        connection_profile,
+        version_id,
+        flags,
+        cancellable,
+        _reapply_callback,
+        user_data,
+    )
+
+
+def _wait_for_active_connection_async(dev, connection_profile):
+    active_conn = connection.get_device_active_connection(dev)
+    if active_conn:
+        act_conn = ac.ActiveConnection(active_conn)
+        if act_conn.is_activating:
+            logging.debug(
+                'Connection activation in progress: dev=%s, state=%s',
+                act_conn.devname, act_conn.state)
+            conn = connection.ConnectionProfile(connection_profile)
+            conn.waitfor_active_connection_async(act_conn)
+            return True
+    return False
+
+
+def _reapply_callback(src_object, result, user_data):
+    mainloop, nmdev, cancellable = user_data
+    mainloop.drop_cancellable(cancellable)
+
+    devname = src_object.get_iface()
+    try:
+        success = src_object.reapply_finish(result)
+    except Exception as e:
+        if mainloop.is_action_canceled(e):
+            logging.debug('Device reapply aborted on %s: error=%s', devname, e)
+        else:
+            mainloop.quit(
+                'Device reapply failed on {}: error={}'.format(devname, e))
+        return
+
+    if success:
+        logging.debug('Device reapply succeeded: dev=%s', devname)
+        mainloop.execute_next_action()
+    else:
+        mainloop.quit(
+            'Device reapply failed: dev={}, error=unknown'.format(devname))
+
+
 def delete_device(nmdev):
     mainloop = nmclient.mainloop()
     mainloop.push_action(_safe_delete_device_async, nmdev)
--- a/libnmstate/nm/nmclient.py
+++ b/libnmstate/nm/nmclient.py
@@ -59,9 +59,17 @@ def client(refresh=False):
     return _nmclient
 
 
-def mainloop():
+def mainloop(refresh=False):
+    """
+    Create a singleton (global) mainloop object.
+
+    When the refresh flag is set, the existing mainloop object is re-created.
+    Use the refresh flag in scenarios where the mainloop is needed but
+    the existing object is no longer usable (in an undefined state),
+    e.g. the mainloop exited with failure.
+    """
     global _mainloop
-    if _mainloop is None:
+    if _mainloop is None or refresh:
         _mainloop = _MainLoop()
     return _mainloop
 
--- a/tests/integration/nm/ipv4_test.py
+++ b/tests/integration/nm/ipv4_test.py
@@ -1,0 +1,82 @@
+#
+# Copyright 2019 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+from libnmstate import nm
+
+from .testlib import mainloop
+
+
+TEST_IFACE = 'eth1'
+
+IPV4_ADDRESS1 = '192.0.2.251'
+
+
+def test_interface_ipv4_change(eth1_up):
+    with mainloop():
+        _modify_interface(
+            ipv4_state={
+                'enabled': True,
+                'dhcp': False,
+                'address': [
+                    {'ip': IPV4_ADDRESS1, 'prefix-length': 24}
+                ]
+            }
+        )
+
+    nm.nmclient.client(refresh=True)
+    ipv4_current_state = _get_ipv4_current_state(TEST_IFACE)
+
+    ip4_expected_state = {
+        'enabled': True,
+        'dhcp': False,
+        'address': [
+            {'ip': IPV4_ADDRESS1, 'prefix-length': 24}
+        ]
+    }
+    assert ip4_expected_state == ipv4_current_state
+
+
+def _modify_interface(ipv4_state):
+    conn = nm.connection.ConnectionProfile()
+    conn.import_by_id(TEST_IFACE)
+    settings = _create_iface_settings(ipv4_state, conn)
+    new_conn = nm.connection.ConnectionProfile()
+    new_conn.create(settings)
+    conn.update(new_conn)
+    conn.commit(save_to_disk=False)
+
+    nmdev = nm.device.get_device_by_name(TEST_IFACE)
+    nm.device.reapply(nmdev, conn.profile)
+
+
+def _get_ipv4_current_state(ifname):
+    nmdev = nm.device.get_device_by_name(ifname)
+    active_connection = nm.connection.get_device_active_connection(nmdev)
+    return nm.ipv4.get_info(active_connection)
+
+
+def _create_iface_settings(ipv4_state, con_profile):
+    con_setting = nm.connection.ConnectionSetting()
+    con_setting.import_by_profile(con_profile)
+
+    # Wired is required due to https://bugzilla.redhat.com/1703960
+    wired_setting = con_profile.profile.get_setting_wired()
+
+    ipv4_setting = nm.ipv4.create_setting(ipv4_state, con_profile.profile)
+    ipv6_setting = nm.ipv6.create_setting({}, None)
+
+    return con_setting.setting, wired_setting, ipv4_setting, ipv6_setting
--- a/tests/integration/nm/wired_test.py
+++ b/tests/integration/nm/wired_test.py
@@ -15,18 +15,23 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
+import pytest
+
 from libnmstate import nm
 from libnmstate import schema
 
 from .testlib import mainloop
+from .testlib import TestMainloopError
 
 
 ETH1 = 'eth1'
 
+MAC0 = '02:FF:FF:FF:FF:00'
 MTU0 = 1200
 
 
-def test_interface_mtu_change(eth1_up):
+@pytest.mark.xfail(reason='https://bugzilla.redhat.com/1702657', strict=True)
+def test_interface_mtu_change_with_reapply(eth1_up):
     wired_base_state = _get_wired_current_state(ETH1)
     with mainloop():
         _modify_interface(wired_state={schema.Interface.MTU: MTU0})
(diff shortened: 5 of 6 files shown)


libnmstate/nm/connection.py
```diff
@@ -151,7 +151,7 @@ def _safe_activate_async(self):
                 logging.debug(
                     'Connection activation in progress: dev=%s, state=%s',
                     ac.devname, ac.state)
-                self._waitfor_active_connection_async(ac)
```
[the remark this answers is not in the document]
EdDev: Changed it, hope it is clearer now.

libnmstate/nm/connection.py
```diff
@@ -205,7 +205,7 @@ def _active_connection_callback(self, src_object, result, user_data):
             if ac.is_active:
                 self._mainloop.execute_next_action()
             elif ac.is_activating:
-                self._waitfor_active_connection_async(ac)
```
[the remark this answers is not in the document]
EdDev: The renaming is just an upgrade of a private method to a public one, originated by the needs in this change.
IMO it is small enough not to need its own commit (a side effect).

libnmstate/nm/device.py
```diff
@@ -49,6 +49,91 @@ def delete(dev):
...
+
+    mainloop = nmclient.mainloop()
+    cancellable = mainloop.new_cancellable()
+
+    active_conn = connection.get_device_active_connection(dev)
+    if active_conn:
+        act_conn = ac.ActiveConnection(active_conn)
+        if act_conn.is_activating:
+            logging.debug(
+                'Connection activation in progress: dev=%s, state=%s',
```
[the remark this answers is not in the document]
EdDev: We have never done that so far, probably because the need was not there.
If there is a decision/need to do it, I would prefer to do it all over.

libnmstate/nm/device.py
```diff
@@ -49,6 +49,91 @@ def delete(dev):
...
+                act_conn.devname, act_conn.state)
+            conn = connection.ConnectionProfile(connection_profile)
+            conn.waitfor_active_connection_async(act_conn)
+            return
+
+    user_data = mainloop, dev, cancellable
+    dev.reapply_async(
+        connection_profile,
+        0,
```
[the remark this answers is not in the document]
EdDev: Right, thanks.

libnmstate/nm/device.py
```diff
@@ -49,6 +49,93 @@ def delete(dev):
         con_profile.delete()
 
 
+def reapply(dev, connection_profile):
```
tyll: nit: Seems this method could be tested in a unit test.

libnmstate/nm/device.py
```diff
@@ -49,6 +49,93 @@ def delete(dev):
...
+    """Reapply the given device or remote connection profile."""
```
tyll: IMHO a device cannot be reapplied... Also the doc string does not really contain more information than the function name itself...

libnmstate/nm/device.py
```diff
@@ -49,6 +49,93 @@ def delete(dev):
...
+    mainloop = nmclient.mainloop()
+    mainloop.push_action(_safe_reapply_async, dev, connection_profile)
+
+
+def _safe_reapply_async(dev, connection_profile):
+    # Special cases exist (bugs) where reapply is not supported and
```
tyll: AFAIU not all special cases are bugs unless you see it as a bug in the Linux kernel. Since reapply will fail in the other cases, the comment is a little bit misleading since we also fall back to apply when reapply fails. This also leads to the question why this special handling is needed at all. Why can this not be handled in the callback?

libnmstate/nm/device.py
tyll: Why is the reapply function part of the device module? It seems it would be a easier to change the activation code in the connection module to try reapply and then fallback to normal activation. Then the caller does not need to decide whether or not to use reapply. This would also have the advantage of not adding so much non object-oriented code.

tests/integration/nm/ipv4_test.py
```diff
@@ -0,0 +1,82 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+from libnmstate import nm
+
+from .testlib import mainloop
+
+
+ETH1 = 'eth1'
```
tyll: IMHO it is actually nicer to give the constant a describing name, such as `TEST_INTERFACE` to simplify changing this to a different name without having to rename the constant everywhere as well since using `ETH1 = 'testveth'` would not be that nice.

tests/integration/nm/ipv4_test.py
```diff
@@ -0,0 +1,82 @@
...
+
+IPV4_ADDRESS1 = '192.0.2.251'
+
+
+def test_interface_ipv4_change(eth1_up):
```
tyll: IMHO a test for reapply should be in its own test module. It is not really clear taht the IPv4 test module would test reapply. Also the test does not really verify that reapply was used. Since there is a fallback mechanism to use normal activation instead, it would be great to actually verify that reapply works as expected.

tests/integration/nm/ipv4_test.py
EdDev: Done

tests/integration/nm/ipv4_test.py
EdDev: Your feedback is related to the "strict" reapply I mentioned in my previous response.
Without a "strict" version, I do not know how to check that reapply worked or not.

I also do not see a great benefit in testing the reapply separately, as I would just have to use a subject like ipvX, wired and so on... When one of those is checked, it implicitly checks activate or reapply.

In the context of nmstate, we will always use reapply (see the following changes/PRs) with the fallback.

libnmstate/nm/device.py
```diff
@@ -50,12 +50,22 @@ def delete(dev):
...
+    activation.
+    """
     mainloop = nmclient.mainloop()
     mainloop.push_action(_safe_reapply_async, dev, connection_profile)
 
 
 def _safe_reapply_async(dev, connection_profile):
+    # Special cases are known to exist (bugs) where reapply is not functioning
+    # correctly. Use activation instead. (https://bugzilla.redhat.com/1702657)
+    if _requires_activation(dev, connection_profile):
```
tyll: Since the bug is already fixed upstream, it would make sense to only do the special handling if it is really necessary. Otherwise we cannot test against the master branch and check if it works there.

Reference: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/commit/4ed72fa658c03790700ba9084e9328fe38afdee9

libnmstate/nm/device.py
tyll: Since you expect the feature in NM, I do not follow why Nmstate should not be better than NM in this case. I did no mean to suggest to try activation first and then reapply. My suggestion is to provide one interface to activate a connection for users of the nm submodule and make this one try a reapply first and fallback to normal activation.

tests/integration/nm/ipv4_test.py
tyll: IMHO it makes sense to explicitly test all supported changes with reapply. In the current test setup, the actual test case (`test_interface_ipv4_change()`) does not contain any indication that reapply is required and also does not verify it. AFAIU, only the line `nm.device.reapply(nmdev, conn.profile)` might make the test use reapply. However, this line could just be changed to issue a normal activation or the reapply method could be accidentally changed to never really do a reapply (for example because the check whether reapply would work accidentally always returns True) and the test case would not fail. So the test case only checks that using `nm.device.reapply()` does not break adding an IPv4 address but it tests nothing about the actual functionality.

Also IMHO a strict method is not the way to check it but some assertion that for example the link did not go down for the interface (which is AFAIU why we want reapply).

tyll: Btw. having a user story like `I can change the IP address for an interface without the link going down.` and developing a test that shows that this user story is supported might help here.

libnmstate/nm/device.py
tyll: My comment is about not requesting users of the nm module to know that they need to use reapply but just make the activate method in the connection module do the right thing (trying to use reapply and activate if necessary).

tests/integration/nm/ipv4_test.py
tyll: The vdsm test is not part of Nmstate. Therefore the current test does not show that the code actually delivers something new (reapply). Even if you split it into three functions, the tests will not show that the functions actually use reapply. This can only be guessed by looking at the actual implementation.

Regarding doubling the tests, if we want to have meaningful tests that check both the normal activation (which would be needed when the profile is created for the first time I guess) and reapply (which would be changing a profile), we need twice as much tests as if we only test one of these scenarios. It might also be ok to do this in one test, however there should be assertions that ensure that when changing the profile, traffic was not disrupted.

But in the end, since this is an integration test, whether or not reapply is used should not matter. What matters is the actual behavior which would be no traffic disruption when changing the IP address (according the the commit messages). If you do not want to do the test, I will not force you to do it. Nevertheless, it is a flaw in the current test and does not seem to hold up to the standards that you seem to have.

libnmstate/nm/device.py
tyll: The separate functions would not solve this problem AFAICS. The code would need to auto-detect based on NM's behavior whether or not to only try normal activation. If reapply does not fail directly then IMHO the correct solution would be to verify whether reapply worked and issue a normal activation if the outcome state after reapply is not the desired one. Not sure how feasible this is.

tyll:
@cathay4t there are some (test)design decisions that @EdDev and I disagree on. Since the changes are probably not worse than the current code, I will not require @EdDev to do the changes. However, I would feel more comfortable if you accept this as well.

libnmstate/nm/device.py
EdDev: I feel we are in a loop here.
NM has two api/s, activate and reapply, it is a fact.
I decided to introduce the reapply one with a fallback addition.
You want me to change activate and not introduce reapply.
I explained that it does not matter how we call it and trying to change activate is a mess because I cannot depend there on the device module. I even explained why the reapply is in the device module and to the connection.

Further down, based on your remarks about the tests and special handling, I suggested we keep the two nm api/s as is and use a 3rd one with the fallback solving the test problem.

Please be explicit in what you suggest now base on the above info.

tests/integration/nm/ipv4_test.py
EdDev: I agree with two points here.
The first, you explicitly gave a scenario where activation and not reapply is needed: create a new interface.
The second, we need to check using monitor, but that is too large imo to implement now.

libnmstate/nm/device.py
tyll: The reason why you want to add the reapply code to nm.device is because NM uses it in a device. IMHO it does not matter what NM does if we want to do something better. If adding the code to the device module creates a problem because of a dependency loop to provide the better API, then IMHO the answer is to move the code elsewhere. It could be put as-is into the connection module or it could be put into its own module.

Also splitting the functions does not solve the test problem, because the test will allow to change the code to make all all two or three functions behave the same and it would not fail as long as there is not actual functional check that traffic is not disturbed. Splitting the functions could be use to keep the code in the matching nm modules and also provide a module with a better API than the low-level modules. Since we are time-constraint it still seems to me to be the better intermediate solution to add the code just to the connection module because creating a new module layer in a nice way would probably take some time. Otherwise it will just be shallow code.

You have your opinion, I have mine. I do not block the review on this but you did not persuade me that your solution is better.

tyll:
Btw. the test is not stable when I run it locally, it fails about 50% of all runs:

```
pytest -k test_interface_ipv4_change -vv tests/integration/; pytest -k test_interface_ipv4_change -vv tests/integration/
=========================================================================================== test session starts ============================================================================================
platform linux -- Python 3.7.3, pytest-3.6.4, py-1.5.4, pluggy-0.6.0 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /workspace/nmstate, inifile: tox.ini
plugins: cov-2.5.1
collected 87 items / 86 deselected                                                                                                                                                                         

tests/integration/nm/ipv4_test.py::test_interface_ipv4_change FAILED                                                                                                                                 [100%]

================================================================================================= FAILURES =================================================================================================
________________________________________________________________________________________ test_interface_ipv4_change ________________________________________________________________________________________

eth1_up = None

    def test_interface_ipv4_change(eth1_up):
        with mainloop():
            _modify_interface(
                ipv4_state={
                    'enabled': True,
                    'dhcp': False,
                    'address': [
                        {'ip': IPV4_ADDRESS1, 'prefix-length': 24}
                    ]
                }
            )
    
        nm.nmclient.client(refresh=True)
        ipv4_current_state = _get_ipv4_current_state(ETH1)
    
        ip4_expected_state = {
            'enabled': True,
            'dhcp': False,
            'address': [
                {'ip': IPV4_ADDRESS1, 'prefix-length': 24}
            ]
        }
>       assert ip4_expected_state == ipv4_current_state
E       AssertionError: assert {'address': [...nabled': True} == {'dhcp': False...abled': False}
E         Common items:
E         {'dhcp': False}
E         Differing items:
E         {'enabled': True} != {'enabled': False}
E         Left contains more items:
E         {'address': [{'ip': '192.0.2.251', 'prefix-length': 24}]}
E         Full diff:
E         + {'dhcp': False, 'enabled': False}
E         - {'address': [{'ip': '192.0.2.251', 'prefix-length': 24}],
E         -  'dhcp': False,
E         -  'enabled': True}

tests/integration/nm/ipv4_test.py:50: AssertionError
================================================================================= 1 failed, 86 deselected in 2.14 seconds ==================================================================================
=========================================================================================== test session starts ============================================================================================
platform linux -- Python 3.7.3, pytest-3.6.4, py-1.5.4, pluggy-0.6.0 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /workspace/nmstate, inifile: tox.ini
plugins: cov-2.5.1
collected 87 items / 86 deselected                                                                                                                                                                         

tests/integration/nm/ipv4_test.py::test_interface_ipv4_change PASSED                                                                                                                                 [100%]

================================================================================= 1 passed, 86 deselected in 2.39 seconds ==================================================================================

```

libnmstate/nm/device.py
tyll: Another idea to at least allow to disable this workaround when testing would be to check for an environment variable, for example `NMSTATE_DISABLE_WORKAROUND_RHBZ1702657` and skip the workaround if the variable is set.

libnmstate/nm/connection.py
```diff
@@ -95,7 +95,7 @@ def commit(self, save_to_disk=True, nmdev=None):
 
     def activate(self):
         self._mainloop.push_action(
-            self._safe_activate_async)
```
tyll: There is a typo in the commit msg: s/siletly/silenty.

tyll: Also `execute a` could be removed from the last sentence.

EdDev:
@tyll , I have changed the PR to focus on `reapply` without any special modifications.
It is not possible to show the bugs and limitations of reapply explicitly.

We can discuss the "fallback" mechanism and usage in the next PR (see #356).
There is also a PoC that uses link monitoring to confirm `reapply` correct operation: #355

EdDev:
> Btw. the test is not stable when I run it locally, it fails about 50% of all runs

I have not seen this on my runs (el7), smells like another bug waiting to be discovered.

tyll:
> @tyll , I have changed the PR to focus on `reapply` without any special modifications.
> It is now possible to show the bugs and limitations of reapply explicitly.
> 
> We can discuss the "fallback" mechanism and usage in the next PR (see #356).
> There is also a PoC that uses link monitoring to confirm `reapply` correct operation: #355

Is it possible to see the improved behavior with `ip monitor`? If so, how?

EdDev:
> Is it possible to see the improved behavior with ip monitor? If so, how?

In theory, you could run the specific test while `ip monitor link dev eth1` is running on a different shell and count the number of events (UP). At least, this is what I saw while running with #355 .

tests/integration/nm/wired_test.py
```diff
@@ -40,6 +45,25 @@ def test_interface_mtu_change(eth1_up):
     }
 
 
+def test_interface_mac_change_with_reapply_fails(eth1_up):
+    with pytest.raises(TestMainloopError):
+        _test_interface_mac_change()
```
tyll: What is the benefit of this extra internal function? It seems to be specific to this test and even shares most of the name.

EdDev: It is a bit more readable but mainly because of the modify PR that follows.

cathay4t:
I ran the CI against `tests/integration/nmipv4_test.py` about 100 times with code from PR #360. All pass.

libnmstate/nm/device.py
```diff
@@ -49,6 +49,68 @@ def delete(dev):
...
+
+    devname = src_object.get_iface()
+    try:
+        success = src_object.reapply_finish(result)
+    except Exception as e:
+        if mainloop.is_action_canceled(e):
+            logging.debug('Device reapply aborted on %s: error=%s', devname, e)
+        else:
+            mainloop.quit(
+                'Device reapply failed on {}: error={}'.format(devname, e))
```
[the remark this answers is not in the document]
EdDev: Inside the mainloop we cannot raise.
The mainloop needs to be gracefully terminated, otherwise only the timeout will end it (and in a non graceful manner).

Perhaps investigating exceptions inside the mainloop is to be investigated further, we should consider it.

libnmstate/nm/device.py
```diff
@@ -49,6 +49,68 @@ def delete(dev):
...
+        return
+
+    if success:
+        logging.debug('Device reapply succeeded: dev=%s', devname)
+        mainloop.execute_next_action()
+    else:
+        mainloop.quit(
+            'Device reapply failed: dev={}, error=unknown'.format(devname))
```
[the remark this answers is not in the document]
EdDev: We cannot raise inside the mainloop.

libnmstate/nm/nmclient.py
```diff
@@ -59,9 +59,9 @@ def client(refresh=False):
     return _nmclient
 
 
-def mainloop():
+def mainloop(refresh=False):
     global _mainloop
-    if _mainloop is None:
+    if _mainloop is None or refresh:
```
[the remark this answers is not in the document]
EdDev: A dedicated commit has been assigned for this change with the information.
I will add a doctext per your request, but for the context one needs to read the commit message to fully understand.

libnmstate/nm/device.py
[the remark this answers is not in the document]
cathay4t: https://nmstate.atlassian.net/browse/NMSTATE-208

PR #335, opened 2019-04-26
Reorganize container creation and add helper

tyll:
Here are various improvements to debug the CI. Please only approve but do not merge because docker hub needs to be re-configured before doing the merge to properly rebuild the container images.


packaging/README-dockerhub.md
```diff
@@ -39,3 +41,8 @@ different commit or tag, specify the `SOURCE_COMMIT` build argument:
 ```shell
 sudo docker build --no-cache --build-arg SOURCE_COMMIT=v0.0.4 -t nmstate/centos7-nmstate -f Dockerfile.centos7-nmstate .
 ```
+
+```shell
+./build-container.sh --extra-args "--build-arg SOURCE_COMMIT=v0.0.6" local/centos7-nmstate
```
EdDev: Could you please clarify how is this change related? What is local/centos7-nmstate used for?

packaging/build-container.sh
```diff
@@ -0,0 +1,67 @@
...
+    # remove leading tag prefix
+    container_name="${1#*/}"
+
+    # remove container name suffix
+    echo "${1}" | grep -q "/" && : tag_prefix="${1%/*}"
+
+    # assign default value in case argument did not contain a tag prefix
+    : ${tag_prefix:=nmstate}
+
+    docker build --no-cache ${extra_args} -t "${tag_prefix}/${container_name}" \
```
EdDev: What happened to `--rm` ?

I would also extract this part as well (into a function) and use a variable/constant to add the default arguments in order to make it clear and less hidden.

packaging/README-dockerhub.md
tyll: The `local` prefix comes from you. I moved the build instructions from the README here:
https://github.com/nmstate/nmstate/commit/a2011b00326c0989834559a7d624061c081068f2

packaging/build-container.sh
tyll: It is not used here: https://github.com/nmstate/nmstate/blob/master/automation/run-tests.sh#L160 and the manpage says it is true by default. I can also add it, I do not care atm.

You want a constant for `--no-cache` and the `tag_prefix`?

tyll:
> For the first commit, the automation/docker_*.sh scripts have been deleted instead of moved.

Not sure what you mean. They are shown as deleted and also git does not track whether a file was moved or deleted but decides this just on the file contents AFAIK.

packaging/build-container.sh
tyll: Now I understood your remark about the prefix and changed it to `nmstate`.

automation/run-tests.sh
```diff
@@ -156,31 +156,11 @@ function is_file_changed {
...
-
 function rebuild_container_images {
-    if (is_file_changed "$PROJECT_PATH/packaging" ||
-        is_file_changed "$PROJECT_PATH/automation"); then
+    if is_file_changed "$PROJECT_PATH/packaging"; then
 
-        rebuild_el7_base_container_image
-        rebuild_el7_container_image
-        rebuild_fed_container_image
+        ${PROJECT_PATH}/packaging/build-container.sh centos7-nmstate-base \
```
EdDev: nit:
```
${PROJECT_PATH}/packaging/build-container.sh \
            centos7-nmstate-base \
            centos7-nmstate-dev \
            fedora-nmstate-dev
```

packaging/README-dockerhub.md
```diff
@@ -39,3 +41,8 @@ different commit or tag, specify the `SOURCE_COMMIT` build argument:
 ```shell
 sudo docker build --no-cache --build-arg SOURCE_COMMIT=v0.0.4 -t nmstate/centos7-nmstate -f Dockerfile.centos7-nmstate .
 ```
+
```
EdDev: Maybe you should add an OR here, as by mistake someone may think they need to run both commands (the one from line 42).

packaging/build-container.sh
```diff
@@ -0,0 +1,70 @@
...
+    local extra_args
+
+    extra_args="${1}"
+    shift
+
+    # remove leading tag prefix
+    container_name="${1#*/}"
+
+    # remove container name suffix
+    echo "${1}" | grep -q "/" && : tag_prefix="${1%/*}"
```
EdDev: I do not understand what this does exactly.

packaging/docker_sys_config.sh
```diff
@@ -1,12 +1,25 @@
...
-    /etc/systemd/journald.conf
-sed -i 's/#RateLimitBurst=1000/RateLimitBurst=0/' \
+
+cat > /etc/NetworkManager/conf.d/97-trace-logging.conf <<EOF
+[logging]
+level=TRACE
+domains=ALL
+EOF
+
+sed -i -e 's/^#RateLimitInterval=/RateLimitInterval=0/' \
```
EdDev: What happened to the existing values? You do not need to replace them as well?

nit: Maybe add an `\` after `sed -i`.

automation/run-tests.sh
```diff
@@ -189,8 +193,8 @@ while true; do
         DOCKER_IMAGE="nmstate/centos7-nmstate-dev"
         ;;
     --help)
-        echo -n "$0 [--copr=...] [--pytest-args=...] [--help] [--debug-shell] "
-        echo -n "[--el7] "
+        echo -n "$0 [--copr=...] [--customize=...] [--debug-shell] [--el7]"
```
EdDev: nit: I think the order should be per importance and usage, placing `copr` and this new one first is not that useful/helpful for a user.

automation/open_container_shell.sh
```diff
@@ -0,0 +1,5 @@
+#! /bin/bash
+
+container_id="$(docker ps --format '{{.ID}} {{.Image}}' | grep nmstate-dev \
```
EdDev: I think it needs to warn or fail if there are more than one instance detected, otherwise it may be dangerous (changing the wrong container).

automation/run-tests.sh
tyll: Why is this better?

packaging/README-dockerhub.md
tyll: Good point, I will just remove the first line. Not sure if I intended it to be there.

packaging/build-container.sh
tyll: If $1 contains a slash, it will store everything before the slash in `tag_prefix`.

tyll:
> Reviewed "containers: Unify container names".
> I still do not understand why two files appear as deleted here (there is no counterpart file that was added instead).

The files already exist in the the target directory for building the container images that are defined there.

automation/run-tests.sh
tyll: I guess or preferences differ. For me a sorted list makes it easier to find an argument that I vaguely remember or assume.

automation/open_container_shell.sh
tyll: Not sure why it is dangerous but I changed it to warn...

packaging/docker_sys_config.sh
tyll: What do you mean with existing values? The file conains only commented values and the previous script did also only change the commented lines.

I do not believe atm that the backslash is needed there.

EdDev: > I do not believe atm that the backslash is needed there.

You are just splitting in an unexpected place. Two `-e`, deserves two lines IMO.
But again, if you do not like it, it is not a must.

The comment has a value, like: `#RateLimitInterval=30`. If youdo not include the `30`, then it will not be replaced/removed. Or do I miss something here?

tyll: It just matches the line prefix since there is no dollar sign to match the end

tyll: Yes, you are right. I confused this with how the ansible lineinfile module works. Will fix this by adding `.*` after the equal sign.

automation/run-tests.sh
tyll: Actually the order reflects the current order of importance for me, too ;-P - If we want to overengineer the script we need to make `--test-type` an argument, since it is not really optional (except for the default of all), so it would be `run-tests.sh integ`. Not really in the mood to do this now, though. I added a new commit to make the help a least readable by disabling xtrace for it.

libnmstate/nm/connection.py
```diff
@@ -156,6 +156,10 @@ def _safe_activate_async(self):
 
         specific_object = None
         user_data = cancellable
+
+        logging.debug(
```
EdDev: Why this additional logging is needed?
We have a log when the _safe_activate_async method is called, when it exits early and when the callback is called.
What scenario required this additional log?

automation/run-tests.sh
EdDev: Please feedback on this so I'll know you have not missed it.

tyll: The line break is there to keep the line under 80 characters. I will move building all images to the build-container.sh scripts since it makes more sense to have the list there.

libnmstate/nm/connection.py
tyll: When activating failed without udev, there was no log message without this about how nmstate tried to activate something (i.e. by device or by name or both). This log message ensures that this is logged.

tyll:
Updated and rebased

libnmstate/nm/connection.py
EdDev: We are logging the execution of this method, but without the arguments:
https://github.com/nmstate/nmstate/blob/master/libnmstate/nm/nmclient.py#L88

At some time in the past the argument logging were removed because they seemed not useful and cause a lot of "noise" in the log.

With this logging addition, a duplication is created and it does not confirm with the other async methods/functions logging. On the other hand, I understand you have been missing a piece of information.

My suggestion is to solve this through the timeout cancellation. In case a timeout occurs or the execution of the callback function is cancelled in general, log the relevant info.
 For this specific case, have you seen a log like the one at line 181? I would expect that in case the udev is not operational, a timeout occurs and the CB is cancelled, outputting an error log.

tyll: Only logging the activation type on errors does not allow to compare successful runs with broken runs because the successful runs do not log the full information. Are the previous commits ready to be merged? I can open a new PR to get at least them merged.

EdDev: But it is logged on success as well.. check the callback method registered.

packaging/Dockerfile.centos7-nmstate
```diff
@@ -2,10 +2,15 @@ FROM nmstate/centos7-nmstate-base
...
 
+# dbus-python is special and does not provide egg files, therefore it is not
+# used by pip and needs to be compiled from PyPi. For this the build
+# dependencies are needed.
+# https://bugzilla.redhat.com/show_bug.cgi?id=1654774
 RUN yum -y install \
         python2-pip \
         git \
     && \
+    yum-builddep -y dbus-python && \
```
EdDev: So this is needed to install nmstate using `pip`?
I had to install `make` and `dbus-devel` for this to work, see https://github.com/nmstate/nmstate/pull/298

tyll: yum-builddep  uses the build dependencies as defined for the dbus-python RPM, so it should ensure that nothing is forgotten and we do not need to track the dependencies by hand.

PR #336, opened 2019-04-29
automation: Fix Fedora29 dev image

EdDev:
Udev has stopped working in the Fedora dev image.
It seems that the source is related to a missing systemd-udev package.

The container dockerfile has been updated to include systemd-udev.
In addition, an update has been included to confirm with the systemd
container enablement instructions:
https://hub.docker.com/r/fedora/systemd-systemd/

systemd-udevd is also being started when the container is run and NM is
restarted in order to confirm it loads with udev activated already.

### DIFF
--- a/automation/Dockerfile.fedora
+++ b/automation/Dockerfile.fedora
@@ -12,6 +12,7 @@ RUN dnf -y install --setopt=install_weak_deps=False \
                    NetworkManager \
                    NetworkManager-ovs \
                    openvswitch \
+                   systemd-udev \
                    \
                    python2-dbus \
                    python2-ipaddress \
--- a/automation/run-tests.sh
+++ b/automation/run-tests.sh
@@ -258,8 +258,11 @@ then
 fi
 
 docker_exec 'while ! systemctl is-active dbus; do sleep 1; done'
+docker_exec 'systemctl start systemd-udevd
+             while ! systemctl is-active systemd-udevd; do sleep 1; done
+'
 docker_exec '
-    systemctl start NetworkManager
+    systemctl restart NetworkManager
     while ! systemctl is-active NetworkManager; do sleep 1; done
 '
 pyclean


tyll:
Since this indicates a packaging/configuration error in NetworkManager, I filed https://bugzilla.redhat.com/show_bug.cgi?id=1704456

I am not sure about what you mean about updating in regard to the systemd docker recommendations - is it the dnf update?

automation/Dockerfile.fedora
```diff
@@ -6,7 +6,12 @@ FROM fedora
 ENV container docker
 COPY docker_enable_systemd.sh docker_sys_config.sh ./
 
-RUN bash ./docker_enable_systemd.sh && rm ./docker_enable_systemd.sh
+RUN \
+    dnf -y update && \
+    dnf install -y systemd-udev && \
```
tyll: IMHO this should be added to the NM section in the other dnf command. Also a reference to https://bugzilla.redhat.com/show_bug.cgi?id=1704456 would be great so it is clear that it can be removed eventually.

automation/run-tests.sh
```diff
@@ -258,8 +258,11 @@ then
 fi
 
 docker_exec 'while ! systemctl is-active dbus; do sleep 1; done'
+docker_exec 'systemctl start systemd-udevd
```
tyll: Is this really needed?

tyll: Also, if it is needed, the CentOS images should be fixed to require systemd-udevd as well.

automation/run-tests.sh
```diff
@@ -258,8 +258,11 @@ then
...
+             while ! systemctl is-active systemd-udevd; do sleep 1; done
+'
 docker_exec '
-    systemctl start NetworkManager
+    systemctl restart NetworkManager
```
tyll: Why is this needed?

automation/Dockerfile.fedora
EdDev: I somehow considered the udev, part of systemd patching, that is why I placed it here.
But I guess it is not related, I will move it.

automation/run-tests.sh
EdDev: On CentoOS it seems to be part of the systemd package and it comes enabled.
On Fedora, it was a requirement of `dracut`, but now they seems to have dropped the requirement.

So yes, it is needed otherwise it does not work.

automation/run-tests.sh
EdDev: Based on my tests, the NM service needs to be restarted after systemd-udevd is running, otherwise it does not work. That is why we need to wait for that daemon to be active before we proceed.

automation/run-tests.sh
EdDev: Correction: `dracut` itself got dropped from the base image, that is why `systemd-udev` is now missing, not because of requirement changes in `dracut`.

automation/run-tests.sh
tyll: this would be strange because we did not have to restart NM before and the breaking change was that systemd-udev was not installed anymore.

tyll:
> > Since this indicates a packaging/configuration error in NetworkManager, I filed https://bugzilla.redhat.com/show_bug.cgi?id=1704456
> 
> Ok, but I am not sure if applications should depend on udevd in this manner. Only `dracut` seemed to have depended on it.

NetworkManager depends on udevd, otherwise we would not have this problem. The problem is that the package does not express this dependency properly. Also I am not sure why it would be wrong for an application to depend on udevd.

> > I am not sure about what you mean about updating in regard to the systemd docker recommendations - is it the dnf update?
> 
> Yes, but I think we should do this change in a separate PR, it is not a direct fix to the current problem. I will try to drop it.

I am very confident that this is not really a requirement to run systemd.

automation/Dockerfile.fedora
```diff
@@ -9,6 +9,7 @@ COPY docker_enable_systemd.sh docker_sys_config.sh ./
 RUN bash ./docker_enable_systemd.sh && rm ./docker_enable_systemd.sh
 
 RUN dnf -y install --setopt=install_weak_deps=False \
+                   systemd-udev \
```
tyll: Please keep the package list sorted.

EdDev: Done

PR #337, opened 2019-05-02
Nm wired return existing

EdDev:
Dependent on #325 
nm.wired: Return existing setting instead of nothing

The wired setting is affecting the reapply action when it is missing
from the connection profile [1].
In order to adjust to this behaviour, in cases where no new wired
parameters are mentioned in the desired state and an existing profile
exists, the current wired setting is returned.

1. https://bugzilla.redhat.com/1703960


EdDev:
> MHO a comment in the code would make it more likely that this special behavior will be remembered when refactoring the code.

I do not like most inline comments, this one included. Such comments , in many cases, rot because no one changes them when they should change.
If someone does not understand why it is implemented as it is, git 'blame' is available.

PR #338, opened 2019-05-02
Netapplier: Optimize add and edit processing

EdDev:
netapplier: Optimize new and existing interface processing

The desired state is processed to determine which interfaces are new and
which ones need editing.
This change optimizes the setup flow by not including the new interfaces
in the editation step.


EdDev:
The schema changes came in a separate commit. It could have been a separate PR indeed, but as it was a decorative pre-step I included it here.

PR #339, opened 2019-05-02
Modify integrated

EdDev:
Integrate modify into nmstate main setup flow.

Depends on #378


EdDev:
> Since this patch enable the reapply at netapplier level, please add a integration test case for it.

The test has been added (Linux bridge) but the results are not good (xfail).
The reapply seems to work fine when using the nm provider directly (see #379 ) which hints that the problem may be cause by an existing bug in NM which nmstate merge reveals.

I would suggest taking this change in, despite the issues seen. Further analysis is required to pin point the problem, but keeping this change in will assure it is tracked (the test and the issue opened).

PR #340, opened 2019-05-05
Canonicalize ip subtree

EdDev:
Complement IPv{4,6} keys in the interface state and assume they are there when processing the state.
The same question should not be asked multiple times.

### DIFF
--- a/libnmstate/state.py
+++ b/libnmstate/state.py
@@ -54,6 +54,7 @@ class State(object):
     def __init__(self, state):
         self._state = copy.deepcopy(state)
         self._ifaces_state = State._index_interfaces_state_by_name(self._state)
+        self._complement_interface_empty_ip_subtrees()
 
     def __eq__(self, other):
         return self.state == other.state
@@ -79,6 +80,13 @@ class State(object):
     def interfaces(self):
         """ Indexed interfaces state """
         return self._ifaces_state
+
+    def _complement_interface_empty_ip_subtrees(self):
+        """ Complement the interfaces states with empty IPv4/IPv6 subtrees. """
+        for iface_state in six.viewvalues(self.interfaces):
+            for family in (Interface.IPV4, Interface.IPV6):
+                if family not in iface_state:
+                    iface_state[family] = {}
 
     def sanitize_ethernet(self, other_state):
         """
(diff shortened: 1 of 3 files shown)


cathay4t:
Not sure I correct or not, but some interface type (bridget) does not support IPv4 or IPv6.

EdDev:
> Not sure I correct or not, but some interface type (bridget) does not support IPv4 or IPv6.

Currently, only the OVS bridge does not support a stack over it... actually it does not support many other properties and there is not kernel interface backing it. (pending changing, as we agreed to use an internal interface for the bridge and represent it through it)

In any case, the fact that we add the key with an empty dict does not mean there is a stack, that is why we have the `enabled` property.

tyll:
IMHO the method does not create a canonical form because it actually just adds a abnormal version since you already said that there should be `enabled: false` in case the stack is not there. Therefore the name is misleading IMHO. In general I would prefer an approach where the empty/default subtrees are added when the state is created and then updated instead of adding it at an arbitrary later time.

EdDev:
It creates a “stable” form which includes the information other steps need later on. If you prefer a different name, please suggest it.
Regarding the place it should be executed, I have no ideologic objection to relocate it, but I also do not think it is worth the effort to recheck it all and see that it does not collide with something else. You could just accept it and propose an optimization for it later.

This change has been triggered by Gris route PR.

tyll:
> It creates a “stable” form which includes the information other steps need later on. If you prefer a different name, please suggest it.

It does not add any information, it just adds empty IP subtrees. Another name for the current state would be `add_empty_ip_subtrees()`.

> Regarding the place it should be executed, I have no ideologic objection to relocate it, but I also do not think it is worth the effort to recheck it all and see that it does not collide with something else. You could just accept it and propose an optimization for it later.

If it does not have a misleading name, I am fine with accepting it. But I do not believe that this is a great design.

EdDev:
@tyll , I have moved the normalization into the constructor and changed the name. Hope it is ok now.

libnmstate/state.py
```diff
@@ -80,6 +81,15 @@ def interfaces(self):
         """ Indexed interfaces state """
         return self._ifaces_state
 
+    def _complement_interface_empty_ip_subtrees(self):
+        """ Complement the state interfaces with:
+          - An empty IPvX subtree.
```
tyll: I suggest the following docstring:
`Complement the interfaces' states with empty IPv4/IPv6 subtrees.`

PR #343, opened 2019-05-08
checkpoints: Properly report missing permissions

tyll:

### DIFF
--- a/libnmstate/netapplier.py
+++ b/libnmstate/netapplier.py
@@ -29,6 +29,7 @@ from libnmstate import state
 from libnmstate import validator
 from libnmstate.error import NmstateConflictError
 from libnmstate.error import NmstateLibnmError
+from libnmstate.error import NmstatePermissionError
 from libnmstate.error import NmstateValueError
 from libnmstate.nm import nmclient
 
@@ -130,6 +131,8 @@ def _apply_ifaces_state(desired_state, verify_change, commit,
                 _verify_change(desired_state)
         if not commit:
             return checkpoint
+    except nm.checkpoint.NMCheckPointPermissionError:
+        raise NmstatePermissionError('Error creating a check point')
     except nm.checkpoint.NMCheckPointCreationError:
         raise NmstateConflictError('Error creating a check point')
 
--- a/libnmstate/nm/checkpoint.py
+++ b/libnmstate/nm/checkpoint.py
@@ -28,6 +28,9 @@ CHECKPOINT_CREATE_FLAG_DESTROY_ALL = 0x01
 CHECKPOINT_CREATE_FLAG_DELETE_NEW_CONNECTIONS = 0x02
 CHECKPOINT_CREATE_FLAG_DISCONNECT_NEW_DEVICES = 0x04
 
+NM_PERMISSION_DENIED = 'org.freedesktop.NetworkManager.PermissionDenied'
+
+
 _nmdbus_manager = None
 
 
@@ -40,6 +43,10 @@ class NMCheckPointError(Exception):
 class NMCheckPointCreationError(NMCheckPointError):
     """ Error creating a Network Manager CheckPoint """
 
+    pass
+
+
+class NMCheckPointPermissionError(NMCheckPointError):
     pass
 
 
@@ -118,6 +125,8 @@ class CheckPoint(object):
                           dbuspath, timeout)
             self._dbuspath = dbuspath
         except dbus.exceptions.DBusException as e:
+            if e.get_dbus_name() == NM_PERMISSION_DENIED:
+                raise NMCheckPointPermissionError(str(e))
             raise NMCheckPointCreationError(str(e))
 
     def destroy(self):
--- a/libnmstate/nm/checkpoint.py
+++ b/libnmstate/nm/checkpoint.py
@@ -35,14 +35,10 @@ _nmdbus_manager = None
 
 
 class NMCheckPointError(Exception):
-    """ Error creating a Network Manager CheckPoint """
-
     pass
 
 
 class NMCheckPointCreationError(NMCheckPointError):
-    """ Error creating a Network Manager CheckPoint """
-
     pass
 
 
--- a/nmstatectl/nmstatectl.py
+++ b/nmstatectl/nmstatectl.py
@@ -32,6 +32,7 @@ import yaml
 from libnmstate import netapplier
 from libnmstate import netinfo
 from libnmstate.error import NmstateConflictError
+from libnmstate.error import NmstatePermissionError
 from libnmstate.error import NmstateValueError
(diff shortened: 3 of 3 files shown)


libnmstate/netapplier.py
```diff
@@ -124,6 +125,8 @@ def _apply_ifaces_state(desired_state, verify_change, commit,
                 _verify_change(desired_state)
         if not commit:
             return checkpoint
+    except nm.checkpoint.NMCheckPointPermissionError:
+        raise NmstatePermissionError('Error creating a check point')
```
EdDev: re-raising in this manner causes a loss of information about the origin of the error.
It is preferred to use `six.reraise`. You may then use the original error text instead of a new one.

You can use `six.reraise` in this manner:
```
except nm.checkpoint.NMCheckPointPermissionError:    
    _, value, tb = sys.exc_info()
    six.reraise(NmstatePermissionError, NmstatePermissionError(*value.args), tb)
```

I would also consider logging the exception using `logging.exception`, mainly because `nmstatectl` serializes it to stderr and losses the traceback.

libnmstate/nm/checkpoint.py
```diff
@@ -43,6 +46,12 @@ class NMCheckPointCreationError(NMCheckPointError):
     pass
 
 
+class NMCheckPointPermissionError(NMCheckPointError):
+    """ Error creating a Network Manager CheckPoint """
```
EdDev: I think the name of the error is informative enough, this extra doctext adds no informative info.

libnmstate/nm/checkpoint.py
```diff
@@ -43,6 +46,12 @@ class NMCheckPointCreationError(NMCheckPointError):
     pass
 
 
+class NMCheckPointPermissionError(NMCheckPointError):
```
EdDev: Isn't this a specific creation error? If so, it needs to be a child of `NMCheckPointCreationError`.

libnmstate/nm/checkpoint.py
```diff
@@ -118,7 +127,10 @@ def create(self):
                           dbuspath, timeout)
             self._dbuspath = dbuspath
         except dbus.exceptions.DBusException as e:
-            raise NMCheckPointCreationError(str(e))
+            if e.get_dbus_name() == NM_PERMISSION_DENIED:
+                raise NMCheckPointPermissionError(str(e))
+            else:
```
EdDev: There is no need to the `else`.

libnmstate/nm/checkpoint.py
tyll: Currently it is only used in the creation step but commit and rollback should probably handle this the same way and I do not think we need a separate exception for each case.

libnmstate/nm/checkpoint.py
tyll: fixed, also adjusted the other docstrings

libnmstate/netapplier.py
tyll: In current Python 3, `raise Exception()` just does the right thing AFAIU. What is the advantage of using the complicated `six.reraise()` call? This method seems more like something to improve python 2 instead of only providing compatibility since `raise Exception()` works fine in both Python versions but Python 3 just keeps the origin.

libnmstate/nm/checkpoint.py
tyll: fixed

tyll:
@EdDev ready for re-review

libnmstate/netapplier.py
EdDev: I was under the impression that we write code to support both py2 and py3 at the moment.
I would prefer to keep this going for the sake of VDSM for the next year.

I do not think it is that horrible to keep the dual support here, but I will not fight over it too much.

tyll: I am fine with supporting Python 2 but this request seems like adding extra features to the Python 2 code that are not needed for compatibility. Also I am not even sure that the code is really the best for Python 3. If you can explain in more detail why this approach is better, I might reconsider. However, in Python 3 just using raise will set the `__cause__` attribute properly which seems to me to be better that mangling with the traceback.

EdDev: `six.reraise` is used to keep the traceback info across py2 and py3. I think that for py3 it uses the same mechanism py3 uses.
The only advantage of using `six.reraise` is to keep the tb when using py2.

tyll: Using six in the way you described makes the default traceback handler in python3 less clear IMHO.

Test script:
```
cat > 'six_reraise.py' <<EOF
#!/usr/bin/python3 -tt
# vim: fileencoding=utf8

import six
import sys

def raise_a():
    raise RuntimeError()

def raise_b():
    try:
        raise_a()
    except:
        raise ValueError()

def reraise_b():
    try:
        raise_a()
    except:
        six.reraise(ValueError, None)


def ed_reraise_b():
    try:
        raise_a()
    except:
        _, value, tb = sys.exc_info()
        six.reraise(ValueError, ValueError(*value.args), tb)

if len(sys.argv) > 1:
    if sys.argv[1] == "e":
        ed_reraise_b()
    else:
        reraise_b()
else:
    raise_b()
EOF
```

Output with your suggested code:

```
python3 six_reraise.py e
  File "six_reraise.py", line 25, in ed_reraise_b
    raise_a()
  File "six_reraise.py", line 8, in raise_a
    raise RuntimeError()
RuntimeError

During handling of the above exception, another exception occurred:

  File "six_reraise.py", line 32, in <module>
    ed_reraise_b()
  File "six_reraise.py", line 28, in ed_reraise_b
    six.reraise(ValueError, ValueError(*value.args), tb)
  File "/usr/lib/python3.7/site-packages/six.py", line 692, in reraise
    raise value.with_traceback(tb)
  File "six_reraise.py", line 25, in ed_reraise_b
    raise_a()
  File "six_reraise.py", line 8, in raise_a
    raise RuntimeError()
ValueError
```

Output with just using raise:

```
python3 six_reraise.py
  File "six_reraise.py", line 12, in raise_b
    raise_a()
  File "six_reraise.py", line 8, in raise_a
    raise RuntimeError()
RuntimeError

During handling of the above exception, another exception occurred:

  File "six_reraise.py", line 36, in <module>
    raise_b()
  File "six_reraise.py", line 14, in raise_b
    raise ValueError()
ValueError
```

Observations:

- with your code, the traceback for the RuntimeError is shown twice
- the traceback for the ValueError ends with a raise for the RuntimeError, which is IMHO irritating. 
- Also the raise of the ValueError is IMHO obfuscated in the traceback instead of clearly shown

I do not see any value that `six.reraise()` brings for Python 3. And for the python 2 output, it still irritates me that the traceback does not end with the code that is actually raising the exception. Therefore I am not convinced that using `six.reraise()` is an improvement.

PR #344, opened 2019-05-12
netapplier, validator: Refactor validator module

EdDev:
Refactor the validator entry point input to accept the desired and
current state objects.

### DIFF
--- a/libnmstate/validator.py
+++ b/libnmstate/validator.py
@@ -50,21 +50,19 @@ def verify_interface_capabilities(ifaces_state, capabilities):
                 "and started")
 
 
-def verify_interfaces_state(ifaces_state, ifaces_desired_state):
-    verify_link_aggregation_state(ifaces_state, ifaces_desired_state)
+def verify_interfaces_state(desired_state, current_state):
+    verify_link_aggregation_state(desired_state, current_state)
 
 
-def verify_link_aggregation_state(ifaces_state, ifaces_desired_state):
+def verify_link_aggregation_state(desired_state, current_state):
     available_ifaces = {
-        iface_state['name'] for iface_state in ifaces_state
-        if iface_state.get('state') != 'absent'
-    }
-    available_ifaces |= {
-        ifname for ifname, ifstate in six.viewitems(ifaces_desired_state)
+        ifname for ifname, ifstate in six.viewitems(desired_state.interfaces)
         if ifstate.get('state') != 'absent'
     }
+    available_ifaces |= set(current_state.interfaces)
+
     specified_slaves = set()
-    for iface_state in ifaces_state:
+    for iface_state in six.viewvalues(desired_state.interfaces):
         if iface_state.get('state') != 'absent':
             link_aggregation = iface_state.get('link-aggregation')
             if link_aggregation:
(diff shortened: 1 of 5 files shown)


tests/lib/validator_test.py
```diff
@@ -18,104 +18,122 @@
...
+                    'name': 'bond1',
+                    'link-aggregation': {
+                        'slaves': []
+                    },
+                }
+            ]
+        })
+
+        libnmstate.validator.verify_link_aggregation_state(desired_state,
+                                                           state.State({}))
```
tyll: How about an empty_state fixture? The parens and curly braces hurt my eyes. (not a must)

libnmstate/validator.py
```diff
@@ -50,21 +50,19 @@ def verify_interface_capabilities(ifaces_state, capabilities):
                 "and started")
 
 
-def verify_interfaces_state(ifaces_state, ifaces_desired_state):
-    verify_link_aggregation_state(ifaces_state, ifaces_desired_state)
+def verify_interfaces_state(desired_state, current_state):
```
tyll: Maybe this is also a good opportunity to rename it to `validate_...` - it is irritating that it is called verify but does not the usual verification.

tyll: strictly speaking: This method is not covered by the unit tests AFAICS because the unit tests only cover the one function that this function calls. For KISS it would make sense to drop this function IMHO.

EdDev: Agreed, changing to `validate`.

This function is an abstract layer, supposed to handle interfaces state verification in general.
At the moment we have only one, and it makes sense to name that code based on what it does (link aggregation validation).
I think it is fine as it is now, but if you can think of a quick additional validation, maybe we can just add it and make it more clear.

tests/lib/validator_test.py
EdDev: Using a fixture here feels an overuse of fixtures.
These are the options, please let me know what you prefer (the first are locals to the test module):
1. state.State(EMPTY_STATE)
2. empty_state()
3. state.empty_state()
4. state.State.empty()

libnmstate/validator.py
tyll: IMHO we can add the abstraction when we actually need it/when we have multiple things to validate. (not a must)

tests/lib/validator_test.py
tyll: `state.State()` could just return an empty state, i.e. use an empty dict by default in the constructor.

EdDev: I am trying to avoid defaults as much as possible as a rule, in this case it is pretty obvious this is a corner case and making it implicit does not make sense.
I took option 2 for now.

BTW, this now shouts that additional unit tests are needed.

libnmstate/validator.py
EdDev: I think we can add the dhcp validation, will check this later on.

tests/lib/validator_test.py
tyll: Why are defaults bad?

tyll: Also IMHO most Python developers would expect `state.State()` to be an empty state just as `dict()` is an empty dictionary or `list()` an empty list...

EdDev: > Why are defaults bad?

Explicit vs implicit.
Default hide assumptions the coder had. Sometimes they make sense to most, sometimes it makes sense only to some, sometime it makes sense only for one.
As someone who reads the code, I would not expect anyone to create an 'empty' state.

PR #345, opened 2019-05-13
nm.wired: Introduce WiredSetting

EdDev:


libnmstate/nm/wired.py
```diff
@@ -23,61 +23,93 @@
...
 
-def create_setting(iface_state, base_con_profile):
-    mtu = iface_state.get('mtu')
-    mac = iface_state.get(Interface.MAC)
+class WiredSetting(object):
+    def __init__(self, state):
+        self.mtu = state.get(Interface.MTU)
+        self.mac = state.get(Interface.MAC)
+
+        ethernet = state.get('ethernet', {})
```
tyll: You get bonus points from me if you would also add a commit to convert these strings to constants. :-)

EdDev: Yes, it crossed my mind.

EdDev: Will post it in a separate PR, makes sense to change it all over the project.
BTW, It also revealed a bug we have: in the schema and some tests we have the ethernet specific properties under the interface subtree (flat) and not under the `ethernet` subtree.

EdDev:
> The commit message is not explaining why this is a good idea.

Done

EdDev:
@tyll , unit tests added with + fixes.

tests/lib/nm/wired_test.py
```diff
@@ -113,3 +114,118 @@ def test_get_info_with_invalid_duplex(ethtool_mock, NM_mock):
...
+        state = {
+            schema.Interface.MTU: 1500,
+            schema.Interface.MAC: 'abc',
+        }
+        obj = nm.wired.WiredSetting(state)
+
+        assert obj
+
+    def test_equality_for_empty_states(self):
+        state = {}
```
tyll: I wonder if it would be better to use two different dicts here as well...

EdDev: The identity of the dict should not matter.
Different dicts with the same content is checked later on, so I guess it is ok to exercise both.

tests/lib/nm/wired_test.py
```diff
@@ -113,3 +114,118 @@ def test_get_info_with_invalid_duplex(ethtool_mock, NM_mock):
...
     }
+
+
+class TestWiredSetting(object):
+
+    def test_identity(self):
+        state = {}
+        obj1 = obj2 = nm.wired.WiredSetting(state)
+
+        assert obj1 == obj2
```
tyll: Just wondering, how is it possible to create two objects where the first assert succeeds but the second one does not (or the other way round).

tests/lib/nm/wired_test.py
tyll: Why should it not matter? You do not seem to trust anything in these tests and the test name says "test equality for *empty* states" not "test equality for identical states". It is not a blocker for me, though.

tests/lib/nm/wired_test.py
EdDev: Try to remove `__ne__` and you will find out :)
I `__eq__` has been implemented, usually `__ne__` will call it with a `not`. However, in PY2 it is not done by default (there is a need to implement `__ne__` explicitly) and in some special cases, one may implement the `__ne__` by not negating equality, but implementing something else which is "optimized".

tests/lib/nm/wired_test.py
EdDev: I almost believed you here :)
Identity is between the objects, not the dict it gets as input.
There is no special handling for identical dicts, the data from the dict is extracted to attributes and the attributes values are checked for equality.

PR #346, opened 2019-05-14
Schema ethernet constants

EdDev:
Dependent on #345

### DIFF
--- a/libnmstate/nm/wired.py
+++ b/libnmstate/nm/wired.py
@@ -17,6 +17,7 @@
 
 from libnmstate.ethtool import minimal_ethtool
 from libnmstate.nm import nmclient
+from libnmstate.schema import Ethernet
 from libnmstate.schema import Interface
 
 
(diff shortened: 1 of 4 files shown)


EdDev:
TODO:
- Add the wired test modules to the existing (first) commit.
- Change the other modules (one or more commits).

libnmstate/nm/wired.py
```diff
@@ -92,24 +93,24 @@ def create_setting(iface_state, base_con_profile):
...
+            setting.speed = ethtool_results[Ethernet.SPEED]
         if not setting.duplex:
-            setting.duplex = ethtool_results['duplex']
+            setting.duplex = ethtool_results[Ethernet.DUPLEX]
 
     if setting.speed:
         nm_wired_setting.props.speed = setting.speed
 
-    if setting.duplex in ['half', 'full']:
+    if setting.duplex in [Ethernet.DUPLEX_HALF, Ethernet.DUPLEX_FULL]:
```
tyll: nit: IMHO `HALF_DUPLEX` is more natural. If you like to have a hierachy, `DUPLEX.HALF` would be better IMHO. (not a must)

EdDev: Your proposal feels both right and wrong, and I cannot say which wins. But it is worth changing if it is worth commenting about :)
Done

libnmstate/nm/wired.py
```diff
@@ -92,24 +93,24 @@ def create_setting(iface_state, base_con_profile):
...
+            setting.speed = ethtool_results[Ethernet.SPEED]
         if not setting.duplex:
-            setting.duplex = ethtool_results['duplex']
+            setting.duplex = ethtool_results[Ethernet.DUPLEX]
 
     if setting.speed:
         nm_wired_setting.props.speed = setting.speed
 
-    if setting.duplex in ['half', 'full']:
+    if setting.duplex in [Ethernet.HALF_DUPLEX, Ethernet.FULL_DUPLEX]:
```
tyll: sorry, another nitpick: It is probably better to make full duplex the first setting since I believe it will be used more often. Not sure it is worth the effort but it might be a possible optimization (with little effect probably).

EdDev: Sorry, too late. Please send a patch, it sounds fine to me.

PR #347, opened 2019-05-15
schema constants: Introduce Bond and BondMode

EdDev:
Use the new constants in the integration bond_test module.

Note: Bond options have not been specified in this change. It is left
for a following change.

### DIFF
--- a/libnmstate/schema.py
+++ b/libnmstate/schema.py
@@ -98,6 +98,25 @@ class InterfaceType(object):
     )
 
 
+class Bond(object):
+    KEY = InterfaceType.BOND
+    CONFIG_SUBTREE = 'link-aggregation'
+
+    MODE = 'mode'
+    SLAVES = 'slaves'
+    OPTIONS_SUBTREE = 'options'
+
+
+class BondMode(object):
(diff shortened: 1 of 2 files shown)


libnmstate/schema.py
```diff
@@ -98,6 +98,25 @@ class InterfaceType(object):
     )
```
tyll: The commit message irritates me. It says "Use the new constants" which implies that the constants already exist and are now used but AFAICS the change introduces them.

Suggestion for a better commit message: "Use constants for bond settings and mode"

The bond options seem to require only one more constant for the tests, so IMHO it would be easier just to add it instead of this partial change. Also it would be nicer to just change all instances if you do a dedicated PR just for the constants. The other options is to adjust at least the code that you are touching. But doing a special PR for just some changes seems like a lot of overhead.

EdDev: I find it odd the commit message has such an affect on you, I will try to adjust it.

Regarding the option and usages of the constants, it is too costly to invest at the moment. I prefer to improve in small steps and make it a bit better than before. My first thought about adding all the options was to defer it when one will add them all.

PR #348, opened 2019-05-15
tests integ: Refactor fixture setup of ethX interfaces

EdDev:
The ethX fixtures prepare the base interfaces per test by initializing
their configuration (performing a down-up action).

The existing flow has been initializing the interfaces before a test is
executed with no action taken when it ends. This causes in some
scenarios [1] leftovers from tests, affecting the following tests.

In order to resolve the post test leftovers, the test fixtures have been
adjusted to setup the interface at the test setup and to remove them
(set them `down`) at the test teardown.
In order to treat the initial state of the tests session execution,
where the interfaces may have an unknown state/configuration, a session
scoped fixture is used to initialize them (by setting them `down`).

1. Defining a bond/bridge with a slave/port.

---
This change is [Reviewable](https://reviewable.io/reviews/nmstate/nmstate/348)

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1,5 +1,5 @@
 #
-# Copyright 2018 Red Hat, Inc.
+# Copyright 2018-2019 Red Hat, Inc.
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


tyll:
> What leftovers? There is a reference to the scenario, if this is what you mean.

I mean this part: `1. Defining a bond/bridge with a slave/port.` I do not see how it belongs there.

> The iterative text is good for the topic and summaries first line. I do not think the body needs to be so strict about it.
> I will try to see if I can change it for your taste nevertheless.

Thank you for changing this! I am fine if you keep this in mind for future PRs. IMHO imperative is always better to describe changes in a patch btw.

EdDev:
> I mean this part: 1. Defining a bond/bridge with a slave/port. I do not see how it belongs there.

It is referred in the body:
> This causes in some scenarios [1] leftovers from tests, affecting the following tests.

tyll:
> > I mean this part: 1. Defining a bond/bridge with a slave/port. I do not see how it belongs there.
> 
> It is referred in the body:
> 
> > This causes in some scenarios [1] leftovers from tests, affecting the following tests.

Interesting, it did not look like a footnote to me (usually they are using the same mark...). IMHO there is no need for the footnote. `This causes leftovers when testing bonds/bridges with slaves/ports, affecting the following tests.` would be nicer IMHO. (not a must or something that I will block this PR for)

EdDev:
No problem, will change the commit.

PR #349, opened 2019-05-15
Integ tests nm mainloop

EdDev:
tests, nm.testlib: Introduce mainloop_run decorator

Start using the decorator in the nm.bond_test integration module.

### DIFF
--- a/tests/integration/nm/testlib.py
+++ b/tests/integration/nm/testlib.py
@@ -1,5 +1,5 @@
 #
-# Copyright 2018 Red Hat, Inc.
+# Copyright 2018-2019 Red Hat, Inc.
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -15,6 +15,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 from contextlib import contextmanager
+import functools
 
 from libnmstate import nm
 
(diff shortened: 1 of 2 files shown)


EdDev:
Dependent on #347

tests/integration/nm/testlib.py
```diff
@@ -25,3 +26,11 @@ def mainloop():
     yield
     success = mloop.run(timeout=5)
     assert success, 'Mainloop error: ' + mloop.error
+
+
+def mainloop_run(func):
```
tyll: On first thought, it should be `run_mainloop`

EdDev: I think decorators are not supposed to be verbs.
They are usually used to add a property/behaviour to a function/method/class.

tyll: I see.

tests/integration/nm/testlib.py
```diff
@@ -25,3 +26,11 @@ def mainloop():
...
+    @functools.wraps(func)
+    def wrapper_mainloop(*args, **kwargs):
```
tyll: nitpick/not a must: IMHO `wrap_mainloop` or `mainloop_wrapper` would be better.

EdDev: Hmm.. This is the func that wraps the original func and which will be called every time the original is referenced.
Based on that, it should probably be a verb, like the one you suggested above.
Not sure exactly, I used the pattern usef here: https://realpython.com/primer-on-python-decorators/
Other references do not help much (https://stackoverflow.com/questions/10957409/python-naming-conventions-in-decorators).

PR #350, opened 2019-05-15
Integ tests nm bond

EdDev:
tests integ: Add a nm.bond test for creating a bond with one slave


EdDev:
Depends on #349

EdDev:
Depends on #348 
(The integration tests fail without it)

tests/integration/nm/bond_test.py
```diff
@@ -87,3 +106,32 @@ def _delete_bond(devname):
...
     nm.device.delete(nmdev)
+
+
+@mainloop_run
+def _attach_slave_to_bond(bond, slave):
+    slave_nmdev = nm.device.get_device_by_name(slave)
+    curr_slave_con_profile = nm.connection.ConnectionProfile()
+    curr_slave_con_profile.import_by_device(slave_nmdev)
+
+    slave_con_profile = nm.connection.ConnectionProfile()
```
maiqueb: could you rename 'slave_con_profile' to something that helps understand what's this variable purpose (update the slave's settings - if I understood correctly) ?

Something like `slave_update_settings_connection` maybe.

EdDev: It is a variable representing an object and not a function/method, therefore naming it with a verb is not that good.
It's purpose it to represent a slave`s connection profile. I could add `new` as a prefix, but I do not know why it will be better or more readable (you will still need to read all the lines that follows).

Let me know if you find `new` any better and I'll change it.

maiqueb: can't say I agree, but I don't disagree enough to block this.

EdDev:
@maiqueb , do you still have comments on this PR?

maiqueb:
> @maiqueb , do you still have comments on this PR?

I had one asking for the ```slave_con_profile``` variable to be renamed. Which I still think it should.

> It is a variable representing an object and not a function/method, therefore naming it with a verb is not that good.

I don't believe so. Let's say you have an object called 'key'. That name alone says what it does - opens / locks something - but tells nothing as to where it should be used, thus, is a bad name.

It feels the same way with 'slave_con_profile'. What's the difference between 'curr_slave_con_profile' and 'slave_con_profile' ?... I personally think that renaming the variable to something that hints what's its use for is a good thing.

Having said that, I repeat what I said before - I won't block it.

EdDev:
> I don't believe so. Let's say you have an object called 'key'. That name alone says what it does - opens / locks something - but tells nothing as to where it should be used, thus, is a bad name.

`key` represents an object and how it should be used is dependent on its type and the available methods or function that can work on/with it. In most cases, you can represent what the object is and the context it is in using a proper name.
`slave_con_profile` represents a slave profile, based on how it was created, it is a new one.
`curr_...` represents an existing one (used `import`).

Anyway, as some time has passed already, please revisit this change and re-comment again if something that is a blocker is missing.
Please note that this PR is currently NACK.

PR #355, opened 2019-05-26
[PoC] Monitor link

EdDev:
It is now possible to monitor the interfaces for link changes, confirming that the `reapply` operation indeed does not cause a significant traffic disruption (link level).

Depends on #332

### DIFF
--- a/tests/integration/testlib/cache.py
+++ b/tests/integration/testlib/cache.py
@@ -1,0 +1,54 @@
+# Copyright 2017 Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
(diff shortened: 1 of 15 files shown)


EdDev:
Known issue: The lint checks fail.
This PR is aimed to show that the `reapply` indeed has the expected behaviour. Embedded these changes into the testlib sub-package requires more work.
(It also has some licensing implications if we move to lgpl)

tests/integration/testlib/cache.py
```diff
@@ -0,0 +1,54 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+#
+# Refer to the README and COPYING files for full details of the license
```
tyll: Just noticed this: This part of the header does not seem to be valid anymore. Also the link monitoring seems to be so complex that it should probably be moved into its own module/project and have tests on its own.

EdDev: The code has been transferred from VDSM with small adjustments and is here for the PoC.
Managing it as a stand-alone repo is a great idea, but it requires ownership and a larger amount of effort.
If we will decide it is useful for our needs, we could suggest it upstream.

EdDev:
An alternative is proposed in #360

PR #360, opened 2019-05-29
Reapply ip monitor check

EdDev:
Depends on #332

### DIFF
--- a/tests/integration/testlib/iproutelib.py
+++ b/tests/integration/testlib/iproutelib.py
@@ -1,0 +1,83 @@
+#
+# Copyright 2019 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+from contextlib import contextmanager
+import subprocess
+import threading
+import time
+
+
+TIMEOUT = 10
+
+
+class IpMonitorResult(object):
+    def __init__(self):
+        self.out = None
+        self.err = None
+        self.popen = None
+
+
+@contextmanager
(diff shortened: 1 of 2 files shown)


tests/integration/testlib/iproutelib.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def get_non_up_events(result, dev):
+    """
+    Given a result and device, filter only the non UP events (DOWN, UNKNOWN)
+    and return them as a list.
+    :param result: IpMonitorResult
+    :return: List of non UP events
+    """
+    return [l for l in result.out.split('\n')
```
cathay4t: I got error on python 3.
```
TypeError: a bytes-like object is required, not 'str'
```

tests/integration/testlib/iproutelib.py
```diff
@@ -0,0 +1,81 @@
...
+            if 'state UP' not in l and dev in l]
```
cathay4t: You might need to check NO-CARRIER for software interface.

When adding interface to bond, I noticed this state:
```
bond99: <NO-CARRIER,BROADCAST,MULTICAST,MASTER,UP>
```

tests/integration/testlib/iproutelib.py
cathay4t: You might use `libcmd.exec_cmd` to save you from handling that.

EdDev: Thanks, fixed.
I could not use `exec_cmd` because I needed the `Popen` object to control the process (which is hidden in `exec_cmd`).

tests/integration/testlib/iproutelib.py
EdDev: event example:
`eth1: <BROADCAST,MULTICAST> mtu 1500 qdisc pfifo_fast state DOWN group default`

We are looking for the `state DOWN` in this case.
The `NO-CARRIER` and the other  properties in the `<..>` are about the current flags of the interface. I think one is the current state and the other is the transition event.
Have you seen this affecting the `state ????` value?

cathay4t: The reapply does not allow changes on `connection.master`, so it seems bond cannot benefit on reapply:
```
[fge@fed nmstate]$ sudo nmcli d reapply eth2
Error: Reapplying connection to device 'eth2' (/org/freedesktop/NetworkManager/Devices/5) failed: Can't reapply changes to 'connection.master' setting
```

EdDev: Rebinding a slave to a bond or bridge is disruptive, but changing an existing bond or adding a new slave should probably be supported (while checking the bond/bridge itself).

PR #368, opened 2019-06-03
tests integ: Test Linux bridge uses port's MAC

Nandan307:
Test that when a Linux bridge is added over an interface,
by default it uses interface's MAC address.
    
Signed-off-by: Nandan Kulkarni <nakulkar@redhat.com>

PR #370, opened 2019-06-03
tests integ: Refactor Linux Bridge test module

EdDev:


tests/integration/linux_bridge_test.py
```diff
@@ -42,38 +45,56 @@
...
+def test_create_and_remove_linux_bridge_with_min_desired_state():
+    bridge_name = TEST_BRIDGE0
+    with _linux_bridge(bridge_name, bridge_state=None) as desired_state:
+        assertlib.assert_state(desired_state)
 
-    with linux_bridge(bridge_name, bridge_state) as desired_state:
+    assertlib.assert_absent(bridge_name)
+
+
+def test_create_and_remove_linux_bridge_with_one_port(eth1_up):
```
tyll: Is it really best to use  `eth1_up` as a fixture? Having eth1 down as an initial state seems to make more sense to me actually. In the end it should work for both cases I guess.

EdDev: In general, it depends on where you put the line between the test itself and the setup for it.
The setup you assume and the test you question.

In scenarios where eth1 and eth2 are used as slaves, nmstate implicitly sets them to state `up` as part of the metadata processing. Therefore, I think that for these tests it can work both ways.
It does not work like this for tests which do not implicitly enable the interface.

Therefore, I tend to agree that we can do this here.
How about trying this in a separated PR? This PR has not changed the functionality, only adjusted the creating of the config in order to reuse it later on.

Nandan307: Refactor port names eth1 and eth2 for readability and
to sync with global bridge name variable.

Make Port eth1 to be globa var PORT0 and port eth2 to be PORT1.

EdDev: Looks like you already done it in #368 . Please clarify if you want me to change this PR or you will keep yours.

tests/integration/linux_bridge_test.py
```diff
@@ -29,6 +29,9 @@
 from .testlib.statelib import INTERFACES
 
 
+TEST_BRIDGE0 = 'linux-br0'
+
```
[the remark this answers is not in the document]
EdDev: Looks like you already done it in #368 . Please clarify if you want me to change this PR or you will keep yours.

PR #378, opened 2019-06-08
tests integ: Introduce a link state monitor decorator

EdDev:
Link state monitoring is used to detect link state changes while
executing apply changes. In particular, to detect links that change to
anything but UP.

In order to allow simpler usage in selected tests, a decorator is
introduced in addition to the existing context-manager.

Note: six is used to overcome functools.wraps limitations with pytest
(https://github.com/pytest-dev/pytest/issues/2782)

### DIFF
--- a/tests/integration/testlib/iproutelib.py
+++ b/tests/integration/testlib/iproutelib.py
@@ -20,6 +20,8 @@ import subprocess
 import threading
 import time
 
+import six
+
 
 TIMEOUT = 10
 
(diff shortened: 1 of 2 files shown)


tests/integration/testlib/iproutelib.py
```diff
@@ -31,6 +33,18 @@ def __init__(self):
         self.popen = None
 
 
+def ip_monitor_assert_stable_link_up(dev, timeout=10):
+    def decorator(func):
+        @six.wraps(func)
```
cathay4t: Please refer the pytest bug so that other may know the need of this line.

PR #379, opened 2019-06-08
Test integ nm bridge add port

EdDev:
Introduce a test at the nm subpackage which adds a port to an existing bridge using the `modify` operation. Showing that the bridge link has not been affected.

Note that this PR also includes a refactoring of the test module as a pre-step, allowing the addition of new tests.

Depends on #416

### DIFF
--- a/tests/integration/nm/linux_bridge_test.py
+++ b/tests/integration/nm/linux_bridge_test.py
@@ -19,14 +19,25 @@
 
 from contextlib import contextmanager
 
+import pytest
+
 from libnmstate import nm
 from libnmstate import schema
 from libnmstate.schema import LinuxBridge as LB
 
+from ..testlib import iproutelib
 from .testlib import mainloop_run
 
 
 BRIDGE0 = 'brtest0'
+
+
+@pytest.fixture
+def bridge0_with_port0(port0_up):
+    port_name = port0_up[schema.Interface.KEY][0][schema.Interface.NAME]
+    bridge_desired_state = _create_bridge_config((port_name,))
+    with _bridge_interface(bridge_desired_state):
+        yield _get_bridge_current_state()
 
 
 def test_create_and_remove_minimum_config_bridge():
@@ -50,6 +61,21 @@ def test_create_and_remove_bridge(port0_up):
         assert bridge_desired_state == bridge_current_state
 
     assert not _get_bridge_current_state()
+
+
+@iproutelib.ip_monitor_assert_stable_link_up(BRIDGE0)
+def test_add_port_to_existing_bridge(bridge0_with_port0, port1_up):
+    port_name = port1_up[schema.Interface.KEY][0][schema.Interface.NAME]
+    _add_ports_to_bridge_config(bridge0_with_port0, (port_name,))
+
+    _modify_bridge(bridge0_with_port0)
+
+    assert bridge0_with_port0 == _get_bridge_current_state()
+
+
+def _add_ports_to_bridge_config(bridge_state, ports):
+    ports_config = _create_bridge_ports_config(ports)
+    bridge_state[LB.CONFIG_SUBTREE][LB.PORT_SUBTREE] += ports_config
 
 
 def _create_bridge_config(ports):
@@ -96,6 +122,31 @@ def _bridge_interface(state):
             _delete_iface(p[LB.PORT_NAME])
 
 
+@mainloop_run
+def _modify_bridge(bridge_desired_state):
+    bridge_config = bridge_desired_state[LB.CONFIG_SUBTREE]
+    _modify_bridge_options(bridge_config)
+    _modify_ports(bridge_config[LB.PORT_SUBTREE])
+
+
+def _modify_bridge_options(bridge_config):
+    br_options = bridge_config.get(LB.OPTIONS_SUBTREE)
+    nmdev = nm.device.get_device_by_name(BRIDGE0)
+    conn = nm.connection.ConnectionProfile()
+    conn.import_by_id(BRIDGE0)
+    iface_bridge_settings = _create_iface_bridge_settings(br_options, conn)
+    new_conn = nm.connection.ConnectionProfile()
+    new_conn.create(iface_bridge_settings)
+    conn.update(new_conn)
+    conn.commit(save_to_disk=False, nmdev=nmdev)
+    nm.device.modify(nmdev, conn.profile)
+
+
+def _modify_ports(ports_state):
+    for port_state in ports_state:
+        _attach_port_to_bridge(port_state)
+
+
 def _get_bridge_current_state():
     nm.nmclient.client(refresh=True)
     nmdev = nm.device.get_device_by_name(BRIDGE0)
@@ -126,7 +177,7 @@ def _attach_port_to_bridge(port_state):
 
     curr_port_con_profile.update(port_con_profile)
     curr_port_con_profile.commit(nmdev=port_nmdev)
-    nm.device.activate(connection_id=port_state[LB.PORT_NAME])
+    nm.device.modify(port_nmdev, curr_port_con_profile.profile)
 
 
 def _create_bridge_iface(iface_bridge_settings):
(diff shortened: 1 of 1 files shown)


cathay4t:
I got failure constantly:
```
Device reapply failed on eth1: error=nm-device-error-quark: Can't reapply changes to '802-3-ethernet.cloned-mac-address' setting
```

cathay4t:
The test `test_add_port_to_existing_bridge` failed on my VM(Fedora 30 with [master branch of NM][1]) with IP monitor failure:

```
E       AssertionError: result: 12: brtest0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state DOWN group default 
E           link/ether 36:34:28:e8:a6:e0 brd ff:ff:ff:ff:ff:ff
E       12: brtest0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state DOWN group default 
E           link/ether 36:34:28:e8:a6:e0 brd ff:ff:ff:ff:ff:ff
E       12: brtest0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state DOWN group default 
E           link/ether 36:34:28:e8:a6:e0 brd ff:ff:ff:ff:ff:ff

../testlib/iproutelib.py:43: AssertionError
```

[1]: NetworkManager-1.19.3-1.3c54b5eb2b.fc30.x86_64 from [NM copr](https://copr.fedorainfracloud.org/coprs/networkmanager/NetworkManager-master/)

tests/integration/nm/linux_bridge_test.py
```diff
@@ -22,24 +22,65 @@
...
 from libnmstate.schema import LinuxBridge as LB
 
-from .testlib import mainloop
+from ..testlib import iproutelib
+from .testlib import mainloop_run
 
 
 BRIDGE0 = 'brtest0'
-ETH1 = 'eth1'
+PORT0 = 'eth1'
```
tyll: how about also adding `port0_up`  = `eth1_up` to abstract the fixture for the port as well?

tests/integration/nm/linux_bridge_test.py
```diff
@@ -22,24 +22,65 @@
...
+    with _bridge_interface(bridge_desired_state):
+
+        bridge_current_state = _get_bridge_current_state()
+        assert bridge_desired_state == bridge_current_state
+
+    assert not _get_bridge_current_state()
+
+
+@iproutelib.ip_monitor_assert_stable_link_up(BRIDGE0)
+def test_add_port_to_existing_bridge(bridge0_with_port0, eth2_up):
```
tyll: Why is this using `eth2_up`?

tyll: I see, it is because `PORT1` is `eth2` - IMHO this is just too confusing and aliasing the fixtures with the abstract names would make it a lot better.

tests/integration/nm/linux_bridge_test.py
```diff
@@ -53,6 +64,20 @@ def test_create_and_remove_bridge(eth1_up):
...
+    _add_port_to_bridge_config(bridge0_with_port0, (PORT1,))
+
+    _modify_bridge(bridge0_with_port0)
+
+    assert bridge0_with_port0 == _get_bridge_current_state()
+
+
+def _add_port_to_bridge_config(bridge0_with_port0, ports):
```
tyll: IMHO reusing the name of the fixture, or is it the fixture here is confusing as well. If it is the fixture, the  function should be renamed to reflect this, if not, the argument should just be bridge. Also the functions  says `add_port` (singular) but the argument is `ports` (plural). Does it add only one port or also multiple ports?

tests/integration/nm/linux_bridge_test.py
EdDev: https://github.com/nmstate/nmstate/pull/415

tests/integration/nm/linux_bridge_test.py
```diff
@@ -25,19 +25,62 @@
...
+    assert bridge0_with_port0 == _get_bridge_current_state()
+
+
+def _add_ports_to_bridge_config(bridge_state, ports):
+    ports_config = _create_bridge_ports_config(ports)
+    bridge_state[LB.CONFIG_SUBTREE][LB.PORT_SUBTREE] += ports_config
+
+
+def _create_bridge_config(ports):
+
```
maiqueb: please remove this empty line (or is there some deeper meaning to it 0_o)

tests/integration/nm/linux_bridge_test.py
```diff
@@ -52,42 +95,21 @@ def bridge_default_config():
...
-
-        bridge_current_state = _get_bridge_current_state()
-        assert bridge_desired_state == bridge_current_state
-
-    assert not _get_bridge_current_state()
+def _create_bridge_ports_config(ports):
+    return [
+        {
+            LB.PORT_NAME: port,
+            LB.PORT_STP_PRIORITY: 32,
```
maiqueb: wouldn't we want to parameterize  these other values ? (just asking, got no need for it).

tests/integration/nm/linux_bridge_test.py
```diff
@@ -101,22 +123,47 @@ def _bridge_interface(state):
...
+def _modify_bridge_options(bridge_config):
+    br_options = bridge_config.get(LB.OPTIONS_SUBTREE)
+    nmdev = nm.device.get_device_by_name(BRIDGE0)
+    conn = nm.connection.ConnectionProfile()
+    conn.import_by_id(BRIDGE0)
+    iface_bridge_settings = _create_iface_bridge_settings(br_options, conn)
+    new_conn = nm.connection.ConnectionProfile()
+    new_conn.create(iface_bridge_settings)
+    conn.update(new_conn)
+    conn.commit(save_to_disk=False, nmdev=nmdev)
```
maiqueb: why don't we want this to persist  ?

tests/integration/nm/linux_bridge_test.py
```diff
@@ -101,22 +123,47 @@ def _bridge_interface(state):
...
-        ports_state = bridge_desired_state[LB.CONFIG_SUBTREE][LB.PORT_SUBTREE]
-        for port_state in ports_state:
-            _attach_port_to_bridge(port_state)
+    _create_bridge_iface(iface_bridge_settings)
+    ports_state = bridge_desired_state[LB.CONFIG_SUBTREE][LB.PORT_SUBTREE]
+    for port_state in ports_state:
+        _attach_port_to_bridge(port_state)
 
 
 def _attach_port_to_bridge(port_state):
```
maiqueb: feels very weird that this does not receive the bridge as well, but if I recall your words correctly, adding a port to a bridge, is setting the port to be a slave of bridge - hence I think it would be a property of the bridge. 

Am I right?

tests/integration/nm/linux_bridge_test.py
EdDev: Done

tests/integration/nm/linux_bridge_test.py
EdDev: Not sure what you are asking.
At the moment this is not changing the logic, just refactoring the code without changes to behaviour.

If we will need to change the ports properties, we will probably need to add options to do so in the future.

tests/integration/nm/linux_bridge_test.py
EdDev: Because we do not need it. We are not testing NM persistent capabilities here.

tests/integration/nm/linux_bridge_test.py
EdDev: Yes, you are right. However, this refactoring was aimed to serve the following patch and not fix all issues.
The problem is with the usage of `BRIDGE0` all over the place. It should have been used only at the top most level (test) and not in the helpers.
I would prefer not to add this task on this change, but I think it is reasonable to add a patch that fixes it in general.

PR #389, opened 2019-06-12
State route index with object

EdDev:
Depends on #388


libnmstate/nm/__init__.py
```diff
@@ -36,6 +37,7 @@
 checkpoint
 connection
 device
+dns
```
cathay4t: This change seems irrelevant to this PR. Can we remove it?

EdDev: metadata module needs it as it uses nm.dns.
I do not know how it worked without it. It is not related to my changes.. strange.

PR #391, opened 2019-06-12
dns: Remove dns module dependency on ipv{4,6} modules

EdDev:
Use existing generators from the ip modules to fetch the DNS
configuration and by that removing the existing dependency.

### DIFF
--- a/libnmstate/nm/dns.py
+++ b/libnmstate/nm/dns.py
@@ -14,13 +14,12 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
+from itertools import chain
 from operator import itemgetter
 
 from libnmstate import iplib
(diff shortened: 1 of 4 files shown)


libnmstate/nm/dns.py
```diff
@@ -56,37 +55,32 @@ def get_running():
...
-                    priority = DNS_DEFAULT_PRIORITY_OTHER
 
-            tmp_dns_confs.append(
-                {
-                    'server': ip_profile.props.dns,
-                    'priority': priority,
-                    'search': ip_profile.props.dns_search,
-                }
-            )
+    for ac, ip_profile in chain(ipv4_profiles, ipv6_profiles):
```
cathay4t: Should prefer ipv6 over IPv4. Please place ipv4_profiles before ipv4.

EdDev: Done

PR #397, opened 2019-06-12
tests integ: Test bond iface and slaves have same MAC

Nandan307:
Test that when one or more slaves are added, the bond interface
and its slaves have the same MAC-address. Confirm that the order
of the mentioned slaves is not affecting the MAC-address.

Signed-off-by: Nandan Kulkarni <nakulkar@redhat.com>

### DIFF
--- a/tests/integration/bond_test.py
+++ b/tests/integration/bond_test.py
@@ -266,3 +266,32 @@ def test_set_bond_mac_address(eth1_up):
         assert bond99_cur_state[Interface.MAC] == MAC1.upper()
         assert bond99_cur_state[Interface.MAC] == eth1_cur_state[Interface.MAC]
 
+
+@pytest.fixture
+def bond99():
+    with bond_interface(name='bond99', slaves=['eth1', 'eth2']) as bond_state:
+        yield bond_state
+
+
+def test_reordering_the_slaves_does_not_change_the_mac(bond99):
+    state = statelib.show_only(('bond99', 'eth1', 'eth2'))
+    bond99_state = state[Interface.KEY][0]
+    eth1_state = state[Interface.KEY][1]
+    eth2_state = state[Interface.KEY][2]
+
+    assert (bond99_state[Interface.MAC] ==
+            eth1_state[Interface.MAC] ==
+            eth2_state[Interface.MAC])
+
+    bond99[INTERFACES][0]['link-aggregation']['slaves'].reverse()
+    libnmstate.apply(bond99)
+
+    modified_state = statelib.show_only(('bond99', 'eth1', 'eth2'))
+    bond99_modified_state = modified_state[Interface.KEY][0]
+    eth1_modified_state = modified_state[Interface.KEY][1]
+    eth2_modified_state = modified_state[Interface.KEY][2]
+
+    assert (bond99_modified_state[Interface.MAC] ==
+            eth1_modified_state[Interface.MAC] ==
+            eth2_modified_state[Interface.MAC] ==
+            bond99_state[Interface.MAC])
--- a/tests/integration/bond_test.py
+++ b/tests/integration/bond_test.py
@@ -251,13 +251,18 @@ def test_set_bond_mac_address(eth1_up):
         bond_state[Interface.KEY][0][Interface.MAC] = MAC0
         libnmstate.apply(bond_state)
 
-        current_state = statelib.show_only(('bond99',))
+        current_state = statelib.show_only(('bond99', 'eth1'))
         bond99_cur_state = current_state[INTERFACES][0]
+        eth1_cur_state = current_state[INTERFACES][1]
         assert bond99_cur_state[Interface.MAC] == MAC0.upper()
+        assert bond99_cur_state[Interface.MAC] == eth1_cur_state[Interface.MAC]
 
         bond_state[Interface.KEY][0][Interface.MAC] = MAC1
         libnmstate.apply(bond_state)
 
-        current_state = statelib.show_only(('bond99',))
+        current_state = statelib.show_only(('bond99', 'eth1'))
         bond99_cur_state = current_state[INTERFACES][0]
+        eth1_cur_state = current_state[INTERFACES][1]
         assert bond99_cur_state[Interface.MAC] == MAC1.upper()
+        assert bond99_cur_state[Interface.MAC] == eth1_cur_state[Interface.MAC]
+


ovirt-infra:

tyll:
Thank you. It seems like you copied most of the bond test file again into the file. Please take a look it clean up the commit.

tyll:
Thank you. It seems like you copied most of the bond test file again into the file. Please take a look it clean up the commit.

tests/integration/bond_test.py
```diff
@@ -262,3 +262,46 @@ def test_set_bond_mac_address(eth1_up):
...
+    eth1_slaves_mac = eth1_slaves_state[Interface.MAC]
+    eth2_slaves_mac = eth2_slaves_state[Interface.MAC]
+
+    assert (bond99_slaves_mac ==
+            eth1_slaves_mac ==
+            eth2_slaves_mac)
+
+
+def test_bond_with_no_slaves_have_same_mac(bond0):
+    assert_mac_address_behavior(bond0)
```
tyll: Which interfaces should have the same MAC in this case if there are no slaves?

tests/integration/bond_test.py
```diff
@@ -262,3 +262,46 @@ def test_set_bond_mac_address(eth1_up):
...
         assert bond99_cur_state[Interface.MAC] == MAC1.upper()
+
+
+@pytest.fixture
+def bond0():
+    with bond_interface(name='bond99', slaves=[]) as bond_state:
+        yield bond_state
+
+
+def assert_mac_address_behavior(bond0):
```
tyll: The assertion should be named according to what it asserts. Also it should get the state it should assert as an argument. Why does it have `bond0` as an argument?

tests/integration/bond_test.py
```diff
@@ -262,3 +262,46 @@ def test_set_bond_mac_address(eth1_up):
...
+    bond99_slaves_mac = bond99_slaves_state[Interface.MAC]
+    eth1_slaves_mac = eth1_slaves_state[Interface.MAC]
+    eth2_slaves_mac = eth2_slaves_state[Interface.MAC]
+
+    assert (bond99_slaves_mac ==
+            eth1_slaves_mac ==
+            eth2_slaves_mac)
+
+
+def test_bond_with_no_slaves_have_same_mac(bond0):
```
tyll: If the bond does not have any slaves, this test does not make sense.

tests/integration/bond_test.py
```diff
@@ -262,3 +262,46 @@ def test_set_bond_mac_address(eth1_up):
...
+    assert_mac_address_behavior(bond0)
+
+
+def test_bond_with_two_slaves_has_same_mac(bond0):
```
tyll: It should be `have` not `has`

tests/integration/bond_test.py
```diff
@@ -262,3 +262,46 @@ def test_set_bond_mac_address(eth1_up):
...
+    bond0[INTERFACES][0]['link-aggregation']['slaves'] = ['eth1', 'eth2']
+    netapplier.apply(bond0)
+    assert_mac_address_behavior(bond0)
+
+
+def test_bond_with_all_removed_slaves_has_same_mac(bond0):
```
tyll: If all slaves are removed, why should they have the same MAC address?

tests/integration/bond_test.py
```diff
@@ -262,3 +262,46 @@ def test_set_bond_mac_address(eth1_up):
...
+    bond0[INTERFACES][0]['link-aggregation']['slaves'] = []
+    netapplier.apply(bond0)
+    assert_mac_address_behavior(bond0)
+
+
+def test_bond_with_reversed_two_slaves_has_same_mac(bond0):
```
tyll: You need to combine this test with the other test. The fixture creates a new bond interface for each test function. So if you want to test that changes to a bond interface do not change the MAC address, you need to do it in the same test function. Also you need to store the initial MAC address.

tyll:
The lint CI job also found errors that need to be addressed.

tests/integration/bond_test.py
tyll: you need to assert that the slavees and the bond all have different MAC addresses here.

tests/integration/bond_test.py
```diff
@@ -262,3 +262,46 @@ def test_set_bond_mac_address(eth1_up):
         current_state = statelib.show_only(('bond99',))
         bond99_cur_state = current_state[INTERFACES][0]
         assert bond99_cur_state[Interface.MAC] == MAC1.upper()
+
+
+@pytest.fixture
+def bond0():
```
tyll: The fixture should be called `bond99` since this is the bond's name.

tests/integration/bond_test.py
```diff
@@ -255,9 +255,26 @@ def test_set_bond_mac_address(eth1_up):
...
+
+def return_bond_and_slaves_mac():
+    bond_and_slaves_state = statelib.show_only(('bond99', 'eth1', 'eth2'))
+    bond99_bond_and_slaves_state = bond_and_slaves_state[Interface.KEY][0]
+    eth1_bond_and_slaves_state = bond_and_slaves_state[Interface.KEY][1]
+    eth2_bond_and_slaves_state = bond_and_slaves_state[Interface.KEY][2]
+
+    return bond99_bond_and_slaves_state[Interface.MAC], \
+        eth1_bond_and_slaves_state[Interface.MAC], \
+        eth2_bond_and_slaves_state[Interface.MAC]
```
tyll: There is a newline missing at the end of the file

tests/integration/bond_test.py
```diff
@@ -255,9 +255,26 @@ def test_set_bond_mac_address(eth1_up):
...
         current_state = statelib.show_only(('bond99',))
         bond99_cur_state = current_state[INTERFACES][0]
         assert bond99_cur_state[Interface.MAC] == MAC1.upper()
+
+        bond99_slaves_mac, eth1_slaves_mac, _ = return_bond_and_slaves_mac()
+        assert bond99_slaves_mac == eth1_slaves_mac
+
+
+def return_bond_and_slaves_mac():
+    bond_and_slaves_state = statelib.show_only(('bond99', 'eth1', 'eth2'))
```
tyll: The show needs to happen outside this function and the function should get a parameter like `state` and it could then just return the mac addresses of all interfaces in the state. Then it could be `get_macs()`.

tests/integration/bond_test.py
```diff
@@ -255,9 +255,26 @@ def test_set_bond_mac_address(eth1_up):
...
+        assert bond99_slaves_mac == eth1_slaves_mac
+
         bond_state[Interface.KEY][0][Interface.MAC] = MAC1
         libnmstate.apply(bond_state)
 
         current_state = statelib.show_only(('bond99',))
         bond99_cur_state = current_state[INTERFACES][0]
         assert bond99_cur_state[Interface.MAC] == MAC1.upper()
+
+        bond99_slaves_mac, eth1_slaves_mac, _ = return_bond_and_slaves_mac()
```
tyll: IMHO it would be good if you use `get_macs()` here.

tests/integration/bond_test.py
```diff
@@ -255,9 +255,26 @@ def test_set_bond_mac_address(eth1_up):
         bond99_cur_state = current_state[INTERFACES][0]
         assert bond99_cur_state[Interface.MAC] == MAC0.upper()
 
+        bond99_slaves_mac, eth1_slaves_mac, _ = return_bond_and_slaves_mac()
```
tyll: IMHO it would be good if you implement `get_macs()`, so can use it like this here:

```
bond_mac, slave_mac = get_macs(statelib.show_only(('bond99', 'eth1')))
assert bond_mac == slave_mac == MAC0.upper()
```

Then you can also adjust the lines before this line.

tests/integration/bond_test.py
```diff
@@ -277,4 +277,49 @@ def return_bond_and_slaves_mac():
...
+def bond99():
+    with bond_interface(name='bond99', slaves=[]) as bond_state:
+        yield bond_state
+
+
+def test_bond_and_no_slaves_have_same_mac(bond99):
+    bond99_slaves_mac, eth1_slaves_mac, eth2_slaves_mac = \
+        return_bond_and_slaves_mac()
+
+    assert not bond99_slaves_mac == eth1_slaves_mac
```
tyll: Use `!=` please (might be a duplicate comment, but I cannot see the other comment currently).

tests/integration/bond_test.py
```diff
@@ -277,4 +277,49 @@ def return_bond_and_slaves_mac():
...
+        eth2_bond_and_slaves_state[Interface.MAC]
+
+
+@pytest.fixture
+def bond99():
+    with bond_interface(name='bond99', slaves=[]) as bond_state:
+        yield bond_state
+
+
+def test_bond_and_no_slaves_have_same_mac(bond99):
```
tyll: How about `test_bond_and_test_inferfaces_have_different_macs(bond99)`?

tests/integration/bond_test.py
```diff
@@ -277,4 +277,49 @@ def return_bond_and_slaves_mac():
...
+
+    assert not bond99_slaves_mac == eth1_slaves_mac
+    assert not eth1_slaves_mac == eth2_slaves_mac
+    assert not eth2_slaves_mac == bond99_slaves_mac
+
+
+def test_bond_and_all_removed_slaves_have_same_mac(bond99):
+    bond99[INTERFACES][0]['link-aggregation']['slaves'] = []
+    libnmstate.apply(bond99)
+    test_bond_and_no_slaves_have_same_mac(bond99)
```
tyll: I do not understand this test. Please do not call the test function again. You can create a helper for the assert, for example `assert_all_different()` or something like this and then re-use it.

tests/integration/bond_test.py
```diff
@@ -277,4 +277,49 @@ def return_bond_and_slaves_mac():
...
+
+
+def test_bond_and_two_slaves_have_same_mac(bond99):
+    bond99[INTERFACES][0]['link-aggregation']['slaves'] = ['eth1', 'eth2']
+    libnmstate.apply(bond99)
+    prev_bond99_slaves_mac, prev_eth1_slaves_mac, prev_eth2_slaves_mac = \
+        return_bond_and_slaves_mac()
+
+    assert prev_bond99_slaves_mac == prev_eth1_slaves_mac
```
tyll: You can assert for equality in a single assert.

tests/integration/bond_test.py
Nandan307: It is on a different branch and as spoken earlier, I'll keep this function and this commit and then create a PR from `get_all_macs branch` for all the previously submitted commits refactored with get_all_macs

EdDev: At line 254, we get the state from the system. You now need to get the slaves as well:
`current_state = statelib.show_only(('bond99', 'eth1', 'eth2'))`

Then you assert like we assert the bond mac (line 256).

Save below in the 2nd stage.

tests/integration/bond_test.py
```diff
@@ -265,3 +265,81 @@ def test_set_bond_mac_address(eth1_up):
...
+def test_bond_and_two_slaves_have_same_mac(bond99):
+    bond99[INTERFACES][0]['link-aggregation']['slaves'] = ['eth1', 'eth2']
+    libnmstate.apply(bond99)
+
+    slaves_state = statelib.show_only(('bond99', 'eth1', 'eth2'))
+    bond99_slaves_state = slaves_state[Interface.KEY][0]
+    eth1_slaves_state = slaves_state[Interface.KEY][1]
+    eth2_slaves_state = slaves_state[Interface.KEY][2]
+
+    prev_bond99_slaves_mac = bond99_slaves_state[Interface.MAC]
```
EdDev: Do you really need these `prev_*` variables? Looks like you can use the `*__state[Interface.MAC]` directly.

tests/integration/bond_test.py
```diff
@@ -265,3 +265,81 @@ def test_set_bond_mac_address(eth1_up):
...
+    prev_eth1_slaves_mac = eth1_slaves_state[Interface.MAC]
+    prev_eth2_slaves_mac = eth2_slaves_state[Interface.MAC]
+
+    assert (prev_bond99_slaves_mac ==
+            prev_eth1_slaves_mac ==
+            prev_eth2_slaves_mac)
+
+    bond99[INTERFACES][0]['link-aggregation']['slaves'] = ['eth2', 'eth1']
```
EdDev: Better user `.reverse()`

tests/integration/bond_test.py
```diff
@@ -265,3 +265,81 @@ def test_set_bond_mac_address(eth1_up):
...
+    libnmstate.apply(bond99)
+
+    slaves_state = statelib.show_only(('bond99', 'eth1', 'eth2'))
```
EdDev: `modified_state` or `new_state`

tests/integration/bond_test.py
```diff
@@ -265,3 +265,81 @@ def test_set_bond_mac_address(eth1_up):
...
+    bond99_slaves_state = slaves_state[Interface.KEY][0]
+    eth1_slaves_state = slaves_state[Interface.KEY][1]
+    eth2_slaves_state = slaves_state[Interface.KEY][2]
+
+    curr_bond99_slaves_mac = bond99_slaves_state[Interface.MAC]
+    curr_eth1_slaves_mac = eth1_slaves_state[Interface.MAC]
+    curr_eth2_slaves_mac = eth2_slaves_state[Interface.MAC]
+
+    assert (curr_bond99_slaves_mac ==
```
EdDev: Can't you just add the old bond mac here as a 4th equality? (instead of the new assert block on line 342)

tests/integration/bond_test.py
Nandan307: `>>> slaves = ['eth1', 'eth2']
>>> slaves
['eth1', 'eth2']
>>> slaves = ['eth1', 'eth2'].reverse()
>>> slaves
>>> print(slaves)
None
`

Since I cannot directly use the reverse() function, I used reversed(list_name)

tests/integration/bond_test.py
```diff
@@ -265,3 +265,81 @@ def test_set_bond_mac_address(eth1_up):
...
+    bond99_slaves_mac = bond99_slaves_state[Interface.MAC]
+    eth1_slaves_mac = eth1_slaves_state[Interface.MAC]
+    eth2_slaves_mac = eth2_slaves_state[Interface.MAC]
+
+    assert bond99_slaves_mac != eth1_slaves_mac
+    assert eth1_slaves_mac != eth2_slaves_mac
+    assert eth2_slaves_mac != bond99_slaves_mac
+
+
+def test_bond_and_two_slaves_have_same_mac(bond99):
```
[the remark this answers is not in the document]
Nandan307: `

@pytest.fixture
def bond99():
    with bond_interface(name='bond99', slaves=['eth1', 'eth2']) as bond_state:
        yield bond_state
`
something like this?

tests/integration/bond_test.py
```diff
@@ -265,3 +265,81 @@ def test_set_bond_mac_address(eth1_up):
...
+    bond99_slaves_mac = bond99_slaves_state[Interface.MAC]
+    eth1_slaves_mac = eth1_slaves_state[Interface.MAC]
+    eth2_slaves_mac = eth2_slaves_state[Interface.MAC]
+
+    assert bond99_slaves_mac != eth1_slaves_mac
+    assert eth1_slaves_mac != eth2_slaves_mac
+    assert eth2_slaves_mac != bond99_slaves_mac
+
+
+def test_bond_and_all_removed_slaves_have_same_mac(bond99):
```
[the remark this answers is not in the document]
Nandan307: Don't have this test anymore, it was written so comply with your steps(in comments) on Jira issue 198 and to add eth2 since  test_remove_all_slaves_from_bond didn't consider eth2

tests/integration/bond_test.py
```diff
@@ -265,3 +265,81 @@ def test_set_bond_mac_address(eth1_up):
...
         assert bond99_cur_state[Interface.MAC] == eth1_cur_state[Interface.MAC]
+
+
+@pytest.fixture
+def bond99():
+    with bond_interface(name='bond99', slaves=[]) as bond_state:
+        yield bond_state
+
+
+def test_bond_and_test_inferfaces_have_different_macs(bond99):
```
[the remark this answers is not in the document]
Nandan307: Had to get rid of it completely since I am attaching slaves eth1 and eth2 in the fixture itself making them have the same mac right there so assert macs not equal to each other fails since the bond and its slaves have same MAC right from the fixture

Nandan307:
> Please make sure you update the commit message based on what you exactly do in the test.
> In case you only leave the ordering test, you should express that and not a general `same mac`.
> 
> The commit message is important, it allows us to understand what was the intention when the code was added.

I tried making my commit title and messages as descriptive as possible all the time. Hope the updated one
is much clearer than before.

tests/integration/bond_test.py
```diff
@@ -265,3 +265,40 @@ def test_set_bond_mac_address(eth1_up):
...
+
+@pytest.fixture
+def bond99():
+    with bond_interface(name='bond99', slaves=['eth1', 'eth2']) as bond_state:
+        yield bond_state
+
+
+def test_reordering_the_slaves_does_not_change_the_mac(bond99):
+    # Adding slaves to the bond
+    bond99[INTERFACES][0]['link-aggregation']['slaves'] = ['eth1', 'eth2']
```
EdDev: Bit the fixture has already done that for you, why are you doing it again? (including the apply part)

tests/integration/bond_test.py
```diff
@@ -265,3 +265,40 @@ def test_set_bond_mac_address(eth1_up):
...
+
+
+@pytest.fixture
+def bond99():
+    with bond_interface(name='bond99', slaves=['eth1', 'eth2']) as bond_state:
+        yield bond_state
+
+
+def test_reordering_the_slaves_does_not_change_the_mac(bond99):
+    # Adding slaves to the bond
```
EdDev: You can drop the comments, I consider comments the last resort.
If you need a comment, it means the code is not clear enough: Merits extracting to a function that expresses what the comment said.
In this case, I think it is pretty clear, but you can do it if you want.

tests/integration/bond_test.py
```diff
@@ -265,3 +265,40 @@ def test_set_bond_mac_address(eth1_up):
...
+    eth1_state = state[Interface.KEY][1]
+    eth2_state = state[Interface.KEY][2]
+
+    assert (bond99_state[Interface.MAC] ==
+            eth1_state[Interface.MAC] ==
+            eth2_state[Interface.MAC])
+
+    # Adding slaves to the bond in a reverse order
+    bond99[INTERFACES][0]['link-aggregation']['slaves'] = \
+        list(reversed(['eth1', 'eth2']))
```
EdDev: I think it is safer to do: `bond99[INTERFACES][0]['link-aggregation']['slaves'].reverse()`

Nandan307:
All changes incorporated. Please review and merge

Nandan307:
> Thank you!

Thanks for merging :)

PR #402, opened 2019-06-14
tests: Fix metadata unit test dependency

EdDev:


tests/lib/metadata_test.py
```diff
@@ -33,6 +37,21 @@
...
 
+@pytest.fixture(autouse=True)
+def nm_mock():
+    with mock.patch.object(metadata, 'nm') as m:
+        yield m
+
+
+@pytest.fixture
+def nm_dns_mock(nm_mock):
+    nm_mock.dns.find_interfaces_for_name_servers.return_value = ('eth1',
```
tyll: IMHO these are good candidates for constants to ensure that the interface names are in sync with the actual test.

EdDev: You are right, but this is a hot-fix and the metadata dependency on the nm provider should drop.
We will take care of this after we merge the route refactoring PR/s.
But if you insist, I will change it.

tyll: AFAIU it just needs changes in the three test functions, therefore I would prefer the constant directly. But I am not really sure what is really the test in the functions, Seems like an opportunity for a fixture was missed earlier. If you really do not want to introduce the constant, please add a card for it and then it is ok enough as well (however, fixing it seems to be less effort).

EdDev: This change is fixing a dependency problem and is not supposed to fix the lack of constants use. However, I need the unit tests working for me so I will do whatever is needed to get this in.

EdDev: Done

tyll: You did not need to change the strings in the unrelated test functions, just the ones that depend on the new mock IMHO. Since you introduced the dependency between the fixture and the test, it is IMHO a valid request to introduce constants (or something better) for this.

PR #416, opened 2019-06-21
Test integ get port name from fixture

EdDev:
Instead of using hardcoded port names, extract them from the provided fixtures (portX-up).

Depends on #415


tests/integration/linux_bridge_test.py
```diff
@@ -58,9 +56,10 @@
 
 
 @pytest.fixture
-def bridge0_with_port0(eth1_up):
+def bridge0_with_port0(port0_up):
     bridge_name = TEST_BRIDGE0
-    bridge_state = _create_bridge_subtree_config(('eth1',))
+    port_name = port0_up[Interface.KEY][0][Interface.NAME]
```
tyll: This is nice! :+1:

PR #417, opened 2019-06-21
nmstatectl: Control the log level

EdDev:
Control the log level through an argument (`--log-level`) or an environment
variable (`NMSTATE_LOG_LEVEL`).

Usage example:
```
nmstatectl --log-level=INFO edit eth1
NMSTATE_LOG_LEVEL=INFO nmstatectl edit eth1
```

### DIFF
--- a/nmstatectl/nmstatectl.py
+++ b/nmstatectl/nmstatectl.py
@@ -38,13 +38,28 @@ from libnmstate.schema import Interface
 from libnmstate.schema import Route
 
 
-def main():
-    logging.basicConfig(
-        format='%(asctime)s %(name)-12s %(levelname)-8s %(message)s',
-        level=logging.DEBUG,
-    )
+class NmstateEnvVar(object):
+    LOG_LEVEL = 'NMSTATE_LOG_LEVEL'
 
+
+LOG_LEVELS = {
+    'CRITICAL': logging.CRITICAL,
+    'ERROR': logging.ERROR,
+    'WARN': logging.WARNING,
+    'WARNING': logging.WARNING,
+    'INFO': logging.INFO,
+    'DEBUG': logging.DEBUG,
+    'NOTSET': logging.NOTSET,
+}
+
+
+LOG_LEVEL_DEFAULT = 'DEBUG'
+
+
+def main():
     parser = argparse.ArgumentParser()
+
+    _setup_argument_log_level(parser)
 
     subparsers = parser.add_subparsers()
     setup_subcommand_commit(subparsers)
@@ -57,7 +72,28 @@ def main():
         parser.print_usage()
         return errno.EINVAL
     args = parser.parse_args()
+
+    try:
+        _setup_logging(args.log_level)
+    except ValueError:
+        return os.EX_USAGE
+
     return args.func(args)
+
+
+def _setup_argument_log_level(parser):
+    """
+    Setup the log level argument.
+    Default to the value passed through the environment variable.
+    """
+    log_level = os.getenv(NmstateEnvVar.LOG_LEVEL, LOG_LEVEL_DEFAULT)
+    parser.add_argument(
+        '--log-level',
+        type=str,
+        choices=LOG_LEVELS,
+        default=log_level,
+        help='Set log level.',
+    )
 
 
 def setup_subcommand_commit(subparsers):
@@ -372,3 +408,21 @@ def _filter_routes(state, patterns):
                 if fnmatch.fnmatch(route[Route.NEXT_HOP_INTERFACE], pattern):
                     routes[route_type].append(route)
     return routes
+
+
+def _setup_logging(log_level):
+    try:
+        log_level = log_level.upper()
+        log_level_id = LOG_LEVELS[log_level]
+    except (AttributeError, KeyError):
+        sys.stderr.write(
+            'ERROR: {} is not recognized as a logging level name\n'.format(
+                log_level
+            )
+        )
+        raise ValueError
+
+    logging.basicConfig(
+        format='%(asctime)s %(name)-12s %(levelname)-8s %(message)s',
+        level=log_level_id,
+    )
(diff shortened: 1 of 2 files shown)


nmstatectl/nmstatectl.py
```diff
@@ -57,9 +72,29 @@ def main():
...
 
 
+def _setup_argument_log_level(parser):
+    """
+    Setup the log level argument.
+    Default to the value passed through the environment variable.
+    """
+    log_level = os.getenv(NmstateEnvVar.LOG_LEVEL, LOG_LEVEL_DEFAULT)
+    parser.add_argument(
+        '--log-level',
```
tyll: Why do you prefer setting the log-level instead of using a reasonable log level by default and allowing to increase/decrease it with -q/-v?

nmstatectl/nmstatectl.py
```diff
@@ -38,14 +38,29 @@
 from libnmstate.schema import Route
 
 
-def main():
-    logging.basicConfig(
-        format='%(asctime)s %(name)-12s %(levelname)-8s %(message)s',
-        level=logging.DEBUG,
-    )
+class NmstateEnvVar(object):
+    LOG_LEVEL = 'NMSTATE_LOG_LEVEL'
```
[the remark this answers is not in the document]
EdDev: I think it is pretty common to use env variables for scripts which may use multiple times a tool. Then they do not need to re-specify the argument on each call.

We can also use this in the future to create a non-root logger for libnmstate (for it to initialize). We do not have it yet, but it may be useful to have one.

nmstatectl/nmstatectl.py
EdDev: Per what I know, `-v` or `-q` are usually controlling stdout/sdterr and not the loggers. We can log by default to a file and use `-q/v` to control if we want it on the console as well or not.

nmstatectl/nmstatectl.py
```diff
@@ -57,9 +72,29 @@ def main():
...
+    """
+    Setup the log level argument.
+    Default to the value passed through the environment variable.
+    """
+    log_level = os.getenv(NmstateEnvVar.LOG_LEVEL, LOG_LEVEL_DEFAULT)
+    parser.add_argument(
+        '--log-level',
+        type=str,
+        default=log_level,
+        help='Set log level (' + str(list(LOG_LEVELS)) + ').',
```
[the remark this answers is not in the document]
EdDev: Looking into this, thanks.

tyll:
Just noticed, that there are no tests included in this PR.

EdDev:
> Just noticed, that there are no tests included in this PR.

This is challenging, looking into this.

nmstatectl/nmstatectl.py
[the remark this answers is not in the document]
EdDev: Done

EdDev:
Tests added.

nmstatectl/nmstatectl.py
tyll: Currently, even though the argument is called `log-level`, it actually configures how much information is outputted at the console because nmstatectl is just using the logging module to output to the console. So you could also add an extra feature to log to a logfile (which I think is not useful for nmstatectl) that could be controlled as well.

nmstatectl/nmstatectl.py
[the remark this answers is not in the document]
tyll: You mean to use `import logging; log = logging.getLogger(__name__); log.debug(...)` in libnmstate? Makes sense to me. However, we still do not need the environment variable then.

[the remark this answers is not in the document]
EdDev: You may have missed the first reasoning, which is practically a de-facto standard in all applications I know of. Environment variables are used for scripting the usage of the app, such that the user does not need to repeat itself each time it uses it, add it to the bash profile and so on. Any reasonable app has this option, it costs very little and I see no reasoning not to have it.
In our case, my next suggestion would have been to propose changing the default to `INFO` and enabling `DEBUG` for ourself and the CI through this environment variable.

For the second part, my focus was on the log-level context, meaning: `logging.setLevel()`.
In the future, an app may set the log level for the lib in the code, or the lib providing an env variable which can be used to do the same without any code from the app part (moving it to the app setup/deployment/configuration).

nmstatectl/nmstatectl.py
EdDev: You are pointing our a consequence of the change, not the change itself. The commit message and the change implementation is controlling the *log-level*, it does not mention anything about changing what is shown on the console.

Attempting to couple `-v/q` to the log-level is IMO improper, imposing limitations which are not useful.
`-q` is usually used to redirect stdout/stderr to `/dev/null`.
`-v` is usually used to choose what to display on the console, including the logging but not limited to it. For example, should the state *diff* be presented, at what depth, should the final state be presented at the end, etc...

cathay4t: I don't think the user of `nmstatectl` care about the difference between `controling stdout/stderr` vs `logging`.

For me, I use `-vvv` or `-d9` to get all debug output from a CLI and I don't want to type things like `--log-level NOTSET` where I need to know which log level can get all the logs.

EdDev: You are using a reasoning of "not needed", where the original reasoning is both common usage (see other projects) and flexibility to separate between verbose or quite output and the logging mechanism.

You both are suggesting to use a non standard translation between `-vvvvv` to the log level and mix the ability to output diffs or other output which has nothing to do with the logging in one argument/place.
I do not understand how this is making any sense.

cathay4t: My argument is user of nmstatectl does not care about log level and `-q/-v` is enough for current design.

We can add this complex-long-typing argument `--log-level` when a actual user asked about it.

EdDev: Sounds like a deadlock to me.
I'm ok with `-q` / `-v` controlling the stdout/stderr, not with it controlling the logging level.
So if you want a solution that does not print anything to the console, please propose a different change, which redirects logging to `/dev/null` or to the logging service.

PR #422, opened 2019-06-25
netapplier: Initialize mainloop global object on failure

EdDev:
The mainloop object is a singletone.
When libnmstate is used by an application which calls it multiple times
(e.g. running as a daemon), on mainloop (run) failure, the instance is kept
and reused in the following calls. Unfortunately, when the mainloop run fails,
its state may no longer be usable and may require initialization.

Therefore, to support such usages, the mainloop singletone is initialize
in case a mainloop failure is detected.

### DIFF
--- a/libnmstate/netapplier.py
+++ b/libnmstate/netapplier.py
@@ -182,6 +182,7 @@ def _setup_providers():
     yield
     success = mainloop.run(timeout=20)
     if not success:
+        nmclient.mainloop(refresh=True)
         raise NmstateLibnmError(
             'Unexpected failure of libnm when running the mainloop: {}'.format(
                 mainloop.error


cathay4t:
This seems like a thread safe issue where you should never use static variable in a C library.
For this PR, yes, we should do clean up when failure on mainloop.

How we handle when other libnm user does not do clean up in mainloop?

I think we should drop the singletone for the mainloop and initialize it everytime of `show()` or `apply()`?

EdDev:
> How we handle when other libnm user does not do clean up in mainloop?

I'm not sure I understood the question. At the moment, we have only two users for the mainloop object: The `apply` function and the tests which cover the `nm` subpackage. The tests already cover this issue.

> I think we should drop the singletone for the mainloop and initialize it everytime of show() or apply()?

I agree. But the implementation does not allow us to do it easily. I do not think we should pass the mainloop object through all the functions. Probably after we objectize the code/flow, it will be easier to do.
Initializing it on every `apply` will still keep it a singletone. The problem seen with this approach was that its state may not be in sync with the `show`. Having one throughout the transaction is preferable (for both apply and show).

cathay4t:
> > How we handle when other libnm user does not do clean up in mainloop?
> 
> I'm not sure I understood the question. At the moment, we have only two users for the mainloop object: The `apply` function and the tests which cover the `nm` subpackage. The tests already cover this issue.

Oh. My understanding of mainloop is incorrect, I through it exists after process quit. Yes. Nmstate only has two user of it. It should be fine now.

PR #429, opened 2019-06-30
state: Make interface type optional

EdDev:
Defining a desired state for an interface required specifying its type.

When an interface is edited, there is no need to specify its type, as it
may be concluded from checking the current state.


EdDev:
This change is needed for VDSM integration (encountered during tests): When VDSM translates its own desired state to the Nmstate one, it needs to specify the interface type. However, VDSM is not distinguishing between base nic types: dummy, ethernet, veth. This behaviour is wanted in order to allow tests to run on dummy interfaces.

With this PR, we can run VDSM tests without specifying the type, allowing Nmstate to reuse whatever existing.

PR #431, opened 2019-07-02
metadata: Remove profile for detached slaves

EdDev:
When a slave is removed from a master, it needs to be marked as `down`
in order to remove its configuration completly.
In case of the NM provider, its profile is removed.


EdDev:
This issue has been detected when testing with VDSM.
Non active bond slaves profiles are detected after the tests.

EdDev:
Depends now on on #434

PR #434, opened 2019-07-03
metadata: Slaves swap between masters

EdDev:
The existing metadata processing for master-slaves has not handled
correctly scenarios where the slaves are swap between masters.

This change introduces the fix and provides the relevant unit test that
covers it.


EdDev:
Depends on #436

EdDev:
> Can you add integration test using libnmstate.apply() to move slave between masters?

Done, thank you for noticing.

PR #435, opened 2019-07-03
nm.connection: Add 100ms in between retry on missing con

EdDev:


cathay4t:
@EdDev Any backgroup story on this change? What will happen without that 100ms delay?

EdDev:
@cathay4t , I updated the commit message.

PR #437, opened 2019-07-11
Fix DHCP enabled criteria for interfaces.

maiqueb:
When an existent interface was setup to have DHCP enabled, its
'enabled' criteria was not correct, since it required the interface
to have been assigned an IP addresses.

This patch updates the criteria, setting the interface to enabled
as long as it reports back being configured via DHCP by
NetworkManager.

### DIFF
--- a/libnmstate/nm/ipv4.py
+++ b/libnmstate/nm/ipv4.py
@@ -101,18 +101,14 @@ def get_info(active_connection):
             info['auto-routes'] = not ip_profile.props.ignore_auto_routes
             info['auto-gateway'] = not ip_profile.props.never_default
             info['auto-dns'] = not ip_profile.props.ignore_auto_dns
+            info['enabled'] = True
+            info['address'] = []
     else:
         info['dhcp'] = False
 
     ip4config = active_connection.get_ip4_config()
     if ip4config is None:
-        # When DHCP is enable, it might be possible, the active_connection does
-        # not got IP address yet. In that case, we still mark
-        # info['enabled'] as True.
-        if info['dhcp']:
-            info['enabled'] = True
-            info['address'] = []
-        else:
+        if not info['dhcp']:
             del info['dhcp']
         return info
 
--- a/tests/integration/nm/ipv4_test.py
+++ b/tests/integration/nm/ipv4_test.py
@@ -50,6 +50,25 @@ def test_interface_ipv4_change(eth1_up):
     assert ip4_expected_state == ipv4_current_state
 
 
+def test_enable_dhcp_with_no_server(eth1_up):
+    with mainloop():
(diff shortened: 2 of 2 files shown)


ovirt-infra:

maiqueb:
ci add to whitelist

maiqueb:
ci test please

libnmstate/nm/ipv4.py
```diff
@@ -104,15 +104,13 @@ def get_info(active_connection):
     else:
         info['dhcp'] = False
 
+    if info['dhcp']:
```
EdDev: Looks like this has been already asked on line 100, therefore you can place its block there.

maiqueb: Right, I didn't consider that the 'else' on line 105 enforces the same behavior.

Done.

maiqueb:
> nm.ipv4: Fix IPv4 enabled criteria when DHCP is enabled

> * Note that we do not use the `merge` option on the PR form. It generates a merge from master to your branch and not a rebase, making the history messy.

I didn't intend on making this, I just saw a button saying 'update branch' and hit it without thinking too much.

Won't happen again.

Force-pushed a version that rebases the code instead.

> * commit message:
>   
>   * Please use the <title>:  format on the first line.
>   * The first part of the message I find over complex, could you please try to simplify it?
> 
> ```
> nm.ipv4: Fix IPv4 enabled criteria when DHCP is enabled
> 
> When the existing configuration has DHCP enabled but the response is not yet received from the server, the IP stack is wrongly considered disabled.
> 
> Fix such scenarios by considered the IPv4 stack 'enabled' when DHCP is enabled.

Nice; Updated the commit message accordingly.

> ```

tests/integration/nm/ipv4_test.py
```diff
@@ -50,6 +50,19 @@ def test_interface_ipv4_change(eth1_up):
...
+def test_activate_dhcp_on_existing_interface(eth1_up):
+    expected_ipv4_state = {'enabled': True, 'dhcp': True, 'address': []}
+    with mainloop():
+        _modify_interface(ipv4_state=expected_ipv4_state)
+
+    nm.nmclient.client(refresh=True)
+    ipv4_current_state = _get_ipv4_current_state(TEST_IFACE)
+    assert ipv4_current_state == dict(
+        expected_ipv4_state,
+        **{'auto-dns': True, 'auto-gateway': True, 'auto-routes': True}
```
EdDev: This is a bit strange. You could prepare the full expected state in advance in one shot.
I think that the problem is the attempt to share the "expected" state between the desired and expect.
But even if you decide to keep it, you can just update the expected using `update` one line before.

PR #439, opened 2019-07-14
ipv6: Disable the IPv6 stack when explicitly requested

EdDev:
NetworkManager is currently not supporting the ability to disable the
IPv6 stack on the kernel [1].

As an intermediate solution, sysfs is used to control the IPv6 stack.

[1] https://bugzilla.redhat.com/1643841

### DIFF
--- a/libnmstate/netapplier.py
+++ b/libnmstate/netapplier.py
@@ -28,6 +28,7 @@ from libnmstate import netinfo
 from libnmstate import nm
 from libnmstate import schema
 from libnmstate import state
+from libnmstate import sysctl
 from libnmstate import validator
 from libnmstate.error import NmstateConflictError
 from libnmstate.error import NmstateLibnmError
@@ -140,6 +141,7 @@ def _apply_ifaces_state(
                     ifaces2add + ifaces2edit,
                     con_profiles=ifaces_add_configs + ifaces_edit_configs,
                 )
+            _disable_ipv6(desired_state)
             if verify_change:
                 _verify_change(desired_state)
         if not commit:
@@ -232,3 +234,16 @@ def _edit_interfaces(state2edit):
 
 def _index_by_name(ifaces_state):
     return {iface['name']: iface for iface in ifaces_state}
+
+
+def _disable_ipv6(desired_state):
+    """
+    Identify in the desired state all interfaces that explicitly disable
+    the IPv6 stack and apply it through sysfs.
+
+    This is an intermediate workaround for https://bugzilla.redhat.com/1643841.
+    """
+    for ifstate in six.viewvalues(desired_state.interfaces):
+        ipv6_state = ifstate.get(schema.Interface.IPV6, {})
+        if ipv6_state.get('enabled') is False:
+            sysctl.disable_ipv6(ifstate[schema.Interface.NAME])
--- a/libnmstate/sysctl.py
+++ b/libnmstate/sysctl.py
@@ -1,0 +1,56 @@
+#
+# Copyright (c) 2019 Red Hat, Inc.
+#
+# This file is part of nmstate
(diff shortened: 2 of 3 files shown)


libnmstate/sysctl.py
```diff
@@ -0,0 +1,46 @@
...
+#
+
+import errno
+
+
+def enable_ipv6(dev):
+    disable_ipv6(dev, disable=False)
+
+
+def disable_ipv6(dev, disable=True):
```
cathay4t: Can we rename it to `change_ipv6_state(dev, IPV6_STATE_UP)`?

The `disable_ipv6(dev, disable=False)` is confusing.

libnmstate/sysctl.py
```diff
@@ -0,0 +1,46 @@
...
+    try:
+        with open('/proc/sys/net/ipv6/conf/%s/disable_ipv6' % dev, 'w') as f:
+            f.write('1' if disable else '0')
+    except IOError as e:
+        if e.errno == errno.ENOENT and disable:
+            # IPv6 stack is (already) not available on this device
+            return
+        raise
```
cathay4t: We promised libnmstate only raise exception based on `NmstateError`. So please wrap it into child exception of `NmstateError`.

libnmstate/sysctl.py
```diff
@@ -0,0 +1,46 @@
...
+
+def is_disabled_ipv6(dev='default'):
+    try:
+        with open('/proc/sys/net/ipv6/conf/%s/disable_ipv6' % dev) as f:
+            return int(f.read())
+    except IOError as e:
+        if e.errno == errno.ENOENT:
+            return 1
+        else:
+            raise
```
cathay4t: We promised libnmstate only raise exception based on `NmstateError`. So please wrap it into child exception of `NmstateError`.

libnmstate/netapplier.py
```diff
@@ -140,6 +141,7 @@ def _apply_ifaces_state(
                     ifaces2add + ifaces2edit,
                     con_profiles=ifaces_add_configs + ifaces_edit_configs,
                 )
+            _disable_ipv6(desired_state)
```
cathay4t: Please save the current ipv6 state of all interfaces, so that we could rollback on failure.

libnmstate/sysctl.py
maiqueb: +1 on this.

I think @cathay4t suggestion improves readability.

EdDev: Done

libnmstate/sysctl.py
EdDev: Done

libnmstate/sysctl.py
EdDev: Done

libnmstate/netapplier.py
EdDev: Is expect the rollback to succeed when moving from `enable` to `disable`.. NM should manage to rollback to `enable`.

The other way around may not work, but we can mark it as a limitation for now (until NM provides the solution). I guess we can manage it as well, but just for this small think I would prefer to avoid it if possible.

I will add the relevant tests to check this out and reference the limitation if they are seen.

EdDev:
@cathay4t , @maiqueb can you please revisit this?
Taking care of the rollback support requires another PR, I'll work on it next.

EdDev:
Yes, it fails because sometime the rollback from ipv6 enabled to disabled works.
When I run it with all the suite like the ci, it works. When I run only the static ip tests, it fails (so the test passes).
No idea what is going on.

EdDev:
Looks like rollback to ipv6 disable works fine (see test)

PR #440, opened 2019-07-16
schema: add linux bridge vlan filtering

maiqueb:
schema: add linux bridge vlan filtering
    
The defined model allows for setting trunk and access ports.
Furthermore, it allows defining a native vlan on a trunk port.
    
It is possible to independently define the trunks/access vlans per each port, and while NM / linux bridges allows it, at the moment, we will not allow setting allowed vlans on the bridge port.
It will be hardcoded with the linux bridge default: having vlan id 1 as native vlan.
Defining the schema of the uplink port will be done in a separate subsequent PR.
    
This patch adds to the model:
  - a 'vlan' map to the bridge ports, that will hold a list trunk tag (that can be individual tags, or tag ranges), an access tag - to be used for access ports - and an 'enable-native' flag, intended to select one of the trunk tags as native vlan.
    
    An example of a bridge configuration that allows for all vlans on a trunk port:
```yaml
    - name: linux-br0
      type: linux-bridge
      state: up
      bridge:
        options:
          stp:
            enabled: true
        port:
          - name: eth1
            vlans:
              type: trunk
              trunk-tags:
                - id-range:
                    min: 1
                    max: 4095
```
An example of a bridge configuration with an access port:
```yaml
    - name: linux-br0
      type: linux-bridge
      state: up
      bridge:
        options:
          stp:
            enabled: true
        port:
          - name: eth1
            vlan:
              type: access
              tag: 101
```
An example of a bridge configuration having a port with a native vlan trunk tag:
```yaml
    - name: linux-br0
      type: linux-bridge
      state: up
      bridge:
        options:
          stp:
            enabled: true
        port:
          - name: eth1
            vlan:
              type: trunk
              trunk-tags:
                - id: 101
                - id-range:
                    min: 200
                    max: 4095
              tag: 100
              enable-native: true
```
    
The bug that tracks the feature in NetworkManager can be found at [0].
    
More information on this feature can be found in [1].
    
[0] - https://bugzilla.redhat.com/show_bug.cgi?id=1652910#c11
[1] - https://developers.redhat.com/blog/2017/09/14/vlan-filter-support-on-bridge/
    
### DIFF
--- a/libnmstate/schema.py
+++ b/libnmstate/schema.py
@@ -171,6 +171,23 @@ class LinuxBridge(object):
     PORT_STP_HAIRPIN_MODE = 'stp-hairpin-mode'
     PORT_STP_PATH_COST = 'stp-path-cost'
 
+    class Port(object):
+        VLAN_SUBTREE = 'vlan'
+
+        class Vlan(object):
+            TRUNK_TAGS = 'trunk-tags'
+            TAG = 'tag'
+            ENABLE_NATIVE = 'enable-native'
+            TYPE = 'type'
+            ACCESS_TYPE = 'access'
+            TRUNK_TYPE = 'trunk'
+
+            class TrunkTags(object):
+                ID = 'id'
+                ID_RANGE = 'id-range'
+                MIN_RANGE = 'min'
+                MAX_RANGE = 'max'
+
 
 class Ethernet(object):
     TYPE = InterfaceType.ETHERNET
--- a/libnmstate/schemas/operational-state.yaml
+++ b/libnmstate/schemas/operational-state.yaml
@@ -56,6 +56,10 @@ definitions:
     mac-address:
       type: string
       pattern: "^([a-fA-F0-9]{2}:){5}[a-fA-F0-9]{2}$"
+    bridge-vlan-tag:
+      type: integer
+      minimum: 0
+      maximum: 4095
 
   # Interface types
   interface-base:
@@ -240,6 +244,22 @@ definitions:
                     type: integer
                   stp-hairpin-mode:
                     type: boolean
+                  vlan:
+                    type: object
+                    properties:
+                      type:
+                        type: string
+                        enum:
+                          - trunk
+                          - access
+                      trunk-tags:
+                        type: array
+                        items:
+                          $ref: "#/definitions/bridge-port-vlan"
+                      tag:
+                        $ref: "#/definitions/types/bridge-vlan-tag"
+                      enable-native:
+                        type: boolean
             options:
               type: object
               properties:
@@ -476,3 +496,15 @@ definitions:
         type: array
         items:
           type: string
+  bridge-port-vlan:
+    type: object
+    properties:
+      id:
+        $ref: "#/definitions/types/bridge-vlan-tag"
+      id-range:
+        type: object
+        properties:
+          min:
+            $ref: "#/definitions/types/bridge-vlan-tag"
+          max:
+            $ref: "#/definitions/types/bridge-vlan-tag"
--- a/libnmstate/validator.py
+++ b/libnmstate/validator.py
@@ -29,6 +29,7 @@ from .schema import Constants
 from libnmstate.schema import DNS
 from libnmstate.schema import InterfaceIP
 from libnmstate.schema import InterfaceIPv6
+from libnmstate.schema import LinuxBridge as LB
 from libnmstate.schema import VXLAN
 from libnmstate.error import NmstateDependencyError
 from libnmstate.error import NmstateNotImplementedError
@@ -176,6 +177,14 @@ def validate_vxlan(state):
             )
 
 
+def validate_bridge(state):
+    for iface_state in state.get(schema.Interface.KEY, []):
+        if iface_state.get(schema.Interface.TYPE) == LB.TYPE:
+            _assert_vlan_filtering_trunk_tags(
+                iface_state.get(LB.PORT_SUBTREE, [])
+            )
+
+
 def _assert_iface_is_up(desired_iface_state, current_iface_state):
     """
     Validates that the interface has an UP state.
@@ -238,3 +247,37 @@ def _assert_vxlan_has_missing_attribute(state, *attributes):
                 state,
             )
         )
+
+
+def _assert_vlan_filtering_trunk_tags(ports_state):
+    for port_state in ports_state:
+        trunk_tags = port_state.get(LB.Port.VLAN_SUBTREE, {}).get(
+            LB.Port.Vlan.TRUNK_TAGS, []
+        )
+        for trunk_tag in trunk_tags:
+            _assert_vlan_filtering_trunk_tag(trunk_tag)
+
+
+def _assert_vlan_filtering_trunk_tag(trunk_tag_state):
+    vlan_id = trunk_tag_state.get(LB.Port.Vlan.TrunkTags.ID)
+    vlan_id_range = trunk_tag_state.get(LB.Port.Vlan.TrunkTags.ID_RANGE)
+
+    if vlan_id and vlan_id_range:
+        raise NmstateValueError(
+            'Trunk port cannot be configured by both id and range: {}'.format(
+                trunk_tag_state
+            )
+        )
+    elif vlan_id_range:
+        if not (
+            {
+                LB.Port.Vlan.TrunkTags.MIN_RANGE,
+                LB.Port.Vlan.TrunkTags.MAX_RANGE,
+            }
+            <= set(vlan_id_range)
+        ):
+            raise NmstateValueError(
+                'Trunk port range requires min / max keys: {}'.format(
+                    vlan_id_range
+                )
+            )
--- a/tests/lib/schema_validation_test.py
+++ b/tests/lib/schema_validation_test.py
@@ -27,11 +27,13 @@ import jsonschema as js
 import libnmstate
 from libnmstate.schema import Constants
 from libnmstate.schema import DNS
+from libnmstate.schema import LinuxBridge as LB
 from libnmstate.schema import VXLAN
 from libnmstate.schema import Interface
 
 INTERFACES = Constants.INTERFACES
 ROUTES = Constants.ROUTES
+THE_BRIDGE = 'br0'
 VXLAN0 = 'vxlan0'
 
 COMMON_DATA = {
@@ -104,6 +106,23 @@ COMMON_DATA = {
 @pytest.fixture
 def default_data():
     return copy.deepcopy(COMMON_DATA)
+
+
+@pytest.fixture
+def portless_bridge_state():
+    return {
+        Interface.NAME: THE_BRIDGE,
+        Interface.STATE: 'up',
+        Interface.TYPE: LB.TYPE,
+        LB.CONFIG_SUBTREE: {LB.PORT_SUBTREE: []},
+    }
+
+
+@pytest.fixture
+def bridge_state(portless_bridge_state):
+    port = {LB.PORT_NAME: 'eth1', LB.Port.VLAN_SUBTREE: {}}
+    portless_bridge_state[LB.CONFIG_SUBTREE][LB.PORT_SUBTREE].append(port)
+    return portless_bridge_state
 
 
 class TestIfaceCommon(object):
@@ -211,3 +230,132 @@ class TestRoutes(object):
         with pytest.raises(js.ValidationError) as err:
             libnmstate.validator.validate(default_data)
         assert 'bad-state' in err.value.args[0]
+
+
+class TestLinuxBridgeVlanFiltering(object):
+    @pytest.mark.parametrize('port_type', argvalues=['trunk', 'access'])
+    def test_vlan_port_types(self, default_data, bridge_state, port_type):
+        valid_port_type = self._generate_vlan_filtering_config(port_type)
+        the_port = bridge_state[LB.CONFIG_SUBTREE][LB.PORT_SUBTREE][0]
+        the_port.update(valid_port_type)
+        default_data[Interface.KEY].append(bridge_state)
+
+        libnmstate.validator.validate(default_data)
+
+    def test_invalid_vlan_port_type(self, default_data, bridge_state):
+        invalid_port_type = self._generate_vlan_filtering_config('fake-type')
+        the_port = bridge_state[LB.CONFIG_SUBTREE][LB.PORT_SUBTREE][0]
+        the_port.update(invalid_port_type)
+        default_data[Interface.KEY].append(bridge_state)
+
+        with pytest.raises(js.ValidationError) as err:
+            libnmstate.validator.validate(default_data)
+        expected_error_msg = "'fake-type' is not one of ['trunk', 'access']"
+        assert expected_error_msg in err.value.args[0]
+
+    def test_access_port_accepted(self, default_data, bridge_state):
+        vlan_access_port_state = self._generate_vlan_filtering_config(
+            'access', access_tag=101
+        )
+        the_port = bridge_state[LB.CONFIG_SUBTREE][LB.PORT_SUBTREE][0]
+        the_port.update(vlan_access_port_state)
+        default_data[Interface.KEY].append(bridge_state)
+
+        libnmstate.validator.validate(default_data)
+
+    def test_wrong_access_port_tag_type(self, default_data, bridge_state):
+        invalid_access_port_tag_type = self._generate_vlan_filtering_config(
+            'access', access_tag='holy-guacamole!'
+        )
+        the_port = bridge_state[LB.CONFIG_SUBTREE][LB.PORT_SUBTREE][0]
+        the_port.update(invalid_access_port_tag_type)
+        default_data[Interface.KEY].append(bridge_state)
+
+        with pytest.raises(js.ValidationError) as err:
+            libnmstate.validator.validate(default_data)
+        expected_error_msg = "'holy-guacamole!' is not of type 'integer'"
+        assert expected_error_msg in err.value.args[0]
+
+    def test_wrong_access_tag_range(self, default_data, bridge_state):
+        invalid_vlan_id_range = self._generate_vlan_filtering_config(
+            'access', access_tag=48000
+        )
+        the_port = bridge_state[LB.CONFIG_SUBTREE][LB.PORT_SUBTREE][0]
+        the_port.update(invalid_vlan_id_range)
+        default_data[Interface.KEY].append(bridge_state)
+
+        with pytest.raises(js.ValidationError) as err:
+            libnmstate.validator.validate(default_data)
+        expected_error_msg = '48000 is greater than the maximum of 4095'
+        assert expected_error_msg in err.value.args[0]
+
+    @pytest.mark.parametrize(
+        'is_native_vlan', argvalues=[True, False], ids=['native', 'not-native']
+    )
+    def test_trunk_port_native_vlan(
+        self, default_data, bridge_state, is_native_vlan
+    ):
+        vlan_access_port_state = self._generate_vlan_filtering_config(
+            'trunk',
+            access_tag=101 if is_native_vlan else None,
+            native_vlan=is_native_vlan,
+        )
+        the_port = bridge_state[LB.CONFIG_SUBTREE][LB.PORT_SUBTREE][0]
+        the_port.update(vlan_access_port_state)
+        default_data[Interface.KEY].append(bridge_state)
+
+        libnmstate.validator.validate(default_data)
+
+    def test_trunk_ports(self, default_data, bridge_state):
+        trunk_tags = self._generate_vlan_id_config(101, 102, 103)
+        trunk_tags.append(self._generate_vlan_id_range_config(500, 1000))
+        vlan_trunk_tags_port_state = self._generate_vlan_filtering_config(
+            'trunk', trunk_tags=trunk_tags
+        )
+        the_port = bridge_state[LB.CONFIG_SUBTREE][LB.PORT_SUBTREE][0]
+        the_port.update(vlan_trunk_tags_port_state)
+        default_data[Interface.KEY].append(bridge_state)
+
+        libnmstate.validator.validate(default_data)
+
+    def test_invalid_trunk_port_vlan_range(self, default_data, bridge_state):
+        invalid_port_vlan_configuration = self._generate_vlan_filtering_config(
+            'trunk',
+            trunk_tags=[self._generate_vlan_id_range_config(100, 5000)],
+        )
+        the_port = bridge_state[LB.CONFIG_SUBTREE][LB.PORT_SUBTREE][0]
+        the_port.update(invalid_port_vlan_configuration)
+        default_data[Interface.KEY].append(bridge_state)
+
+        with pytest.raises(js.ValidationError) as err:
+            libnmstate.validator.validate(default_data)
+        assert '5000 is greater than the maximum of 4095' in err.value.args[0]
+
+    @staticmethod
+    def _generate_vlan_filtering_config(
+        port_type, trunk_tags=None, access_tag=None, native_vlan=None
+    ):
+        vlan_filtering_state = {
+            LB.Port.Vlan.TYPE: port_type,
+            LB.Port.Vlan.TRUNK_TAGS: trunk_tags or [],
+        }
+
+        if access_tag:
+            vlan_filtering_state[LB.Port.Vlan.TAG] = access_tag
+        if native_vlan:
+            vlan_filtering_state[LB.Port.Vlan.ENABLE_NATIVE] = native_vlan
+
+        return {LB.Port.VLAN_SUBTREE: vlan_filtering_state}
+
+    @staticmethod
+    def _generate_vlan_id_config(*vlan_ids):
+        return [{LB.Port.Vlan.TrunkTags.ID: vlan_id} for vlan_id in vlan_ids]
+
+    @staticmethod
+    def _generate_vlan_id_range_config(min_vlan_id, max_vlan_id):
+        return {
+            LB.Port.Vlan.TrunkTags.ID_RANGE: {
+                LB.Port.Vlan.TrunkTags.MIN_RANGE: min_vlan_id,
+                LB.Port.Vlan.TrunkTags.MAX_RANGE: max_vlan_id,
+            }
+        }
--- a/tests/lib/validator_test.py
+++ b/tests/lib/validator_test.py
@@ -26,6 +26,7 @@ from libnmstate import validator
 from libnmstate.schema import DNS
 from libnmstate.schema import InterfaceIPv4
 from libnmstate.schema import InterfaceIPv6
+from libnmstate.schema import LinuxBridge as LB
 from libnmstate.schema import VXLAN
 from libnmstate.error import NmstateNotImplementedError
 from libnmstate.error import NmstateValueError
@@ -337,6 +338,78 @@ class TestVxlanValidation(object):
             ]
         }
         libnmstate.validator.validate_vxlan(desired_state)
+
+
+class TestVlanFilteringValidation(object):
+    def test_specify_both_vlan_id_and_id_range_is_invalid(self):
+        invalid_vlan_config = {
+            LB.Port.Vlan.TYPE: LB.Port.Vlan.TRUNK_TYPE,
+            LB.Port.Vlan.TRUNK_TAGS: [
+                {
+                    LB.Port.Vlan.TrunkTags.ID: 101,
+                    LB.Port.Vlan.TrunkTags.ID_RANGE: {
+                        LB.Port.Vlan.TrunkTags.MIN_RANGE: 200,
+                        LB.Port.Vlan.TrunkTags.MAX_RANGE: 299,
+                    },
+                }
+            ],
+        }
+        desired_state = {
+            schema.Interface.KEY: [
+                {
+                    schema.Interface.NAME: 'br0',
+                    schema.Interface.TYPE: LB.TYPE,
+                    schema.Interface.STATE: schema.InterfaceState.UP,
+                    LB.PORT_SUBTREE: [
+                        {
+                            LB.PORT_NAME: 'eth1',
+                            LB.Port.VLAN_SUBTREE: invalid_vlan_config,
+                        }
+                    ],
+                }
+            ]
+        }
+        with pytest.raises(NmstateValueError) as err:
+            libnmstate.validator.validate_bridge(desired_state)
+        assert (
+            'Trunk port cannot be configured by both id and range'
+            in err.value.args[0]
+        )
+
+    @pytest.mark.parametrize(
+        'range_key',
+        [LB.Port.Vlan.TrunkTags.MIN_RANGE, LB.Port.Vlan.TrunkTags.MAX_RANGE],
+        ids=['only_min', 'only_max'],
+    )
+    def test_vlan_ranges_must_have_min_and_max(self, range_key):
+        vlan_tag = 101
+        desired_state = {
+            schema.Interface.KEY: [
+                {
+                    schema.Interface.NAME: 'br0',
+                    schema.Interface.TYPE: LB.TYPE,
+                    schema.Interface.STATE: schema.InterfaceState.UP,
+                    LB.PORT_SUBTREE: [
+                        {
+                            LB.PORT_NAME: 'eth1',
+                            LB.Port.VLAN_SUBTREE: {
+                                LB.Port.Vlan.TYPE: LB.Port.Vlan.TRUNK_TYPE,
+                                LB.Port.Vlan.TRUNK_TAGS: [
+                                    {
+                                        LB.Port.Vlan.TrunkTags.ID_RANGE: {
+                                            range_key: vlan_tag
+                                        }
+                                    }
+                                ],
+                            },
+                        }
+                    ],
+                }
+            ]
+        }
+        with pytest.raises(NmstateValueError) as err:
+            libnmstate.validator.validate_bridge(desired_state)
+        assert 'Trunk port range requires min / max keys' in err.value.args[0]
 
 
 def _create_interface_state(


dankenigsberg:
@maiqueb can you provide the example in `preformatted` form? I don't understand the meaning of multiple `vlan-range-min`, but maybe bad indentation is the cause of my confusion. Can you add an example with a second port entry in the bridge?

maiqueb:
> @maiqueb can you provide the example in `preformatted` form? I don't understand the meaning of multiple `vlan-range-min`, but maybe bad indentation is the cause of my confusion. Can you add an example with a second port entry in the bridge?

I don't get what you mean w/ preformatted form, sorry.

This is how an array is expressed in yaml, the indentation is correct.

The following yaml excerpt:
```yaml
vlans:
    - vlan-range-min: 20
      pvid: true
    - vlan-range-min: 21
    - vlan-range-min: 22
    - vlan-range-min: 100
      vlan-range-max: 199
      untagged: true
```

translated into the following json (for example..) code:
```python
{
  "vlans": [
    {
      "vlan-range-min": 20,
      "pvid": true
    },
    {
      "vlan-range-min": 21
    },
    {
      "vlan-range-min": 22
    },
    {
      "vlan-range-min": 100,
      "vlan-range-max": 199,
      "untagged": true
    }
  ]
}
```
Since network manager accepts ranges, I followed what - to me - is the principle of least astonishment, while using fewer constructs - everything is a range, and if you only specify the 'min' of the range, well, it count's as a single item.

It's exactly that the networking API does, for port ranges in security groups, for instance.

dankenigsberg:
> I don't get what you mean w/ preformatted form, sorry.

Apparently you have! Your edited https://github.com/nmstate/nmstate/pull/440#issue-298010056 is much more readable to me. I now understand and can live with multiple `vlan-range-min`.

libnmstate/schemas/operational-state.yaml
```diff
@@ -241,6 +254,23 @@ definitions:
                       type: integer
                     max-age:
                       type: integer
+                vlan-filtering:
+                  type: boolean
+                vlan-default-pvid:
```
maiqueb: The simplest course of action is to not expose this setting in the model, and set it to 0 (which indicates NetworkManager to **not** tag untagged frames) in every created BridgeSettings object.

It can always be set via the boolean 'pvid'.

It would simplify the implementation effort quite a lot.

libnmstate/schemas/operational-state.yaml
```diff
@@ -219,6 +219,19 @@ definitions:
...
+                  vlans:
+                    type: array
+                    items:
+                      type: object
+                      properties:
+                        vlan-range-min:
+                          type: integer
+                        vlan-range-max:
+                          type: integer
+                        pvid:
```
cathay4t: How about `use-default-pvid: True/False`?

libnmstate/schemas/operational-state.yaml
```diff
@@ -241,6 +254,23 @@ definitions:
...
+                    type: object
+                    properties:
+                      vlan-range-min:
+                        type: integer
+                      vlan-range-max:
+                        type: integer
+                      pvid:
+                        type: boolean
+                      untagged:
+                        type: boolean
```
cathay4t: These are duplicated lines. Please create reference to a new definition.

cathay4t:
And schema test is also required. Please check `test/lib/schema_validation_test.py` for detail.

libnmstate/schemas/operational-state.yaml
maiqueb: Check [this](https://github.com/nmstate/nmstate/pull/440#discussion_r303965117) comment.

I think the better way is to hard code the default to 0 (which means leave untagged frames untagged). If the user wants to set a pvid, it can be done via the 'pvid' attribute.

There's an ongoing discussion with the layered product(s) - currently CNV - since it seems they are not interested in the 'pvid' and 'untagged' attributes.

libnmstate/schemas/operational-state.yaml
maiqueb: Done.

maiqueb:
> And schema test is also required. Please check `test/lib/schema_validation_test.py` for detail.

Done.

maiqueb:
> Please include test case for `interface-vlan` schema also.

I don't get it; how does that relate to what I'm doing ? I am not using `interface-vlan` (nor plan to).

cathay4t:
> > Please include test case for `interface-vlan` schema also.
> 
> I don't get it; how does that relate to what I'm doing ? I am not using `interface-vlan` (nor plan to).

My fault. I incorrectly read the line in schema on `interface-vlan` section.

libnmstate/schemas/operational-state.yaml
cathay4t: From the schema, my understanding is:
```
The untagged ingress package will be tagged to `vlan-default-pvid`
when `port['pvid']` is True.
```
If so, the `use-default-pvid` could be a better name.

And I don't know what you are referring in `The simplest course of action is to not expose this setting in the model`.

cathay4t:
@maiqueb We still need some clarification on https://github.com/nmstate/nmstate/pull/440#discussion_r304689068

libnmstate/schemas/operational-state.yaml
maiqueb: > The untagged ingress package will be tagged to `vlan-default-pvid` when `port['pvid']` is True.

No, it's not like that. Untagged frames will be tagged with the value of the default-pvid. For port that have a vlan with the 'pvid' flag set, that vlan will be used to tag the ingress frame, instead of the default one.

It overrides the default pvid.

Regarding the default pvid, I propose to not implement it, and hard-code it to 0 - meaning that by default, the untagged ingress frames will *not* be tagged.

If the user so desires, he can set the 'pvid' flag in one of the vlans, but it has to be done per port.

maiqueb:
> @maiqueb We still need some clarification on [#440 (comment)](https://github.com/nmstate/nmstate/pull/440#discussion_r304689068)

I'll update the commit message, to try to shed some light into the open questions.

libnmstate/schemas/operational-state.yaml
cathay4t: Thanks for the detailed explanation.

I am you remove `vlan-default-pvid` from schema, hard-code it to 0 and let each port define it's pvid.

Another concern:

Which VLAN ID should be choose as PVID when enabled?

maiqueb: 0, which instructs network manager to not tag untagged frames.

maiqueb:
@cathay4t , @EdDev I've modified the schema to comply strictly with CNV requirements - it now only allows for trunk ports.

This is simpler to reason about, and there is no need for access ports at the moment, which enables us to bypass the pvid / untagged properties of the vlan attributes.

Please review.

tests/lib/schema_validation_test.py
```diff
@@ -65,7 +65,22 @@
...
+        {
+            'name': 'br0',
+            'state': 'up',
+            'type': 'linux-bridge',
+            'bridge': {
+                'options': {
+                    'vlan-filtering': True,
+                    'vlans': [
+                        {'vlan-range-min': 100},
+                        {'vlan-range-min': 100, 'vlan-range-max': 200},
```
cathay4t: Seems a overlapping here on VLAN 100.

maiqueb: > If possible, please elaborate the use case for this schema for me to better understand it.
> 
> Thanks.

What kubevirt needs is currently quite simple:
- enable vlan_filtering
- attach a host NIC/bond in trunk mode
- accept it when command-line tools attach a port in trunk or a single
vlan mode.

libnmstate/schemas/operational-state.yaml
```diff
@@ -457,3 +467,14 @@ definitions:
         type: array
         items:
           type: string
+  vlan-filter:
+    type: object
+    properties:
+      vlan-range-min:
+        type: integer
+        minimum: 0
```
qinqon: There is no vlan 0 I think

libnmstate/schemas/operational-state.yaml
```diff
@@ -457,3 +467,14 @@ definitions:
...
+        maximum: 4095
+      vlan-range-max:
+        type: integer
+        minimum: 0
```
qinqon: ditto

qinqon:
ci test please

phoracek:
I think we should make the API way easier. I had issues understanding what's going on. Somebody raised that splitting ingress/egress does not really have a use-case.

@maiqueb would you please consider following simplicity of OVS vlan handling? i.e. vlan-filtering is always enabled, full trunk by default on all ports, can be overridden with specific VLAN (and I believe even a list of trunk ports).

libnmstate/schemas/operational-state.yaml
```diff
@@ -456,3 +470,27 @@ definitions:
         type: array
         items:
           type: string
+  vlan-filter:
+    type: object
+    properties:
+      id:
+        type: integer
```
qinqon: Can we use a yaml reference here to reuse type ?

libnmstate/schemas/operational-state.yaml
```diff
@@ -456,3 +470,27 @@ definitions:
         type: array
         items:
           type: string
+  vlan-filter:
+    type: object
+    properties:
+      id:
```
qinqon: Id is the one that is going to be pvid egress untagged is that it ?

libnmstate/schemas/operational-state.yaml
maiqueb: sure, nice suggestion.

maiqueb:
@phoracek please review.

(I think) I've implemented it as we discussed yesterday. Or, that was my intention :)

libnmstate/schema.py
```diff
@@ -162,11 +162,20 @@ class LinuxBridge(object):
...
+
     PORT_SUBTREE = 'port'
     PORT_NAME = 'name'
     PORT_STP_PRIORITY = 'stp-priority'
     PORT_STP_HAIRPIN_MODE = 'stp-hairpin-mode'
     PORT_STP_PATH_COST = 'stp-path-cost'
+    PORT_VLAN_FILTERING_WHITELIST = 'whitelist'
+
+    VLAN_FILTERING_ENTRIES_VLAN_ID = 'id'
+    VLAN_FILTERING_ENTRIES_VLAN_ID_RANGE = 'id_range'
```
maiqueb: id-range

tests/lib/schema_validation_test.py
```diff
@@ -162,3 +191,158 @@ def test_invalid_state(self, default_data):
...
+        bridge_state[LB.CONFIG_SUBTREE][LB.PORT_SUBTREE] = [
+            port_with_vlan_filter
+        ]
+        with pytest.raises(js.ValidationError) as err:
+            libnmstate.validator.validate(default_data)
+        assert '5000 is greater than the maximum of 4095' in err.value.args[0]
+
+    @staticmethod
+    def _get_bridge_iface_state(interfaces):
+        return interfaces[1]
```
phoracek: Would you please at least find it based on the name or type?

tests/lib/schema_validation_test.py
```diff
@@ -162,3 +191,158 @@ def test_invalid_state(self, default_data):
...
+
+    @staticmethod
+    def _get_bridge_vlan_filtering_state(vlan_filtering_enabled, **kwargs):
```
phoracek: Could you please give a name to "kwargs"? So it is easier to understand what is this method good for

tests/lib/schema_validation_test.py
```diff
@@ -162,3 +191,158 @@ def test_invalid_state(self, default_data):
...
+        valid_bridge_vlan_filtering_state = {
+            'vlan-filtering': {'enabled': vlan_filtering_enabled}
```
phoracek: I like strings instead of constants, but I like uniformity in code even more. And since I was bugged to use constants from schema.py everywhere, would you please do the same?

tests/lib/schema_validation_test.py
```diff
@@ -162,3 +191,158 @@ def test_invalid_state(self, default_data):
...
+class TestVlanFiltering(object):
+    def test_basic_configuration(self, default_data):
+        libnmstate.validator.validate(default_data)
+
+    def test_bridge_vlan_filtering_only_values_on_schema(self, default_data):
+        bridge_state = self._get_bridge_iface_state(default_data[INTERFACES])
+        invalid_vlan_filtering_state = self._get_bridge_vlan_filtering_state(
+            vlan_filtering_enabled=True, invalid_key='ka-boom!'
+        )
+        bridge_state[LB.CONFIG_SUBTREE][
```
phoracek: We have methods to update the whitelist, maybe we can create helper methods for all assignments?

tests/lib/schema_validation_test.py
```diff
@@ -162,3 +191,158 @@ def test_invalid_state(self, default_data):
...
+            LB.OPTIONS_SUBTREE
+        ] = invalid_vlan_filtering_state
+        with pytest.raises(js.ValidationError) as err:
+            libnmstate.validator.validate(default_data)
+        assert '(\'invalid_key\' was unexpected)' in err.value.args[0]
```
phoracek: PEP8 tells us to keep uniform `'` or `"` with one exception: It is better to use the other quotation mark than to escape. Could you please change it to `"('invalid_key' was unexpected)"`

tests/lib/schema_validation_test.py
```diff
@@ -162,3 +191,158 @@ def test_invalid_state(self, default_data):
...
+        self._update_port_vlan_filtering_state(
+            port_with_vlan_filter, [invalid_allow_vlan_range]
+        )
+        bridge_state = self._get_bridge_iface_state(default_data[INTERFACES])
+        bridge_state[LB.CONFIG_SUBTREE][LB.PORT_SUBTREE] = [
+            port_with_vlan_filter
+        ]
+        with pytest.raises(js.ValidationError) as err:
+            libnmstate.validator.validate(default_data)
+        assert '(\'invalid_key\' was unexpected)' in err.value.args[0]
```
phoracek: ditto

tests/lib/schema_validation_test.py
```diff
@@ -162,3 +191,158 @@ def test_invalid_state(self, default_data):
...
+        self._update_port_vlan_filtering_state(
+            port_with_vlan_filter, [invalid_allow_vlan_id]
+        )
+        bridge_state = self._get_bridge_iface_state(default_data[INTERFACES])
+        bridge_state[LB.CONFIG_SUBTREE][LB.PORT_SUBTREE] = [
+            port_with_vlan_filter
+        ]
+        with pytest.raises(js.ValidationError) as err:
+            libnmstate.validator.validate(default_data)
+        assert '(\'invalid_key\' was unexpected)' in err.value.args[0]
```
phoracek: ditto

tests/lib/schema_validation_test.py
```diff
@@ -162,3 +191,158 @@ def test_invalid_state(self, default_data):
...
+        bridge_state = self._get_bridge_iface_state(default_data[INTERFACES])
+        bridge_state[LB.CONFIG_SUBTREE][LB.PORT_SUBTREE] = [
+            invalid_port_config
+        ]
+        with pytest.raises(js.ValidationError) as err:
+            libnmstate.validator.validate(default_data)
+        assert 'is valid under each of' in err.value.args[0]
+
+    def test_invalid_linux_bridge_vlan_range(self, default_data):
+        invalid_vlan_id_type = self._get_vlan_filtering_entry('boom!')
```
phoracek: Is building these dictionaries really worth it? I find it hard to follow the code to see what is going on and to imagine what would be the result. If I saw a mere dictionary, it would be obvious right away.

EdDev:
This PR started as a schema and was aimed for discussing the API before investing further.
I am late on adding my feedback indeed, but this has taken from my point of view a wrong direction.

Please open a new PR with the schema only so we can discuss it next week.
And take this into consideration: I would prefer to see an abstraction of the linux-bridge vlan-filter, not the implementation of it. I would be interested to see something that we can use for the OVS as well, or at least very close to something that makes sense there too.
Just to show the direction I am looking for (but please lets discuss it on a less noisy PR):
```
port:
  - name: eth0
    vlan:
      type: <trunk, access>
      trunk-tags: [1-100, 300-100, 2000, 3000, 4000-4095] // Used if type==trunk
      access-tag: 101  // Used if type==access or if type==trunk and enable-native==true
      enable-native: true // Used if type==trunk
```

If someone is interested to have an 1:1 mapping between the Linux bridge vlan-filter API and nmstate, I will make sure to provide a pluggable way to expose a 3rd party implementation in Nmstate.

phoracek:
@EdDev late indeed. Looks ok, except for ranges encoded in strings.

EdDev:
@phoracek , it was not a priority until now but you guys managed to raise my attention.

maiqueb:
> This PR started as a schema and was aimed for discussing the API before investing further.
> I am late on adding my feedback indeed, but this has taken from my point of view a wrong direction.
> 
> Please open a new PR with the schema only so we can discuss it next week.

What's the advantage for that ? There's relevant information here. 

> And take this into consideration: I would prefer to see an abstraction of the linux-bridge vlan-filter, not the implementation of it. I would be interested to see something that we can use for the OVS as well, or at least very close to something that makes sense there too.

I can't say I'm fond of abstracting the vlan filtering instead of just wrapping over the NM's vlan filtering functionality. I sincerely think that model is a lot more straight-forward that the abstraction you're proposing below. 

Also, this is something that will go under the 'linux-bridge' object, and I think that  whoever uses vlan-filtering will faster look for what it is in the linux bridge documentation (it's in the linux-bridge object) rather that what does cisco swtiches do. 

But I'll go with the flow; you're the maintainer, and @phoracek is the consumer, so, if he and his team actually prefer this, I'll comply.

Finally, and not taking into account my personal preference, I can't pretend that having something that can be leveraged also for OVS is not nice to have. 

> Just to show the direction I am looking for (but please lets discuss it on a less noisy PR):
> 
> ```
> port:
>   - name: eth0
>     vlan:
>       type: <trunk, access>
>       trunk-tags: [1-100, 300-100, 2000, 3000, 4000-4095] // Used if type==trunk
>       access-tag: 101  // Used if type==access or if type==trunk and enable-native==true
>       enable-native: true // Used if type==trunk
> ```
> 

Looking at your proposal in a constructive way, I think it could be improved:
  - remove the type attribute. Feel's redundant, and leaves room for plenty of 'WTF' moments. Like, type=trunk, but having the 'access-tag' set up.
  - presence of trunk-tags makes the port a trunk port.
  - absence of trunk-tags **and** presence of access-tag makes it an access port.
  - absence of trunk-tags **and** absence of access-tag makes it a nonsense port, which the schema must reject. 
  - enable-native should be an integer. This is literally the PVID of a linux bridge port. I frown upon your suggestion because when you have a trunk-tag list, and set enable-native = True, what's gonna happen ?  Set the first ? Set the last ? ... I'd also rename the attribute to 'native-vlan'.
  - native-vlan on a trunk port must have an ID specified on the trunk tags. (alternatively, we could auto-add the native vlan to the list when provided.)
  - native-vlan on an access port **must** have the vlan-id specified in the 'access-tag' attribute.

> If someone is interested to have an 1:1 mapping between the Linux bridge vlan-filter API and nmstate, I will make sure to provide a pluggable way to expose a 3rd party implementation in Nmstate.

@EdDev / @phoracek what do you think ?

phoracek:
@maiqueb I think that your comments on @EdDev's proposal ilustrate its drawback - although it is easier to read, it may be problematic to extend it in case we have users demanding better control.

As a consumer of this feature, all I care about that by default vlan-filtering is enabled and all ports are treated as trunks with PVID 1.

As a user of nmstate, I prefer @EdDev's API because of it requires me to think less. This is personal though since I'm familiar with OVS API.

As a developer, I prefer @maiqueb's proposal, since it keeps door open for future extending.

EdDev:
> Also, this is something that will go under the 'linux-bridge' object, and I think that whoever uses vlan-filtering will faster look for what it is in the linux bridge documentation

If the user is good with that, then he should use ansible or go directly and configure his network through NM and its ifcfg files.
The vlan-filtering feature in the linux-bridge is very hard to understand and very foreign to a network engineer. Not only that, it is built for maximum configurability, including squashing vlan/s and all kind of extreme implementations.
An API is suppose to serve and protect the user, not to expose it to the little bits.
One implementation you could do with vlan-filtering is to implement private VLAN/s. I prefer to expose it as such and not through the bits.

>   - remove the type attribute. Feel's redundant, and leaves room for plenty of 'WTF' moments. Like, type=trunk, but having the 'access-tag' set up.

I do not think it is possible unless we must specify the full port-vlan state, see below.

>   - presence of trunk-tags makes the port a trunk port.
>   - absence of trunk-tags **and** presence of access-tag makes it an access port.
>   - absence of trunk-tags **and** absence of access-tag makes it a nonsense port, which the schema must reject. 

This is not supporting partial desired state definition. Nmstate allows partial states, meaning that absence of a property means it should not change, not that it needs to be removed.
We could require full port-vlan definition, but that is the special thing and not the opposite.
I personally like partial states, but only if it is not over-complex (which is not in this case IMO).

>   - enable-native should be an integer. This is literally the PVID of a linux bridge port. I frown upon your suggestion because when you have a trunk-tag list, and set enable-native = True, what's gonna happen ?  Set the first ? Set the last ? ... I'd also rename the attribute to 'native-vlan'.

It means that the access-tag (you can name it differently, whatever is clear... OVS calls it `tag`) contains the ID. It is very unintuitive to me (native vlan is in practice a merge between the access tag and the trunk tags).

>   - native-vlan on a trunk port must have an ID specified on the trunk tags. (alternatively, we could auto-add the native vlan to the list when provided.)
>   - native-vlan on an access port **must** have the vlan-id specified in the 'access-tag' attribute.

If you refer to a native vlan that is both untag and tag in parallel then no,
it is not a must, it is actually an exception. If this will need to be managed, you can add a a flag to control it. (OVS uses the mode for that, which is sooo ugly and non intuitive.)

> ilustrate its drawback - although it is easier to read, it may be problematic to extend it in case we have users demanding better control.
> As a developer, I prefer @maiqueb's proposal, since it keeps door open for future extending.

Could you please provide concrete examples?
Obviously, with any abstraction there is a risk to loose flexibility. I am in favour of trying to challenge the abstractions, it will make them better.

maiqueb:
Thanks for this detailed reply, I appreciate the time. 

Unfortunately, I still don't agree, and think you've dismissed my suggestion too soon.

> > Also, this is something that will go under the 'linux-bridge' object, and I think that whoever uses vlan-filtering will faster look for what it is in the linux bridge documentation
> 
> If the user is good with that, then he should use ansible or go directly and configure his network through NM and its ifcfg files.
> The vlan-filtering feature in the linux-bridge is very hard to understand and very foreign to a network engineer. Not only that, it is built for maximum configurability, including squashing vlan/s and all kind of extreme implementations.
> An API is suppose to serve and protect the user, not to expose it to the little bits.
> One implementation you could do with vlan-filtering is to implement private VLAN/s. I prefer to expose it as such and not through the bits.
> 
> > * remove the type attribute. Feel's redundant, and leaves room for plenty of 'WTF' moments. Like, type=trunk, but having the 'access-tag' set up.
> 
> I do not think it is possible unless we must specify the full port-vlan state, see below.
> 
> > * presence of trunk-tags makes the port a trunk port.
> > * absence of trunk-tags **and** presence of access-tag makes it an access port.
> > * absence of trunk-tags **and** absence of access-tag makes it a nonsense port, which the schema must reject.
> 
> This is not supporting partial desired state definition. Nmstate allows partial states, meaning that absence of a property means it should not change, not that it needs to be removed.

What I mean here by absence of trunk tags is:

```yaml
port:
  - name: eth0
    vlan:
      trunk-tags: []
      access-tag: 101  
```
The above example represents an access port.

Below example represents a trunk port: 
```yaml
port:
  - name: eth0
    vlan:
      trunk-tags: [14, 100-200, ...]
```

Finally, the following yaml would leave the trunks as they are - meaning it would set the 101 VLAN ID as the PVID of the port.
```yaml
port:
  - name: eth0
    vlan:
      access-tag: 101  
```

> We could require full port-vlan definition, but that is the special thing and not the opposite.
> I personally like partial states, but only if it is not over-complex (which is not in this case IMO).
> 
> > * enable-native should be an integer. This is literally the PVID of a linux bridge port. I frown upon your suggestion because when you have a trunk-tag list, and set enable-native = True, what's gonna happen ?  Set the first ? Set the last ? ... I'd also rename the attribute to 'native-vlan'.
> 
> It means that the access-tag (you can name it differently, whatever is clear... OVS calls it `tag`) contains the ID. It is very unintuitive to me (native vlan is in practice a merge between the access tag and the trunk tags).

And that's intuitive ? I fail to see how. But, could be me that's broken.

If I understand correctly, if you want to define a trunk port having native vlan as 101, you'd have to:
```yaml
port:
  - name: eth0
    vlan:
      type: trunk
      trunk-tags: [1-100, ...] 
      access-tag: 101 
      enable-native: true
```

Again, it could be me that's broken, but, IMHO, the above seems miles away from intuitive. There are dependencies between the attributes - e.g. the presence of one attribute, changes the meaning of a previous attribute. 

I still think that your enable-native flag should be the vlan ID of the native VLAN instead. 

> 
> > * native-vlan on a trunk port must have an ID specified on the trunk tags. (alternatively, we could auto-add the native vlan to the list when provided.)
> > * native-vlan on an access port **must** have the vlan-id specified in the 'access-tag' attribute.
> 
> If you refer to a native vlan that is both untag and tag in parallel then no,

Isn't that what a native vlan is ? I mean, on ingress, untagged frames get tagged with that ID, and ingress traffic tagged with that ID becomes untagged.

My mind starts racing when an **access port** with tag 101 has ID 1000 as its native ID. I think following the principle of least astonishment is a good thing. And that, is astonishing to me. I sincerely feel that's like wanting to turn left and right at the same time.

> it is not a must, it is actually an exception. If this will need to be managed, you can add a a flag to control it. (OVS uses the mode for that, which is sooo ugly and non intuitive.)
> 
> > ilustrate its drawback - although it is easier to read, it may be problematic to extend it in case we have users demanding better control.
> > As a developer, I prefer @maiqueb's proposal, since it keeps door open for future extending.
> 
> Could you please provide concrete examples?
> Obviously, with any abstraction there is a risk to loose flexibility. I am in favour of trying to challenge the abstractions, it will make them better.

EdDev:
> Isn't that what a native vlan is ? I mean, on ingress, untagged frames get tagged with that ID, and ingress traffic tagged with that ID becomes untagged.

“Native” is about untagged frames arriving untag on a trunk, getting associated to the mentioned ID. On egress, a symmetric setup will tx frames associated to the native vlan id untagged. There is no conversion involved.
  
> 
> My mind starts racing when an access port with tag 101 has ID 1000 as its native ID. I think following the principle of least astonishment is a good thing. And that, is astonishing to me. I sincerely feel that's like wanting to turn left and right at the same time.

No idea how you reached this state, sorry. Where exactly do you spot a scenario where we have both a native id and an access id?
We have one vlan id that can refer to an access vlan mode (no tagging) or a trunk mode, in which both trunk tags and non tag frames are supported.

maiqueb:
> > Isn't that what a native vlan is ? I mean, on ingress, untagged frames get tagged with that ID, and ingress traffic tagged with that ID becomes untagged.
> 
> “Native” is about untagged frames arriving untag on a trunk, getting associated to the mentioned ID. On egress, a symmetric setup will tx frames associated to the native vlan id untagged. There is no conversion involved.

I meant to say "... on ingress, untagged frames get tagged with that ID, and ***egress*** traffic tagged with that ID becomes untagged". My bad. With that, I think we both mean the same thing whenever we mention native vlan.

> 
> > My mind starts racing when an access port with tag 101 has ID 1000 as its native ID. I think following the principle of least astonishment is a good thing. And that, is astonishing to me. I sincerely feel that's like wanting to turn left and right at the same time.
> 
> No idea how you reached this state, sorry. Where exactly do you spot a scenario where we have both a native id and an access id?
> We have one vlan id that can refer to an access vlan mode (no tagging) or a trunk mode, in which both trunk tags and non tag frames are supported.

That scenario is not real - I was only 'challenging' my own suggested model of having your **enable-native** parameter represented as an integer, **but** on an access port where the 'native' vlan would be **different** from the access tag. From my narrow understanding, that would somehow represent a trunk port, and the model would stop making sense. Which was why I stated that "native-vlan on an access port must have the vlan-id specified in the 'access-tag' attribute" . 

What about your claim that my suggestion would make partial states impossible ? Did you take a look into the examples I provided ?

EdDev:
I think that the best way to proceed here is to provide snippets of the scenarios.
- define access port vlan 101.
- define trunk port, 102-200, no native.
- define trunk port, 102-200, native vlan 100.

Then show an editation of:
- vlan id of an access port.
- vlan id of an native vlan on a trunk port.
- add a tag id to a trunk.
- change an access port to trunk, keep the untag vlan working.

I will add it to my todo.

EdDev:
- define access port vlan 101.
```
port:
  - name: eth0
    vlan:
      type: access
      access-tag: 101
```
- define trunk port, 102-200, no native.
```
port:
  - name: eth0
    vlan:
      type: trunk
      trunk-tags: ['102-200']
      enable-native: false
```
- define trunk port, 102-200, native vlan 100.
```
  - name: eth0
    vlan:
      type: trunk
      trunk-tags: ['102-200']
      access-tag: 100
      enable-native: true
```

Editation: (based on the previous snippets)
- vlan id of an access port.
OR
- vlan id of an native vlan on a trunk port.
```
port:
  - name: eth0
    vlan:
      access-tag: 201
```
- add a tag id to a trunk.
```
port:
  - name: eth0
    vlan:
      trunk-tags: ['102-200', '300']
```
- change an access port to trunk, keep the untag vlan working.
```
  - name: eth0
    vlan:
      type: trunk
      trunk-tags: ['102-200']
```

phoracek:
@EdDev would you consider using:

```
range:
  min: X
  max: Y
```

Instead of `["X-Y"]`. Why would you implement your own syntax to handle ranges? I miss the benefit, it just smells like scripted API that makes the handling harder for machines while making it only slightly more readable for humans.

EdDev:
> Finally, the following yaml would leave the trunks as they are - meaning it would set the 101 VLAN ID as the PVID of the port.
> ```yaml
> port:
>   - name: eth0
>     vlan:
>       access-tag: 101  
> ```

How do you differentiate between the user wanted to change from trunk to access and this native vlan editation?

EdDev:
> would you consider using:
> 
> ```
> range:
>   min: X
>   max: Y
> ```
> 
> Instead of `["X-Y"]`. Why would you implement your own syntax to handle ranges? I miss the benefit, it just smells like scripted API that makes the handling harder for machines while making it only slightly more readable for humans.

You mean like this: (?)
```
trunk-tags:
  - single: 101
  - range:
      min: 110
      max: 200
```

Or do you want to use range with min==max?

phoracek:
@EdDev something like that. I don't care that much about naming as long as it would be reasonably consistent. What I care about is that it is structured on YAML level.

```yaml
trunk-tags:
  - id: 101
  - id-range:
      min: 110
      max: 200
```

EdDev:
@phoracek I’m fine with that.
The string can be formatted using a regex and it will fit a yaml, but I like your option as well.

tests/lib/schema_validation_test.py
maiqueb: Let's see if you buy it as is.

maiqueb:
> > Finally, the following yaml would leave the trunks as they are - meaning it would set the 101 VLAN ID as the PVID of the port.
> > ```yaml
> > port:
> >   - name: eth0
> >     vlan:
> >       access-tag: 101  
> > ```
> 
> How do you differentiate between the user wanted to change from trunk to access and this native vlan editation?

Right. After considering further, I've followed your model, to allow partial states. I still am not convinced that's needed - or even a good thing to have here - but, it's consistent with all the other features.

Hope this is closer to what you were looking for.

tests/lib/schema_validation_test.py
```diff
@@ -162,3 +205,178 @@ def test_invalid_state(self, default_data):
...
+            libnmstate.validator.validate(default_data)
+        assert "('invalid_key' was unexpected)" in err.value.args[0]
+
+        valid_vlan_filtering_state = self._get_bridge_vlan_filtering_state(
+            vlan_filtering_enabled=False
+        )
+        bridge_state[LB.CONFIG_SUBTREE][
+            LB.OPTIONS_SUBTREE
+        ] = valid_vlan_filtering_state
+        libnmstate.validator.validate(default_data)
```
maiqueb: This code is probably not needed, since it's covered in [test_trunk_port_native_vlan](https://github.com/nmstate/nmstate/pull/440/files#diff-d4e1478b9ed666021932b2c056397a85R305)

phoracek:
> at the moment, we will not allow setting allowed vlans on the bridge port. It will be hardcoded with the full range - 1 <= vlan_id <= 4095.

It should be 2-4095 for trunk and 1 for native

phoracek:
@maiqueb I don't think I understand the last example:

```yaml
            vlan:
              type: trunk
              trunk-tags:
                - id: 101
                - id-range:
                    min: 200
                    max: 4095
              tag: 100
              enable-native: true
```

What is the `tag` good for when used with `type: trunk`?

What does `enable-native` do? Is it bound to the `tag`?

maiqueb:
> @maiqueb I don't think I understand the last example:
> 
> ```yaml
>             vlan:
>               type: trunk
>               trunk-tags:
>                 - id: 101
>                 - id-range:
>                     min: 200
>                     max: 4095
>               tag: 100
>               enable-native: true
> ```
> 
> What is the `tag` good for when used with `type: trunk`?

It's meant to be used with the enable-native flag, see below.

> 
> What does `enable-native` do? Is it bound to the `tag`?

I also don't like it, but couldn't come up w/ anything simpler.

This sets vlan 100 (the one specified in tag) as the native vlan on that trunk. So yes, it's bound to `tag`.

maiqueb:
> > at the moment, we will not allow setting allowed vlans on the bridge port. It will be hardcoded with the full range - 1 <= vlan_id <= 4095.
> 
> It should be 2-4095 for trunk and 1 for native

Why ? id 1 is the default pvid, but, it could 1333. I don't follow.

phoracek:
> > > at the moment, we will not allow setting allowed vlans on the bridge port. It will be hardcoded with the full range - 1 <= vlan_id <= 4095.
> > 
> > 
> > It should be 2-4095 for trunk and 1 for native
> 
> Why ? id 1 is the default pvid, but, it could be 1333- e.g. any other int in the range. I don't follow.

But we are speaking about the default there. I would expect that the bridge's default VLAN setting would be full trunk and native untagged on ID 1. I'm trying to make sure we are on the same page there. If you create a bridge and don't set the native ID explicitly, it would not be possible to use the bridge for the default traffic.

maiqueb:
> > > > at the moment, we will not allow setting allowed vlans on the bridge port. It will be hardcoded with the full range - 1 <= vlan_id <= 4095.
> > > 
> > > 
> > > It should be 2-4095 for trunk and 1 for native
> > 
> > 
> > Why ? id 1 is the default pvid, but, it could be 1333- e.g. any other int in the range. I don't follow.
> 
> But we are speaking about the default there. I would expect that the bridge's default VLAN setting would be full trunk and native untagged on ID 1. I'm trying to make sure we are on the same page there. If you create a bridge and don't set the native ID explicitly, it would not be possible to use the bridge for the default traffic.

You're right, I'm wrong - I missed that you were mentioning the default, sorry for that. 

I'll update the commit msg and the PR description accordingly.

libnmstate/schemas/operational-state.yaml
```diff
@@ -240,6 +261,12 @@ definitions:
                       type: integer
                     max-age:
                       type: integer
+                vlan-filtering:
```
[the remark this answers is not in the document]
maiqueb: We can re-use this definition for the uplink / self bridge port. 

I see no reason at all to come up w/ yet another representation.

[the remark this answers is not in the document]
maiqueb: I'm *just* providing a way for the user to disable vlan filtering altogether. I think it's useful.

Do you need an example for that ? 

Furthermore, if you really want the uplink port defined, we can set the vlan port definition also here; it would apply to the uplink / self port .

libnmstate/schemas/operational-state.yaml
```diff
@@ -456,3 +483,24 @@ definitions:
         type: array
         items:
           type: string
+  vlan-filter:
```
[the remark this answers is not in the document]
maiqueb: In the reply to the last patch set, you indicate your desire to also model the uplink port - despite that not being needed for kubevirt. I'm OK with that.

The implementation for that would re-use the vlan filtering definition.

This can be called multiple things, but I'd really rather call it what it is. A vlan filter. Please suggest an alternate name if you dislike this one so much.

Yes, it could be re-used by OVS , the uplink linux bridge port, etc. Since we've agreed a model, I think re-using it whenever possible is preferable.

libnmstate/schemas/operational-state.yaml
```diff
@@ -456,3 +483,24 @@ definitions:
...
+      id-range:
+        type: object
+        properties:
+          min:
+            $ref: "#/definitions/types/vlan-id"
+          max:
+            $ref: "#/definitions/types/vlan-id"
+        required:
+          - min
+          - max
```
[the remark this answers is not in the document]
maiqueb: > Could you please impose this in the code (`validator` module)? The schema errors are not that easy to read when something fails.

You know my opinion about this - e.g. what's the point of using the schema then ?...

Also, I think the tests I've added show that it really is easy to read, since it features very specific strings about what has happened.

Never the less, I'll bend, since that's consistent with the reminder of the project.

libnmstate/schemas/operational-state.yaml
```diff
@@ -456,3 +483,24 @@ definitions:
...
+    additionalProperties: false
```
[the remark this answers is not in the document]
maiqueb: Noted.

libnmstate/schemas/operational-state.yaml
```diff
@@ -456,3 +483,24 @@ definitions:
...
+    oneOf:
+      - required:
+          - id
+      - required:
+          - id-range
```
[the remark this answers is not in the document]
maiqueb: Noted.

libnmstate/schemas/operational-state.yaml
```diff
@@ -240,6 +261,12 @@ definitions:
                       type: integer
                     max-age:
                       type: integer
+                vlan-filtering:
+                  type: object
+                  properties:
+                    enabled:
+                      type: boolean
+                  additionalProperties: false
```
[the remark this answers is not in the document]
maiqueb: Noted.

libnmstate/schemas/operational-state.yaml
```diff
@@ -218,6 +222,23 @@ definitions:
...
+                          - access
+                      trunk-tags:
+                        type: array
+                        items:
+                          $ref: "#/definitions/vlan-filter"
+                      tag:
+                        $ref: "#/definitions/types/vlan-id"
+                      enable-native:
+                        type: boolean
+                    additionalProperties: false
```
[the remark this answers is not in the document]
maiqueb: Good point.

libnmstate/schemas/operational-state.yaml
```diff
@@ -55,6 +55,10 @@ definitions:
     mac-address:
       type: string
       pattern: "^([a-fA-F0-9]{2}:){5}[a-fA-F0-9]{2}$"
+    vlan-id:
+      type: integer
+      minimum: 1
+      maximum: 4095
```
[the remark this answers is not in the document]
maiqueb: I don't want to extract the existing definition - they are indeed different things - as clearly shown  by one's minimum being 0 and the other's minimum being 1. 

I see no reason to consolidate them.

What if I just rename this 'vlan-filter-id' ?

maiqueb:
> > at the moment,
> > we will not allow setting allowed vlans on the bridge port.
> > It will be hardcoded with the full range - 2 <= vlan_id <= 4095,
> > having vlan id 1 as native.
> 
> This sounds like all trunk tag traffic and the non tagged traffic will have access to the bridge IP. I would claim this is illegal, or at least, an exception in a common configuration.
> 
> Originally I thought we can skip this, but the subject that the text raised is showing to be too important to ignore. The bridge interface in many cases is used for management purposes and defining on which VLAN the management is to be enabled is a basic need.
> 
> I can think of two options to model this:
> 
> * Define a new property on the bridge options that defines the VLAN on which the interface will be associated with. Like `interface-vlan-id=100`.
>   Based on the VLAN value, the configuration will be done.
> * If the default is not good (native/access, e.g. the bridge interface is communicating on the non-tag vlan), then specify explicitly to what VLAN it belongs by adding a "self" port definition. We can identify the self port by the name (if the name is the same as the bridge).
> 
> What do you think? (personally, I prefer the 2nd option because it avoids the need to add another property that is non intuitive IMO)

I'd prefer to just emulate the linux bridge API and have it explicitly defined on the vlan filtering configuration, but, can accept 'modulating' it as another port in the port list whose name is the bridge name. 

I'm assuming that other options - hairpin / stp - would be equally available on this 'self' port.

libnmstate/schemas/operational-state.yaml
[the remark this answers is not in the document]
EdDev: status, mac and vlan-id are common to many interfaces, this one is specific on a single type.
But if you prefer not to consolidate, then 'vlan-filter-id' is better than the current name.

I am also not exactly sure if `0` is an invalid value for a trunk vlan ID, in normal switches that indicates a frame with 802.1q header, using only the priority fields and associating the the frames to the native vlan.

libnmstate/schemas/operational-state.yaml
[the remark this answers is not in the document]
EdDev: > I'm just providing a way for the user to disable vlan filtering altogether. I think it's useful.

I'm asking about the two level part, not about the need to have this enable/disable option.
Looking for the reason this simpler options is not enough: `vlan-filtering: False`

> Do you need an example for that ?

I need an example of a future property that you think may fit under this hierarchy (therefore, explaining its need).

> Furthermore, if you really want the uplink port defined, we can set the vlan port definition also here; it would apply to the uplink / self port .

We had two options, one to set it perhaps here or flatted under options, or mention it in the port list. If you prefer the first, then it could be the example I was looking for.

libnmstate/schemas/operational-state.yaml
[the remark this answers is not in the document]
EdDev: In my view, vlan-filter is the implementation and the model is port vlan support.
`bridge-port-vlan` sounds more appropriate, but you can suggest anything that describes what this represents.

libnmstate/schemas/operational-state.yaml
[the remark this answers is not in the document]
EdDev: > You know my opinion about this - e.g. what's the point of using the schema then ?

This has been answered already. It is used to describe the API in a common format/language. It allows other tools to generate client code automatically and so on.

> Also, I think the tests I've added show that it really is easy to read, since it features very specific strings about what has happened.

If this is readable I am ok to keep it.
Could you please paster here an output from the shell when this happens?

libnmstate/schemas/operational-state.yaml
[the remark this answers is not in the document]
maiqueb: > status, mac and vlan-id are common to many interfaces, this one is specific on a single type.
> But if you prefer not to consolidate, then 'vlan-filter-id' is better than the current name.

ACK. 

> 
> I am also not exactly sure if `0` is an invalid value for a trunk vlan ID, in normal switches that indicates a frame with 802.1q header, using only the priority fields and associating the the frames to the native vlan.

I can only provide the output of attempting to do it, please let me know if I'm making any wrong assumption:

```bash
[root@vdsm-rhel81 ~]# bridge vlan
port	vlan ids
;vdsmdummy;
	 1 PVID untagged
linux-br0	None

[root@vdsm-rhel81 ~]# bridge vlan add vid 0 dev linux-br0 self
RTNETLINK answers: Invalid argument

[root@vdsm-rhel81 ~]# bridge vlan add vid 10 dev linux-br0 self

[root@vdsm-rhel81 ~]# bridge vlan
port	vlan ids
;vdsmdummy;
	 1 PVID untagged
linux-br0
	 10

```

libnmstate/schemas/operational-state.yaml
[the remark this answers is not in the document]
maiqueb: > > I'm just providing a way for the user to disable vlan filtering altogether. I think it's useful.
> 
> I'm asking about the two level part, not about the need to have this enable/disable option.
> Looking for the reason this simpler options is not enough: `vlan-filtering: False`

Oh, I see. From past conversation w/ @phoracek, we chose the option of having it as a map to possibly incorporate further data - e.g. the 'self' port vlan filtering configuration.

Choosing to model that port as 'just another port' would indeed lessen the requirement for having that property within a map.

It also feels more in line with what's on STP, for instance (again, **only if** there are more attributes).

> 
> > Do you need an example for that ?
> 
> I need an example of a future property that you think may fit under this hierarchy (therefore, explaining its need).

Roger that.

> 
> > Furthermore, if you really want the uplink port defined, we can set the vlan port definition also here; it would apply to the uplink / self port .
> 
> We had two options, one to set it perhaps here or flatted under options, or mention it in the port list. If you prefer the first, then it could be the example I was looking for.

Will consider both options, then come up with a proposal. Too much uncertainty at the moment.

libnmstate/schemas/operational-state.yaml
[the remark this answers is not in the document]
maiqueb: Fair enough.

libnmstate/schemas/operational-state.yaml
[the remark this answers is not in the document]
maiqueb: Will do.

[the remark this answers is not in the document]
maiqueb: Ended up following your clue, and implementing such things in the validator module.

libnmstate/schema.py
```diff
@@ -165,11 +165,24 @@ class LinuxBridge(object):
...
+
     PORT_SUBTREE = 'port'
     PORT_NAME = 'name'
     PORT_STP_PRIORITY = 'stp-priority'
     PORT_STP_HAIRPIN_MODE = 'stp-hairpin-mode'
     PORT_STP_PATH_COST = 'stp-path-cost'
+    PORT_VLAN_SUBTREE = 'vlan'
+    PORT_VLAN_TYPE = 'type'
+    PORT_VLAN_TRUNK_TAGS = 'trunk-tags'
+    PORT_VLAN_ACCESS_TAG = 'tag'
```
EdDev: You left `ACCESS` intentionally?

libnmstate/schema.py
```diff
@@ -165,11 +165,24 @@ class LinuxBridge(object):
...
+    VLAN_FILTERING_SUBTREE = 'vlan-filtering'
+    VLAN_FILTERING_ENABLED = 'enabled'
+
     PORT_SUBTREE = 'port'
     PORT_NAME = 'name'
     PORT_STP_PRIORITY = 'stp-priority'
     PORT_STP_HAIRPIN_MODE = 'stp-hairpin-mode'
     PORT_STP_PATH_COST = 'stp-path-cost'
+    PORT_VLAN_SUBTREE = 'vlan'
+    PORT_VLAN_TYPE = 'type'
```
EdDev: What about the values?

libnmstate/schema.py
```diff
@@ -165,11 +165,24 @@ class LinuxBridge(object):
...
 
+    VLAN_FILTERING_SUBTREE = 'vlan-filtering'
+    VLAN_FILTERING_ENABLED = 'enabled'
+
     PORT_SUBTREE = 'port'
     PORT_NAME = 'name'
     PORT_STP_PRIORITY = 'stp-priority'
     PORT_STP_HAIRPIN_MODE = 'stp-hairpin-mode'
     PORT_STP_PATH_COST = 'stp-path-cost'
+    PORT_VLAN_SUBTREE = 'vlan'
```
EdDev: The repeating `PORT_VLAN` prefix hints for a separate enum/class.

libnmstate/schema.py
```diff
@@ -165,11 +165,24 @@ class LinuxBridge(object):
...
+    PORT_VLAN_TYPE = 'type'
+    PORT_VLAN_TRUNK_TAGS = 'trunk-tags'
+    PORT_VLAN_ACCESS_TAG = 'tag'
+    PORT_VLAN_ENABLE_NATIVE = 'enable-native'
+
+    PORT_VLAN_ENTRIES_ID = 'id'
+    PORT_VLAN_ENTRIES_ID_RANGE = 'id-range'
+    PORT_VLAN_ENTRIES_MIN_RANGE = 'min'
+    PORT_VLAN_ENTRIES_MAX_RANGE = 'max'
```
EdDev: It is unclear that this belongs below the IP_RANGE.
How about:
`LinuxBridge.Port.Vlan.TrunkTags.ID_RANGE_MAX`

We should start using sub-classes in general, it looks more readable.

libnmstate/schemas/operational-state.yaml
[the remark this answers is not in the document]
EdDev: I am a bit lost.
This one has been left as is (the name) and is not used with regular VLANs?

Regarding ID 0, please separate between the implementation and the abstraction. With Linux-Bridge vlan-filtering, how is a VLAN ID of 0 frame handled on ingress? What about sending frames with tag 0? Nothing like that is supported?

libnmstate/schemas/operational-state.yaml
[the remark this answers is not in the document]
EdDev: Ok, so lets not implement this here as you said. We can interpret if it is enabled or not for now.
If vlans are set, then this is implicitly enabled, otherwise, it is disabled. What do you think?

libnmstate/schemas/operational-state.yaml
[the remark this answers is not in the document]
EdDev: Did I missed something?

libnmstate/validator.py
```diff
@@ -238,3 +247,37 @@ def _assert_vxlan_has_missing_attribute(state, *attributes):
...
+                trunk_tag_state
+            )
+        )
+    elif vlan_id_range:
+        if not (
+            {
+                LB.Port.Vlan.TrunkTags.MIN_RANGE,
+                LB.Port.Vlan.TrunkTags.MAX_RANGE,
+            }
+            <= set(vlan_id_range)
```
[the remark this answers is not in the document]
EdDev: I think that the difference is that here the `not` is far from the `<=`.
But your reasoning makes sense.
You can extract those attributed/keys to a helper variable but it is not a must, up to you.

libnmstate/schemas/operational-state.yaml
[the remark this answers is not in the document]
maiqueb: > I would prefer something more generic, like: 'bridge-vlan-id' or 'bridge-vlan-tag', such that we can reuse it (possibly) with OVS later on.

Renamed it to 'bridge-vlan-tag'.

> For example, in trunk mode with a trunk tag id of 0 can be interpreted as "accept packets with tag 0 and associate them to the vlan specified by the tag". Just an example.

I don't follow your example, would you explain it better ? Got confused on 'accept packets with tag 0 and associate them to the vlan specified by the tag'. What tag ? Furthermore, implementation wise, how can I configure the linux bridge to accept packets tagged w/ vlan tag 0 ? All I can do is try *and* watch it fail.

Independently, I'll allow vlan tag 0. In the implementation, we can always open up a bug on NM, write a test, and xfail it.

PR #445, opened 2019-07-21
Delete master without slave

EdDev:
Support the scenario where the master interface is removed but one of its slaves is explicitly left UP.
Currently this scenario results in the deletion of the slave and therefore the transaction fails on the verification step.

The PR solves this at both the metadata level (such slaves are not updated with master metadata) and the admin state update (not removing these slaves).

### DIFF
--- a/tests/integration/linux_bridge_test.py
+++ b/tests/integration/linux_bridge_test.py
@@ -105,6 +105,38 @@ def test_create_and_remove_linux_bridge_with_two_ports(port0_up, port1_up):
     assertlib.assert_absent(bridge_name)
 
 
+def test_remove_bridge_and_keep_slave_up(bridge0_with_port0, port0_up):
+    bridge_name = bridge0_with_port0[Interface.KEY][0][Interface.NAME]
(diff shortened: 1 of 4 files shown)


EdDev:
@cathay4t , @maiqueb , this is resolving one of the blockers we have in the vdsm-nmstate integration, please prioritize its review.

tests/integration/linux_bridge_test.py
```diff
@@ -105,6 +105,41 @@ def test_create_and_remove_linux_bridge_with_two_ports(port0_up, port1_up):
...
+            {
+                Interface.NAME: bridge_name,
+                Interface.STATE: InterfaceState.ABSENT,
+            },
+            {
+                Interface.NAME: port_name,
+                Interface.STATE: InterfaceState.UP,
+                Interface.IPV4: {'enabled': False},
+                Interface.IPV6: {'enabled': False},
+            },
```
cathay4t: You can reuse `port_desired_state` here. Not a must.

EdDev: Right, thanks.

PR #446, opened 2019-07-21
Show with nmclient refresh

EdDev:
Refresh nmclient for each `show` call

NM client represents the interface to NM.
In most cases, it is up to date. However, immediately after the mainloop
is completed, the verification step kicks in and calls `show`, at which
point the nmclient seems to be out-of-sync.

For consistency, on each call to `show`, the NM client is refreshed
(object recreated).

Note: `show` should not be called from within the mainloop as the NM
client must remain unchanged.

Depends on #445


EdDev:
This issue has been detected during vdsm-nmstate tests (basic bridge tests).

PR #451, opened 2019-07-29
Fix ipv6 related issues

EdDev:
Fix several issues related to IPv6 in both the production code and integration tests.

### DIFF
--- a/libnmstate/netapplier.py
+++ b/libnmstate/netapplier.py
@@ -22,6 +22,7 @@ from contextlib import contextmanager
 import copy
 import logging
 import six
+import time
 
 from libnmstate import metadata
 from libnmstate import netinfo
@@ -31,6 +32,7 @@ from libnmstate import state
 from libnmstate import sysctl
 from libnmstate import validator
 from libnmstate.error import NmstateConflictError
+from libnmstate.error import NmstateError
 from libnmstate.error import NmstateLibnmError
 from libnmstate.error import NmstatePermissionError
 from libnmstate.error import NmstateValueError
@@ -150,6 +152,14 @@ def _apply_ifaces_state(
         raise NmstatePermissionError('Error creating a check point')
     except nm.checkpoint.NMCheckPointCreationError:
         raise NmstateConflictError('Error creating a check point')
+    except NmstateError:
+        # Assume rollback occured, revert IPv6 stack state.
+        # Checkpoint rollback is async, there is a need to wait for it to
+        # finish before proceeding with other actions.
+        # TODO: https://nmstate.atlassian.net/browse/NMSTATE-103
+        time.sleep(5)
+        _disable_ipv6(current_state)
+        raise
 
 
 def _create_editable_desired_state(
--- a/tests/integration/dhcp_test.py
+++ b/tests/integration/dhcp_test.py
@@ -32,6 +32,7 @@ from libnmstate.error import NmstateNotImplementedError
 
 from .testlib import assertlib
 from .testlib import cmd as libcmd
+from .testlib import ifacelib
 from .testlib import statelib
 from .testlib.statelib import INTERFACES
 
@@ -130,6 +131,12 @@ def dhcp_env():
 
 
 @pytest.fixture
+def dhcpcli_up(dhcp_env):
+    with ifacelib.iface_up(DHCP_CLI_NIC) as ifstate:
+        yield ifstate
+
+
+@pytest.fixture
 def setup_remove_bond99():
     yield
     remove_bond = {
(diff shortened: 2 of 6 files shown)


libnmstate/netapplier.py
```diff
@@ -150,6 +152,13 @@ def _apply_ifaces_state(
         raise NmstatePermissionError('Error creating a check point')
     except nm.checkpoint.NMCheckPointCreationError:
         raise NmstateConflictError('Error creating a check point')
+    except NmstateError:
+        # Assume rollback occured, revert IPv6 stack state.
+        # Checkpoint rollback is async, there is a need to wait for it to
+        # finish before proceeding with other actions.
+        time.sleep(5)
```
cathay4t: Maybe we need a `TODO` tag here to indicate we should replace sleep with rollback wait function once we have.

libnmstate/netapplier.py
```diff
@@ -150,6 +152,13 @@ def _apply_ifaces_state(
         raise NmstatePermissionError('Error creating a check point')
     except nm.checkpoint.NMCheckPointCreationError:
         raise NmstateConflictError('Error creating a check point')
+    except NmstateError:
+        # Assume rollback occured, revert IPv6 stack state.
```
maiqueb: I'm confused with the wording here; does this roll back or not ?

If not I agree whit @cathay4t comments, if so, I don't understand them.

tests/integration/dhcp_test.py
```diff
@@ -683,6 +678,17 @@ def test_slave_ipaddr_learned_via_dhcp_added_as_static_to_linux_bridge(
     assertlib.assert_state(bridge_desired_state)
 
 
+@pytest.mark.xfail(raises=NmstateNotImplementedError)
```
maiqueb: I think re-ordering the test is unrelated to the other issue fixed here, and that it deserves a separate commit.

Not that picky though.

libnmstate/netapplier.py
cathay4t: @maiqueb Currently, we just send a request to NM asking a rollback when failure. But we don't check on whether rollback finished. Hence @EdDev put a `sleep(5)` for that.

Once we have rollback wait support, we can remove this sleep.

libnmstate/netapplier.py
EdDev: I replaced them after your findings of the test fail cause.
Basically, I used your #449 idea with some changes.

tests/integration/dhcp_test.py
EdDev: It is related because without it all tests below it fail. Relocating it in its own commit is possible, but then it will have no context/reason.

EdDev:
The tests are passing on CentOS7 but not on Fedora. The failure is always on the nameservers assertion. I cannot figure out what is wrong.

maiqueb:
> The tests are passing on CentOS7 but not on Fedora. The failure is always on the nameservers assertion. I cannot figure out what is wrong.

The error you're facing is that you have extra nameserver(s) in the *running* section , right ? 

I've seen that issue also in the vdsm static_ip tests, for the ```TestNetworkIPDefaultGateway```. I guess it comes from the nameserver libvirt configures on the interface that gets an IP address via dhcp, on my VM. Meaning, meaning that depending on your config, you could be seeing different symptoms.

I got around it by removing that nameserver from the configuration (actually I went neanderthal on it and just ```echo "" > /etc/resolv.conf ``` . 

I think we could / should **not** report back in the 'running' field nameservers / routes that were not configured. *Or*, at least have an option to do that for all entities having the *config*/*running* logical separation. It would be kind of a simpler implementation of the **owner** field - i.e. anything that is **not** listed in the 'config' section is **not** owned by nmstate.

cathay4t:
@maiqueb 
> I think we could / should **not** report back in the 'running' field nameservers / routes that were not configured. _Or_, at least have an option to do that for all entities having the _config_/_running_ logical separation. It would be kind of a simpler implementation of the **owner** field - i.e. anything that is **not** listed in the 'config' section is **not** owned by nmstate.

For DHCP, we cannot determine which nameserver is owned.

maiqueb:
> @maiqueb
> 
> > I think we could / should **not** report back in the 'running' field nameservers / routes that were not configured. _Or_, at least have an option to do that for all entities having the _config_/_running_ logical separation. It would be kind of a simpler implementation of the **owner** field - i.e. anything that is **not** listed in the 'config' section is **not** owned by nmstate.
> 
> For DHCP, we cannot determine which nameserver is owned.

Hm, can't we list the nameservers that are **only** listed in the *running* section ? Those are the one's not nmstate related.

cathay4t:
> > @maiqueb
> > > I think we could / should **not** report back in the 'running' field nameservers / routes that were not configured. _Or_, at least have an option to do that for all entities having the _config_/_running_ logical separation. It would be kind of a simpler implementation of the **owner** field - i.e. anything that is **not** listed in the 'config' section is **not** owned by nmstate.
> > 
> > 
> > For DHCP, we cannot determine which nameserver is owned.
> 
> Hm, can't we list the nameservers that are **only** listed in the _running_ section ? Those are the one's not nmstate related.

For this PR, root cause is the [reapply bug of NM][1], if you want to discuss the DNS schema, we
can move it to github issue or github PR or jira card.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1734470

EdDev:
Depends on #453, #454 .

libnmstate/netapplier.py
maiqueb: Ah ok. Got it, tks.

tests/integration/dhcp_test.py
maiqueb: Fair enough, as I said, won't block this because of that.

maiqueb:
Shouldn't CI pass now that you've got the other 2 PRs (#453, #454 ) rebased here ?

EdDev:
> Shouldn't CI pass now that you've got the other 2 PRs (#453, #454 ) rebased here ?

We fail randomly on OVS tests. Gris already has a change that takes them out using `xfail`.

maiqueb:
> > Shouldn't CI pass now that you've got the other 2 PRs (#453, #454 ) rebased here ?
> 
> We fail randomly on OVS tests. @cathay4t already has a change that takes them out using `xfail`.

Right, my bad; didn't look at the failed test list, sorry for that.

Issue #461, opened 2019-08-02
VLAN cannot be activated if underlying interface is not owned by NM


maiqueb:
## Summary
Trying to create a VLAN on top of a dummy interface not owned by network manager fails.

## Version
nmstate==0.0.8

## More info
The following script can be used to reproduce the issue:
```bash
▶ sudo ip link add dummy_1 type dummy    
▶ cat reproducer.yaml 
---
interfaces:
  - name: eth0
    type: ethernet
    state: up
  - name: dummy_1
    type: dummy
    state: up
  - name: dummy_1.101
    type: vlan
    state: up
    vlan:
      base-iface: dummy_1
      id: 101

▶ sudo nmstatectl set reproducer.yaml 
2019-08-02 12:47:25,247 root         DEBUG    Checkpoint /org/freedesktop/NetworkManager/Checkpoint/154 created for all devices: 60
2019-08-02 12:47:25,247 root         DEBUG    Adding new interfaces: ['dummy_1.101']
2019-08-02 12:47:25,247 root         DEBUG    Connection settings for ConnectionSetting.create:
id: dummy_1.101
iface: dummy_1.101
uuid: 65cebeaf-e908-42a6-b73e-95e2e8b854bc
type: vlan
autoconnect: True
autoconnect_slaves: <enum NM_SETTING_CONNECTION_AUTOCONNECT_SLAVES_YES of type NM.SettingConnectionAutoconnectSlaves>
2019-08-02 12:47:25,248 root         DEBUG    Editing interfaces: ['dummy_1', 'eth0']
2019-08-02 12:47:25,249 root         DEBUG    Connection settings for ConnectionSetting.create:
id: dummy_1
iface: dummy_1
uuid: 46417305-6868-4004-92e8-da724e669094
type: dummy
autoconnect: True
autoconnect_slaves: <enum NM_SETTING_CONNECTION_AUTOCONNECT_SLAVES_YES of type NM.SettingConnectionAutoconnectSlaves>
2019-08-02 12:47:25,249 root         WARNING  IPv6 link local address fe80::b7a2:5bed:7e0c:e27f/64 is ignored when applying desired state
2019-08-02 12:47:25,249 root         DEBUG    Connection settings for ConnectionSetting.import_by_profile:
id: System eth0
iface: eth0
uuid: 5fb06bd0-0bb0-7ffb-45f1-d6edd65f3e03
type: 802-3-ethernet
autoconnect: True
[...]

PR #464, opened 2019-08-04
nm.applier: Prioritize base iface activation before the vlan

EdDev:
When both the base interface and the VLAN interface are created in one
transaction, the base interface must be activated before the vlan one,
otherwise the VLAN activation fails.

The new VLAN interfaces are now activated after all other new
interfaces.

Solves #461

### DIFF
--- a/libnmstate/nm/applier.py
+++ b/libnmstate/nm/applier.py
@@ -21,6 +21,8 @@ import itertools
 import six
 
 from libnmstate.error import NmstateValueError
+from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceType
 from libnmstate.schema import LinuxBridge as LB
 
 from . import bond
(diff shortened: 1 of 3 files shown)


EdDev:
@maiqueb , can you please check if this solves the problem you see with the VLAN interfaces?

tests/integration/vlan_test.py
```diff
@@ -110,6 +110,30 @@ def test_set_vlan_iface_down(eth1_up):
...
+            },
+            {'name': iface_base, 'type': 'dummy', 'state': 'up'},
+        ]
+    }
+    try:
+        libnmstate.apply(desired_state)
+    finally:
+        for ifstate in desired_state[INTERFACES]:
+            ifstate['state'] = 'absent'
+        libnmstate.apply(desired_state)
```
cathay4t: Maybe we should use `verify_change=False` here.

libnmstate/nm/applier.py
```diff
@@ -174,6 +179,9 @@ def set_ifaces_admin_state(ifaces_desired_state, con_profiles=()):
     for ifname in new_ifaces_to_activate:
         device.activate(dev=None, connection_id=ifname)
 
+    for ifname in new_vlan_ifaces_to_activate:
```
maiqueb: These lines have to be moved below lines https://github.com/nmstate/nmstate/pull/464/files#diff-de1f01201a749955840472d0101bf589R194 - https://github.com/nmstate/nmstate/pull/464/files#diff-de1f01201a749955840472d0101bf589R196 , since the dummy interfaces **not** owned by NM are present within the `ifaces_to_edit` set. 

That fixes issue #461 , and unlocks the VLAN related test cases in vdsm.

PR #466, opened 2019-08-06
Relicense last files to LGPL2.1+

maiqueb:
IIUC, the subnet_of function was added *only* in python 3.7, which means we'll need to stick to using the `ipaddress_extender` for a bit longer (RHEL 8 is fedora 28 based, which ships py3.6).

If we don't care about that - e.g. we want to only run on py3.7, we should update the spec file to require py3.7.

### DIFF
--- a/automation/open_container_shell.sh
+++ b/automation/open_container_shell.sh
@@ -2,17 +2,19 @@
 #
 # Copyright 2019 Red Hat, Inc.
 #
+# This file is part of nmstate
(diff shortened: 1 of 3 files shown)


cathay4t:
Current code base does not require `ipaddress` anymore. We can drop that in another PR.

cathay4t:
@maiqueb Please rebase. Thank you.

PR #480, opened 2019-08-15
metadata: Detach slaves without deleting them

EdDev:


EdDev:
> I miss information as to why this is needed, and problem it solves in the commit msg.

IMO the subject says it all, but if a longer story makes it clearer, hope the updated one does the job.

EdDev:
@cathay4t , I have added the test.

PR #482, opened 2019-08-18
tests integ: Introduce the bridgelib module

EdDev:
Extract the Linux Bridge helpers from the linux_bridge_test module in
order to reuse them in other tests.

### DIFF
--- a/tests/integration/testlib/bridgelib.py
+++ b/tests/integration/testlib/bridgelib.py
@@ -1,0 +1,84 @@
+#
+# Copyright (c) 2019 Red Hat, Inc.
+#
+# This file is part of nmstate
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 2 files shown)


EdDev:
@Nandan307 , this may help you with your other tests.

tests/integration/testlib/bridgelib.py
```diff
@@ -0,0 +1,83 @@
...
+    }
+    if extra_iface_state:
+        desired_state[Interface.KEY][0].update(extra_iface_state)
+
+    libnmstate.apply(desired_state)
+
+    try:
+        yield desired_state
+    finally:
+        libnmstate.apply(
```
cathay4t: Please use `verify_change=False` when removing the bridge.

tests/integration/testlib/bridgelib.py
```diff
@@ -0,0 +1,83 @@
...
+    if port_state is None:
+        port_state = {}
+
+    port_state[LinuxBridge.PORT_NAME] = port_name
+    bridge_subtree_state[LinuxBridge.PORT_SUBTREE].append(port_state)
+
+    return bridge_subtree_state
+
+
+def create_bridge_subtree_state(options_state=None):
```
cathay4t: No one is using this function, is it required?

EdDev: Yes, it is used in the next change: https://github.com/nmstate/nmstate/pull/485
I think it belongs in this commit just because it completes the needs for defining a bridge.

tests/integration/testlib/bridgelib.py
EdDev: Done

PR #485, opened 2019-08-19
Test dhcp over bridge

EdDev:
Depends on #482

### DIFF
--- a/tests/integration/dhcp_test.py
+++ b/tests/integration/dhcp_test.py
@@ -16,6 +16,7 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 #
+from contextlib import contextmanager
 import os
 import time
 
@@ -24,15 +25,16 @@ import pytest
 import libnmstate
 from libnmstate.schema import Constants
 from libnmstate.schema import DNS
+from libnmstate.schema import Interface
 from libnmstate.schema import InterfaceIPv4
 from libnmstate.schema import InterfaceIPv6
+from libnmstate.schema import InterfaceState
 from libnmstate.schema import Route as RT
 
 from libnmstate.error import NmstateNotImplementedError
 
 from .testlib import assertlib
 from .testlib import cmd as libcmd
-from .testlib import ifacelib
 from .testlib import statelib
 from .testlib.statelib import INTERFACES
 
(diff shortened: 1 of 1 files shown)


tests/integration/dhcp_test.py
```diff
@@ -114,8 +118,37 @@ def dhcp_env():
 
 @pytest.fixture
 def dhcpcli_up(dhcp_env):
```
cathay4t: Please create new fixture like `dhcpcli_up_auto` and place the sleep(5) in it.

tests/integration/dhcp_test.py
```diff
@@ -492,6 +525,63 @@ def test_ipv6_dhcp_switch_on_to_off(dhcpcli_up):
...
+
+    time.sleep(5)  # wait to get IP addresses
+    origin_port_state = statelib.show_only((port_name,))
+
+    bridge_state = create_bridge_subtree_state()
+    port_state = {
+        'stp-hairpin-mode': False,
+        'stp-path-cost': 100,
+        'stp-priority': 32,
+    }
```
cathay4t: I think we don't required to this anymore. The verification only check defined data.

tests/integration/dhcp_test.py
EdDev: The current fixture is inconsistent between ipv4 and ipv6. For ipv4 it is set to auto but for ipv6 is somehow does not work. I think we need to be explicit here to avoid these inconsistencies now and in the future.

I will split this change so it will be clear.

tests/integration/dhcp_test.py
EdDev: I tried it but it did not work. I was not focusing on this so I left it like this.
I will recheck so I'll have a better answer. (maybe the assert here is not including this)

tests/integration/dhcp_test.py
EdDev: Done. One commit for changing the way we start dhcpcli interface and another for the test with the addition of the delay.

tests/integration/dhcp_test.py
EdDev: Confirmed, we have a bug in the test assertion. I will take care of that later.

EdDev: Fixed by changing the assert function.

EdDev:
@cathay4t , could you please revisit this?

PR #489, opened 2019-08-22
nm.applier: Prioritize master interfaces activaction

maiqueb:
The master interfaces must be activated first, otherwise
NetworkManager fails with errors such as:

"""
NetworkManager[858]: <debug>
[1565960861.2223] manager: Activation of <slave> requires master
device <master device>
"""

Depends on: #491 
Fixes: https://bugzilla.redhat.com/1749314

### DIFF
--- a/libnmstate/nm/applier.py
+++ b/libnmstate/nm/applier.py
@@ -41,6 +41,7 @@ from . import wired
 
 MASTER_METADATA = '_master'
 MASTER_TYPE_METADATA = '_master_type'
+MASTER_IFACE_TYPES = ovs.BRIDGE_TYPE, bond.BOND_TYPE, LB.TYPE
 
 BRPORT_OPTIONS_METADATA = '_brport_options'
 
@@ -111,6 +112,7 @@ def set_ifaces_admin_state(ifaces_desired_state, con_profiles=()):
     leaving it to choose the correct profile.
 
     In order to activate correctly the interfaces, the order is significant:
+    - Master-less master interfaces.
     - New interfaces (virtual interfaces, but not OVS ones).
     - Master interfaces.
     - OVS ports.
@@ -123,6 +125,7 @@ def set_ifaces_admin_state(ifaces_desired_state, con_profiles=()):
     new_vlan_ifaces_to_activate = set()
     new_ovs_interface_to_activate = set()
     new_ovs_port_to_activate = set()
+    new_master_not_enslaved_ifaces = set()
     master_ifaces_to_edit = set()
     ifaces_to_edit = set()
     remove_devs_actions = {}
@@ -132,7 +135,13 @@ def set_ifaces_admin_state(ifaces_desired_state, con_profiles=()):
         nmdev = device.get_device_by_name(ifname)
         if not nmdev:
             if ifname in new_ifaces and iface_desired_state['state'] == 'up':
-                if iface_desired_state['type'] == ovs.INTERNAL_INTERFACE_TYPE:
+                if _is_master_iface(
+                    iface_desired_state
+                ) and not _is_slave_iface(iface_desired_state):
+                    new_master_not_enslaved_ifaces.add(ifname)
+                elif (
+                    iface_desired_state['type'] == ovs.INTERNAL_INTERFACE_TYPE
+                ):
                     new_ovs_interface_to_activate.add(ifname)
                 elif iface_desired_state['type'] == ovs.PORT_TYPE:
                     new_ovs_port_to_activate.add(ifname)
@@ -142,8 +151,7 @@ def set_ifaces_admin_state(ifaces_desired_state, con_profiles=()):
                     new_ifaces_to_activate.add(ifname)
         else:
             if iface_desired_state['state'] == 'up':
-                master_iface_types = ovs.BRIDGE_TYPE, bond.BOND_TYPE, LB.TYPE
-                if iface_desired_state['type'] in master_iface_types:
+                if _is_master_iface(iface_desired_state):
                     master_ifaces_to_edit.add(
                         (nmdev, con_profiles_by_devname[ifname].profile)
                     )
@@ -175,6 +183,9 @@ def set_ifaces_admin_state(ifaces_desired_state, con_profiles=()):
     # Do not remove devices that are marked for editing.
     for dev, _ in itertools.chain(master_ifaces_to_edit, ifaces_to_edit):
         remove_devs_actions.pop(dev, None)
+
+    for ifname in new_master_not_enslaved_ifaces:
+        device.activate(dev=None, connection_id=ifname)
 
     for ifname in new_ifaces_to_activate:
         device.activate(dev=None, connection_id=ifname)
@@ -211,6 +222,14 @@ def _get_new_ifaces(con_profiles):
         if not nmdev:
             ifaces_without_device.add(ifname)
     return ifaces_without_device
+
+
+def _is_master_iface(iface_state):
+    return iface_state[Interface.TYPE] in MASTER_IFACE_TYPES
+
+
+def _is_slave_iface(iface_state):
+    return iface_state.get(MASTER_METADATA)
 
 
 def _get_affected_devices(iface_state):
--- a/tests/integration/linux_bridge_test.py
+++ b/tests/integration/linux_bridge_test.py
@@ -29,6 +29,7 @@ from libnmstate.schema import InterfaceState
 from libnmstate.schema import LinuxBridge
 
 from .testlib import assertlib
+from .testlib.bondlib import bond_interface
 from .testlib.bridgelib import add_port_to_bridge
 from .testlib.bridgelib import linux_bridge
 from .testlib.iproutelib import ip_monitor_assert_stable_link_up
@@ -84,6 +85,14 @@ def port0_vlan101(port0_up):
     with vlan_interface(port_name, vlan_id, vlan_base_iface):
         state = show_only((port_name,))
         yield state
+
+
+@pytest.fixture
+def bond0(port0_up):
+    bond_name = 'testbond0'
+    port_name = port0_up[Interface.KEY][0][Interface.NAME]
+    with bond_interface(bond_name, [port_name], create=False) as bond0:
+        yield bond0
 
 
 def test_create_and_remove_linux_bridge_with_min_desired_state():
@@ -267,6 +276,23 @@ def test_replace_port_on_linux_bridge(port0_vlan101, port1_up):
         assert (
             InterfaceState.UP == port_state[Interface.KEY][0][Interface.STATE]
         )
+
+
+def test_linux_bridge_over_bond_over_slave_in_one_transaction(bond0):
+    bridge_name = TEST_BRIDGE0
+    bond_name = bond0[Interface.KEY][0][Interface.NAME]
+    bridge_config_state = _create_bridge_subtree_config((bond_name,))
+    with linux_bridge(
+        bridge_name, bridge_config_state, create=False
+    ) as bridge0:
+        desired_state = bond0
+        bridge_state = bridge0[Interface.KEY][0]
+        desired_state[Interface.KEY].append(bridge_state)
+        libnmstate.apply(desired_state)
+
+        assertlib.assert_state_match(desired_state)
+
+    assertlib.assert_absent(bridge_name)
 
 
 def _add_port_to_bridge(bridge_state, ifname):
--- a/tests/integration/testlib/bondlib.py
+++ b/tests/integration/testlib/bondlib.py
@@ -27,7 +27,7 @@ from libnmstate.schema import InterfaceType
 
 
 @contextmanager
-def bond_interface(name, slaves, extra_iface_state=None):
+def bond_interface(name, slaves, extra_iface_state=None, create=True):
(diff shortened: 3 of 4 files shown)


maiqueb:
@EdDev should I replicate the vdsm functional test that triggered this as an nmstate unit test? 

It would:
  - create a nic w/ IP configuration
  - create a bond w/out IP configuration as master of nic create in the previous step 
  - create a bridge having the bond created in the previous step as its only port

libnmstate/nm/applier.py
```diff
@@ -132,6 +133,12 @@ def set_ifaces_admin_state(ifaces_desired_state, con_profiles=()):
         nmdev = device.get_device_by_name(ifname)
         if not nmdev:
             if ifname in new_ifaces and iface_desired_state['state'] == 'up':
+                master_iface_types = ovs.BRIDGE_TYPE, bond.BOND_TYPE, LB.TYPE
```
EdDev: Now that this is used at least twice here they merit a helper/constant.

But a more important point is to name this check correctly through a helper function.
Like `if _is_master_iface() and not _is_slave_iface():`

libnmstate/nm/applier.py
```diff
@@ -123,6 +123,7 @@ def set_ifaces_admin_state(ifaces_desired_state, con_profiles=()):
     new_vlan_ifaces_to_activate = set()
     new_ovs_interface_to_activate = set()
     new_ovs_port_to_activate = set()
+    master_ifaces_to_add = set()
```
EdDev: The missing info here is: `new`, `not enslaved/slave`.
`new_master_non_slave_ifaces` OR
`new_master_not_enslaved_ifaces`?

tests/integration/linux_bridge_test.py
```diff
@@ -86,6 +87,13 @@ def port0_vlan101(port0_up):
         yield state
 
 
+@pytest.fixture
+def bond0(port0_up):
+    port_name = port0_up[Interface.KEY][0][Interface.NAME]
+    with bond_interface('bond0', [port_name]) as bond:
```
EdDev: `testbond0`?

tests/integration/linux_bridge_test.py
```diff
@@ -269,6 +277,15 @@ def test_replace_port_on_linux_bridge(port0_vlan101, port1_up):
         )
 
 
+def test_linux_bridge_over_bond_over_slave(bond0):
```
EdDev: Could you please confirm that this test will fail without this fix?

libnmstate/nm/applier.py
maiqueb: Done.

libnmstate/nm/applier.py
maiqueb: Done.

tests/integration/linux_bridge_test.py

tests/integration/linux_bridge_test.py
maiqueb: In this new updated version, it does - it is not deterministic; it fails around 50% of the times.

I need to think how to update the commit msg of this patch to reflect that.

tests/integration/linux_bridge_test.py
```diff
@@ -269,6 +304,15 @@ def test_replace_port_on_linux_bridge(port0_vlan101, port1_up):
         )
 
 
+def test_linux_bridge_over_bond_over_slave(bond0, bridge0):
+    bridge_state = bridge0[Interface.KEY][0]
+    desired_state = deepcopy(bond0)
+    desired_state[Interface.KEY].append(bridge_state)
```
maiqueb: I can also perform this 'state merge' in an added fixture, which would:
  - combine/merge states
  - apply desired state
  - yield desired state

I could actually move the 'finally' clauses of the bridge0 and bond0 fixtures to the finally clause of that new fixture.

What do you think @EdDev  ?

maiqueb:
ci test please

libnmstate/nm/applier.py
```diff
@@ -132,6 +134,10 @@ def set_ifaces_admin_state(ifaces_desired_state, con_profiles=()):
         nmdev = device.get_device_by_name(ifname)
         if not nmdev:
             if ifname in new_ifaces and iface_desired_state['state'] == 'up':
+                if _is_master_iface(
+                    iface_desired_state
+                ) and not iface_desired_state.get(MASTER_METADATA):
+                    new_master_not_enslaved_ifaces.add(ifname)
                 if iface_desired_state['type'] == ovs.INTERNAL_INTERFACE_TYPE:
```
EdDev: It will be safer to change this to an `elif`.

libnmstate/nm/applier.py
```diff
@@ -132,6 +134,10 @@ def set_ifaces_admin_state(ifaces_desired_state, con_profiles=()):
         nmdev = device.get_device_by_name(ifname)
         if not nmdev:
             if ifname in new_ifaces and iface_desired_state['state'] == 'up':
+                if _is_master_iface(
```
EdDev: Consider for readability (not a must): `if is_master_iface() and not is_slave_iface():`

tests/integration/linux_bridge_test.py
```diff
@@ -269,6 +303,15 @@ def test_replace_port_on_linux_bridge(port0_vlan101, port1_up):
         )
 
 
+def test_linux_bridge_over_bond_over_slave(bond0, bridge0):
+    bridge_state = bridge0[Interface.KEY][0]
+    desired_state = deepcopy(bond0)
```
EdDev: I think this is the responsibility of the fixture.

tests/integration/linux_bridge_test.py
EdDev: The fixtures represent the test setup and teardown. They are not part of the test.
Therefore, I think it is fine as it is.

tests/integration/linux_bridge_test.py
```diff
@@ -269,6 +303,15 @@ def test_replace_port_on_linux_bridge(port0_vlan101, port1_up):
...
+    desired_state[Interface.KEY].append(bridge_state)
+
+    libnmstate.apply(desired_state, verify_change=False)
```
EdDev: Why `verify_change=False` is added here?

tests/integration/testlib/bondlib.py
```diff
@@ -28,6 +28,17 @@
...
+
+    libnmstate.apply(desired_state)
+
+    try:
+        yield desired_state
+    finally:
+        libnmstate.apply(generate_remove_bond_state(name), verify_change=False)
+
+
+def generate_create_bond_state(name, slaves, extra_iface_state=None):
```
EdDev: I find the usage of two verbs here problematic (generate_create).
`create_bond_state` and `remove_bond_state` is clear enough.

tests/integration/linux_bridge_test.py
```diff
@@ -86,6 +90,36 @@ def port0_vlan101(port0_up):
         yield state
 
 
+@pytest.fixture
+def bond0(port0_up):
+    bond_name = 'testbond0'
+    port_name = port0_up[Interface.KEY][0][Interface.NAME]
+    bond0 = generate_create_bond_state(bond_name, [port_name])
+    try:
```
EdDev: The `yeld` part is already covered by try/finally in a fixture, you do not need to wrap it.

tests/integration/linux_bridge_test.py
maiqueb: > The fixtures represent the test setup and teardown. They are not part of the test.
Therefore, I think it is fine as it is.

touchè.

libnmstate/nm/applier.py
maiqueb: 'black' - the code formatter - made me do it! 

Since we do also assert the code format in the tests, this has to stay like this. Right?..

libnmstate/nm/applier.py
maiqueb: Done.

tests/integration/linux_bridge_test.py
maiqueb: The linux bridge is no longer a fixture, and instead it's state is generated within the test itself; as such, I think this no longer applies.

tests/integration/linux_bridge_test.py
maiqueb: Slipped through :) 

Removed it.

tests/integration/testlib/bondlib.py
maiqueb: No longer applies.

tests/integration/linux_bridge_test.py
maiqueb: right ! 

After the latest refactor, this no longer applies.

tests/integration/linux_bridge_test.py
```diff
@@ -86,6 +87,14 @@ def port0_vlan101(port0_up):
         yield state
 
 
+@pytest.fixture
+def bond0(port0_up):
```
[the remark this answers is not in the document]
maiqueb: If you prefer, what I can do is use the `eth1_up` alias. I think adding aliasing this to another type does not add much.

tests/integration/linux_bridge_test.py
```diff
@@ -269,6 +278,23 @@ def test_replace_port_on_linux_bridge(port0_vlan101, port1_up):
...
 
 
+def test_linux_bridge_over_bond_over_slave(bond0):
+    bridge_name = TEST_BRIDGE0
+    bond_name = bond0[Interface.KEY][0][Interface.NAME]
+    bridge_config_state = _create_bridge_subtree_config((bond_name,))
+    with linux_bridge(
+        bridge_name, bridge_config_state, create=False
+    ) as bridge0:
+        desired_state = deepcopy(bond0)
```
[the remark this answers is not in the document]
maiqueb: I just want to be able to represent that the desired state *is* a composition of the bond0 with the bridge0 entities - imho, easiest way to do that is:

```python
desired_state = bond0
desired_state[Interface.KEY].append(bridge0[Interface.KEY][0])
```

That's why the deepcopy is done outside the fixture - so I can have a variable called `desired_state`.

libnmstate/nm/applier.py
```diff
@@ -132,7 +135,13 @@ def set_ifaces_admin_state(ifaces_desired_state, con_profiles=()):
         nmdev = device.get_device_by_name(ifname)
         if not nmdev:
             if ifname in new_ifaces and iface_desired_state['state'] == 'up':
-                if iface_desired_state['type'] == ovs.INTERNAL_INTERFACE_TYPE:
+                if _is_master_iface(
+                    iface_desired_state
+                ) and not iface_desired_state.get(MASTER_METADATA):
+                    new_master_not_enslaved_ifaces.add(ifname)
+                elif (
+                    iface_desired_state['type'] == ovs.INTERNAL_INTERFACE_TYPE
```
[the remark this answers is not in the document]
maiqueb: This is unrelated to this PR. 

I can make another updating the keys used in this file to the Interface constants though.

tests/integration/linux_bridge_test.py
```diff
@@ -269,6 +278,23 @@ def test_replace_port_on_linux_bridge(port0_vlan101, port1_up):
         )
 
 
+def test_linux_bridge_over_bond_over_slave(bond0):
```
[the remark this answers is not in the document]
maiqueb: longest. test. name. ever. 

Done.

tests/integration/linux_bridge_test.py
```diff
@@ -269,6 +278,23 @@ def test_replace_port_on_linux_bridge(port0_vlan101, port1_up):
...
+        bridge_name, bridge_config_state, create=False
+    ) as bridge0:
+        desired_state = deepcopy(bond0)
+        bridge_state = bridge0[Interface.KEY][0]
+        desired_state[Interface.KEY].append(bridge_state)
+        libnmstate.apply(desired_state)
+
+        assertlib.assert_state_match(desired_state)
+
+    assertlib.assert_absent(bridge_name)
```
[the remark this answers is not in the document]
maiqueb: I guess it is as needed there as in https://github.com/nmstate/nmstate/pull/489/files#diff-41e570df111ba3cf748cf4f22fb0e3c1R240 . 

Just say the word, I'll remove it. But, I think it can't hurt to assert that the bridge is gone - but I understand the reasoning that it might lead the reader into thinking that's somehow important to the test.

libnmstate/nm/applier.py
```diff
@@ -213,6 +224,10 @@ def _get_new_ifaces(con_profiles):
     return ifaces_without_device
 
 
+def _is_master_iface(iface_state):
+    return iface_state['type'] in MASTER_IFACE_TYPES
```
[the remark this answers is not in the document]
maiqueb: Done.

libnmstate/nm/applier.py
maiqueb: Sorry, misinterpreted your comment. Followed your suggestion, and added that method.

Done.

tests/integration/linux_bridge_test.py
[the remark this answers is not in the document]
maiqueb: Replace ```deepcopy``` with a simple variable assignment.

PR #491, opened 2019-08-22
tests integ: Introduce the bondlib module

maiqueb:
Extract the bond creation function to a dedicated module, enabling
it to be re-used in other tests.

### DIFF
--- a/tests/integration/testlib/bondlib.py
+++ b/tests/integration/testlib/bondlib.py
@@ -1,0 +1,62 @@
+#
+# Copyright (c) 2019 Red Hat, Inc.
+#
+# This file is part of nmstate
+#
+# This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


tests/integration/testlib/bondlib.py
```diff
@@ -0,0 +1,61 @@
...
+        ]
+    }
+    if extra_iface_state:
+        desired_state[Interface.KEY][0].update(extra_iface_state)
+
+    libnmstate.apply(desired_state)
+    try:
+        yield desired_state
+    finally:
+        libnmstate.apply(
```
EdDev: Please add `verify_change=False` to the `apply`. On cleanup we should not fail at verification.

maiqueb: Done.

PR #502, opened 2019-09-11
tests, integ: Test MTU change with two VLAN/s

EdDev:
Define two VLAN/s over a shared base interface and edit their MTU.
The transaction fails and therfore marked as xfail.

Ref: https://bugzilla.redhat.com/1751079

### DIFF
--- a/tests/integration/ethernet_mtu_test.py
+++ b/tests/integration/ethernet_mtu_test.py
@@ -31,6 +31,7 @@ from libnmstate.error import NmstateVerificationError
 from .testlib import assertlib
 from .testlib import statelib
 from .testlib.statelib import INTERFACES
+from .testlib.vlan import vlan_interface
 
 
 @pytest.fixture(scope='function', autouse=True)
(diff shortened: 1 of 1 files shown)


tests/integration/ethernet_mtu_test.py
```diff
@@ -133,3 +134,23 @@ def test_decrease_to_lower_than_min_ipv6_iface_mtu(eth1_with_ipv6):
...
+    v102 = vlan_interface('eth1.102', 102, base_ifname)
+    with v101 as v101_state, v102 as v102_state:
+        desired_state = {
+            schema.Interface.KEY: [
+                v101_state[schema.Interface.KEY][0],
+                v102_state[schema.Interface.KEY][0],
+            ]
+        }
+        for iface_state in desired_state[schema.Interface.KEY]:
+            iface_state[schema.Interface.MTU] = 2000
```
cathay4t: As base interface should hold MTU bigger or equal to its slaves. Please also set base interface MTU to 2000.

EdDev: Done

EdDev:
Thanks @cathay4t , done.

tests/integration/ethernet_mtu_test.py
```diff
@@ -133,3 +134,24 @@ def test_decrease_to_lower_than_min_ipv6_iface_mtu(eth1_with_ipv6):
...
+            schema.Interface.KEY: [
+                base_ifname[schema.Interface.KEY][0],
+                v101_state[schema.Interface.KEY][0],
+                v102_state[schema.Interface.KEY][0],
+            ]
+        }
+        for iface_state in desired_state[schema.Interface.KEY]:
+            iface_state[schema.Interface.MTU] = 2000
+
+        libnmstate.apply(desired_state)
```
almusil: I think this won't trigger the bug that is currently in NM. To trigger it we have to change vlan101 and base interface to MTU e.g. 2000 in first transaction. Then vlan102 and base interface to MTU e.g. 2200 in second. The result should be that the vlan101 does not persist the 2000 and goes back to 1500. 

So IMO this is wrong to test this.

maiqueb: > I think this won't trigger the bug that is currently in NM. To trigger it we have to change vlan101 and base interface to MTU e.g. 2000 in first transaction. Then vlan102 and base interface to MTU e.g. 2200 in second. The result should be that the vlan101 does not persist the 2000 and goes back to 1500.

While you're right in the sense that this test does not represent the bugzilla referenced in the xfail reason, this test alone is helpful, since apparently it also uncovers a scenario we're not handling properly.

Another test should be added, which would represent the scenario you describe.

Not exactly sure as to what to do in the xfail reason. Perhaps a new bug should be opened ?..

> 
> So IMO this is wrong to test this.

From my understanding, the scenario introduced in this PR is also valid, and as such, worthy of a test.

EdDev: I do not see the big difference here. This test eventually does this: set the base mtu, then one of the vlans mtu and lastly the second vlan mtu, which in turn resets the other vlan mtu back to the default.
Starting from a non default mtu does not change a lot here.

PR #503, opened 2019-09-12
nm.applier: Activate slave interfaces last

maiqueb:
The slave interfaces - that are not masters themselves - must be
activated last, otherwise NetworkManager fails with errors such
as:

```
NetworkManager[858]: <debug>
[1565960861.2223] manager: Activation of <slave> requires master
device <master device>
```

This change addresses the scenario introduced by provisioning the
following yaml:

```yaml
---
interfaces:
  - name: bond0
    type: bond
    state: up
    link-aggregation:
      slaves:
        - eth0
        - eth1
      options:
        miimon: "100"
        xmit_hash_policy: "2"
      mode: 802.3ad
    ipv4:
      enabled: false
    ipv6:
      enabled: false
  - name: bond0.166
    type: vlan
    state: up
    vlan:
      id: 166
      base-iface: bond0
    ipv4:
      enabled: false
    ipv6:
      enabled: false
  - name: bridge0
    type: linux-bridge
    state: up
    bridge:
      port:
        - name: bond0.166
      options:
        stp:
          enabled: false
    ipv4:
      enabled: true
      dhcp: true
      auto-dns: false
      auto-gateway: false
      auto-routes: false
    ipv6:
      enabled: false
```

Provisioning the above yaml without this change fails, since it
tries to activate the eth[0|1] interfaces before the bond.

### DIFF
--- a/libnmstate/nm/applier.py
+++ b/libnmstate/nm/applier.py
@@ -123,11 +123,12 @@ def set_ifaces_admin_state(ifaces_desired_state, con_profiles=()):
     """
     con_profiles_by_devname = _index_profiles_by_devname(con_profiles)
     new_ifaces = _get_new_ifaces(con_profiles)
-    new_ifaces_to_activate = set()
+    new_master_enslaved_ifaces = set()
     new_vlan_ifaces_to_activate = set()
     new_ovs_interface_to_activate = set()
     new_ovs_port_to_activate = set()
     new_master_not_enslaved_ifaces = set()
+    new_slave_not_master_ifaces = set()
     master_ifaces_to_edit = set()
     ifaces_to_edit = set()
     remove_devs_actions = {}
@@ -144,6 +145,10 @@ def set_ifaces_admin_state(ifaces_desired_state, con_profiles=()):
                     iface_desired_state
                 ) and not _is_slave_iface(iface_desired_state):
                     new_master_not_enslaved_ifaces.add(ifname)
+                elif _is_slave_iface(
+                    iface_desired_state
+                ) and not _is_master_iface(iface_desired_state):
+                    new_slave_not_master_ifaces.add(ifname)
                 elif (
                     iface_desired_state[Interface.TYPE]
                     == ovs.INTERNAL_INTERFACE_TYPE
@@ -154,7 +159,7 @@ def set_ifaces_admin_state(ifaces_desired_state, con_profiles=()):
                 elif iface_desired_state[Interface.TYPE] == InterfaceType.VLAN:
                     new_vlan_ifaces_to_activate.add(ifname)
                 else:
-                    new_ifaces_to_activate.add(ifname)
+                    new_master_enslaved_ifaces.add(ifname)
         else:
             if iface_desired_state[Interface.STATE] == InterfaceState.UP:
                 if _is_master_iface(iface_desired_state):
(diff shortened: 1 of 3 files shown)


EdDev:
> Even with `libnmstate/nm/applier.py` from master branch, the test case `test_linux_bridge_over_vlan_over_bond_over_slave_in_one_transaction` still pass.
> So the test actually does not recreate the original issue.

@maiqueb claimed that the problem is not recreated all the time, seems to be a race between the autoconnect/autoconnect-slaves and the explicit activation.
Logically, it makes sense. If you disable autoconnect* this should get recreated easily.

maiqueb:
I am currently unsure this fixes the underlying issue, illustrated on the yaml on the PR description.

Need more time to investigate.

cathay4t:
> > Even with `libnmstate/nm/applier.py` from master branch, the test case `test_linux_bridge_over_vlan_over_bond_over_slave_in_one_transaction` still pass.
> > So the test actually does not recreate the original issue.
> 
> @maiqueb claimed that the problem is not recreated all the time, seems to be a race between the autoconnect/autoconnect-slaves and the explicit activation.
> Logically, it makes sense. If you disable autoconnect* this should get recreated easily.

I tried to `autoconnect= False` and `AutoconnectSlaves.NO`.  The test still pass with master code.

And I don't understand where the race is.

maiqueb:
> > > Even with `libnmstate/nm/applier.py` from master branch, the test case `test_linux_bridge_over_vlan_over_bond_over_slave_in_one_transaction` still pass.
> > > So the test actually does not recreate the original issue.
> > 
> > 
> > @maiqueb claimed that the problem is not recreated all the time, seems to be a race between the autoconnect/autoconnect-slaves and the explicit activation.
> > Logically, it makes sense. If you disable autoconnect* this should get recreated easily.
> 
> I tried to `autoconnect= False` and `AutoconnectSlaves.NO`. The test still pass with master code.

The PR isn't fixing the real issue, I'm afraid.

> 
> And I don't understand where the race is.

@cathay4t , using the nmstate container, try to provision the following yaml:

```yaml
---
interfaces:
  - name: bond0
    type: bond
    state: up
    link-aggregation:
      slaves:
        - eth0
        - eth1
      options:
        miimon: "100"
        xmit_hash_policy: "2"
      mode: 802.3ad
    ipv4:
      enabled: false
    ipv6:
      enabled: false
  - name: bond0.166
    type: vlan
    state: up
    vlan:
      id: 166
      base-iface: bond0
    ipv4:
      enabled: false
    ipv6:
      enabled: false
  - name: bridge0
    type: linux-bridge
    state: up
    bridge:
      port:
        - name: bond0.166
      options:
        stp:
          enabled: false
    ipv4:
      enabled: true
      dhcp: true
      auto-dns: false
      auto-gateway: false
      auto-routes: false
    ipv6:
      enabled: false
```

cathay4t:
> > > > Even with `libnmstate/nm/applier.py` from master branch, the test case `test_linux_bridge_over_vlan_over_bond_over_slave_in_one_transaction` still pass.
> > > > So the test actually does not recreate the original issue.
> > > 
> > > 
> > > @maiqueb claimed that the problem is not recreated all the time, seems to be a race between the autoconnect/autoconnect-slaves and the explicit activation.
> > > Logically, it makes sense. If you disable autoconnect* this should get recreated easily.
> > 
> > 
> > I tried to `autoconnect= False` and `AutoconnectSlaves.NO`. The test still pass with master code.
> 
> The PR isn't fixing the real issue, I'm afraid.
> 
> > And I don't understand where the race is.
> 
> @cathay4t , using the nmstate container, try to provision the following yaml:

@maiqueb 
I got failure as DHCP failure but related to activating order. If I change it to ipv4 ipv6 both disabled,
it works well. I tried 30 times, no activation order problem.

Will it helps if you firstly disable IPv4 and IPv6, then changed to DHCP?

The logic is good in this PR, but I think it's better to confirm the root cause of the problem and then confirm a fix for it.

maiqueb:
Closed because #512 adds the correct tests, and this code is apparently not needed.

PR #508, opened 2019-09-18
Add vxlan schema

qinqon:

### DIFF
--- a/libnmstate/schema.py
+++ b/libnmstate/schema.py
@@ -97,6 +97,7 @@ class InterfaceType(object):
     OVS_PORT = 'ovs-port'
     UNKNOWN = 'unknown'
     VLAN = 'vlan'
+    VXLAN = 'vxlan'
 
     VIRT_TYPES = (
         BOND,
@@ -106,6 +107,7 @@ class InterfaceType(object):
         OVS_PORT,
         OVS_INTERFACE,
         VLAN,
+        VXLAN,
     )
 
 
@@ -188,6 +190,16 @@ class VLAN(object):
 
     ID = 'id'
     BASE_IFACE = 'base-iface'
+
+
+class VXLAN(object):
+    TYPE = InterfaceType.VXLAN
+    CONFIG_SUBTREE = 'vxlan'
+
+    ID = 'id'
+    BASE_IFACE = 'base-iface'
+    REMOTE = 'remote'
+    DESTINATION_PORT = 'destination-port'
 
 
 class OVSBridge(object):
--- a/libnmstate/schemas/operational-state.yaml
+++ b/libnmstate/schemas/operational-state.yaml
@@ -24,6 +24,7 @@ properties:
             - "$ref": "#/definitions/interface-ovs-interface/rw"
             - "$ref": "#/definitions/interface-dummy/rw"
             - "$ref": "#/definitions/interface-vlan/rw"
+            - "$ref": "#/definitions/interface-vxlan/rw"
   routes:
     type: object
     properties:
@@ -177,6 +178,27 @@ definitions:
           required:
             - id
             - base-iface
+  interface-vxlan:
+    rw:
+      properties:
+        type:
+          type: string
+          enum:
+            - vxlan
+        vxlan:
+          type: object
+          properties:
+            id:
+              type: integer
+              minimum: 0
+              maximum: 16777215
+            remote:
+              type: string
+            destination-port:
+              type: integer
+            base-iface:
+              type: string
+
   interface-bond:
     rw:
       properties:
--- a/tests/lib/schema_validation_test.py
+++ b/tests/lib/schema_validation_test.py
@@ -27,11 +27,12 @@ import jsonschema as js
 import libnmstate
 from libnmstate.schema import Constants
 from libnmstate.schema import DNS
-
+from libnmstate.schema import VXLAN
+from libnmstate.schema import Interface
 
 INTERFACES = Constants.INTERFACES
 ROUTES = Constants.ROUTES
-
+VXLAN0 = 'vxlan0'
 
 COMMON_DATA = {
     INTERFACES: [
@@ -147,6 +148,54 @@ class TestIfaceTypeEthernet(object):
             {'type': 'ethernet', 'auto-negotiation': False}
         )
         del default_data[INTERFACES][0]['link-speed']
+
+        libnmstate.validator.validate(default_data)
+
+
+class TestIfaceTypeVxlan(object):
+    def test_bad_id_type_is_invalid(self, default_data):
+        default_data[Interface.KEY].append(
+            {
+                Interface.NAME: VXLAN0,
+                Interface.TYPE: VXLAN.TYPE,
+                VXLAN.CONFIG_SUBTREE: {
+                    VXLAN.ID: 'badtype',
+                    VXLAN.BASE_IFACE: 'eth1',
+                    VXLAN.REMOTE: '192.168.3.3',
+                },
+            }
+        )
+
+        with pytest.raises(js.ValidationError) as err:
+            libnmstate.validator.validate(default_data)
+
+        assert '\'badtype\' is not of type \'integer\'' in err.value.args[0]
+
+    def test_bad_id_range_is_invalid(self, default_data):
+        default_data[Interface.KEY].append(
+            {
+                Interface.NAME: VXLAN0,
+                Interface.TYPE: VXLAN.TYPE,
+                VXLAN.CONFIG_SUBTREE: {
+                    VXLAN.ID: 16777216,
+                    VXLAN.BASE_IFACE: 'eth1',
+                    VXLAN.REMOTE: '192.168.3.3',
+                },
+            }
+        )
+
+        with pytest.raises(js.ValidationError) as err:
+            libnmstate.validator.validate(default_data)
+
+        assert (
+            '16777216 is greater than the maximum of 16777215'
+            in err.value.args[0]
+        )
+
+    def test_no_config_is_valid(self, default_data):
+        default_data[Interface.KEY].append(
+            {Interface.NAME: VXLAN0, Interface.TYPE: VXLAN.TYPE}
+        )
 
         libnmstate.validator.validate(default_data)
 
--- a/tests/lib/validator_test.py
+++ b/tests/lib/validator_test.py
@@ -26,6 +26,7 @@ from libnmstate import validator
 from libnmstate.schema import DNS
 from libnmstate.schema import InterfaceIPv4
 from libnmstate.schema import InterfaceIPv6
+from libnmstate.schema import VXLAN
 from libnmstate.error import NmstateNotImplementedError
 from libnmstate.error import NmstateValueError
 
@@ -288,6 +289,54 @@ class TestRouteValidation(object):
         return _create_route(
             '2001:db8:a::/64', '2001:db8:1::a', 'eth2', 51, 104
         )
+
+
+class TestVxlanValidation(object):
+
+    parametrize_vxlan_req_fields = pytest.mark.parametrize(
+        'required_field', [VXLAN.ID, VXLAN.REMOTE, VXLAN.BASE_IFACE]
+    )
+
+    @parametrize_vxlan_req_fields
+    def test_missing_requiered_field_is_invalid(self, required_field):
+        desired_state = {
+            schema.Interface.KEY: [
+                {
+                    schema.Interface.NAME: 'eth0.101',
+                    schema.Interface.TYPE: VXLAN.TYPE,
+                    VXLAN.CONFIG_SUBTREE: {
+                        VXLAN.ID: 99,
+                        VXLAN.REMOTE: '192.168.3.3',
+                        VXLAN.BASE_IFACE: 'eth0',
+                    },
+                }
+            ]
+        }
+
+        desired_state[schema.Interface.KEY][0][VXLAN.CONFIG_SUBTREE].pop(
+            required_field
+        )
+
+        with pytest.raises(NmstateValueError) as err:
+            libnmstate.validator.validate_vxlan(desired_state)
+
+        assert required_field in err.value.args[0]
+
+    def test_minimal_config_is_valid(self):
+        desired_state = {
+            schema.Interface.KEY: [
+                {
+                    schema.Interface.NAME: 'eth0.101',
+                    schema.Interface.TYPE: VXLAN.TYPE,
+                    VXLAN.CONFIG_SUBTREE: {
+                        VXLAN.ID: 99,
+                        VXLAN.REMOTE: '192.168.3.3',
+                        VXLAN.BASE_IFACE: "eth0",
+                    },
+                }
+            ]
+        }
+        libnmstate.validator.validate_vxlan(desired_state)
 
 
 def _create_interface_state(
--- a/libnmstate/netapplier.py
+++ b/libnmstate/netapplier.py
@@ -61,6 +61,7 @@ def apply(desired_state, verify_change=True, commit=True, rollback_timeout=60):
     validator.validate_capabilities(desired_state, netinfo.capabilities())
     validator.validate_dhcp(desired_state)
     validator.validate_dns(desired_state)
+    validator.validate_vxlan(desired_state)
 
     checkpoint = _apply_ifaces_state(
         state.State(desired_state), verify_change, commit, rollback_timeout
--- a/libnmstate/validator.py
+++ b/libnmstate/validator.py
@@ -29,6 +29,7 @@ from .schema import Constants
 from libnmstate.schema import DNS
 from libnmstate.schema import InterfaceIP
 from libnmstate.schema import InterfaceIPv6
+from libnmstate.schema import VXLAN
 from libnmstate.error import NmstateDependencyError
 from libnmstate.error import NmstateNotImplementedError
 from libnmstate.error import NmstateValueError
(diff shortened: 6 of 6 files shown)


libnmstate/schema.py
```diff
@@ -24,50 +24,50 @@
...
 def load(schema_name):
     return yaml.load(
-        pkgutil.get_data('libnmstate', 'schemas/' + schema_name + '.yaml'),
+        pkgutil.get_data("libnmstate", "schemas/" + schema_name + ".yaml"),
         Loader=yaml.SafeLoader,
     )
 
 
-ifaces_schema = load('operational-state')
+ifaces_schema = load("operational-state")
```
maiqueb: ditto.

libnmstate/schema.py
```diff
@@ -24,50 +24,50 @@
 
 def load(schema_name):
     return yaml.load(
-        pkgutil.get_data('libnmstate', 'schemas/' + schema_name + '.yaml'),
+        pkgutil.get_data("libnmstate", "schemas/" + schema_name + ".yaml"),
```
maiqueb: This is against nmstate's [coding guidelines](https://github.com/nmstate/nmstate/blob/master/CONTRIBUTING.md#coding-and-style-guidelines).

libnmstate/schemas/operational-state.yaml
```diff
@@ -177,6 +178,30 @@ definitions:
...
+    rw:
+      properties:
+        type:
+          type: string
+          enum:
+            - vxlan
+        vxlan:
+          type: object
+          properties:
+            id:
```
maiqueb: shouldn't this be VNI - as per [rfc7348](https://tools.ietf.org/html/rfc7348) ?

libnmstate/schema.py
qinqon: I think 'black' has change this for me

libnmstate/schemas/operational-state.yaml
maiqueb: couldn't it be an int, ranging from 0 to 2^24 ?

libnmstate/schemas/operational-state.yaml
```diff
@@ -177,6 +178,32 @@ definitions:
...
+              type: integer
+              minimum: 0
+              maximum: 16777216
+            group:
+              type: string
+            dstport:
+              type: integer
+            base-iface:
+              type: string
+          required:
```
EdDev: Although it is correct, it will be better to raise an error in the code for required fields. Currently, the python json-schema is not that nice when it fails (the output is hard to interpret by a common user).

qinqon:
@EdDev well I am trying to cover here what we need at kubevirt tests

EdDev:
> well I am trying to cover here what we need at kubevirt tests

Sure thing, but in the context of nmstate we need to explain the limited scope and the limitations.
Other potential users may see vxlan support and expect them to work in various modes/scenarios. We need to express the limitations well.

qinqon:
@EdDev can we put in the documentation that it has "basic" vxlan support with following configuration options:
- multicast / unicast
- destination port 

This is all we need.

EdDev:
>    multicast / unicast

But you only added support for multicast if I'm not mistaken.
Anyway, it needs to be detailed with the fields we support and should be added to the limitation list (see #500).
For documentation, it needs to be added here: https://nmstate.github.io/devel/api

EdDev:
Note that in the `schema.py` module you need to add the constants/enums and use them in the tests.

qinqon:
> But you only added support for multicast if I'm not mistaken.
@EdDev if "group" is not present it's unicast, I will add this stuff to the limitations.

qinqon:
@EdDev we should follow the naming here ? https://developer.gnome.org/NetworkManager/unstable/nm-settings.html

EdDev:
> if "group" is not present it's unicast, I will add this stuff to the limitations.

I thought you need `remote` if you use unicast.

> we should follow the naming here

Th naming of what? If you mean the schema, then no necessary. It may be a good baseline, but you can use whatever makes sense. I usually go over ip-link and NM and try to figure out if it sounds good. We can discuss it afterwards on the PR.

qinqon:
Ok I am going to change the scheam to NM names for VXLAN see how it goes

libnmstate/schema.py
```diff
@@ -189,6 +191,16 @@ class VLAN(object):
     BASE_IFACE = 'base-iface'
 
 
+class VXLAN(object):
+    TYPE = InterfaceType.VXLAN
+    CONFIG_SUBTREE = 'vxlan'
+
+    ID = 'id'
+    PARENT = 'parent'
```
cathay4t: Please use `base-iface` as VLAN did.

libnmstate/schemas/operational-state.yaml
```diff
@@ -177,6 +178,32 @@ definitions:
...
+          type: string
+          enum:
+            - vxlan
+        vxlan:
+          type: object
+          properties:
+            id:
+              type: integer
+              minimum: 0
+              maximum: 16777216
```
cathay4t: Should be `16777215` as JSON schema defined `maximum` as inclusive.

libnmstate/schemas/operational-state.yaml
cathay4t: +1

tests/lib/schema_validation_test.py
```diff
@@ -151,6 +152,82 @@ def test_valid_without_auto_neg_and_missing_speed(self, default_data):
         libnmstate.validator.validate(default_data)
 
 
+class TestIfaceTypeVxlan(object):
+    def test_invalid_vxlan_empty(self, default_data):
+        default_data[INTERFACES].append({'name': 'vxlan0', 'type': VXLAN.TYPE})
```
cathay4t: Please use `Interface.NAME` and `Interface.TYPE` instead of magic string.

libnmstate/schemas/operational-state.yaml
qinqon: This would be done in the implementation PR, that's right ?

maiqueb: This is just an opinion, please take it with plenty of salt.

> Although it is correct, it will be better to raise an error in the code for required fields. Currently, the python json-schema is not that nice when it fails (the output is hard to interpret by a common user).

While I agree that the schema's validation errors could be almost considered obfuscation, they still have somewhere a decent indication of the error. Couldn't we wrap those errors, and provide decent info from them instead ? 

Or do we use the schema just to have conversations about the data model ?

qinqon: @maiqueb I forgot it, let me add it.

tests/lib/schema_validation_test.py
```diff
@@ -151,6 +153,78 @@ def test_valid_without_auto_neg_and_missing_speed(self, default_data):
...
+                Interface.NAME: 'vxlan0',
+                Interface.TYPE: VXLAN.TYPE,
+                VXLAN.CONFIG_SUBTREE: {},
+            }
+        )
+
+        with pytest.raises(js.ValidationError) as err:
+            libnmstate.validator.validate(default_data)
+
+        assert "'vxlan' is not one of ['unknown']" in err.value.args[0]
```
maiqueb: this error msg is a bit cryptic isn't it ? I'm surprised at least.

tests/lib/schema_validation_test.py
```diff
@@ -151,6 +153,78 @@ def test_valid_without_auto_neg_and_missing_speed(self, default_data):
...
+                Interface.NAME: 'vxlan0',
+                Interface.TYPE: VXLAN.TYPE,
+                VXLAN.CONFIG_SUBTREE: {VXLAN.ID: 99},
+            }
+        )
+
+        with pytest.raises(js.ValidationError) as err:
+            libnmstate.validator.validate(default_data)
+
+        assert "'vxlan' is not one of ['unknown']" in err.value.args[0]
```
maiqueb: ditto

tests/lib/schema_validation_test.py
```diff
@@ -151,6 +153,78 @@ def test_valid_without_auto_neg_and_missing_speed(self, default_data):
...
+                Interface.NAME: 'vxlan0',
+                Interface.TYPE: VXLAN.TYPE,
+                VXLAN.CONFIG_SUBTREE: {VXLAN.BASE_IFACE: 'eth1'},
+            }
+        )
+
+        with pytest.raises(js.ValidationError) as err:
+            libnmstate.validator.validate(default_data)
+
+        assert "'vxlan' is not one of ['unknown']" in err.value.args[0]
```
maiqueb: ditto

tests/lib/schema_validation_test.py
```diff
@@ -151,6 +153,110 @@ def test_valid_without_auto_neg_and_missing_speed(self, default_data):
...
+
+    def test_invalid_vxlan_without_parent(self, default_data):
+        default_data[INTERFACES].append(
```
cathay4t: Please use `Interface.KEY` instead of `INTERFACES` which is obsoleted.

tests/lib/schema_validation_test.py
cathay4t: Please use single quote for string.

tests/lib/schema_validation_test.py
```diff
@@ -151,6 +153,110 @@ def test_valid_without_auto_neg_and_missing_speed(self, default_data):
...
+                    VXLAN.BASE_IFACE: 'eth1',
+                    VXLAN.REMOTE: "192.168.3.3",
+                },
+            }
+        )
+
+        with pytest.raises(js.ValidationError) as err:
+            libnmstate.validator.validate(default_data)
+
+        assert "'vxlan' is not one of ['unknown']" in err.value.args[0]
```
cathay4t: Please use single quote for string.

tests/lib/schema_validation_test.py
cathay4t: Please use single quote for string.

tests/lib/schema_validation_test.py
```diff
@@ -151,6 +153,110 @@ def test_valid_without_auto_neg_and_missing_speed(self, default_data):
...
+                    VXLAN.BASE_IFACE: 'eth1',
+                    VXLAN.REMOTE: "192.168.3.3",
+                },
+            }
+        )
+
+        with pytest.raises(js.ValidationError) as err:
+            libnmstate.validator.validate(default_data)
+
+        assert "'badtype' is not of type 'integer'" in err.value.args[0]
```
cathay4t: Please use single quote for string.

tests/lib/schema_validation_test.py
```diff
@@ -151,6 +153,110 @@ def test_valid_without_auto_neg_and_missing_speed(self, default_data):
...
+            }
+        )
+
+        with pytest.raises(js.ValidationError) as err:
+            libnmstate.validator.validate(default_data)
+
+        assert "'vxlan' is not one of ['unknown']" in err.value.args[0]
+
+    def test_invalid_vxlan_without_id(self, default_data):
+        default_data[INTERFACES].append(
```
cathay4t: `Interface.KEY`

tests/lib/schema_validation_test.py
```diff
@@ -151,6 +153,110 @@ def test_valid_without_auto_neg_and_missing_speed(self, default_data):
...
+            }
+        )
+
+        with pytest.raises(js.ValidationError) as err:
+            libnmstate.validator.validate(default_data)
+
+        assert "'vxlan' is not one of ['unknown']" in err.value.args[0]
+
+    def test_invalid_vxlan_without_remote(self, default_data):
+        default_data[INTERFACES].append(
```
cathay4t: `Interface.KEY`

tests/lib/schema_validation_test.py
```diff
@@ -151,6 +153,110 @@ def test_valid_without_auto_neg_and_missing_speed(self, default_data):
         libnmstate.validator.validate(default_data)
 
 
+class TestIfaceTypeVxlan(object):
+    def test_invalid_vxlan_without_options(self, default_data):
+        default_data[INTERFACES].append(
+            {
+                Interface.NAME: 'vxlan0',
```
cathay4t: The `vxlan0` has been used multiple times. Please use a constant for it.

libnmstate/schemas/operational-state.yaml
EdDev: > This would be done in the implementation PR, that's right ?

Yes, we do such things in the `validator` module.

> Couldn't we wrap those errors, and provide decent info from them instead ?

Interpretting the schema error seems more complex that just handling this in the code.

> Or do we use the schema just to have conversations about the data model ?

Schemas can be used for several objectives, one is indeed to describe the data model formally, another is to generate client code (or server code, like for an http service). Documentation generation is another option.
But for the current discussion, only very basic stuff we enforce through the schema.

libnmstate/schemas/operational-state.yaml
```diff
@@ -177,6 +178,31 @@ definitions:
           required:
             - id
             - base-iface
+  interface-vxlan:
```
EdDev: What about `group` for multicast support? Isn't it basic?

libnmstate/schema.py
cathay4t: Please use `alias black='black -S -l 79'`
or in vimrc:

```
let g:black_linelength = 79
let g:black_skip_string_normalization = 1
```

libnmstate/schemas/operational-state.yaml
qinqon: Remote can be unicast or multicast, that's why I don't add it like 'group'

https://developer.gnome.org/NetworkManager/stable/settings-vxlan.html

tests/lib/schema_validation_test.py
qinqon: This is how the error at jsonschema works, since there is no schema that matches it it mark it as unknown, it does not show you the partial errors or the like.

libnmstate/schema.py
qinqon: This is old stuff, now I follow that approach.

tests/lib/schema_validation_test.py
```diff
@@ -143,10 +143,52 @@ def test_valid_without_auto_neg_and_missing_speed(self, default_data):
...
+                    VXLAN.ID: 99,
+                    VXLAN.BASE_IFACE: 'eth1',
+                    VXLAN.REMOTE: '192.168.3.3',
+                },
+            }
+        )
+
+        libnmstate.validator.validate(default_data)
+
+    def test_valid_vxlan_without_subtree(self, default_data):
```
EdDev: I would replace this with "bad-id-range". But it is not a must.

tests/lib/validator_test.py
```diff
@@ -290,6 +291,24 @@ def _create_route1(self):
         )
 
 
+class TestVxlanValidation(object):
+    def test_invalid_vxlan_without_parent(self):
```
EdDev: Reads better I think if you call this `test_missing_parent_is_invalid`.
The use of `vxlan` in the test name is redundant, you have the class name to mention that.

tests/lib/schema_validation_test.py
```diff
@@ -143,10 +143,52 @@ def test_valid_without_auto_neg_and_missing_speed(self, default_data):
...
             {'type': 'ethernet', 'auto-negotiation': False}
         )
-        del default_data[INTERFACES][0]['link-speed']
+        del default_data[Interface.KEY][0]['link-speed']
+
+        libnmstate.validator.validate(default_data)
+
+
+class TestIfaceTypeVxlan(object):
+    def test_invalid_vxlan_with_bad_id_type(self, default_data):
```
EdDev: I would recommend to drop the `vxlan` name from the test name. The class name already has this and provides the needed context.

Also, for a smoother reading of the test name, you can leave the expected outcome at the end:
`test_bad_ip_type_is_invalid`

tests/lib/schema_validation_test.py
```diff
@@ -27,14 +27,14 @@
...
 
-
-INTERFACES = Constants.INTERFACES
 ROUTES = Constants.ROUTES
-
+VXLAN0 = 'vxlan0'
 
 COMMON_DATA = {
-    INTERFACES: [
+    Interface.KEY: [
```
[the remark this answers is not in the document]
qinqon: Agree

tests/lib/validator_test.py
```diff
@@ -290,6 +291,72 @@ def _create_route1(self):
         )
 
 
+class TestVxlanValidation(object):
+    def test_missing_parent_is_invalid(self):
```
EdDev: The scenarios look great.
These tests that check missing required parameters look to have the same template, so you can use pytest parametrize option and do something like this:
```
parametrize_vxlan_req_fields = pytest.mark.parametrize(
    'required_field', [VXLAN.FOO, VXLAN.FOO2, ...]
)

    @parametrize_vxlan_req_fields
    def test_missing_parent_is_invalid(self, required_field):
        vxlan_state = {
            VXLAN.ID: 99,
            VXLAN.REMOTE: '192.168.3.3',
            VXLAN.BASE_IFACE: "eth0",
        }
        vxlan_state.pop(required_fields)

        desired_state = {
            schema.Interface.KEY: [
                {
                    schema.Interface.NAME: 'eth0.101',
                    schema.Interface.TYPE: VXLAN.TYPE,
                    VXLAN.CONFIG_SUBTREE: vxlan_state,
                }
            ]
        }
        with pytest.raises(NmstateValueError):
            libnmstate.validator.validate_vxlan(desired_state)
```

EdDev:
@qinqon , Thank you for the integration tests adjustment.
Could you also answer the others?

qinqon:
@EdDev ohh sorry I miss them, let me fix them too.

EdDev:
@cathay4t previous comments have been addressed, merging.

PR #510, opened 2019-09-22
Refactor dynamic ip tests

EdDev:
- Rename the test module.
- Start all tests that depend on the client interface with IP stack disabled.
- Use IP state helpers
- Use schema constants.
- Use bridgelib and bondlib.
- Place helpers at the module tail (they appeared in between tests).


tests/integration/dynamic_ip_test.py
```diff
@@ -589,6 +527,52 @@ def test_dhcp_on_bridge0(dhcpcli_up_with_delay_status):
...
+):
+    dhcpcli_up[Interface.KEY][0][Interface.IPV4] = create_ipv4_state(
+        enabled=True, dhcp=True
+    )
+
+    libnmstate.apply(dhcpcli_up)
+
+    slave_ifname = dhcpcli_up[Interface.KEY][0][Interface.NAME]
+    current_state = statelib.show_only((slave_ifname,))
+    client_current_state = current_state[INTERFACES][0]
```
maiqueb: client_current_iface_state ?

tests/integration/dynamic_ip_test.py
```diff
@@ -589,6 +527,52 @@ def test_dhcp_on_bridge0(dhcpcli_up_with_delay_status):
...
+    dhcpcli_ip = client_current_state[Interface.IPV4][InterfaceIPv4.ADDRESS]
+
+    bridge_state = create_bridge_subtree_state()
+    bridge_state = add_port_to_bridge(bridge_state, slave_ifname)
```
maiqueb: I think there's no need to store the (initial) bridge state on a variable - which is later on updated.

Why not just ``` bridge_state = add_port_to_bridge(create_bridge_subtree_state(), slave_ifname)```

tests/integration/dynamic_ip_test.py
```diff
@@ -589,6 +527,52 @@ def test_dhcp_on_bridge0(dhcpcli_up_with_delay_status):
...
+
+    ipv4_state = create_ipv4_state(enabled=True, dhcp=False)
+    ipv4_state[InterfaceIPv4.ADDRESS] = dhcpcli_ip
+    bridge_iface_state = {Interface.IPV4: ipv4_state}
```
maiqueb: I miss 'ip' in the name of the variable - this is a bridge ip configuration. 

bridge_ip_iface_state ?

tests/integration/dynamic_ip_test.py
```diff
@@ -589,6 +527,52 @@ def test_dhcp_on_bridge0(dhcpcli_up_with_delay_status):
...
+    with linux_bridge(
+        'brtest0', bridge_state, bridge_iface_state, create=False
```
maiqueb: wouldn't it be 'better' to also indicate the 'extra_iface_state' keyword argument there ? why only for the 'create' parameter?

tests/integration/dynamic_ip_test.py
EdDev: I took it out.

tests/integration/dynamic_ip_test.py
```diff
@@ -245,41 +226,24 @@ def test_ipv6_dhcp_and_autoconf(dhcpcli_up):
 
 
 def test_dhcp_with_addresses(dhcpcli_up):
+    ipv4_state = create_ipv4_state(enabled=True, dhcp=True)
+    ipv4_state[InterfaceIPv4.ADDRESS] = [
+        create_ipv4_address_state(IPV4_ADDRESS1, 24),
+        create_ipv4_address_state(IPV4_ADDRESS2, 24),
+    ]
+    ipv6_state = create_ipv6_state(enabled=True, dhcp=True, autoconf=True)
+    ipv4_state[InterfaceIPv6.ADDRESS] = [
```
cathay4t: Should be `ipv6_state`

EdDev: Done

PR #512, opened 2019-09-25
tests, integ, dhcp: add tests exposing that dhcp on a bridge whose port is a vlan is buggy

maiqueb:
On more complex scenarios, involving VLANs, NetworkManager hangs while assigning an IP address via DHCP to the linux bridge.

This is examplified when creating a VLAN on top of a bond, on top of an ethernet interface. That VLAN is in turn set as a port to a linux bridge.

The 2 added tests show that NM hangs when the linux bridge has a dynamic IP configuration on top of a VLAN - in the first test, this happens within a single transaction, while on the last, it manifests
when multiple transactions are used.

The NM main loop aborts since the interfaces are not assigned a dynamic IP address - since the DHCP server configured in the dynamic IP module is not being ran behind a VLAN interface, and as
such, the L2 traffic (untagged) is discarded before reaching the bridge.

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1755542

### DIFF
--- a/tests/integration/dynamic_ip_test.py
+++ b/tests/integration/dynamic_ip_test.py
@@ -67,37 +67,12 @@ DHCP_SRV_IP6_NETWORK = '{}::/64'.format(DHCP_SRV_IP6_PREFIX)
 IPV6_DEFAULT_GATEWAY = '::/0'
 IPV4_DEFAULT_GATEWAY = '0.0.0.0/0'
 
-DNSMASQ_CONF_STR = """
-interface={iface}
-dhcp-range={ipv4_prefix}.200,{ipv4_prefix}.250,255.255.255.0,48h
-enable-ra
-dhcp-range={ipv6_prefix}::100,{ipv6_prefix}::fff,ra-names,slaac,64,480h
-dhcp-range={ipv6_classless_route}::100,{ipv6_classless_route}::fff,static
-dhcp-option=option:classless-static-route,{classless_rt},{classless_rt_dst}
-dhcp-option=option:dns-server,{v4_dns_server}
-""".format(
-    **{
-        'iface': DHCP_SRV_NIC,
-        'ipv4_prefix': DHCP_SRV_IP4_PREFIX,
-        'ipv6_prefix': DHCP_SRV_IP6_PREFIX,
-        'classless_rt': IPV4_CLASSLESS_ROUTE_DST_NET1,
-        'classless_rt_dst': IPV4_CLASSLESS_ROUTE_NEXT_HOP1,
-        'v4_dns_server': DHCP_SRV_IP4,
-        'ipv6_classless_route': IPV6_CLASSLESS_ROUTE_PREFIX,
-    }
-)
-
 DNSMASQ_CONF_PATH = '/etc/dnsmasq.d/nmstate.conf'
+
 # Docker does not allow NetworkManager to edit /etc/resolv.conf.
 # Have to read NetworkManager internal resolv.conf
 RESOLV_CONF_PATH = '/var/run/NetworkManager/resolv.conf'
 
-SYSFS_DISABLE_IPV6_FILE = '/proc/sys/net/ipv6/conf/{}/disable_ipv6'.format(
-    DHCP_SRV_NIC
-)
-SYSFS_DISABLE_RA_SRV = '/proc/sys/net/ipv6/conf/{}/accept_ra'.format(
-    DHCP_SRV_NIC
-)
 
 TEST_BRIDGE = 'linux-br0'
 BRIDGE_PORT_CONFIG = {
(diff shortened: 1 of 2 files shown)


cathay4t:
The test does not setup a DHCP server on VLAN interface, so we cannot tell it's a nmstate bug or DHCP failure on this issue.

cathay4t:
To cover this case, please consider to create a test case where DHCP server is actually running:

```bash
#!/bin/bash

VLAN_ID=100
SRV_NIC="dhcpsrv"
CLI_NIC="dhcpcli"
SRV_VLAN_NIC="$SRV_NIC.$VLAN_ID"
NS_NAME="ns_dhcp_vlan"

ip link del $CLI_NIC
ip netns del $NS_NAME

ip link add $CLI_NIC type veth peer name $SRV_NIC
ip link set up $CLI_NIC
ip netns add $NS_NAME
ip link set $SRV_NIC netns $NS_NAME
ip netns exec $NS_NAME \
    ip link set up $SRV_NIC
ip netns exec $NS_NAME \
    ip link add link $SRV_NIC name $SRV_VLAN_NIC type vlan id $VLAN_ID
ip netns exec $NS_NAME \
    ip link set up $SRV_VLAN_NIC
ip netns exec $NS_NAME ip addr add 192.0.2.251/24 dev $SRV_VLAN_NIC
ip netns exec $NS_NAME \
    dnsmasq --dhcp-range=192.0.2.100,192.0.2.200,255.255.255.0 \
    --interface=$SRV_VLAN_NIC

nmcli device set $CLI_NIC managed yes
```

maiqueb:
> The test does not setup a DHCP server on VLAN interface, so we cannot tell it's a nmstate bug or DHCP failure on this issue.

This could be useful though to keep an xfailed test to show that hanging DHCP responses block us.

Just an opinion though.

maiqueb:
> To cover this case, please consider to create a test case where DHCP server is actually running:
> 
> ```shell
> #!/bin/bash
> 
> VLAN_ID=100
> SRV_NIC="dhcpsrv"
> CLI_NIC="dhcpcli"
> SRV_VLAN_NIC="$SRV_NIC.$VLAN_ID"
> NS_NAME="ns_dhcp_vlan"
> 
> ip link del $CLI_NIC
> ip netns del $NS_NAME
> 
> ip link add $CLI_NIC type veth peer name $SRV_NIC
> ip link set up $CLI_NIC
> ip netns add $NS_NAME
> ip link set $SRV_NIC netns $NS_NAME
> ip netns exec $NS_NAME \
>     ip link set up $SRV_NIC
> ip netns exec $NS_NAME \
>     ip link add link $SRV_NIC name $SRV_VLAN_NIC type vlan id $VLAN_ID
> ip netns exec $NS_NAME \
>     ip link set up $SRV_VLAN_NIC
> ip netns exec $NS_NAME ip addr add 192.0.2.251/24 dev $SRV_VLAN_NIC
> ip netns exec $NS_NAME \
>     dnsmasq --dhcp-range=192.0.2.100,192.0.2.200,255.255.255.0 \
>     --interface=$SRV_VLAN_NIC
> 
> nmcli device set $CLI_NIC managed yes
> ```

Sure, thanks for this :)

I'll adapt the existing code to optionally run dnsmasq behind a VLAN.

tests/integration/dynamic_ip_test.py
```diff
@@ -134,16 +131,22 @@ def dhcpcli_up(dhcp_env):
         yield ifstate
 
 
+@pytest.fixture
+def dhcp_sb_cli_up(dhcp_env_vlan):
```
cathay4t: What does the `sb` here mean?

maiqueb:
> Please place the server veth and dnsmasq into a namespace. In my test, only with that the DHCP server can work on a VLAN interface.

Why does that happen ? I'll try to get traffic captures of this, to ensure there are dhcp responses from the server on the dhcp_cli interface.

tests/integration/dynamic_ip_test.py
maiqueb: southbound.

would you propose a better name?

tests/integration/dynamic_ip_test.py
```diff
@@ -576,6 +606,71 @@ def test_ipv6_autoconf_only(dhcpcli_up):
...
+        bridge_name,
+        bridge_config_state,
+        extra_iface_state=bridge_ip_config,
+        create=False,
+    ) as bridge0:
+        desired_state = bond0
+        vlan_iface_state = bond0_vlan101[Interface.KEY][0]
+        desired_state[Interface.KEY].append(vlan_iface_state)
+        bridge_iface_state = bridge0[Interface.KEY][0]
+        desired_state[Interface.KEY].append(bridge_iface_state)
```
EdDev: Could you extract it to a `state_merge` helper function?
It will also be helpful when reading to name the state variable with what it includes (e.g. bridge_vlan_bond_with_ipv4_state).

tests/integration/dynamic_ip_test.py
```diff
@@ -576,6 +606,71 @@ def test_ipv6_autoconf_only(dhcpcli_up):
     libnmstate.apply(desired_state)
 
 
+@pytest.mark.xfail(reason='DHCP server does not run behind vlan', strict=True)
+def test_linux_bridge_over_vlan_over_bond_over_slave_single_transaction(
```
EdDev: The name does not express that there is no dhcp server, which is the main point of this test.

tests/integration/dynamic_ip_test.py
```diff
@@ -576,6 +606,71 @@ def test_ipv6_autoconf_only(dhcpcli_up):
...
+        desired_state[Interface.KEY].append(vlan_iface_state)
+        bridge_iface_state = bridge0[Interface.KEY][0]
+        desired_state[Interface.KEY].append(bridge_iface_state)
+        libnmstate.apply(desired_state)
+
+        desired_state = bridge0
+        desired_state[Interface.KEY][0][Interface.IPV4] = create_ipv4_state(
+            enabled=True, dhcp=True
+        )
+        libnmstate.apply(desired_state)
```
EdDev: bridge0 is already supposed to include the IPv4 state and in general I do not understand why this section is needed.

tests/integration/dynamic_ip_test.py
EdDev: southbound of what? This fixture/func does not know how the return value is used, so naming it relative to something that may use it is problematic.

How about `dhcp_client_iface_up`.

EdDev:
@maiqueb , could you please revisit this PR?

maiqueb:
Temporarily closing the PR, to reduce the PR queue, since I won't be able to work on this at the moment.

PR #513, opened 2019-09-26
clean-up OVS integration test

phoracek:
Introduce helper functions to the module and streamline its format.
This will be helpful in a follow-up patch introducing OVS bonding.

### DIFF
--- a/tests/integration/ovs_test.py
+++ b/tests/integration/ovs_test.py
@@ -18,83 +18,72 @@
 #
 
 import pytest
-import yaml
 
-import libnmstate
+from libnmstate.schema import Interface
 from libnmstate.schema import InterfaceIPv4
+from libnmstate.schema import OVSBridge
 from libnmstate.error import NmstateLibnmError
 
-from .testlib import statelib
-from .testlib.statelib import INTERFACES
+from .testlib import assertlib
+from .testlib.ovslib import Bridge
 
 
-OVS_BRIDGE_YAML_BASE = """
-interfaces:
-  - name: ovs-br0
-    type: ovs-bridge
-    state: up
-    bridge:
-      options:
-        fail-mode: ''
-        mcast-snooping-enable: false
-        rstp: false
-        stp: true
-"""
-
-
-@pytest.mark.xfail(
-    raises=NmstateLibnmError, reason='https://bugzilla.redhat.com/1724901'
-)
-def test_create_and_remove_ovs_bridge_with_a_system_port(eth1_up):
-    state = yaml.load(OVS_BRIDGE_YAML_BASE, Loader=yaml.SafeLoader)
-    state[INTERFACES][0]['bridge']['port'] = [
-        {'name': 'eth1', 'type': 'system'}
-    ]
-    libnmstate.apply(state)
-
-    setup_remove_ovs_bridge_state = {
-        INTERFACES: [
-            {'name': 'ovs-br0', 'type': 'ovs-bridge', 'state': 'absent'}
-        ]
-    }
-    libnmstate.apply(setup_remove_ovs_bridge_state)
-    state = statelib.show_only((state[INTERFACES][0]['name'],))
-    assert not state[INTERFACES]
+BRIDGE1 = 'br1'
+PORT1 = 'ovs1'
 
 
 @pytest.mark.xfail(
     raises=NmstateLibnmError, reason='https://bugzilla.redhat.com/1724901'
 )
 def test_create_and_remove_ovs_bridge_with_min_desired_state():
-    desired_state = {
-        INTERFACES: [{'name': 'ovs-br0', 'type': 'ovs-bridge', 'state': 'up'}]
-    }
-    libnmstate.apply(desired_state)
+    with Bridge(BRIDGE1).create() as state:
+        assertlib.assert_state_match(state)
 
-    setup_remove_ovs_bridge_state = {
-        INTERFACES: [
-            {'name': 'ovs-br0', 'type': 'ovs-bridge', 'state': 'absent'}
-        ]
-    }
-    libnmstate.apply(setup_remove_ovs_bridge_state)
-    state = statelib.show_only((desired_state[INTERFACES][0]['name'],))
-    assert not state[INTERFACES]
+    assertlib.assert_absent(BRIDGE1)
 
 
 @pytest.mark.xfail(
     raises=NmstateLibnmError, reason='https://bugzilla.redhat.com/1724901'
 )
-def test_create_and_remove_ovs_bridge_with_an_internal_port():
-    state = yaml.load(OVS_BRIDGE_YAML_BASE, Loader=yaml.SafeLoader)
-    state[INTERFACES][0]['bridge']['port'] = [
-        {'name': 'ovs0', 'type': 'internal'}
-    ]
-    ovs_internal_interface_state = {
-        'name': 'ovs0',
-        'type': 'ovs-interface',
-        'state': 'up',
-        'mtu': 1500,
-        'ipv4': {
+def test_create_and_remove_ovs_bridge_options_specified():
+    bridge = Bridge(BRIDGE1)
+    bridge.set_options(
+        {
+            OVSBridge.FAIL_MODE: '',
+            OVSBridge.MCAST_SNOOPING_ENABLED: False,
+            OVSBridge.RSTP: False,
+            OVSBridge.STP: True,
+        }
+    )
+
+    with bridge.create() as state:
+        assertlib.assert_state_match(state)
+
+    assertlib.assert_absent(BRIDGE1)
+
+
+@pytest.mark.xfail(
+    raises=NmstateLibnmError, reason='https://bugzilla.redhat.com/1724901'
+)
+def test_create_and_remove_ovs_bridge_with_a_system_port(port0_up):
+    bridge = Bridge(BRIDGE1)
+    port0_name = port0_up[Interface.KEY][0][Interface.NAME]
+    bridge.add_system_port(port0_name)
+
+    with bridge.create() as state:
+        assertlib.assert_state_match(state)
+
+    assertlib.assert_absent(BRIDGE1)
+
+
+@pytest.mark.xfail(
+    raises=NmstateLibnmError, reason='https://bugzilla.redhat.com/1724901'
+)
+def test_create_and_remove_ovs_bridge_with_internal_port_and_static_ip():
+    bridge = Bridge(BRIDGE1)
+    bridge.add_internal_port(
+        PORT1,
+        {
             InterfaceIPv4.ENABLED: True,
             InterfaceIPv4.ADDRESS: [
                 {
@@ -103,18 +92,10 @@ def test_create_and_remove_ovs_bridge_with_an_internal_port():
                 }
             ],
         },
-    }
-    state[INTERFACES].append(ovs_internal_interface_state)
-    libnmstate.apply(state, verify_change=False)
-
-    setup_remove_ovs_bridge_state_and_port = {
-        INTERFACES: [
-            {'name': 'ovs-br0', 'type': 'ovs-bridge', 'state': 'absent'},
-            {'name': 'ovs', 'type': 'ovs-interface', 'state': 'absent'},
-        ]
-    }
-    libnmstate.apply(setup_remove_ovs_bridge_state_and_port)
-    state = statelib.show_only(
-        (state[INTERFACES][0]['name'], state[INTERFACES][1]['name'])
     )
-    assert not state[INTERFACES]
+
+    with bridge.create() as state:
+        assertlib.assert_state_match(state)
+
+    assertlib.assert_absent(BRIDGE1)
+    assertlib.assert_absent(PORT1)
--- a/tests/integration/testlib/ovslib.py
+++ b/tests/integration/testlib/ovslib.py
@@ -1,0 +1,87 @@
+#
+# Copyright (c) 2019 Red Hat, Inc.
+#
+# This file is part of nmstate
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+
+from contextlib import contextmanager
+import copy
+
+import libnmstate
+from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceState
+from libnmstate.schema import InterfaceType
+from libnmstate.schema import OVSBridge
+from libnmstate.schema import OVSBridgePortType
+
+
+class Bridge:
+    def __init__(self, name):
+        self._name = name
+        self._ifaces = [
+            {
+                Interface.NAME: name,
+                Interface.TYPE: InterfaceType.OVS_BRIDGE,
+                OVSBridge.CONFIG_SUBTREE: {},
+            }
+        ]
+        self._bridge_iface = self._ifaces[0]
+
+    def set_options(self, options):
+        self._bridge_iface[OVSBridge.CONFIG_SUBTREE][
+            OVSBridge.OPTIONS_SUBTREE
+        ] = options
+
+    def add_system_port(self, name):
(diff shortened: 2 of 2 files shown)


phoracek:
@EdDev @maiqueb @cathay4t could you please take a look? I wanted to do some clean-up before jumping to OVS bonding implementation.

phoracek:
Failed on unrelated tests.

phoracek:
ci test please

phoracek:
@maiqueb fixed. Could you please take another look? Thanks!

tests/integration/ovs_test.py
```diff
@@ -15,106 +14,102 @@
...
 
 @pytest.mark.xfail(
     raises=NmstateLibnmError, reason='https://bugzilla.redhat.com/1724901'
 )
-def test_create_and_remove_ovs_bridge_with_min_desired_state():
-    desired_state = {
-        INTERFACES: [{'name': 'ovs-br0', 'type': 'ovs-bridge', 'state': 'up'}]
-    }
-    libnmstate.apply(desired_state)
+def test_create_and_remove_ovs_bridge_with_a_system_port(eth1_up):
```
EdDev: For bridges we have an alias ‘port0_up’ instead of ‘eth1_up’.
And we fetch from it the actual name (see linux bridge tests).

tests/integration/ovs_test.py
```diff
@@ -15,106 +14,102 @@
...
-        fail-mode: ''
-        mcast-snooping-enable: false
-        rstp: false
-        stp: true
-"""
+@pytest.mark.xfail(
+    raises=NmstateLibnmError, reason='https://bugzilla.redhat.com/1724901'
+)
+def test_create_and_remove_ovs_bridge_with_min_desired_state():
+    with ovs_bridge('br1') as bridge_state:
```
EdDev: Please make ‘br1’ a constant, it is used a lot.

tests/integration/ovs_test.py
```diff
@@ -15,106 +14,102 @@
...
-        {'name': 'ovs0', 'type': 'internal'}
-    ]
-    ovs_internal_interface_state = {
-        'name': 'ovs0',
-        'type': 'ovs-interface',
-        'state': 'up',
-        'mtu': 1500,
-        'ipv4': {
+    bridge = create_bridge_subtree_state()
+    add_port_to_bridge(bridge, 'ovs1', OVSBridgePortType.INTERNAL)
```
EdDev: A constant or a local constant for ‘ovs1’ will be nice.

tests/integration/ovs_test.py
```diff
@@ -15,106 +14,102 @@
...
+            assertlib.assert_state_match(bridge_state)
+
+    assertlib.assert_absent('br1')
+    assertlib.assert_absent('ovs1')
+
+
+@pytest.mark.xfail(
+    raises=NmstateLibnmError, reason='https://bugzilla.redhat.com/1724901'
+)
+def test_create_and_remove_ovs_bridge_with_internal_port_and_static_ip():
```
EdDev: What is the point of having two tests for the internal port in this case? This 2nd one includes the other.

tests/integration/ovs_test.py
```diff
@@ -15,106 +14,102 @@
...
+    bridge = create_bridge_subtree_state()
+    add_port_to_bridge(bridge, 'ovs1', OVSBridgePortType.INTERNAL)
+
+    iface = {
```
EdDev: how about: ‘ipv4_state’?

tests/integration/ovs_test.py
```diff
@@ -15,106 +14,102 @@
 #
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
-#
```
EdDev: The header “fix” better be consistent with the whole project, therefore I suggest you do this in a separate PR project wide.

tests/integration/testlib/ovslib.py
```diff
@@ -0,0 +1,113 @@
...
+                Interface.TYPE: InterfaceType.OVS_BRIDGE,
+                Interface.STATE: InterfaceState.UP,
+                OVSBridge.CONFIG_SUBTREE: bridge_subtree_state or {},
+            }
+        ]
+    }
+    if extra_iface_state:
+        desired_state[Interface.KEY][0].update(extra_iface_state)
+
+    libnmstate.apply(desired_state, verify_change=False)
```
EdDev: We usually disable verification only on removal. Any reason you used it here?

tests/integration/testlib/ovslib.py
```diff
@@ -0,0 +1,113 @@
...
+
+    try:
+        yield desired_state
+    finally:
+        libnmstate.apply(
+            {
+                Interface.KEY: [
+                    {
+                        Interface.NAME: name,
+                        Interface.TYPE: InterfaceType.OVS_BRIDGE,
```
EdDev: The type is not needed.

tests/integration/testlib/ovslib.py
```diff
@@ -0,0 +1,113 @@
...
+        )
+
+
+def add_port_to_bridge(
+    bridge_subtree_state, port_name, port_type, port_state=None
+):
+    if OVSBridge.PORT_SUBTREE not in bridge_subtree_state:
+        bridge_subtree_state[OVSBridge.PORT_SUBTREE] = []
+
+    if port_state is None:
```
EdDev: Editing the caller extra port state internally is surprising. Lets avoid it.
I would also call it ‘extra...’

tests/integration/testlib/ovslib.py
```diff
@@ -0,0 +1,113 @@
...
+                Interface.NAME: name,
+                Interface.TYPE: InterfaceType.OVS_INTERFACE,
+                Interface.STATE: InterfaceState.UP,
+            }
+        ]
+    }
+    if extra_iface_state:
+        desired_state[Interface.KEY][0].update(extra_iface_state)
+
+    libnmstate.apply(desired_state, verify_change=False)
```
EdDev: Why no verification?

tests/integration/testlib/ovslib.py
```diff
@@ -0,0 +1,113 @@
...
+
+    try:
+        yield desired_state
+    finally:
+        libnmstate.apply(
+            {
+                Interface.KEY: [
+                    {
+                        Interface.NAME: name,
+                        Interface.TYPE: InterfaceType.OVS_INTERFACE,
```
EdDev: Type is not needed.

tests/integration/ovs_test.py
phoracek: Hoped to sneak it in... :crying_cat_face:

tests/integration/testlib/ovslib.py
phoracek: Cause verification should happen in the test, not during the preparation. No?

tests/integration/ovs_test.py
phoracek: Weird and ugly. Done

tests/integration/testlib/ovslib.py
phoracek: Done

tests/integration/testlib/ovslib.py
phoracek: Done

tests/integration/testlib/ovslib.py
phoracek: In my opinion, the verification should happen in the test, not here. Also, it would fail if the bridge is created only after this.

tests/integration/testlib/ovslib.py
phoracek: It is merely following the linux bridge helper. I will drop it, since it has no users here.

tests/integration/ovs_test.py
EdDev: You can use `TEST_BRIDGE1` or `BRIDGE1`, I do not mind. When you use it in multiple places, especially in a single test, it is preferred to use a constant in order to avoid mistakes.

tests/integration/ovs_test.py
EdDev: Because you re-use it all over in the test, and defining the name once is better over repeating. It is also clear that writer intended for the same thing and not picked a random name.

tests/integration/ovs_test.py
EdDev: That is correct, but at the same time also costly. If it would have been part of the development of the production code, I would even not notice it. But you do define both of them now, which is a bit wasteful, especially when we are talking about an integration test and not a unit test (which is super fast).

tests/integration/testlib/ovslib.py
EdDev: We want to check that the integral verification works well, so I see no reason not to check it.
The final assert may fail in cases this verification is faulty.

Not sure I follow your last sentence about this.

tests/integration/testlib/ovslib.py
EdDev: See the other response.

tests/integration/ovs_test.py
phoracek: Done

tests/integration/ovs_test.py
phoracek: Done

tests/integration/ovs_test.py
phoracek: Done

tests/integration/testlib/ovslib.py
phoracek: Done

tests/integration/testlib/ovslib.py
phoracek: Done

phoracek:
@EdDev please take another look.

tests/integration/ovs_test.py
```diff
@@ -18,103 +18,90 @@
...
-        ]
+        }
     }
-    libnmstate.apply(setup_remove_ovs_bridge_state_and_port)
-    state = statelib.show_only(
-        (state[INTERFACES][0]['name'], state[INTERFACES][1]['name'])
-    )
-    assert not state[INTERFACES]
+
+    with ovs_interface(PORT1, ipv4_state) as iface_state:
```
EdDev: It does not seem to make a lot of sense to create it in a separate transaction. Consider using the ‘create’ flag as in  https://github.com/nmstate/nmstate/blob/bb9b06287d1e5f04c74c892a1aea14afcc484ab3/tests/integration/testlib/bridgelib.py#L49

tests/integration/testlib/ovslib.py
```diff
@@ -0,0 +1,109 @@
...
+                Interface.TYPE: InterfaceType.OVS_BRIDGE,
+                Interface.STATE: InterfaceState.UP,
+                OVSBridge.CONFIG_SUBTREE: bridge_subtree_state or {},
+            }
+        ]
+    }
+    if extra_iface_state:
+        desired_state[Interface.KEY][0].update(extra_iface_state)
+
+    libnmstate.apply(desired_state, verify_change=True)
```
EdDev: verify change is the default, there is no need to be explicit. But if you like it more like that, I won’t mind.

phoracek: Doen

tests/integration/ovs_test.py
phoracek: Considered, but I prefer it this way. If I use `create=False` I need to add join the configuration and then apply it myself and it lowers readability of the test.

phoracek:
ci test please

tests/integration/ovs_test.py
EdDev: On the other hand, the test now is highly dependent on the NM provider.
This would have never worked with an ovsdb or ovs-vsctl provider.
An internal interface cannot exist without a bridge. I think it is wrong for NM to represent the interface as `UP` (active) without it being actually created.

phoracek: If you want me to change this to create=False, just say a word. I'd prefer not to theorize about corner-cases of a driver that may or may not appear in the future. Should I open a bug on NM/nmstate?

EdDev: I am asking to do this setup in one transaction, especially when they are so coupled anyway.
If you use `create` or something else, it is up to you.

And yes, I think you should do the minimum to be generic and not to depend on a specific implementation. Unless it is too complicated.

Regarding NM, I would check with the NM team before opening the BZ, but I tend to think it is a bug.

phoracek: Done

phoracek:
@cathay4t would you please take a look while @EdDev is gone?

tests/integration/testlib/ovslib.py
```diff
@@ -0,0 +1,118 @@
...
+        libnmstate.apply(
+            {
+                Interface.KEY: [
+                    {
+                        Interface.NAME: name,
+                        Interface.STATE: InterfaceState.ABSENT,
+                    }
+                ]
+            },
+            verify_change=False,
```
cathay4t: Like Edy, I also don't like `verify_change=False` here.  The `libnmstate` has different verification code comparing to the `assertlib` in test. It could be better we can compare both verification results.

You replied with `Done`  after https://github.com/nmstate/nmstate/pull/513#discussion_r331051989 , what does that means?

phoracek: IIUIC Edy wanted to verify it only on the configuration, not the cleanup (the same that is done in the linux bridge helper).

tests/integration/ovs_test.py
```diff
@@ -18,103 +18,92 @@
...
     }
-    libnmstate.apply(setup_remove_ovs_bridge_state_and_port)
-    state = statelib.show_only(
-        (state[INTERFACES][0]['name'], state[INTERFACES][1]['name'])
-    )
-    assert not state[INTERFACES]
+
+    with ovs_interface(PORT1, ipv4_state, create=False) as iface_state:
+        with ovs_bridge(
+            BRIDGE1, bridge, additional_ifaces=iface_state[Interface.KEY]
```
EdDev: `additional_ifaces` is very ugly. This context does more than it is supposed to (creating an ovs bridge).
It also confuses you and the reader... the returning state includes both the ovs bridge and these additional ifaces. On the following asserts you check it twice.
There is also a problem with the teardown which happens in two transactions (which is a waste).

I guess, for OVS, it fits better to create an internal interface as part of the bridge. Creating an internal interface can be done as stand-alone only by passing a bridge to it, and that will touch both the bridge representation and the internal interface.

I do not like it at all but I do not want to block it. I can refactor it later myself.

phoracek:
@EdDev please take another look, master, I tried my very best.

tests/integration/ovs_test.py
phoracek: Done

tests/integration/testlib/ovslib.py
```diff
@@ -0,0 +1,87 @@
...
+            }
+        )
+
+    def _add_port(self, name, _type):
+        self._bridge_iface[OVSBridge.CONFIG_SUBTREE].setdefault(
+            OVSBridge.PORT_SUBTREE, []
+        ).append({OVSBridge.PORT_NAME: name, OVSBridge.PORT_TYPE: _type})
+
+    @contextmanager
+    def created(self):
```
EdDev: Using verbs in a past tense is strange.
Why not `create`?

Another option would be to use `__enter__` and `__exit__` directly.
This will yield:
```
bond = Bond (..)
bond....
with bond as desired_state:
    ...
```
(but I leave it up to you, it is just a suggestion)

phoracek: >  Why not create?
Because in past tense it makes sense as a context "with created bridge, I want to do X and Y". But since Python uses it with present, I will change it.

EdDev: I never heard about a past sense unless it is a question. A context is usually an object, therefore the name should be a noun but in some cases a verb is also file (e.g. ‘open’).
‘enter/exit’ are separated from the class instantiation and fit well with the later creation.

phoracek: Sure, just wanted it to make sense when read. Unrelated to programming. Will change.

Ahaaa, sorry, got a little rusty with Python. Anyway this looks weird, that you treat a value (without `()`) as something that returns context.

```python
with bond as desired_state:
   ...
```

Also, I like to use `@contextmanager` instead of magical methods.

EdDev: I do not mind you using it, but that decorator is the magic not the actual implementation of a python context-manager. The decorator is creating a class that have these “magic” methods.
Anyway, it is not worth debating this, do what makes you comfortable with.

phoracek: Was refering to `__magic_methods__` in the code. Petr doesn't like how they look. That's all. Debate debate!

phoracek: Done :fireworks:

phoracek:
@EdDev done.

EdDev:
Thank you!

PR #514, opened 2019-09-26
add OVS bonding support

phoracek:
**Work in progress, don't bother reading the code**

This PR adds support for OVS bonding.

Note that removal of slaves is not supported yet.

https://bugzilla.redhat.com/show_bug.cgi?id=1730341

## Proposed API

```yaml
name: br1
type: ovs-bridge
state: up
bridge:
  port:
  - name: bond1
    link-aggregation:
      mode: balance-slb
      slaves:
      - name: eth0
      - name: eth1
```

Single-slave bond is indistinguishable from system interface port, therefore at least 2 slaves are mandatory. 

## Current state

When there is a port attached to a bridge with multiple underlying interfaces, only one of them is shown.

### Port with single iface

```yaml
- name: bridge0
  type: ovs-bridge
  state: up
  bridge:
    options:
      fail-mode: ''
      mcast-snooping-enable: false
      rstp: false
      stp: false
    port:
    - name: v1a
      type: system
```

```
ff3015ae-ba93-48c7-8a43-d4f99399841c
    Bridge "bridge0"
        Port "bond0"
            Interface "v1a"
                type: system
    ovs_version: "2.10.1"
```

### Port after attaching second iface

```yaml
- name: bridge0
  type: ovs-bridge
  state: up
  bridge:
    options:
      fail-mode: ''
      mcast-snooping-enable: false
      rstp: false
      stp: false
    port:
    - name: v2a
      type: system
```

```
ff3015ae-ba93-48c7-8a43-d4f99399841c
    Bridge "bridge0"
        Port "bond0"
            Interface "v2a"
                type: system
            Interface "v1a"
                type: system
    ovs_version: "2.10.1"
```

OVS NM tests refactoring: https://github.com/nmstate/nmstate/pull/524
OVS functional tests refactoring: https://github.com/nmstate/nmstate/pull/513

### DIFF
--- a/examples/ovsbridge_bond_create.yml
+++ b/examples/ovsbridge_bond_create.yml
@@ -1,0 +1,15 @@
+---
+interfaces:
+  - name: ovs-br0
+    type: ovs-bridge
+    state: up
+    bridge:
+      options:
+        stp: false
+      port:
+        - name: ovs-bond1
+          link-aggregation:
+            mode: balance-slb
+            slaves:
+              - name: eth1
+              - name: eth2
--- a/libnmstate/metadata.py
+++ b/libnmstate/metadata.py
@@ -62,7 +62,7 @@ def generate_ifaces_metadata(desired_state, current_state):
         desired_state.interfaces,
         current_state.interfaces,
         master_type="ovs-bridge",
-        get_slaves_func=_get_interfaces_from_ovs_bridge_state,
+        get_slaves_func=_get_ovs_interfaces_from_interface_state,
         set_metadata_func=_set_ovs_bridge_interfaces_metadata,
     )
     _generate_link_master_metadata(
@@ -101,54 +101,55 @@ def _set_ovs_bridge_interfaces_metadata(master_state, interface_state):
     OVS bridges.
     """
     _set_common_slaves_metadata(master_state, interface_state)
-    interface_state[BRPORT_OPTIONS] = _find_ovs_bridge_port_for_interface(
+    interface_state[BRPORT_OPTIONS] = _lookup_ovs_port_by_interface(
         master_state, interface_state[Interface.NAME],
     )
 
 
-def _find_ovs_bridge_port_for_interface(bridge_state, interface_name):
+def _lookup_ovs_port_by_interface(bridge_state, interface_name):
     """
-    Find bridge's port for given interface name. For simple ports connected
-    to a single system or internal interface, the interface name matches the
-    port name. When looking for a slave of a link aggregation port, we return
-    the whole parent port.
+    Return the OVS port state given an interface name.
+    For a link aggregation port, any of its slaves will match the port.
     """
     bridge_subtree = bridge_state.get(OVSBridge.CONFIG_SUBTREE, {})
-    ports = bridge_subtree.get(OVSBridge.PORT_SUBTREE, [])
+    ports = bridge_subtree.get(OVSBridge.PORT_SUBTREE, ())
     for port in ports:
-        if OVSBridge.Port.LINK_AGGREGATION_SUBTREE in port:
-            la_subtree = port[OVSBridge.Port.LINK_AGGREGATION_SUBTREE]
-            slaves = la_subtree.get(
-                OVSBridge.Port.LinkAggregation.SLAVES_SUBTREE, []
-            )
-            for slave in slaves:
-                if (
-                    slave[OVSBridge.Port.LinkAggregation.Slave.NAME]
-                    == interface_name
-                ):
-                    return port
-        else:
-            if port[OVSBridge.Port.NAME] == interface_name:
-                return port
+        if _is_iface_a_slave_of(interface_name, port):
+            return port
+        if port[OVSBridge.Port.NAME] == interface_name:
+            return port
     return None
 
 
-def _get_interfaces_from_ovs_bridge_state(bridge_state, default=()):
+def _is_iface_a_slave_of(interface_name, port):
+    la_subtree = port.get(OVSBridge.Port.LINK_AGGREGATION_SUBTREE)
+    if la_subtree:
+        slaves = la_subtree.get(
+            OVSBridge.Port.LinkAggregation.SLAVES_SUBTREE, ()
+        )
+        for slave in slaves:
+            if (
+                slave[OVSBridge.Port.LinkAggregation.Slave.NAME]
+                == interface_name
+            ):
+                return True
+    return False
+
+
+def _get_ovs_interfaces_from_interface_state(interface_state, default=()):
     """
-    Return names of all the interfaces connected to ports of given OVS bridge.
-    For simple ports connected to internal and system interfaces, the interface
-    name matches the port name. For link aggregation ports, the interface names
-    are obtained from listed slaves.
+    Return all interfaces connected to the given bridge, including
+    link-aggregation slaves.
     """
     ovs_slaves = []
 
-    bridge_subtree = bridge_state.get(OVSBridge.CONFIG_SUBTREE, {})
-    ports = bridge_subtree.get(OVSBridge.PORT_SUBTREE, [])
+    bridge_subtree = interface_state.get(OVSBridge.CONFIG_SUBTREE, {})
+    ports = bridge_subtree.get(OVSBridge.PORT_SUBTREE, ())
     for port in ports:
         if OVSBridge.Port.LINK_AGGREGATION_SUBTREE in port:
             la_subtree = port[OVSBridge.Port.LINK_AGGREGATION_SUBTREE]
             slaves = la_subtree.get(
-                OVSBridge.Port.LinkAggregation.SLAVES_SUBTREE, []
+                OVSBridge.Port.LinkAggregation.SLAVES_SUBTREE, ()
             )
             for slave in slaves:
                 ovs_slaves.append(
--- a/libnmstate/metadata.py
+++ b/libnmstate/metadata.py
@@ -26,6 +26,7 @@ from libnmstate.schema import Interface
 from libnmstate.schema import InterfaceIP
 from libnmstate.schema import InterfaceState
 from libnmstate.schema import InterfaceType
+from libnmstate.schema import OVSBridge
 
 
 BRPORT_OPTIONS = "_brport_options"
@@ -61,8 +62,8 @@ def generate_ifaces_metadata(desired_state, current_state):
         desired_state.interfaces,
         current_state.interfaces,
         master_type="ovs-bridge",
-        get_slaves_func=_get_ovs_slaves_from_state,
-        set_metadata_func=_set_ovs_bridge_ports_metadata,
+        get_slaves_func=_get_interfaces_from_ovs_bridge_state,
+        set_metadata_func=_set_ovs_bridge_interfaces_metadata,
     )
     _generate_link_master_metadata(
         desired_state.interfaces,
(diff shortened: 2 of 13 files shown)


phoracek:
@EdDev @cathay4t @maiqueb could you please check posted examples and tell me whether it is a good direction? Thanks.

EdDev:
The schema looks fine to me, except of one point: Do we really need the slaves to have a type?
Is there a scenario where the slaves are not system type?
Are there any other properties that are interface/slave specific?

libnmstate/nm/applier.py
```diff
@@ -262,19 +262,24 @@ def _get_ovs_bridge_port_devices(iface_state):
...
-            OvsB.PORT_SUBTREE, []
-        )
-    ]
-    ports = [ovs.PORT_PROFILE_PREFIX + iface for iface in ifaces]
+    ports = []
+    ifaces = []
+    for port in iface_state.get('bridge', {}).get('port', []):
+        ports.append(ovs.PORT_PROFILE_PREFIX + port['name'])
+        if port['type'] == 'bond':
+            for slave in port.get('link-aggregation', {}).get('slaves', []):
```
oshoval: in case port.get('link-aggregation', {}) returns {} 
(if i understand right, it will return that in case it didnt find anything)
would the get('slaves', []) operates on {} and will assert?

phoracek:
@EdDev internal device types cover:
- system
- internal
- tap
- geneve
- gre
- ipsec_gre
- gre64
- ipsec_gre64
- vxlan
- lisp
- patch

I think that at least VXLAN and patch would make sense as bond slaves.

libnmstate/nm/applier.py
phoracek: If it returns `{}` we call the second `.get('slaves', [])`, where `[]` is the default (in case the key `'slaves'` is not found).

What do you mean by "assert"?
```python
>>> bridge = {}
>>> bridge.get('link-aggregation', {})
{}
>>> bridge.get('link-aggregation', {}).get('slaves', [])
[]
```

oshoval: throwing exception,

ok but it doesn't throw exception as you shown
thanks

EdDev:
@phoracek , looks like `internal` is only one of those. These are not a subset of the `internal` one.

It is unclear to me which of these are just managed through OVS but created in the kernel and which are just logical entities of OVS itself (like the `patch` one).
The ones like `internal` and `patch` are simpler to manage in this context, but the ones that have both a kernel representation and the ability to manage through OVS is less trivial (how will they appear in the interfaces state?).

I am now starting to question if it was correct to set a port type parameter at all. We could have use a heuristic to lookup the correct type from the system (lookup for an existing interface with the same name, if it exists, it is a system one, otherwise, look for the desired state for the type, if not existing, use `internal` as the default). Not sure if it is better or not, perhaps you can suggest what seems easier to you and others.

EdDev:
I guess there is also an option to set the bond as an aggregation of ports, this way:
```YAML
name: br1
type: ovs-bridge
state: up
bridge:
  port:
  - name: eth0
    type: system
  - name: eth1
    type: system
  - name: bond1
    type: bond
    link-aggregation:
      mode: balance-slb
      slaves: [eth0, eth1]
```

This way we can extend it nicely on the port level or as a stand-alone interface and not define the schema of a slave as a duplicate of the port one.

phoracek:
@EdDev 

> I guess there is also an option to set the bond as an aggregation of ports

I would be against this option, there are several reasons:
- Linking between items on the same level is confusing and requires more validation
- IMO it would be harder for user which is used to regular bonding to grasp this
- Most importantly, the list of ports is really a list of OVS ports (or abstraction of switch ports in general). For instance, it will be used to configure VLAN. If we keep interface-only ports here (slaves of a bond), we would need to introduce special logic preventing users from using some of port options there.

The code change to support link-aggregation is not that drastic IMO. This PR should be reviewable soon.

---

> It is unclear to me which of these are just managed through OVS but created in the kernel and which are just logical entities of OVS itself (like the patch one).

Yeah, I don't think it would be a good way for us to care about the nature of those interface types.

> The ones like internal and patch are simpler to manage in this context, but the ones that have both a kernel representation and the ability to manage through OVS is less trivial (how will they appear in the interfaces state?).

Understood.

> I am now starting to question if it was correct to set a port type parameter at all. We could have use a heuristic to lookup the correct type from the system (lookup for an existing interface with the same name, if it exists, it is a system one, otherwise, look for the desired state for the type, if not existing, use internal as the default). Not sure if it is better or not, perhaps you can suggest what seems easier to you and others.

This PR is heavily reusing existing code handling ports to handle slaves (since they are sharing a lot of logic). Can we stick to that and consider removing the type in a follow-up? It would be dropped from both ports and slaves. It you agree, I will open a Jira task to track that.

phoracek:
@EdDev @qinqon @oshoval  the PR passed a manual test and successfully created an OVS bonding. Could you please take a look? I will work on test coverage in the meantime.

maiqueb:
I think the unit tests on this PR would pass if you rebased #519 .

examples/ovsbridge_bond_create.yml
```diff
@@ -0,0 +1,18 @@
...
+      options:
+        stp: false
+      port:
+        - name: ovs-bond1
+          type: bond
+          link-aggregation:
+            mode: balance-slb
+            slaves:
+              - name: eth1
+                type: system
```
qinqon: Do we need type here ?

libnmstate/metadata.py
```diff
@@ -92,12 +92,23 @@ def _get_bond_slaves_from_state(iface_state, default=()):
...
-    port = next(
-        six.moves.filter(lambda n: n['name'] == slave_state['name'], ports), {}
+    slave_state[BRPORT_OPTIONS] = _find_ovs_bridge_port(
+        master_state, slave_state['name']
     )
-    slave_state[BRPORT_OPTIONS] = port
+
+
+def _find_ovs_bridge_port(bridge_state, port_name):
+    for port_state in bridge_state.get('bridge', {}).get('port', []):
```
qinqon: Use schema constants for "bridge" and "port"

libnmstate/metadata.py
```diff
@@ -92,12 +92,23 @@ def _get_bond_slaves_from_state(iface_state, default=()):
...
+        if port_state['type'] == 'bond':
```
qinqon: Use schema constants for "bond" type

libnmstate/metadata.py
```diff
@@ -92,12 +92,23 @@ def _get_bond_slaves_from_state(iface_state, default=()):
...
+            for slave in port_state.get('link-aggregation', {}).get(
```
qinqon: Use schema constants for "link-aggregation"

examples/ovsbridge_bond_create.yml
phoracek: We discussed it with Edy aboved. I want to keep it identical to regular ports. Then we can see if we can get a rid of "type" from both slaves and ports.

libnmstate/metadata.py
phoracek: Petr hates constants, but will do.

examples/ovsbridge_bond_create.yml
qinqon: ack

cathay4t:
@phoracek  Does OVS plan to support other port type other than `system` for bond?

If not, we may just use the schema of Bond:

```
      slaves:
      - eth0
      - eth1
```

phoracek:
@cathay4t please see https://github.com/nmstate/nmstate/pull/514#issuecomment-537436369.

EdDev:
> This PR is heavily reusing existing code handling ports to handle slaves (since they are sharing a lot of logic). Can we stick to that and consider removing the type in a follow-up? It would be dropped from both ports and slaves.

This is an api change, if we decide to embrace the idea, it better be done sooner than latter.
On the other hand, you can make this fully working with tests and all, giving the base for the change. But consider that the current api will not be supported.

> It you agree, I will open a Jira task to track that.

Yes, please open the request.
Please raise this with @cathay4t and put it on writing in the nmstate-devel mailing list.

When I think about it now, I find it irrational that we need to define the type twice. It also sounds not that complex to lookup for the type from the current and desired states.

phoracek:
@EdDev 
The story: https://nmstate.atlassian.net/browse/NMSTATE-248
Draft PR: https://github.com/nmstate/nmstate/pull/525

libnmstate/schema.py
```diff
@@ -205,7 +205,23 @@ class OVSBridge(object):
     PORT_VLAN_MODE = 'vlan-mode'
     PORT_ACCESS_TAG = 'access-tag'
 
+    LA_SUBTREE = 'link-aggregation'
+    LA_MODE = 'mode'
```
tyll: This should probably be moved into its own class:

```
class LinkAggregation:
    MODE = 'mode'
...
```
See #608

libnmstate/schema.py
```diff
@@ -205,7 +205,23 @@ class OVSBridge(object):
     PORT_VLAN_MODE = 'vlan-mode'
     PORT_ACCESS_TAG = 'access-tag'
 
+    LA_SUBTREE = 'link-aggregation'
```
tyll: `LA` is not a good abbreviation IMHO. Using the full name would match the other constants better.

phoracek:
There are some flaws, but the main stuff is in place and reviewable.

libnmstate/state.py
```diff
@@ -32,6 +32,8 @@
 from ipaddress import ip_address
 from operator import itemgetter
 
+import six
```
qinqon: I think nmstate no longer support python2 we can use proper python3.

libnmstate/state.py
```diff
@@ -575,6 +579,17 @@ def _sort_lag_slaves(self):
         for ifstate in self.interfaces.values():
             ifstate.get("link-aggregation", {}).get("slaves", []).sort()
 
+    def _sort_ovs_lag_slaves(self):
+        for ifstate in six.viewvalues(self.interfaces):
```
qinqon: ditto

tests/integration/ovs_test.py
```diff
@@ -141,3 +142,86 @@ def disable_nm_ovs_plugin():
     finally:
         libcmd.exec_cmd(DNF_INSTALL_NM_OVS_CMD)
         libcmd.exec_cmd(SYSTEMCTL_RESTART_NM_CMD)
+
+
+@pytest.mark.xfail(
+    raises=(NmstateLibnmError, AssertionError),
```
EdDev: When you see `AssertionError` failing? On which assertion?

tests/integration/ovs_test.py
```diff
@@ -141,3 +142,86 @@ def disable_nm_ovs_plugin():
...
+):
+    port0_name = port0_up[Interface.KEY][0][Interface.NAME]
+    port1_name = port1_up[Interface.KEY][0][Interface.NAME]
+    port2_name = port2_up[Interface.KEY][0][Interface.NAME]
+
+    bridge = Bridge(BRIDGE1)
+    bridge.add_link_aggregation_port(BOND1, [port0_name, port1_name])
+
+    with bridge.create() as state:
+        # Plug in additional slave
```
EdDev: Redundant.

tests/integration/ovs_test.py
```diff
@@ -141,3 +142,86 @@ def disable_nm_ovs_plugin():
...
+
+    assertlib.assert_absent(BRIDGE1)
+    assertlib.assert_absent(BOND1)
+
+
+@pytest.mark.xfail(
+    raises=(NmstateLibnmError, AssertionError),
+    reason="https://bugzilla.redhat.com/1724901",
+)
+def test_ovs_bridge_link_aggregation_and_specific_mode(port0_up, port1_up):
```
EdDev: Seems wasteful, can you just use the mode in the  the first bond test? The integration tests are costly.

tests/integration/ovs_test.py
```diff
@@ -141,3 +142,86 @@ def disable_nm_ovs_plugin():
...
+    )
+
+    with bridge.create() as state:
+        assertlib.assert_state_match(state)
+
+    assertlib.assert_absent(BRIDGE1)
+    assertlib.assert_absent(BOND1)
+
+
+def test_ovs_bridge_link_aggregation_and_invalid_mode(port0_up, port1_up):
```
EdDev: `and` ==> `with_an`

tests/integration/testlib/ovslib.py
```diff
@@ -84,6 +87,18 @@ def _get_port(self, name):
             (port for port in ports if port[OVSBridge.Port.NAME] == name), None
         )
 
+    def del_port(self, name):
+        new_ports = [
```
EdDev: Can you use here `filter`?

Also, for readability (reduce the line breaking), could you please place `self._bridge_iface[OVSBridge.CONFIG_SUBTREE][OVSBridge.PORT_SUBTREE]` into a temp variable?

tests/lib/metadata_test.py
```diff
@@ -609,6 +609,122 @@ def test_ovs_reusing_slave_used_by_existing_bridge(self):
         assert desired_state == expected_dstate
         assert current_state == expected_cstate
 
+    def test_ovs_creation_with_new_link_aggregation_port(self):
+        desired_state = state.State(
+            {
+                Interface.KEY: [
+                    {
+                        "name": OVS_NAME,
```
EdDev: Please use constants for the keys.

tests/lib/metadata_test.py
```diff
@@ -609,6 +609,122 @@ def test_ovs_reusing_slave_used_by_existing_bridge(self):
...
+            expected_dstate.interfaces[iface][
+                metadata.BRPORT_OPTIONS
+            ] = desired_state.interfaces[OVS_NAME]["bridge"]["port"][0]
+
+        metadata.generate_ifaces_metadata(desired_state, current_state)
+
+        assert desired_state == expected_dstate
+        assert current_state == expected_cstate
+
+    def test_ovs_adding_slaves(self):
```
EdDev: ..`link_aggregation_slave`? Note that the test adds a single slave.

tests/lib/metadata_test.py
```diff
@@ -609,6 +609,122 @@ def test_ovs_reusing_slave_used_by_existing_bridge(self):
...
+            ] = TYPE_OVS_BR
+            expected_dstate.interfaces[iface][
+                metadata.BRPORT_OPTIONS
+            ] = desired_state.interfaces[OVS_NAME]["bridge"]["port"][0]
+
+        metadata.generate_ifaces_metadata(desired_state, current_state)
+
+        assert desired_state == expected_dstate
+        assert current_state == expected_cstate
+
```
EdDev: What about these cases:
- `test_ovs_removing_link_aggregation_slave`
- `test_ovs_removing_link_aggregation_port`

tests/lib/nm/ovs_test.py
```diff
@@ -122,8 +122,6 @@ def test_get_ovs_info_with_ports_with_interfaces(
 
     assert len(info["port"]) == 1
     assert "name" in info["port"][0]
-    assert "vlan-mode" in info["port"][0]
```
EdDev: These removals fit a separate commit/PR. I think you cleaned the implementation as well.

EdDev:
Replaced by #795

PR #515, opened 2019-09-27
Vxlan implementation

qinqon:


libnmstate/nm/vxlan.py
```diff
@@ -0,0 +1,63 @@
...
+    vxlan_destination_port = vxlan[VXLAN.DESTINATION_PORT]
+
+    vxlan_setting = None
+    if base_con_profile:
+        vxlan_setting = base_con_profile.get_setting_vxlan()
+        if vxlan_setting:
+            vxlan_setting = vxlan_setting.duplicate()
+
+    if not vxlan_setting:
+        vxlan_setting = nmclient.NM.SettingVlan.new()
```
maiqueb: setting vlan ?

qinqon: thanks

qinqon:
NM has the `remote` [property](https://valadoc.org/libnm/NM.SettingVxlan.html), don't know what's the issue.

```
autoconnect_slaves: <enum NM_SETTING_CONNECTION_AUTOCONNECT_SLAVES_YES of type NM.SettingConnectionAutoconnectSlaves>
2019-09-30 13:29:52 netapplier.py:228 DEBUG Editing interfaces: []
2019-09-30 13:29:52 nmclient.py:98 DEBUG Executing NM action: func=add_connection_async
2019-09-30 13:29:52 nmclient.py:161 ERROR NM main-loop aborted: Connection adding failed: error=nm-connection-error-quark: vxlan.remote: property is missing (6)
2019-09-30 13:29:52 checkpoint.py:146 DEBUG Checkpoint /org/freedesktop/NetworkManager/Checkpoint/4 rollback executed: dbus.Dictionary({dbus.String('/org/freedesktop/NetworkManager/Devices/2'): dbus.UInt32(0), dbus.String('/org/freedesktop/NetworkManager/Devices/4'): dbus.UInt32(0), dbus.String('/org/freedesktop/NetworkManager/Devices/1'): dbus.UInt32(0), dbus.String('/org/freedesktop/NetworkManager/Devices/3'): dbus.UInt32(0)}, signature=dbus.Signature('su'))
----------------------------------------------------------------------------------------------------------- Captured log teardown -----------------------------------------------------------------------------------------------------------
2019
```

maiqueb:
> NM has the `remote` [property](https://valadoc.org/libnm/NM.SettingVxlan.html), don't know what's the issue. @cathay4t @EdDev do you know why ?
> 
> ```
> autoconnect_slaves: <enum NM_SETTING_CONNECTION_AUTOCONNECT_SLAVES_YES of type NM.SettingConnectionAutoconnectSlaves>
> 2019-09-30 13:29:52 netapplier.py:228 DEBUG Editing interfaces: []
> 2019-09-30 13:29:52 nmclient.py:98 DEBUG Executing NM action: func=add_connection_async
> 2019-09-30 13:29:52 nmclient.py:161 ERROR NM main-loop aborted: Connection adding failed: error=nm-connection-error-quark: vxlan.remote: property is missing (6)
> 2019-09-30 13:29:52 checkpoint.py:146 DEBUG Checkpoint /org/freedesktop/NetworkManager/Checkpoint/4 rollback executed: dbus.Dictionary({dbus.String('/org/freedesktop/NetworkManager/Devices/2'): dbus.UInt32(0), dbus.String('/org/freedesktop/NetworkManager/Devices/4'): dbus.UInt32(0), dbus.String('/org/freedesktop/NetworkManager/Devices/1'): dbus.UInt32(0), dbus.String('/org/freedesktop/NetworkManager/Devices/3'): dbus.UInt32(0)}, signature=dbus.Signature('su'))
> ----------------------------------------------------------------------------------------------------------- Captured log teardown -----------------------------------------------------------------------------------------------------------
> 2019
> ```

I can't be sure, since you don't say what you did that triggered that, but I think that in https://github.com/nmstate/nmstate/pull/515/files#diff-2808139b5c039d9782e3f7fb3ab11d39R56 you're handling the 'remote' property as mandatory, while I think it is optional.

Let me know how it went.

libnmstate/schemas/operational-state.yaml
```diff
@@ -177,6 +178,33 @@ definitions:
...
+          properties:
+            id:
+              type: integer
+              minimum: 0
+              maximum: 16777216
+            remote:
+              type: string
+            destination-port:
+              type: integer
+            parent:
```
cathay4t: Please use `base-iface` as `VLAN` did.

libnmstate/schemas/operational-state.yaml
```diff
@@ -177,6 +178,33 @@ definitions:
...
+          type: string
+          enum:
+            - vxlan
+        vxlan:
+          type: object
+          properties:
+            id:
+              type: integer
+              minimum: 0
+              maximum: 16777216
```
cathay4t: The `maximum` here means inclusive maximum, so the number should be `16777215`

libnmstate/schemas/operational-state.yaml
```diff
@@ -177,6 +178,33 @@ definitions:
...
+            remote:
+              type: string
+            destination-port:
+              type: integer
+            parent:
+              type: string
+          required:
+            - id
+            - parent
+            - remote
```
cathay4t: `remote` should be optional.

tests/integration/testlib/vxlan.py
```diff
@@ -0,0 +1,44 @@
...
+
+@contextmanager
+def vxlan_interface(ifname, vxlan_id, parent):
+    desired_state = {
+        INTERFACES: [
+            {
+                'name': ifname,
+                'type': 'vxlan',
+                'state': 'up',
+                'vxlan': {'id': vxlan_id, 'parent': parent, 'remote': '192.168.100.1'},
```
cathay4t: Please use constants from schema.

tests/integration/testlib/vxlan.py
```diff
@@ -0,0 +1,44 @@
...
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+
+from contextlib import contextmanager
+
+import libnmstate
+from .statelib import INTERFACES
```
cathay4t: `INTERFACES` is deprecated(undocumented).

Please use `schema.Interface.KEY` instead.

tests/integration/vxlan_test.py
```diff
@@ -0,0 +1,165 @@
...
+from .testlib.assertlib import assert_mac_address
+from .testlib.vxlan import vxlan_interface
+
+VXLAN_IFNAME = 'eth1.101'
+VXLAN2_IFNAME = 'eth1.102'
+
+TWO_VXLANS_STATE = {
+    INTERFACES: [
+        {
+            'name': VXLAN_IFNAME,
```
cathay4t: Please use `Interface.NAME` instead of `name`.

tests/integration/vxlan_test.py
```diff
@@ -0,0 +1,165 @@
...
+            'state': 'up',
+            VXLAN.CONFIG_SUBTREE: {VXLAN.ID: 102, VXLAN.PARENT: 'eth1', VXLAN.REMOTE: '192.168.100.2'},
+        },
+    ]
+}
+
+
+def test_add_and_remove_vxlan(eth1_up):
+    with vxlan_interface(
+        VXLAN_IFNAME, 101, eth1_up[Interface.KEY][0][Interface.NAME]
```
cathay4t: As `101` been used multiple times, please use constant.

tests/integration/testlib/vxlan.py
```diff
@@ -0,0 +1,44 @@
...
+            }
+        ]
+    }
+    libnmstate.apply(desired_state)
+    try:
+        yield desired_state
+    finally:
+        libnmstate.apply(
+            {INTERFACES: [{'name': ifname, 'type': 'vxlan', 'state': 'absent'}]}
+        )
```
cathay4t: `verify_change=False` is suggested when cleaning up.

tests/integration/vxlan_test.py
```diff
@@ -0,0 +1,165 @@
...
+        yield desired_state
+    finally:
+        libnmstate.apply(
+            {
+                INTERFACES: [
+                    {'name': VXLAN_IFNAME, 'type': 'vxlan', 'state': 'absent'},
+                    {'name': VXLAN2_IFNAME, 'type': 'vxlan', 'state': 'absent'},
+                ]
+            }
+        )
```
cathay4t: `verify_change=False` on clean up.

libnmstate/schemas/operational-state.yaml
qinqon: libnm fails if we don't pass it, looks like is mandatory at NetworkManager

qinqon:
@EdDev this one is ready too.

tests/integration/vxlan_test.py
```diff
@@ -0,0 +1,179 @@
...
+from .testlib import assertlib
+from .testlib import statelib
+from .testlib.vxlan import vxlan_interface
+
+VXLAN1_ID = 101
+VXLAN2_ID = 102
+VXLAN1_IFNAME = 'eth1.{}'.format(VXLAN1_ID)
+VXLAN2_IFNAME = 'eth1.{}'.format(VXLAN2_ID)
+
+TWO_VXLANS_STATE = {
```
EdDev: This looks like a type of duplication with the `vxlan_interface` helper.

tests/integration/vxlan_test.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def test_add_and_remove_vxlan(eth1_up):
+    with vxlan_interface(
+        VXLAN1_IFNAME, VXLAN1_ID, eth1_up[Interface.KEY][0][Interface.NAME]
+    ) as desired_state:
+        assertlib.assert_state(desired_state)
+
+    current_state = statelib.show_only((VXLAN1_IFNAME,))
+    assert not current_state[Interface.KEY]
```
EdDev: Please use `assertlib.assert_absent`

tests/integration/vxlan_test.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+@pytest.fixture
+def vxlan_on_eth1(eth1_up):
```
EdDev: nit: The convention in our tests is to place the fixtures at the top, before the tests.

It seems no one uses it.

But if you intend to use it, please do not specify the name of the base in the fixture name... we can abstract it by not mentioning it at all. Calling it `vxlan0_up` should be enough.

tests/integration/vxlan_test.py
```diff
@@ -0,0 +1,179 @@
...
+    try:
+        libnmstate.apply(desired_state)
+    finally:
+        for ifstate in desired_state[Interface.KEY]:
+            ifstate[Interface.STATE] = InterfaceState.ABSENT
+        libnmstate.apply(desired_state, verify_change=False)
+
+
+@contextmanager
+def two_vxlans_on_eth1():
```
EdDev: This smells duplication with the `vxlan_interface` helper.
You can create to states without `apply` and then simply merge them. OR you can adjust the helper to create multiple vxlan ifaces on the same base. The first sounds nicer to me, especially when the later is used only once.

libnmstate/nm/vxlan.py
```diff
@@ -0,0 +1,71 @@
...
+    return {}
+
+
+def get_info(device):
+    """
+    Provides the current active values for a device
+    """
+    info = {}
+    if device.get_device_type() == nmclient.NM.DeviceType.VXLAN:
+        settings = get_settings(device)
```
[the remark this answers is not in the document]
qinqon: Reading destination port from kernel is not working, although nm show it alrigth.

I am using `device.get_dst_port` and always return 0

tests/integration/testlib/vxlan.py
```diff
@@ -0,0 +1,84 @@
...
+
+from contextlib import contextmanager
+
+import libnmstate
+
+from libnmstate.schema import VXLAN
+from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceState
+
+VXLAN_IFNAME = '{}.{}'
```
EdDev: It is unclear why you need this one. It is used once and has no special value as far as I can understand.
(BTW, it is a name template, the current name is misleading)

tests/integration/testlib/vxlan.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def vxlan_desired_state(vxlans):
+    vxlans_up = []
+    for vxlan in vxlans:
+        vxlans_up.append(vxlan_up(*vxlan))
```
EdDev: Fits a list comprehension.

tests/integration/testlib/vxlan.py
```diff
@@ -0,0 +1,84 @@
...
+import libnmstate
+
+from libnmstate.schema import VXLAN
+from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceState
+
+VXLAN_IFNAME = '{}.{}'
+
+
+def vxlan_desired_state(vxlans):
```
EdDev: The name does not say what this one does and the argument is not self explanatory.
I suggest for the name to add a verb and for the argument to extract the tuple to multiple arguments.

tests/integration/testlib/vxlan.py
```diff
@@ -0,0 +1,84 @@
...
+    vxlans_up = []
+    for vxlan in vxlans:
+        vxlans_up.append(vxlan_up(*vxlan))
+    return {Interface.KEY: vxlans_up}
+
+
+@contextmanager
+def vxlan_interfaces(vxlans, extra_interfaces=[]):
```
EdDev: In python you cannot use mutable default values.
But it is also not that nice to ask this function to control other interfaces (extra ones).

Try to reach a design where each function/contextmanager does one thing and then just combine them together here or in the test that calls it to combine them.
For example, helpers create the state, then the test merges them and apply it. we do things like that with other tests.

tests/integration/testlib/vxlan.py
```diff
@@ -0,0 +1,84 @@
...
+def vxlan_absent(vlan_id, base_if):
+    vxlan_name = vxlan_ifname(vlan_id, base_if)
+    return {Interface.NAME: vxlan_name, Interface.STATE: InterfaceState.ABSENT}
+
+
+def vxlan_down(vlan_id, base_if):
+    vxlan_name = vxlan_ifname(vlan_id, base_if)
+    return {
+        Interface.NAME: vxlan_name,
+        Interface.TYPE: VXLAN.TYPE,
```
EdDev: Editing an iface does not require the `type` field.

tests/integration/nm/vxlan_test.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def _delete_vxlan(devname):
+    nmdev = nm.device.get_device_by_name(devname)
+    with mainloop():
+        nm.device.deactivate(nmdev)
+        nm.device.delete(nmdev)
+
+
+def _get_vxlan_ifname(state):
```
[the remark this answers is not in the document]
qinqon: I am removing the 'get' preffix but keeping the function since we don't want to harcode the "eth1" interface name we get from the fixture and we are using that name as suffix of vxlan interface name.

tests/integration/testlib/vxlan.py
qinqon: It fails if I don't add type here.

qinqon:
@EdDev is there anything else missing here ?

tests/integration/nm/vxlan_test.py
```diff
@@ -0,0 +1,128 @@
...
+            VXLAN.ID: 101,
+            VXLAN.BASE_IFACE: ifname,
+            VXLAN.REMOTE: '192.168.1.18',
+            VXLAN.DESTINATION_PORT: 4789,
+        }
+    }
+
+    with _vxlan_interface(vxlan_desired_state):
+        vxlan_ifname = _vxlan_ifname(vxlan_desired_state)
+        vxlan_current_state = _get_vxlan_current_state(vxlan_ifname)
```
EdDev: How about returning this from `_vxlan_interface`?
`with _vxlan_interface(...) as vxlan_current_state:` fits nicely.

tests/integration/nm/vxlan_test.py
```diff
@@ -0,0 +1,128 @@
...
+    vxlan_desired_state = {
+        VXLAN.CONFIG_SUBTREE: {
+            VXLAN.ID: 101,
+            VXLAN.BASE_IFACE: ifname,
+            VXLAN.REMOTE: '192.168.1.18',
+            VXLAN.DESTINATION_PORT: 4789,
+        }
+    }
+
+    with _vxlan_interface(vxlan_desired_state):
```
EdDev: In this test you can use a fixture to prepare the vxlan interface and then just show the problem.
It fits well as a fixture because the creation of the vxlan interface is not part of the test, but a given.. the test just shows that we can read (or not) the specific property.
BTW, failing on the fixture or in the test results in different reports (one is an error and the later a failure).

tests/integration/nm/vxlan_test.py
```diff
@@ -0,0 +1,128 @@
...
+        vxlan_ifname = _vxlan_ifname(vxlan_desired_state)
+        obtained_destination_port = _get_vxlan_device(
+            vxlan_ifname
+        ).props.dst_port
```
EdDev: nit: It will read nicer if you split these (get the device, then query it).
Not a must.

tests/integration/testlib/vxlan.py
```diff
@@ -0,0 +1,102 @@
...
+
+from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceState
+from libnmstate.schema import VXLAN
+
+
+def vxlans_up(vxlans):
+    vxlans_state = []
+    for vxlan in vxlans:
+        vxlans_state.append(vxlan_up(*vxlan))
```
EdDev: This fits a list comprehension: `vxlans_state = [vxlan_up(*vxlan) for vxlan in vxlans]`

tests/integration/testlib/vxlan.py
```diff
@@ -0,0 +1,102 @@
...
+from contextlib import contextmanager
+
+import libnmstate
+
+from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceState
+from libnmstate.schema import VXLAN
+
+
+def vxlans_up(vxlans):
```
EdDev: Please order the functions/methods such that the caller appears before the callee. It makes it easier for me to read from top to bottom, from abstract to details.

tests/integration/testlib/vxlan.py
```diff
@@ -0,0 +1,102 @@
...
+    vxlans_state = []
+    for vxlan in vxlans:
+        vxlans_state.append(vxlan_absent(*vxlan))
+    return {Interface.KEY: vxlans_state}
+
+
+@contextmanager
+def apply_state(setup_state, teardown_state):
+    try:
+        libnmstate.apply(setup_state)
```
EdDev: I think this should be outside the `try` block. If `apply` fails, it should clean-up itself.

tests/integration/testlib/vxlan.py
```diff
@@ -0,0 +1,102 @@
...
+    finally:
+        libnmstate.apply(teardown_state, verify_change=False)
+
+
+@contextmanager
+def vxlan_interfaces(vxlans):
+    vxlans_interfaces_absent = []
+    try:
+        setup_state = vxlans_up(vxlans)
+        libnmstate.apply(setup_state)
```
EdDev: Same here, these should probably be outside the `try` block.

tests/integration/testlib/vxlan.py
```diff
@@ -0,0 +1,102 @@
...
+    try:
+        libnmstate.apply(setup_state)
+        yield setup_state
+    finally:
+        libnmstate.apply(teardown_state, verify_change=False)
+
+
+@contextmanager
+def vxlan_interfaces(vxlans):
+    vxlans_interfaces_absent = []
```
EdDev: Please define it near its usage.

tests/integration/testlib/vxlan.py
```diff
@@ -0,0 +1,102 @@
...
+        setup_state = vxlans_up(vxlans)
+        libnmstate.apply(setup_state)
+        yield setup_state
+    finally:
+
+        # We need a little adaptation to remove remote from the tuple
+        # since it's not needed for absent
+        for vxlan in vxlans:
+            vxlan_id, parent, _ = vxlan
+            vxlans_interfaces_absent.append((vxlan_id, parent))
```
EdDev: Fits a list comprehension.

tests/integration/testlib/vxlan.py
```diff
@@ -0,0 +1,102 @@
...
+
+        libnmstate.apply(
+            vxlans_absent(vxlans_interfaces_absent), verify_change=False
+        )
+
+
+def vxlan_up(vxlan_id, base_if, remote):
```
EdDev: How about `create_vxlan_state`?

tests/integration/testlib/vxlan.py
EdDev: You are not supposed to need it once it exists.
Can you check this also for `absent`?

tests/integration/testlib/vxlan.py
```diff
@@ -0,0 +1,102 @@
...
+def vxlan_down(vlan_id, base_if):
+    vxlan_name = vxlan_ifname(vlan_id, base_if)
+    return {
+        Interface.NAME: vxlan_name,
+        Interface.TYPE: VXLAN.TYPE,
+        Interface.STATE: InterfaceState.DOWN,
+    }
+
+
+def vxlan_ifname(vxlan_id, base_if):
```
EdDev: By itself this function works for me, but it seems overused in the above code.
I would expect the name to be passed once on creation and then either re-passed to the relevant functions or parsed from the data itself. Recreating the name make the functions that uses it dependent in this module, we will not be able to reuse/copy it easily (it is more like a smell than a real problem).

tests/integration/vxlan_test.py
```diff
@@ -0,0 +1,118 @@
...
+from libnmstate.schema import InterfaceState
+from libnmstate.schema import InterfaceType
+
+from .testlib import assertlib
+from .testlib.vxlan import apply_state
+from .testlib.vxlan import vxlan_down
+from .testlib.vxlan import vxlan_interfaces
+from .testlib.vxlan import vxlan_up
+from .testlib.vxlan import vxlans_absent
+from .testlib.vxlan import vxlans_up
```
EdDev: This shouts "I am an object" :)
Not a must, but you should consider it.

tests/integration/vxlan_test.py
```diff
@@ -0,0 +1,118 @@
...
+
+VXLAN1_ID = 101
+VXLAN2_ID = 102
+
+
+def test_add_and_remove_vxlan(eth1_up):
+    ifname = eth1_up[Interface.KEY][0][Interface.NAME]
+    with vxlan_interfaces(
+        [(VXLAN1_ID, ifname, '192.168.100.1')]
```
EdDev: I suggest using regular arguments here or a data structure.
When defining an API between modules (this test module and its helper), it is preferred to avoid tuples/lists which include multiple types... you cannot track this easily and it is not structural visible to the reader.

tests/integration/vxlan_test.py
```diff
@@ -0,0 +1,118 @@
...
+
+def test_set_vxlan_iface_down(eth1_up):
+    ifname = eth1_up[Interface.KEY][0][Interface.NAME]
+    with vxlan_interfaces(
+        [(VXLAN1_ID, ifname, '192.168.100.1')]
+    ) as desired_state:
+        libnmstate.apply({Interface.KEY: [vxlan_down(VXLAN1_ID, ifname)]})
+
+    vxlan_ifname = desired_state[Interface.KEY][0][Interface.NAME]
+    assertlib.assert_absent(vxlan_ifname)
```
EdDev: Don't you need to check this inside the contextmanager block? At this point, it already clean-up.

tests/integration/vxlan_test.py
```diff
@@ -0,0 +1,118 @@
...
+    )
+    absent_interfaces = vxlans_absent([(VXLAN1_ID, parent)])
+    absent_interfaces[Interface.KEY].append(
+        {
+            Interface.NAME: parent,
+            Interface.TYPE: InterfaceType.DUMMY,
+            Interface.STATE: InterfaceState.ABSENT,
+        }
+    )
+    with apply_state(up_interfaces, absent_interfaces) as desired_state:
```
EdDev: This one seem generic, maybe too generic.
Perhaps move it to this test module, it does not seem to belong to vxlan test helper.

Also, the absent state can be auto-generated from the "up" state by just changing the state.

tests/integration/nm/vxlan_test.py
```diff
@@ -0,0 +1,128 @@
...
+import pytest
+
+from libnmstate import nm
+from libnmstate.schema import Interface
+from libnmstate.schema import VXLAN
+
+from .testlib import mainloop_run
+
+
+def _vxlan_ifname(state):
```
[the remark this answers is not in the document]
qinqon: Did the opposite like defining before use, sorry about that.

tests/integration/nm/vxlan_test.py
```diff
@@ -0,0 +1,128 @@
...
+        VXLAN.CONFIG_SUBTREE: {
+            VXLAN.ID: 101,
+            VXLAN.BASE_IFACE: ifname,
+            VXLAN.REMOTE: '192.168.1.18',
+            VXLAN.DESTINATION_PORT: 4789,
+        }
+    }
+
+    with _vxlan_interface(vxlan_desired_state):
+        vxlan_ifname = _vxlan_ifname(vxlan_desired_state)
```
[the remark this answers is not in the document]
qinqon: I am going to return the name at _vxlan_interface ctx manager function

tests/integration/nm/vxlan_test.py
qinqon: I am going to return just the vxlan_ifname, it does not fit so well at test_read_destionation_port_from_libnm

tests/integration/testlib/vxlan.py
qinqon: create_vxlans would be since name looks like an adjetive instead of verb.

tests/integration/testlib/vxlan.py
qinqon: Ok, let's pass and don't expect the name to always be parent+id

tests/integration/vxlan_test.py
qinqon: Quite the oppsite, it checks that interface is not there after it's out of the scope of context manager since it will call absent out of context.

tests/integration/vxlan_test.py
qinqon: About absent, you may want to just put as absent some of them.

tests/integration/vxlan_test.py
qinqon: Going to refactor with OOP, this is growing out of proportions.

tests/integration/vxlan_test.py
qinqon: Have refactor it with Vxlan class and proper methods.

tests/integration/testlib/vxlan.py
qinqon: After the OOP refactor it's just up_state since it's a simple factory to genereate the state

tests/integration/nm/vxlan_test.py
qinqon: I am just generating the vxlan name inside the new Vxlan class so it get encapsulated there.

qinqon:
@EdDev have refactor it to use a little of OOP also down state need the type but absent state does not.

qinqon:
integration test test_add_new_base_iface_with_vxlan has transitory failures

libnmstate/nm/vxlan.py
```diff
@@ -0,0 +1,85 @@
...
+
+def _get_destination_port(device):
+    """
+    Retrieve the destination port.
+
+    The destination port is retrieved from the profile settings instead
+    of the device (which represents the kernel state)
+    due to an existing issue [1].
+
+    [1] https://bugzilla.redhat.com/show_bug.cgi?id=1768388
```
cathay4t: I noticed the API has `NM.DeviceVxlan.get_dst_port()`. No work?

qinqon: @thom311 fixed it upstream, there where an issue with libnm cache, the fix is at bz.

cathay4t: OK. If we know for sure NM 1.22 will fix it, we can add a check on NM version there in case we forgot.
Just a suggestion, not a mast.

qinqon: Well, I don't know when is going to land that, but if you guys know is 1.22 let's check version here with xfail

Thanks!

tests/integration/nm/vxlan_test.py
```diff
@@ -0,0 +1,132 @@
...
+from libnmstate.schema import VXLAN
+
+from .testlib import mainloop_run
+
+
+@pytest.mark.xfail(
+    condition=StrictVersion(nm.nmclient.nm_version()) < StrictVersion('1.20'),
+    reason="Failing on device not found is known issue",
+    strict=False,
+)
```
cathay4t: You may remove above check, we only support NM 1.20+ now.

libnmstate/nm/vxlan.py
cathay4t: I am not referring to a test case, but:

```
if nm_version >= 1.22:
    NM.DeviceVxlan.get_dst_port()
else:
    use ConnectionProfile

```

tests/integration/testlib/statelib.py
```diff
@@ -39,6 +40,15 @@
 from libnmstate.state import canonicalize_ipv6_addr
 
 
+@contextmanager
+def apply_state(setup_state, teardown_state):
```
cathay4t: This is not used by any tests

libnmstate/nm/vxlan.py
cathay4t: Yes. It will be fixed in NM 1.22.

tests/integration/testlib/statelib.py
qinqon: Done

tests/integration/nm/vxlan_test.py
qinqon: Done

libnmstate/nm/vxlan.py
qinqon: Done

tests/integration/testlib/vxlan.py
```diff
@@ -0,0 +1,87 @@
...
+
+class Vxlan:
+    def __init__(self, id, base_if, remote, destination_port=4789):
+        self.name = '{}.{}'.format(id, base_if)
+        self.id = id
+        self.base_if = base_if
+        self.remote = remote
+        self.destination_port = destination_port
+
+    def up_state(self):
```
EdDev: nit: These could be made `properties` as they just format and expose the inner state (like getters).
```
@propery
def up_state(self):
    ...

v = Vxlan(...)
state = v.up_state
```

tests/integration/testlib/vxlan.py
```diff
@@ -0,0 +1,87 @@
...
+import libnmstate
+
+from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceState
+from libnmstate.schema import VXLAN
+
+
+class Vxlan:
+    def __init__(self, id, base_if, remote, destination_port=4789):
+        self.name = '{}.{}'.format(id, base_if)
```
EdDev: Isn't this the other way around? (base_if.id)
BTW, now that we are python3-only, you can use the neat new formatting: `f'{base_if}.{id}'`

tests/integration/testlib/vxlan.py
```diff
@@ -0,0 +1,87 @@
...
+from contextlib import contextmanager
+
+import libnmstate
+
+from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceState
+from libnmstate.schema import VXLAN
+
+
+class Vxlan:
```
EdDev: Very nice, it can even be (in a following change) moved to the production code.
If you call this `VxlanState` , you can drop all the `state` wording in the methods.

tests/integration/testlib/vxlan.py
```diff
@@ -0,0 +1,87 @@
...
+    def down_state(self):
+        return {
+            Interface.NAME: self.name,
+            Interface.TYPE: VXLAN.TYPE,
+            Interface.STATE: InterfaceState.DOWN,
+        }
+
+
+@contextmanager
+def vxlan_interfaces(vxlans, create=True):
```
EdDev: Now, with python3, you can do this:
`def vxlan_interfaces(*vxlans, create=True):`

And in the caller you do not need the list/tuple:
`with vxlan_interfaces(Vxlan(...), Vxlan(...))`

tests/integration/nm/vxlan_test.py
```diff
@@ -0,0 +1,127 @@
...
+    vxlan_desired_state = {
+        VXLAN.CONFIG_SUBTREE: {
+            VXLAN.ID: 101,
+            VXLAN.BASE_IFACE: ifname,
+            VXLAN.REMOTE: '192.168.1.18',
+            VXLAN.DESTINATION_PORT: 4789,
+        }
+    }
+
+    yield _vxlan_interface(vxlan_desired_state)
```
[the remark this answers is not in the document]
qinqon: In the test we do :
```
    with vxlan_101 as vxlan_desired_state:                                      
        vxlan_name = _vxlan_ifname(vxlan_desired_state)                         
        vxlan_current_state = _get_vxlan_current_state(vxlan_name)              
        assert vxlan_desired_state == vxlan_current_state                       
                                                                                
    assert not _get_vxlan_current_state(vxlan_name)                   
```

So we need it to be a context manager and not the state.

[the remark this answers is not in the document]
EdDev: Ok, so then the fixture is not needed, you just need a helper function.
Something like:
```
vxlan_desired_state = _create_vxlan_state()
with _vxlan_interface(vxlan_desired_state):
    ...
```

EdDev:
@qinqon , CI fails: https://travis-ci.com/nmstate/nmstate/jobs/258153028

qinqon:
Ci is passing here for vxlan

```bash
./automation/run-tests.sh te--test-type integ --pytest-args='-x -k vxlan' --el8
tests/integration/vxlan_test.py::test_add_and_remove_vxlan PASSED                                                            [ 14%]
tests/integration/vxlan_test.py::test_add_and_remove_two_vxlans_on_same_iface PASSED                                         [ 28%]
tests/integration/vxlan_test.py::test_rollback_for_vxlans PASSED                                                             [ 42%]
tests/integration/vxlan_test.py::test_set_vxlan_iface_down PASSED                                                            [ 57%]
tests/integration/vxlan_test.py::test_add_new_base_iface_with_vxlan XPASS                                                    [ 71%]
tests/integration/nm/vxlan_test.py::test_create_and_remove_vxlan PASSED                                                      [ 85%]
tests/integration/nm/vxlan_test.py::test_read_destionation_port_from_libnm XFAIL                                             [100%]
```

cathay4t:
@qinqon The CI is failling at 
`tests/integration/nm/vxlan_test.py::test_create_and_remove_vxlan FAILED  [ 96%]`

with error:

```
2019-11-19 11:48:45 nmclient.py:180 ERROR NM main-loop aborted: Connection activation failed on connection_id eth1.101: error=nm-device-error-quark: eth1.101 failed to create resources: Failed to create VXLAN interface 'eth1.101' for 'eth1.101': NME_UNSPEC (1)
```

tests/integration/nm/vxlan_test.py
```diff
@@ -0,0 +1,126 @@
...
+    )
+    con_profile.add(save_to_disk=False)
+    nm.device.activate(connection_id=ifname)
+
+
+@mainloop_run
+def _delete_vxlan(devname):
+    nmdev = nm.device.get_device_by_name(devname)
+    nm.device.deactivate(nmdev)
+    nm.device.delete(nmdev)
```
EdDev: The removal of software devices seems to be buggy as you found out. Please use here also ‘delete_device’ (in addition to the current 2): https://github.com/nmstate/nmstate/blob/780e0833dfb4a9cc72679463affc9ce9c9ea7f3f/libnmstate/nm/device.py#L245

tests/integration/vxlan_test.py
```diff
@@ -0,0 +1,110 @@
...
+    with vxlan_interfaces(vxlan):
+        libnmstate.apply(vxlans_down([vxlan]))
+
+    assertlib.assert_absent(vxlan.name)
+
+
+@pytest.mark.xfail(
+    reason='https://bugzilla.redhat.com/show_bug.cgi?id=1772382', strict=False
+)
+def test_add_new_base_iface_with_vxlan(eth1_up):
```
EdDev: Maybe replace ‘base’ with ‘bond’?

libnmstate/nm/device.py
```diff
@@ -285,7 +285,7 @@ def _delete_device_callback(src_object, result, user_data):
             mainloop.execute_next_action()
         else:
             mainloop.quit(
-                'Device deletion failed on {}: error={}'.format(
+                'Device deletion failed on {}: unexpected error={}'.format(
```
EdDev: Seems to be an unrelated change.

tests/integration/nm/vxlan_test.py
```diff
@@ -0,0 +1,127 @@
...
+
+@pytest.mark.xfail(
+    condition=StrictVersion(nm.nmclient.nm_version()) < StrictVersion('1.22'),
+    strict=True,
+    reason='https://bugzilla.redhat.com/show_bug.cgi?id=1768388',
+)
+def test_read_destionation_port_from_libnm(eth1_up):
+    vxlan_desired_state = _create_vxlan_state(eth1_up)
+    with _vxlan_interface(vxlan_desired_state):
+        vxlan_name = vxlan_desired_state[Interface.KEY][0][Interface.NAME]
```
EdDev: This seems to be wrong.
The vxlan_desired_state is pointing to the vxlan subtree, not to a full state.

tests/integration/vxlan_test.py
```diff
@@ -0,0 +1,110 @@
...
+    assert current_state == current_state_after_apply
+
+
+def test_set_vxlan_iface_down(eth1_up):
+    ifname = eth1_up[Interface.KEY][0][Interface.NAME]
+    vxlan = VxlanState(id=VXLAN1_ID, base_if=ifname, remote='192.168.100.1')
+    with vxlan_interfaces(vxlan):
+        libnmstate.apply(vxlans_down([vxlan]))
+
+    assertlib.assert_absent(vxlan.name)
```
EdDev: I think you want to check that the ‘down’ was applied (i.e. removed). If so, this assert needs to be indented into the ‘with’ block.

tests/integration/vxlan_test.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+@pytest.mark.xfail(
+    reason='https://bugzilla.redhat.com/show_bug.cgi?id=1772382', strict=False
+)
+def test_add_new_bond_iface_with_vxlan(eth1_up):
+    eth_name = eth1_up[Interface.KEY][0][Interface.NAME]
+    bond_name = "bond1"
```
EdDev: Single quotes please.

libnmstate/nm/device.py
qinqon: Is good to debug the difference of not matching exception instead of exception that we don't catch.

tests/integration/vxlan_test.py
qinqon: absent is good here to check that vxlan no longer exists, we can also add a down check within with.

libnmstate/nm/device.py
EdDev: Good argument, but please propose it through its own PR with this reasoning in the message.

tests/integration/vxlan_test.py
EdDev: The ‘down’ should remove it.
There is no need to assert what was already tested elsewhere, although I do not mind too much if you do.

The test says it checks setting the iface down, so it makes sense to assert it IMO. How will you know if it worked?

qinqon: ack, asserting within context manager.

tests/integration/nm/vxlan_test.py
```diff
@@ -0,0 +1,127 @@
...
+
+    assert not _get_vxlan_current_state(vxlan_name)
+
+
+@pytest.mark.xfail(
+    condition=StrictVersion(nm.nmclient.nm_version()) < StrictVersion('1.22'),
+    strict=True,
+    reason='https://bugzilla.redhat.com/show_bug.cgi?id=1768388',
+)
+def test_read_destionation_port_from_libnm(eth1_up):
```
EdDev: Spelling

EdDev:
@cathay4t , All comments have been answered as far as I can see. If I missed anything, please let me know.

Issue #516, opened 2019-09-27
Bridge's link stays in the system after setup fails with `device is strictly unmanaged`


phoracek:
When a bridge setup fails, it successfully rolls back the NetworkManager connection, but the netlink device stays in the kernel:

```yaml
interfaces:
 - name: br4
   type: linux-bridge
   state: up
   bridge:
     options:
       stp:
         enabled: false
     port:
     - name: petr
```
-> Fails with `device is strictly unmanaged`

```
[root@t470s phoracek]# ip l | grep br4
20: br4: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
[root@t470s phoracek]# nmcli c | grep br4
```

On top of that, nmstate is not able to remove the bridge in this state:

```yaml
interfaces:
- name: br4
  type: linux-bridge
  state: absent
```

-> Succeeds, br4 remains.

When I trigger the setup again, it fails due to non-existent port and cleans up the bridge.

https://paste.fedoraproject.org/paste/U-7cflqpwzFmyGT15WibCw

So I think we have two bugs here.

1. We are unable to rollback partially created interface
2. We have a race when creating virtual interfaces

We think we fixed the first one by applying the same workaround as is done for OVS interfaces - adding it to `remove_devs_actions`:

```diff
diff --git a/libnmstate/nm/applier.py b/libnmstate/nm/applier.py
index f67f948..09efb96 100644
--- a/libnmstate/nm/applier.py
+++ b/libnmstate/nm/applier.py
@@ -179,6 +179,7 @@ def set_ifaces_admin_state(ifaces_desired_state, con_profiles=()):
                         nmclient.NM.DeviceType.OVS_BRIDGE,
                         nmclient.NM.DeviceType.OVS_PORT,
                         nmclient.NM.DeviceType.OVS_INTERFACE,
+                        nmclient.NM.DeviceType.BRIDGE,
                     ):
                         remove_devs_actions[nmdev].append(device.delete_device)
             else:
```

phoracek:
@maiqueb @EdDev I think it could affect you in VDSM too. It made our CI flaky.

qinqon:
@EdDev do we have ci tests for rollbacks ?

EdDev:
[...]

PR #518, opened 2019-09-30
Add rollback test for linux-bridge

qinqon:
To verify #516

### DIFF
--- a/tests/integration/linux_bridge_test.py
+++ b/tests/integration/linux_bridge_test.py
@@ -18,10 +18,12 @@
 #
 from copy import deepcopy
 
+import time
 import pytest
 import yaml
 
 import libnmstate
+from libnmstate.error import NmstateVerificationError
 from libnmstate.schema import Interface
 from libnmstate.schema import InterfaceIPv4
 from libnmstate.schema import InterfaceIPv6
@@ -293,6 +295,20 @@ def test_linux_bridge_over_bond_over_slave_in_one_transaction(bond0):
         assertlib.assert_state_match(desired_state)
 
     assertlib.assert_absent(bridge_name)
+
+
+def test_rollback_for_linux_bridge():
+    original_state = libnmstate.show()
+    bridge_name = TEST_BRIDGE0
+    bridge_state = _create_bridge_subtree_config(())
+    with pytest.raises(NmstateVerificationError):
+        with linux_bridge(bridge_name, bridge_state) as desired_state:
+            desired_state[Interface.KEY][0]['invalid_key'] = 'foo'
+            libnmstate.apply(desired_state)
+
+    time.sleep(5)  # Give some time for NetworkManager to rollback
+    current_state = libnmstate.show()
+    assert original_state == current_state
 
 
 def _add_port_to_bridge(bridge_state, ifname):


tests/integration/linux_bridge_test.py
```diff
@@ -295,6 +297,19 @@ def test_linux_bridge_over_bond_over_slave_in_one_transaction(bond0):
...
+    current_state = libnmstate.show()
+    bridge_name = TEST_BRIDGE0
+    bridge_state = _create_bridge_subtree_config(("invalid_port",))
+    with pytest.raises(NmstateVerificationError):
+        with linux_bridge(bridge_name, bridge_state) as desired_state:
+            libnmstate.apply(desired_state)
+
+    time.sleep(5)  # Give some time for NetworkManager to rollback
+    current_state_after_apply = libnmstate.show()
+    assert current_state == current_state_after_apply
```
phoracek: It's not current state anymore. Please rename it to `original_state` or something alike.

cathay4t:
@qinqon I tried to reproduce it in RHEL 8.1 and seems works well. And the CI also show the works well.

Which  system(Fedora or RHEL) are you using to reproduce the problem?

The `device is strictly unmanaged` issue is fixed in NM 1.20 which was related to the race between NM and udev.

tests/integration/linux_bridge_test.py
```diff
@@ -295,6 +297,19 @@ def test_linux_bridge_over_bond_over_slave_in_one_transaction(bond0):
...
+    original_state = libnmstate.show()
+    bridge_name = TEST_BRIDGE0
+    bridge_state = _create_bridge_subtree_config(("invalid_port",))
+    with pytest.raises(NmstateVerificationError):
+        with linux_bridge(bridge_name, bridge_state) as desired_state:
+            libnmstate.apply(desired_state)
+
+    time.sleep(5)  # Give some time for NetworkManager to rollback
+    current_state = libnmstate.show()
+    assert original_state == current_state
```
cathay4t: In the issue, you mentioned there is leftover bridge interface. Please consider to add a check on output of `ip link show <br_name>` to confirm there is no leftover bridge.

EdDev: We are supposed to report *all* interfaces, even if there are no profiles for it. (they will just show up as `DOWN`)
Therefore, using `ip link` should not be needed.

tests/integration/linux_bridge_test.py
```diff
@@ -295,6 +297,19 @@ def test_linux_bridge_over_bond_over_slave_in_one_transaction(bond0):
     assertlib.assert_absent(bridge_name)
 
 
+def test_rollback_for_linux_bridge():
+    original_state = libnmstate.show()
+    bridge_name = TEST_BRIDGE0
+    bridge_state = _create_bridge_subtree_config(("invalid_port",))
```
EdDev: The invalid port name is confusing and may fail the transaction before anything was created (which will not check what you intended to check).
Please add a key-value that is invalid at the `desired_state` end result, just before you apply.

BTW: please always use single quotes (`).

qinqon: Done

tests/integration/linux_bridge_test.py
phoracek: That is the issue of the bug we've found AFAIK. The interface is not reported in state after rollback, but it is in kernel. (But not 100% sure here)

https://github.com/nmstate/nmstate/issues/516

EdDev: Where did you see that the interface is not reported? I do not see it in the issue description.

phoracek: I'm not sure actually. nmstate was not able to remove the interface from system, but I don't know whether it was reported in `show` or not.

EdDev:
@qinqon , you merged instead of rebasing. Please rebase over master such that this PR will include only you changes and no merge commits.

qinqon:
@EdDev rebased

EdDev:
@qinqon , any reason you reseted my ack on this pr? I did not see any changes since the last ack.

qinqon:
@EdDev nop, all good, travis failure is unrelated I think.

EdDev:
@qinqon , please rebase.

PR #524, opened 2019-10-12
ovs: nm tests refactoring

phoracek:
Current NetworkManager tests of OVS are rather hard to understand. It is not
clear how are NM devices built up and they can break in future when internal
implementation of either NM or nmstate changes. This patch moves these
heavy-mocked tests under integration/nm/ovs_test.py, making sure all existing
scenarios are still covered.

There is also was bug in mocking, where we set master of an interface to itself.
This issue manifests itself when we connect multiple interfaces to a port. In
the current state it works because of the way we tread OVS ports and interfaces
as 1:1.

### DIFF
--- a/tests/integration/nm/ovs_test.py
+++ b/tests/integration/nm/ovs_test.py
@@ -72,16 +72,24 @@ def test_create_and_remove_minimum_config_bridge(
 @pytest.mark.xfail(
     raises=MainloopTestError, reason='https://bugzilla.redhat.com/1724901'
 )
+def test_bridge_without_ports(bridge_default_config):
+    bridge_desired_state = bridge_default_config
+
+    with _bridge_interface(bridge_desired_state):
+        bridge_current_state = _get_bridge_current_state()
+        assert bridge_desired_state == bridge_current_state
+        assert bridge_current_state[OB.CONFIG_SUBTREE][OB.PORT_SUBTREE] == []
+
+    assert not _get_bridge_current_state()
+
+
+@pytest.mark.xfail(
+    raises=MainloopTestError, reason='https://bugzilla.redhat.com/1724901'
+)
 def test_bridge_with_system_port(eth1_up, bridge_default_config):
     bridge_desired_state = bridge_default_config
 
-    eth1_port = {
-        OB.PORT_NAME: 'eth1',
-        OB.PORT_TYPE: OBPortType.SYSTEM,
-        # OVS vlan/s are not yet supported.
-        # OB.PORT_VLAN_MODE: None,
-        # OB.PORT_ACCESS_TAG: 0,
-    }
+    eth1_port = {OB.PORT_NAME: 'eth1', OB.PORT_TYPE: OBPortType.SYSTEM}
 
     bridge_desired_state[OB.CONFIG_SUBTREE][OB.PORT_SUBTREE].append(eth1_port)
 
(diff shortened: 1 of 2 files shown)


phoracek:
@cathay4t @maiqueb @EdDev could you please take a look? I want to do this all around refactorings while we discuss OVS bond proposal.

EdDev:
At some point in time we realized that unit testing the nm subpackage is not useful and we better test it through the nm integration tests.
Over-mocking is not maintainable.

phoracek:
@EdDev i agree that mimicking NM internals is not the happiest solution. However, this PR fixes bugs in the test (the fact that it passed was a coincidence) and it merely moves those mock so it makes sense to casual reader like me.

I did these refactorings to support my follow-up bonding PR. In my opinion we should either accept these fixes or drop it all together.

EdDev:
I suggest we make sure we cover everything that exists here in https://github.com/nmstate/nmstate/blob/master/tests/integration/nm/ovs_test.py and drop this file.

If we’ll manage to separate business logic from the nm api stuff, we can bring back the unit tests.

I’m also ok with taking fixes here, but I will not want to maintain it and future changes you’ll need to add here when new ovs stuff gets in production.

phoracek:
@EdDev I'd still keep this file to test simple functions without side-effects. Lemme do it in this PR.

EdDev:
@phoracek , I went over this change and then went over the production code.
This change by itself is fine, but my original conclusion that these unit tests are not maintainable is stronger than ever.

In order to unit test the nm.ovs module, there is a need to completely separate the logic from the NM api. Then we will also have very little to test.

Assuming that we will not accept any new unit tests on any nm subpackage (unless it is pure logic == no mocks), do you see any worth in taking these changes?
I personally cannot trust them as I need to understand them first, and that will take me too much time. I feel like if we change anything in the production code it will break all its tests and to fix them it will cost us too much.

phoracek:
@EdDev well, there are bugs in these tests, so you might as well drop them completely. It gives you no additional value.

I'm pretty confident those tests are fine after the refactoring and no change in your production code should not affect them, unless NM drastically changes their API.

It took me a lot a of time to understand this pile of mocks, but I don't really care if you close this PR. If you do though, please drop those mocking tests completely. Having bugged tests is more harmful than having complicated ones.

EdDev:
> It took me a lot a of time to understand this pile of mocks,

This is why I think we better drop tests that have too many mocks... the test become unmaintainable.

> but I don't really care if you close this PR. If you do though, please drop those mocking tests completely. Having bugged tests is more harmful than having complicated ones.

Obviously that is the alternative... to drop them all and depend on the integration tests (integration/nm).

It was in my plan to go over all NM unit tests and make sure they are covered in the integration/nm tests.

phoracek:
Let's settle it. Petr will make sure we have integration for the OVS NM tests, drop all mocks. All as part of this PR. Should get to it later today.

phoracek:
@EdDev please take a look. Was easier than expected, most of those scenarios was already in place, I just had to add test of both internal and system iface connected to the bridge.

phoracek:
@EdDev done

EdDev:
@phoracek I just noticed bugs in the commit message, can you have a look?

phoracek:
@EdDev not sure what were you talking about, maybe it is fixed now.

EdDev:
@phoracek , yes, now it looks fine. Thanks.

EdDev:
@phoracek , seems like the rebase failed. Probably a rename was missed.

phoracek:
Passed locally.

EdDev:
@phoracek , based on (travis-ci)[https://travis-ci.org/nmstate/nmstate/jobs/601368305#L971] , the assert fails on the ordering. I think we need to sort the desired state port list (post apply) and the report of the ports should be sorted in the production code.

phoracek:
@EdDev I will add the fix as a separate commit within this PR, is that fine?

phoracek:
@EdDev they are already being sorted for comparison in `state.py`. I will just do the sorting local to the test (as it is already done in `route_test.py`).

EdDev:
@phoracek , I did not understand your last two comments. The tests still fail and I have not seen any sorting done in the tests themselves.

EdDev:
@phoracek , could you please update on the status of this PR?

phoracek:
I don't have the capacity to finish this. If anyone else is interested, feel free to re-open.

PR #525, opened 2019-10-12
drop OVS port's "type" attribute

phoracek:
OVS port "type" attribute is redundant. It can be currently set to "system" or "internal", but these can be easily deduced from desired and current state. I propose we drop this attribute from the API, so we don't keep redundant information and make schema more streamlined.

No other nmstate interface type requires users to specify slave type (both bond and linux-bridge require the name only).

OVS bridge schema would change from:

```yaml
interfaces:
- name: ovs-br0
  type: ovs-bridge
  state: up
  bridge:
    port:
    - name: eth3
      type: system
```

To:

```yaml
interfaces:
- name: ovs-br0
  type: ovs-bridge
  state: up
  bridge:
    port:
    - name: eth3
```

https://nmstate.atlassian.net/browse/NMSTATE-248

Note that no changes were needed to implement this, apart from dropping the "type" from reporting, schema and tests.

### DIFF
--- a/tests/integration/nm/ovs_test.py
+++ b/tests/integration/nm/ovs_test.py
@@ -23,7 +23,6 @@ import pytest
 
 from libnmstate import nm
 from libnmstate.schema import OVSBridge as OB
-from libnmstate.schema import OVSBridgePortType as OBPortType
 
 from .testlib import mainloop
 from .testlib import MainloopTestError
@@ -77,7 +76,6 @@ def test_bridge_with_system_port(eth1_up, bridge_default_config):
 
     eth1_port = {
         OB.PORT_NAME: 'eth1',
-        OB.PORT_TYPE: OBPortType.SYSTEM,
         # OVS vlan/s are not yet supported.
         # OB.PORT_VLAN_MODE: None,
         # OB.PORT_ACCESS_TAG: 0,
@@ -95,10 +93,10 @@ def test_bridge_with_system_port(eth1_up, bridge_default_config):
 @pytest.mark.xfail(
     raises=MainloopTestError, reason='https://bugzilla.redhat.com/1724901'
 )
-def test_bridge_with_internal_interface(eth1_up, bridge_default_config):
+def test_bridge_with_internal_interface(bridge_default_config):
     bridge_desired_state = bridge_default_config
 
-    ovs_port = {OB.PORT_NAME: 'ovs0', OB.PORT_TYPE: OBPortType.INTERNAL}
+    ovs_port = {OB.PORT_NAME: 'ovs0'}
 
     bridge_desired_state[OB.CONFIG_SUBTREE][OB.PORT_SUBTREE].append(ovs_port)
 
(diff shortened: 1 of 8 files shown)


cathay4t:
Please don't change `libnmstate/schema.py` which is a part of API.

EdDev:
> Please don't change libnmstate/schema.py which is a part of API.

If this patch will do what it is intended to do, then it will be correct to remove it and change the formal API.
We are supposed to be free to change the API, especially for OVS. But more important, if the type is getting interpreted, the users will not notice this (and will not care). The change seems to be backward compatible.

phoracek:
@cathay4t @EdDev please take a look

phoracek:
nmstate.github.io update https://github.com/nmstate/nmstate.github.io/pull/21

phoracek:
Passed locally. @EdDev could you please take a look? We will need to rebase it on https://github.com/nmstate/nmstate/pull/524, but it should be ready to get reviews.

tests/integration/nm/ovs_test.py
```diff
@@ -95,10 +93,10 @@ def test_bridge_with_system_port(eth1_up, bridge_default_config):
 @pytest.mark.xfail(
     raises=MainloopTestError, reason='https://bugzilla.redhat.com/1724901'
 )
-def test_bridge_with_internal_interface(eth1_up, bridge_default_config):
+def test_bridge_with_internal_interface(bridge_default_config):
```
EdDev: I am not looking... ;)

tests/integration/nm/ovs_test.py
```diff
@@ -149,13 +147,23 @@ def _attach_port_to_bridge(port_state):
...
-    if port_state[OB.PORT_TYPE] == OBPortType.SYSTEM:
+    if _is_ethernet(port_state[OB.PORT_NAME]):
         _connect_interface(port_profile_name, port_state)
-    elif port_state[OB.PORT_TYPE] == OBPortType.INTERNAL:
+    else:
         iface_name = port_state[OB.PORT_NAME]
         _create_internal_interface(iface_name, master_name=port_profile_name)
 
 
+def _is_ethernet(iface_name):
```
EdDev: Won't it be more correct to name it `_is_system_interface`?

tests/integration/nm/ovs_test.py
```diff
@@ -149,13 +147,23 @@ def _attach_port_to_bridge(port_state):
...
+    dev = nm.device.get_device_by_name(iface_name)
+    if not dev:
+        return False
+    return dev.get_device_type() in (
+        nm.nmclient.NM.DeviceType.ETHERNET,
+        nm.nmclient.NM.DeviceType.VETH,  # simulates ethernet in tests
```
EdDev: Wouldn't it be more correct to specify the types that are not "system" interfaces, which in practice means they are OVS ones.

Non system ifaces: UNKNOWN, GENERIC, OVS_*

tests/integration/testlib/ovslib.py
```diff
@@ -25,7 +25,6 @@
 from libnmstate.schema import InterfaceState
 from libnmstate.schema import InterfaceType
 from libnmstate.schema import OVSBridge
-from libnmstate.schema import OVSBridgePortType
 
 
 class Bridge:
```
EdDev: Ohhh... this is a bug :/

tests/integration/nm/ovs_test.py
phoracek: :eyes:

tests/integration/nm/ovs_test.py
phoracek: Done

tests/integration/nm/ovs_test.py
phoracek: I prefer to do positive checks that assuming that if something is not X it is Y. I can change it, but I think the current form is easier to understand.

tests/integration/testlib/ovslib.py
phoracek: I don't get it, what is the bug?

phoracek: You mean having Bridge without (object)?

phoracek:
@EdDev fixed the commit message and answered your comments. Please take another look.

tests/integration/nm/ovs_test.py
EdDev: Currently the code assumes (i.e. depends) on which interface types are used in the test env, which seems wrong to me.

I am good with a positive check here, how about checking against ‘is_internal_interface’? This should drop the assumptions on the env.

tests/integration/testlib/ovslib.py
EdDev: Yes. It will be fine when we drop py2. (Should happen soon)

phoracek: Fixed, don't look

tests/integration/nm/ovs_test.py
phoracek: Makes sense. Should be fixed now.

Issue #131, opened 2018-10-30
verification breaks removing a description from an interface


tyll:
After removing a description by setting an empty description, verification fails because it expects an empty description in the resulting state:

```Applying the following state: 
interfaces:
- name: eth2
  type: ethernet
  state: up
  description: ''
  ipv4:
    enabled: false
  ipv6:
    address:
    - ip: fe80::5054:ff:fe0d:e7ae
      prefix-length: 64
    enabled: true
  mtu: 1500

2018-10-30 13:38:07,326 root         DEBUG    Checkpoint /org/freedesktop/NetworkManager/Checkpoint/30 created for all devices: 60
2018-10-30 13:38:07,356 root         DEBUG    Executing NM action: func=commit_changes_async, args=(True, <Gio.Cancellable object at 0x7fbb54014d20 (GCancellable at 0x7fbb440340e0)>, <function _commit_changes_callback at 0x7fbb55c785f0>, (<libnmstate.nm.nmclient._MainLoop object at 0x7fbb53b59250>, <NM.DeviceEthernet object at 0x7fbb54014370 (NMDeviceEthernet at 0x2490310)>))
2018-10-30 13:38:07,368 root         DEBUG    Connection update succeeded: dev=eth2
2018-10-30 13:38:07,368 root         DEBUG    Executing NM action: func=activate_connection_async, args=(None, <NM.DeviceEthernet object at 0x7fbb54014370 (NMDeviceEthernet at 0x2490310)>, None, <Gio.Cancellable object at 0x7fbb54014d20 (GCancellable at 0x7fbb440340e0)>, <function _active_connection_callback at 0x7fbb55c78cf8>, <libnmstate.nm.nmclient._MainLoop object at 0x7fbb53b59250>)
2018-10-30 13:38:07,410 root         DEBUG    Connection activation succeeded: dev=eth2, con-state=<enum NM_ACTIVE_CONNECTION_STATE_ACTIVATING of type NM.ActiveConnectionState>
2018-10-30 13:38:07,410 root         DEBUG    NM action queue exhausted, quiting mainloop
[...]

PR #527, opened 2019-10-14
netapplier: allow description removal

ffmancera:
Allows description removal by setting an empty description. We need to
delete the description element from the desired state after being
applied.

### DIFF
--- a/libnmstate/state.py
+++ b/libnmstate/state.py
@@ -284,6 +284,7 @@ class State(object):
         self._remove_iface_ipv6_link_local_addr()
         self._sort_ip_addresses()
         self._capitalize_mac()
+        self._remove_empty_description()
 
     def merge_interfaces(self, other_state):
         """
@@ -473,6 +474,11 @@ class State(object):
             mac = ifstate.get(Interface.MAC)
             if mac:
                 ifstate[Interface.MAC] = mac.upper()
+
+    def _remove_empty_description(self):
+        for ifstate in six.viewvalues(self.interfaces):
+            if ifstate.get(Interface.DESCRIPTION) == '':
+                del ifstate[Interface.DESCRIPTION]
 
     def _assert_interfaces_equal(self, current_state):
         for ifname in self.interfaces:
--- a/tests/integration/interface_common_test.py
+++ b/tests/integration/interface_common_test.py
@@ -1,0 +1,36 @@
+#
+# Copyright (c) 2019 Red Hat, Inc.
+#
+# This file is part of nmstate
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+
+import libnmstate
+
+from .testlib import statelib
+from libnmstate.schema import Interface
+
+
+def test_iface_description_removal(eth1_up):
+    desired_state = eth1_up
+    desired_state[Interface.KEY][0][Interface.DESCRIPTION] = 'bar'
+    libnmstate.apply(desired_state)
+    current_state = statelib.show_only(('eth1',))
+    assert current_state[Interface.KEY][0][Interface.DESCRIPTION] == 'bar'
+
+    desired_state[Interface.KEY][0][Interface.DESCRIPTION] = ''
+    libnmstate.apply(desired_state)
+    current_state = statelib.show_only(('eth1',))
+    assert Interface.DESCRIPTION not in current_state[Interface.KEY][0]
--- a/tests/lib/state_test.py
+++ b/tests/lib/state_test.py
@@ -132,6 +132,21 @@ class TestAssertIfaceState(object):
 
         desired_state.verify_interfaces(current_state)
 
+    def test_description_is_empty(self):
+        desired_state = self._base_state
+        current_state = self._base_state
+        desired_state.interfaces['foo-name'][Interface.DESCRIPTION] = ''
+
+        desired_state.verify_interfaces(current_state)
+
+    def test_description_is_not_empty(self):
+        desired_state = self._base_state
+        current_state = self._base_state
+        desired_state.interfaces['foo-name'][Interface.DESCRIPTION] = 'bar'
+        current_state.interfaces['foo-name'][Interface.DESCRIPTION] = 'bar'
+
+        desired_state.verify_interfaces(current_state)
+
     @property
     def _base_state(self):
         return state.State(
--- a/libnmstate/schema.py
+++ b/libnmstate/schema.py
@@ -38,6 +38,7 @@ class Interface(object):
     NAME = 'name'
     TYPE = 'type'
     STATE = 'state'
+    DESCRIPTION = 'description'
 
     IPV4 = 'ipv4'
     IPV6 = 'ipv6'


ffmancera:
Fixes: #131

ffmancera:
> 
> 
> I do not understand the reasoning for this change.
> Please explain why we need it.
> 
> In addition, any change done in the production code needs to be covered by tests. We need unit and integration ones, depending on what is changed.

If the description of an interface is set, the way to remove it is by setting an empty description. Currently, this is not possible if we activate the verification.

When the verification is performed, the "description" field appears empty in the desired state, however in the current state the "description" field does not appear so they do not match. This causes an error.

@tyll reported that in #131 

My proposal is to remove the "description" field from the desired state once the changes have been applied. This way, the verification will be successful and the interface description will be removed.

Sorry if I am missing something, thanks!

EdDev:
> If the description of an interface is set, the way to remove it is by setting an empty description. 

> When the verification is performed, the "description" field appears empty in the desired state, however in the current state the "description" field does not appear so they do not match. This causes an error.

Please include this info in the commit message.

> My proposal is to remove the "description" field from the desired state once the changes have been applied. This way, the verification will be successful and the interface description will be removed.

There is a dedicated location where we do these kind of things, please review https://github.com/nmstate/nmstate/blob/925f695f8de1cb4a467094ab8c623dd3eacca2e0/libnmstate/state.py#L279

Please make sure to cover this change in the unit test: https://github.com/nmstate/nmstate/blob/925f695f8de1cb4a467094ab8c623dd3eacca2e0/tests/lib/state_test.py#L45

We should probably also add a general interface integration test module under `tests/integration/` and include a test that covers the `description` option. We can call it `interface_common_test`.
It may be appropiate to promote `iface_admin_state_test` module to this common one, but I am not that sure... we can decide on this later on.

tests/lib/state_test.py
```diff
@@ -132,6 +132,22 @@ def test_sort_multiple_ip(self):
...
+        desired_state.verify_interfaces(current_state)
+
+    def test_description_is_not_empty(self):
+        desired_state = self._base_state
+        current_state = self._base_state
+        desired_state.interfaces['foo-name'][Interface.DESCRIPTION] = 'bar'
+        current_state.interfaces['foo-name'][Interface.DESCRIPTION] = 'bar'
+
+        desired_state.verify_interfaces(current_state)
+        desired_state.interfaces['foo-name'][Interface.DESCRIPTION]
```
EdDev: What is this line all about?

ffmancera:
As we are removing the description when is empty that should arise an exception if the description field have been removed in the verification phase.

EdDev:
> As we are removing the description when is empty that should arise an exception if the description field have been removed in the verification phase.

What?

ffmancera:
I'm sorry, I wrote the message quickly. 

In the negative case test, we checked that we have not removed the field "description" as it is not empty. The last line causes an exception if the field "description" does not exist. 

Another option could be to use an assert statement.

EdDev:
> Another option could be to use an assert statement.

Lets move the discussion to the inline comment, discussing it here is not convenient.

tests/lib/state_test.py
EdDev: > In the negative case test, we are checking that we have not removed the field "description" as it is not empty. The last line causes an exception if the field "description" does not exist.

In python, this is causing the function/method to return the value.
But in general, it looks like a hack/abuse of the language options.

> 
> Another option could be to use an assert statement.

If you want to check if a key exists, check it with an assert please. `assert k in d`.

But.. checking if the description key exists or not is an implementation detail which the test should not care about.
It is an assumption that `verify_interfaces` is mutable, it could have been immutable (or changed to be so in the future).

ffmancera: Seems fine, I am going to remove that line. Thanks!

EdDev:
All good, but unfortunately we forgot the integration test.
Could you please add now or as a following PR an integration test that recreates this issue?

tests/integration/interface_common_test.py
```diff
@@ -0,0 +1,32 @@
+#
+# Copyright (c) 2018-2019 Red Hat, Inc.
```
EdDev: Please drop the 2018, this is a new file.

tests/integration/interface_common_test.py
```diff
@@ -0,0 +1,32 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+
+import libnmstate
+
+from .testlib.statelib import INTERFACES
```
EdDev: This is deprecated, please use ‘Interface.KEY’ from the schema module.

tests/integration/interface_common_test.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def test_set_empty_iface_description(eth1_up):
+    desired_state = {INTERFACES: [{'name': 'eth1', 'description': ''}]}
```
EdDev: Please fetch the name from eth1_up or just edit the state it returns.

tests/integration/interface_common_test.py
```diff
@@ -0,0 +1,32 @@
...
+    libnmstate.apply(desired_state)
+
```
EdDev: You need to check something so it will be clear what is the scenario expectation.
How about reporting back the current state and asserting the expected missing key?

tests/integration/interface_common_test.py
```diff
@@ -0,0 +1,32 @@
...
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+
+import libnmstate
+
+from .testlib.statelib import INTERFACES
+
+
+def test_set_empty_iface_description(eth1_up):
```
EdDev: It will be more interesting to check its removal, correlating to the original issue found.
- add a non empty description.
- remove the description.

As these are integration tests which are slow (unlike unit tests) I would just merge these two tests you added into one, but still assert each step to check that the outcome is correct.

tests/integration/interface_common_test.py
```diff
@@ -0,0 +1,35 @@
...
+import libnmstate
+
+from libnmstate.schema import Interface
+
+
+def test_iface_description_removal(eth1_up):
+    desired_state = eth1_up
+    desired_state[Interface.KEY][0][Interface.DESCRIPTION] = 'bar'
+    libnmstate.apply(desired_state)
+    eth1 = libnmstate.show()
```
EdDev: Unfortunately this is incorrect. A plain ‘show’ will return the full state with all interfaces, not only for eth1.
This means you cannot know which interface exists at index 0.
Also, it is unlikely that the value of the key in question has the same identity as ‘bar’. You need to use equality and not identity.

ffmancera: Oh sorry, I knew it is wrong. I just pushed it for my personal use but the PR got updated too. Tomorrow morning I am pushing the fixed version. Thanks!

libnmstate/state.py
```diff
@@ -474,6 +475,11 @@ def _capitalize_mac(self):
             if mac:
                 ifstate[Interface.MAC] = mac.upper()
 
+    def _remove_empty_description(self):
+        for ifstate in six.viewvalues(self.interfaces):
+            if not ifstate.get(Interface.DESCRIPTION):
+                ifstate.pop(Interface.DESCRIPTION, None)
```
EdDev: A non-must one: The first asks if there is a key or it is empty, the second cannot trust the first so it needs to check again.
This is a direct approach:
```
if ifstate.get(Interface.DESCRIPTION) == ''
    del ifstate[Interface.DESCRIPTION]
```

tests/integration/interface_common_test.py
```diff
@@ -0,0 +1,50 @@
...
+from libnmstate.schema import Interface
+
+
+def test_iface_description_removal(eth1_up):
+    desired_state = eth1_up
+    eth1_desired = next(
+        ifstate
+        for ifstate in desired_state[Interface.KEY]
+        if ifstate[Interface.NAME] == 'eth1'
+    )
```
EdDev: This is not needed. The `eth1_up` fixture returns the state of `eth1` only, therefore, there is only one interface in the report. Accessing it can be done directly.

tests/integration/interface_common_test.py
```diff
@@ -0,0 +1,50 @@
...
+    eth1_desired[Interface.DESCRIPTION] = 'bar'
+    libnmstate.apply(desired_state)
+    current_state = libnmstate.show()
+    eth1_current = next(
+        ifstate
+        for ifstate in current_state[Interface.KEY]
+        if ifstate[Interface.NAME] == 'eth1'
+    )
```
EdDev: Please use `show_only`: https://github.com/nmstate/nmstate/blob/5701bbe97e23e66f8966993fbe60b99e1796168b/tests/integration/testlib/statelib.py#L41

Then you can access it directly (as there is only one interface).

tests/integration/interface_common_test.py
```diff
@@ -0,0 +1,50 @@
...
+    assert eth1_current[Interface.DESCRIPTION] == 'bar'
+
+    eth1_desired[Interface.DESCRIPTION] = ''
+    libnmstate.apply(desired_state)
+    current_state = libnmstate.show()
+    eth1_current = next(
+        ifstate
+        for ifstate in current_state[Interface.KEY]
+        if ifstate[Interface.NAME] == 'eth1'
+    )
```
EdDev: Same here

EdDev:
Thank you @ffmancera !

PR #532, opened 2019-10-17
packaging, tox: Upgrade pytest and pytest-cov

EdDev:
The existing pytest* packages used by the container images are broken,
therefore, they are upgraded with this change to the latest (py2
compatible) version.

### DIFF
--- a/packaging/Dockerfile.centos7-nmstate-dev
+++ b/packaging/Dockerfile.centos7-nmstate-dev
@@ -11,7 +11,7 @@ RUN yum -y install \
         dhclient \
     && yum clean all
 
-RUN pip install --upgrade pip pytest==4.2.1 pytest-cov==2.6.1
+RUN pip install --upgrade pip pytest==4.6.6 pytest-cov==2.8.1
 
 RUN pip3 install --user python-coveralls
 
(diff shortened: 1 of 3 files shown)


tyll:
There is now this warning for CentOS 7: Is it related with the new pytest version?

```
=============================== warnings summary ===============================
7241tests/integration/nm/testlib.py:25
7242  /tmp/nmstate/tests/integration/nm/testlib.py:25: PytestCollectionWarning: cannot collect test class 'TestMainloopError' because it has a __init__ constructor (from: tests/integration/nm/ovs_test.py)
7243    class TestMainloopError(Exception):
7244
7245tests/integration/nm/testlib.py:25
7246  /tmp/nmstate/tests/integration/nm/testlib.py:25: PytestCollectionWarning: cannot collect test class 'TestMainloopError' because it has a __init__ constructor (from: tests/integration/nm/wired_test.py)
7247    class TestMainloopError(Exception):
7248
7249-- Docs: https://docs.pytest.org/en/latest/warnings.html

```

tyll:
I created #535 to address the class naming.

EdDev:
> There is now this warning for CentOS 7: Is it related with the new pytest version?

I guess it is. But we can easily handle it later and it does not seem to do any harm (not collecting it in the tests is correct).

> I created 535 to address the class naming.

Lets try to avoid opening github issues internally, we do not manage them well.
I prefer to leave the github issues to external users which are infrequently feedbacking.

We are managing the Jira ones at the moment (barely, we need to improve).

PR #534, opened 2019-10-17
VERSION: Bump to 0.0.9

tyll:

Since 0.0.8 was already released, bump to 0.0.9

---
This change is [Reviewable](https://reviewable.io/reviews/nmstate/nmstate/534)


EdDev:
I prefer to do this:
- Mark the last release version which we released from the current (0.0.8) stable branch with 0.1.0. We have not done a release here.
- Change master to 0.1.1 (instead of 0.0.9).

tyll:
> I prefer to do this:
> 
> * Mark the last release version which we released from the current (0.0.8) stable branch with 0.1.0. We have not done a release here.

I do not understand what you mean there.

> * Change master to 0.1.1 (instead of 0.0.9).

There is a discussion about future branches in #522, maybe we can discuss the future version release scheme there. Using 0.0.9 can easily be changed in the future but using 0.0.8 now makes the COPR RPM packages from git master older than the stable package.

EdDev:
- I want to tag and release the existing stable branch.
- I prefer master from now to be 0.1.1

phoracek:
@EdDev I wonder, why don't you just follow the de-facto standard of semver https://semver.org/? Then you would bump the middle part when you release, the last part of the version should be used when you backport fixes to released version.

EdDev:
@phoracek , this is why I want to mark the existing stable to 0.1.0 and continue on with 0.1.x until we will finally move on to 0.2.x (which is in practice targeted for RHEL/CentOS 8.2)
At the moment I hope to release more frequently for QA and the various projects to consume and test in a tight loop. The mid number can currently be bound to the distribution release (CentOS/RHEL).

I think @cathay4t had a document prepared on this, we should continue the discussion there.

EdDev:
@phoracek , the rules apply to major > 0 versions, i.e. stable API projects.
We have not reached that point yet, we are still adapting to the needs and therefore change the API frequently.
After 1.0.0 the API will have to follow these rules for sure.

My request is to move now and use the minor and patch versions and not only the patch ones. This will allow us to manage the development for additions/fixes to 8.1 and the effort for the next release (targeting 8.2).
I am not very concern about API at the moment as we try not to break backwards compatibility (but at the same time we will not commit to it formally).

EdDev:
@phoracek , I mean exactly that. We are in major == 0, which implies `The public API SHOULD NOT be considered stable`.

phoracek:
@EdDev I'm just trying to make sure to convince you guys (on a wrong PR), to use semver, and not reinvent the wheel. Just place link to semver to docs, done.

EdDev:
@phoracek it is not only the wrong PR, but also too early to decide on.

tyll:
To clarify the problem:

```
[root@fef716f26b49 /]# dnf copr enable nmstate/nmstate-git-fedora

You are about to enable a Copr repository. Please note that this
repository is not part of the main distribution, and quality may vary.

The Fedora Project does not exercise any power over the contents of
this repository beyond the rules outlined in the Copr FAQ at
<https://docs.pagure.org/copr.copr/user_documentation.html#what-i-can-build-in-copr>,
and packages are not held to any quality or security level.

Please do not file bug reports about these packages in Fedora
Bugzilla. In case of problems, contact the owner of this repository.

Do you really want to enable copr.fedorainfracloud.org/nmstate/nmstate-git-fedora? [y/N]: y
Repository successfully enabled.
[root@fef716f26b49 /]# yum install nmstate
Copr repo for nmstate-git-fedora owned by nmstate                                                                                                                           8.6 kB/s |  10 kB     00:01    
Fedora Modular 30 - x86_64                                                                                                                                                   81 kB/s |  25 kB     00:00    
Fedora Modular 30 - x86_64 - Updates                                                                                                                                         38 kB/s |  21 kB     00:00    
Fedora 30 - x86_64 - Updates                                                                                                                                                138 kB/s |  20 kB     00:00    
Fedora 30 - x86_64                                                                                                                                                           50 kB/s |  25 kB     00:00    
Dependencies resolved.
============================================================================================================================================================================================================
 Package                                                   Architecture                               Version                                             Repository                                   Size
============================================================================================================================================================================================================
Installing:
 nmstate       
[...]
```

Instead of installing nmstate from COPR, yum/dnf uses the stable version, since it is newer than the pre-release in COPR (which is a pre-release for 0.0.8 since the VERSION was not bumped).

EdDev:
Please bump it to `0.1.1`.
`0.0.8` was a mistake, we should have use `0.1.0` back then instead and now is a good time to set it correctly.

phoracek:
/me _complains about abuse of the PATCH part of version number_

EdDev:
> /me complains about abuse of the PATCH part of version number

It was noted, but as we are currently managing our work with only two numbers (minor and patch) I see no better solution.
Progressing through the minor release while we are the major==0 is not reasonable, especially when we change the API frequently.

VERSION
```diff
@@ -1 +1 @@
-0.0.8
+0.0.9
```
EdDev: Lets use 0.1.1 please.
0.0.8 was a mistake, we should have use 0.1.0 back then instead and now is a good time to set it correctly.

tyll: You wrote yourself that this is not the place to discuss/implement a new versioning scheme, therefore this is IMHO not a valid objection.

EdDev: It is valid if I consider it wrong.

tyll: You did not explain at all why following the previously agreed upon procedure is suddenly wrong or why even using `0.0.9` for now is wrong. You only wrote that you would prefer an even bigger version increment which continues to be possible even after merging the current PR as-is. And deciding on a new versioning scheme is as you agreed discussed in #522.

Valid objections iMHO would be if the version would be too high (because we cannot easily go back to lower versions) or if the file would be syntactically broken.

EdDev: I will try to explain why `0.0` versions are no longer valid, although I feel that it is a replay from other discussions.

`0.0.8` was a special version, it got added into RHEL 8.1 and became the first version we need to support.
We did a mistake with its naming, as we should have called it 0.1.0 and open the option to have fixes (or enhances) that are tailored to it. After we release 0.0.8 we continued with bug fixes and enhancements, but we could not backport all of them (rebase) to be supported by 0.0.8, only specific changed got backported. At that point the `stable_0.0.8` was created to manage those backports and to be able and release build from it into RHEL. We should have released intermediate builds upstream so the "community" could test it, but we have not.

Consider these platform distributions our milestones, which we need to support them for some time on upstream (at least until the next version of the platform is released).

If such a "milestone" is marked using a minor version ID (e.g. 0.1), we can fix stuff or add backported enhancement from master and at the same time continue with master for the next milestone (e.g. 0.2).

With my nack I claim that `0.0.9` is not a valid version ID for master, and I think it is valid to ask not to see such a version there.

I understand if you cannot trust only my proposal and will prefer to wait for @cathay4t to finalize this. I am ok with waiting.

tyll: > With my nack I claim that `0.0.9` is not a valid version ID for master, and I think it is valid to ask not to see such a version there.

After `0.0.8` was released/tagged, it became an invalid version for the master branch, so before any new commits, `VERSION` needed to be bumped. Since it did not happen, dozens of invalid packages were already published as an invalid pre-release for `0.0.8` but they are all post releases, see:

https://copr.fedorainfracloud.org/coprs/nmstate/nmstate-git-fedora/builds/

So, even if master is never released as 0.0.9, it is not worse than the current situation, since master is also not going to be released as 0.0.8 (again).

Also with every other versioning scheme, there is no guarantee that the version in `VERSION` will be the next tagged released since if we would be using semver, it might be that we bump the patch level after a release but need to bump the minor release as well before doing the next release.

The discussion about which version to use belongs in the other PR and I consider it more important to fix the versioning for COPR for now than to discuss the next versioning before doing this.

PR #538, opened 2019-10-18
packaging: Clarify post-release version bump

tyll:

---
This change is [Reviewable](https://reviewable.io/reviews/nmstate/nmstate/538)

### DIFF
--- a/packaging/README.release.md
+++ b/packaging/README.release.md
@@ -74,7 +74,9 @@ python3 -m twine upload dist/*
 
 ## Post Release
 
-* Create a pull request with increased version number in the `VERSION` file.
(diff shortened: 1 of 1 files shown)


EdDev:
Ahh... everytime we touch the `packaging` folder, images are created :/
Could we exclude specific files from this automatic imaging? I do not like new images on docker-hub to be created just because a README was edited... it causes a mess.

tyll:
> Ahh... everytime we touch the `packaging` folder, images are created :/
> Could we exclude specific files from this automatic imaging? I do not like new images on docker-hub to be created just because a README was edited... it causes a mess.

The easiest solution would be to move the documentation to a different place (we could still add symlinks to make them easier to discover).

PR #542, opened 2019-10-21
schema: Add team interface schema

ffmancera:

### DIFF
--- a/libnmstate/schema.py
+++ b/libnmstate/schema.py
@@ -111,6 +111,7 @@ class InterfaceType(object):
     UNKNOWN = 'unknown'
     VLAN = 'vlan'
     VXLAN = 'vxlan'
+    TEAM = 'team'
 
     VIRT_TYPES = (
         BOND,
@@ -119,6 +120,7 @@ class InterfaceType(object):
         OVS_BRIDGE,
         OVS_PORT,
         OVS_INTERFACE,
+        TEAM,
         VLAN,
         VXLAN,
     )
@@ -316,3 +318,20 @@ class OVSBridge(object):
                 ID_RANGE = 'id-range'
                 MIN_RANGE = 'min'
                 MAX_RANGE = 'max'
+
+
+class Team:
+    TYPE = InterfaceType.TEAM
+    CONFIG_SUBTREE = InterfaceType.TEAM
+
+    PORT_SUBTREE = 'ports'
+    RUNNER_SUBTREE = 'runner'
+
+    class Port:
+        NAME = 'name'
+
+    class Runner:
+        NAME = 'name'
+
+        class RunnerMode:
+            LOAD_BALANCE = 'loadbalance'
--- a/libnmstate/schemas/operational-state.yaml
+++ b/libnmstate/schemas/operational-state.yaml
@@ -25,6 +25,7 @@ properties:
             - "$ref": "#/definitions/interface-dummy/rw"
             - "$ref": "#/definitions/interface-vlan/rw"
             - "$ref": "#/definitions/interface-vxlan/rw"
+            - "$ref": "#/definitions/interface-team/rw"
   routes:
     type: object
     properties:
@@ -492,6 +493,28 @@ definitions:
                     type: boolean
                   state:
                     type: string
+  interface-team:
+    rw:
+      properties:
+        type:
+          type: string
+          enum:
+          - team
+        team:
+          type: object
+          properties:
+            ports:
+              type: array
+              items:
+                type: object
+                properties:
+                  name:
+                    type: string
+            runner:
+              type: object
+              properties:
+                name:
+                  type: string
 
   route:
     type: object
--- a/tests/lib/schema_validation_test.py
+++ b/tests/lib/schema_validation_test.py
@@ -31,12 +31,15 @@ from libnmstate.schema import InterfaceType
 from libnmstate.schema import LinuxBridge as LB
 from libnmstate.schema import OVSBridge
 from libnmstate.schema import RouteRule
+from libnmstate.schema import Team
 from libnmstate.schema import VXLAN
+
 
 INTERFACES = Constants.INTERFACES
 ROUTES = Constants.ROUTES
 THE_BRIDGE = 'br0'
 VXLAN0 = 'vxlan0'
+TEAM0 = 'team0'
 
 COMMON_DATA = {
     INTERFACES: [
@@ -274,6 +277,64 @@ class TestIfaceTypeVxlan(object):
     def test_no_config_is_valid(self, default_data):
         default_data[Interface.KEY].append(
             {Interface.NAME: VXLAN0, Interface.TYPE: VXLAN.TYPE}
+        )
+
+        libnmstate.validator.validate(default_data)
+
+
+class TestIfaceTypeTeam(object):
+    def test_valid_team_without_options(self, default_data):
+        default_data[Interface.KEY].append(
+            {
+                Interface.NAME: TEAM0,
+                Interface.TYPE: Team.TYPE,
+                Team.CONFIG_SUBTREE: {},
+            }
+        )
+
+        libnmstate.validator.validate(default_data)
+
+    def test_valid_team_with_ports(self, default_data):
+        default_data[Interface.KEY].append(
+            {
+                Interface.NAME: TEAM0,
+                Interface.TYPE: Team.TYPE,
+                Team.CONFIG_SUBTREE: {
+                    Team.PORT_SUBTREE: [{Team.Port.NAME: 'eth1'}]
+                },
+            }
+        )
+
+        libnmstate.validator.validate(default_data)
+
+    def test_valid_team_with_runner(self, default_data):
+        default_data[Interface.KEY].append(
+            {
+                Interface.NAME: TEAM0,
+                Interface.TYPE: Team.TYPE,
+                Team.CONFIG_SUBTREE: {
+                    Team.PORT_SUBTREE: [],
+                    Team.RUNNER_SUBTREE: {
+                        Team.Runner.NAME: Team.Runner.RunnerMode.LOAD_BALANCE
+                    },
+                },
+            }
+        )
+
+        libnmstate.validator.validate(default_data)
+
+    def test_valid_team_with_ports_and_runner(self, default_data):
+        default_data[Interface.KEY].append(
+            {
+                Interface.NAME: TEAM0,
+                Interface.TYPE: Team.TYPE,
+                Team.CONFIG_SUBTREE: {
+                    Team.PORT_SUBTREE: [{Team.Port.NAME: 'eth1'}],
+                    Team.RUNNER_SUBTREE: {
+                        Team.Runner.NAME: Team.Runner.RunnerMode.LOAD_BALANCE
+                    },
+                },
+            }
         )
 
         libnmstate.validator.validate(default_data)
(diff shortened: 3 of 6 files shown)


tests/lib/schema_validation_test.py
```diff
@@ -151,6 +154,57 @@ def test_valid_without_auto_neg_and_missing_speed(self, default_data):
...
+
+        libnmstate.validator.validate(default_data)
+
+    def test_valid_team_without_slaves(self, default_data):
+        default_data[Interface.KEY].append(
+            {
+                Interface.NAME: TEAM0,
+                Interface.TYPE: Team.TYPE,
+                Team.CONFIG_SUBTREE: {
+                    Team.CONFIG: '{\"runner\": {\"name\": \"broadcast\"}}'
```
phoracek: I know that team API uses plain JSON to carry configuration, however, that should not be exposed in nmstate.
The API here should be well defined and part of the YAML schema.

Would it be possible to expose these JSON options (or a subset of them) explicitly in the schema?

ffmancera: Yes, it is possible but then in the implementation we would need to convert it to plain JSON. Is that fine? If it is, I will work on that.

phoracek: I think that's all right. We should make it simple and pretty for nmstate users, hide JSON config from them. Selecting specific configuration options could also help us identify "how" will be team interfaces typically used.

ffmancera: Yes, this approach is more user friendly. Thanks!

libnmstate/schema.py
```diff
@@ -210,3 +212,80 @@ class OVSBridge(object):
...
+
+    SLAVES = 'slaves'
+    CONFIG = 'config'
+
+
+class TeamConfig(object):
+    CONFIG = Team.CONFIG
+
+    DEVICE = 'device'
+    DEBUG_LEVEL = 'debug_level'
```
phoracek: Please use `-` instead of `_`.

libnmstate/schema.py
```diff
@@ -210,3 +212,80 @@ class OVSBridge(object):
...
+    HWADDR = 'hwaddr'
+
+    RUNNER = 'runner'
```
phoracek: That is a lot of parameters.

@EdDev what do you usually do in nmstate, do you keep minimal useful subset or is it fine to have all those?

ffmancera: Yes it is. If we split them into subsets it will result in a lot of classes. Teamd JSON API have a lot of fields and possible values.

See also: man 5 teamd.conf

libnmstate/schema.py
ffmancera: We are using `_` for all the other constants. Why should we use `-` for team interface constants?

EdDev: Which API key has an underscore?

ffmancera: Not the API key string but the constant name. If we replace `_` by `-` in all API key strings, then we would need to replace `-` by `_` again before send the plain text JSON to NetworkManager. Is that okay with you?

EdDev: Sorry, I do not understand what you are referring to. A constant is python is a variable name, which cannot use the `-` character. The original comment was on the value of the variable/constant, which is the string used on the json-schema as a key.

ffmancera: Yes, I was meaning the value of the variable. The teamd JSON API uses underscore for all the keys. e.g `debug_level`. In order to set the team configuration we need to send the plain text JSON to NetworkManager so, we would need to replace all `-` by `_`.  This is why I've decided to use underscore in first place. All teamd json keys are listed here `man  5 teamd.conf`.

I hope it's more clear now. Thanks!

EdDev: Ok, now I understand, thank you for the clarification.
The Nmstate API is not trying to mimic NM API as a general rule. We may mimic or follow naming and models from NM if it fits the needs.

In this case, we use all over the `-` alternative, which is more user friendly than the `_` option (which is developer friendly). We should be consistent here.
In addition, the exposed API should fit the declarative nature, so please do not assume every parameter deserves exposing.

ffmancera: That seems fine to me. Thanks for explaining

ffmancera: Removed debug stuff and added more tests in order to cover all the configuration parameters.

libnmstate/schema.py
cathay4t: You may use the way we did for bond options.

This is not a good practice to include every options in schema, just allow arbitrary dictionary and depend on backend(NM) to check it.

libnmstate/schema.py
```diff
@@ -222,3 +224,82 @@ class OVSBridge(object):
...
     SYSTEM = 'system'
     INTERNAL = 'internal'
+
+
+class Team(object):
+    TYPE = InterfaceType.TEAM
+    CONFIG_SUBTREE = 'team'
+
+    SLAVES = 'slaves'
+    CONFIG = 'config'
```
cathay4t: Since we don't have actual use case on Team yet, maybe we just start with the minimum `slaves` as the `json config`
is too complex do design without use case.

Will `slaves` only works?

libnmstate/schema.py
tyll: Why is it not a good practice? It seems to be more user-friendly to fail as quick as possibly if there is an error and the first checks are the schema validation.

@EdDev what is your opinion on this?

libnmstate/schema.py
ffmancera: Yes it works, done!

libnmstate/schema.py
cathay4t: Whenever team changed their option or add more, we need to sync with that list.

If team decide to break back compatibility, with that hard coded in schema, we also need to
break our back compatibility.

Anyway, we don't need to discuss on this as this PR should only support `slaves` option in
the team interface.

ffmancera:
Rebase done. Thanks for reviewing, now working on the implementation.

tests/lib/schema_validation_test.py
```diff
@@ -247,6 +250,30 @@ def test_no_config_is_valid(self, default_data):
         libnmstate.validator.validate(default_data)
 
 
+class TestIfaceTypeTeam(object):
+    def test_valid_team_without_options(self, default_data):
```
EdDev: You mean "without slaves"?

libnmstate/schemas/operational-state.yaml
```diff
@@ -473,6 +474,20 @@ definitions:
...
+          type: string
+          enum:
+          - team
+        team:
+          type: object
+          properties:
+            slaves:
+              type: array
+              items:
+                type: string
```
EdDev: I think it needs to be an object and not a string.
Team ports have multiple options base on `nm-settings` man page.

ffmancera:
We have scheduled a meeting to discuss this.

libnmstate/schema.py
```diff
@@ -117,6 +118,7 @@ class InterfaceType(object):
         OVS_INTERFACE,
         VLAN,
         VXLAN,
+        TEAM,
```
tyll: IMHO it woudl be nicer to  keep this list sorted.

libnmstate/schema.py
```diff
@@ -247,3 +249,18 @@ class OVSBridge(object):
...
+
+    PORTS_SUBTREE = 'ports'
+    RUNNER_SUBTREE = 'runner'
+
+    class Ports(object):
+        NAME = 'name'
+
+    class Runner(object):
+        NAME = 'name'
+        LOAD_BALANCE = 'loadbalance'
```
tyll: Since it is one of several possible values, it might make more sense to move this into a `TeamRunner` class or similar like we do it for interface types. It is hard to say with this kind of partially defined schema IMHO. What do you think?

libnmstate/schema.py
```diff
@@ -247,3 +249,18 @@ class OVSBridge(object):
     PORT_NAME = 'name'
     PORT_VLAN_MODE = 'vlan-mode'
     PORT_ACCESS_TAG = 'access-tag'
+
+
+class Team(object):
+    TYPE = InterfaceType.TEAM
+    CONFIG_SUBTREE = InterfaceType.TEAM
+
+    PORTS_SUBTREE = 'ports'
```
tyll: OVSBridge uses singular `port` instead of `ports`, I guess we cannot change the value since it comes from the schema but since the constant names can be different, maybe use `PORT_SUBTREE` and `class Port` to keep it uniform for Nmstate. What do you think?

libnmstate/schemas/operational-state.yaml
```diff
@@ -480,6 +481,30 @@ definitions:
...
+  interface-team:
+    rw:
+      properties:
+        type:
+          type: string
+          enum:
+          - team
+        team:
+          type: object
+          properties:
```
tyll: Just for the record, my preference would be not to define some of the team options if we do not want to define all of them because this might depend too much on the team implementation. Instead moving this information to capabilities seems like a better idea so clients can still check which team features are supported. But this is probably too much now.

libnmstate/schema.py
ffmancera: Sure!

libnmstate/schema.py
ffmancera: You are right, do you like something like `RunnerMode`?

libnmstate/schema.py
ffmancera: That seems fine. Thanks!

libnmstate/schemas/operational-state.yaml
ffmancera: I will create a card on Jira with that task. I think we should offer a way for the users to check which team features are supported. Would you like to discuss that on the future?

libnmstate/schema.py
ffmancera: Done!

ffmancera:
Rebased.

libnmstate/schema.py
```diff
@@ -290,3 +292,20 @@ class OVSBridge(object):
     PORT_NAME = 'name'
     PORT_VLAN_MODE = 'vlan-mode'
     PORT_ACCESS_TAG = 'access-tag'
+
+
+class Team(object):
```
EdDev: No need for `object`.

libnmstate/schemas/operational-state.yaml
```diff
@@ -480,6 +481,30 @@ definitions:
...
+                properties:
+                  name:
+                    type: string
+            runner:
+              type: object
+              properties:
+                name:
+                  type: string
+                  enum:
+                  - loadbalance
```
EdDev: I see no reason to limit it in the schema.

libnmstate/schema.py
ffmancera: If there is no need for object, why are we using it in the other schema classes? Thanks!

libnmstate/schemas/operational-state.yaml
ffmancera: I am fine with removing this for now. I am preparing a little patch to discuss about this (this patch is not going to break API compatibility).

Thanks!

libnmstate/schema.py
ffmancera: Okay, this is a leftover from python2 support. I am removing it. Thanks!

cathay4t: Just leftover of python2 stuff.

PR #545, opened 2019-10-22
Compress static IPv6 addresses

maiqueb:
Setting a static IP address on an interface fails, since it is reported by nmstate in the canonical format, as per [0].

The failure happens in nmstate's *verify* stage - since the desired state is expressed in the expanded notation, and the current state in the canonical notation, nmstate considers the IPs to be different, and an automatic rollback is used.

This patch addresses the aforementioned issue by canonicalizing (compressing) the static ipv6 address in the desired state.

The issue can be triggered w/ the following yaml:
```yaml
interfaces:
- name: eth1
  type: ethernet
  state: up
  ipv4:
    enabled: false
  ipv6:
    address:
    - ip: 2001:0db8:85a3:0000:0000:8a2e:0370:7331
      prefix-length: 64
    enabled: true
```

Ref: https://bugzilla.redhat.com/1760800

[0] - https://tools.ietf.org/html/rfc5952#section-4

### DIFF
--- a/libnmstate/state.py
+++ b/libnmstate/state.py
@@ -27,6 +27,7 @@ except ImportError:
 from collections import defaultdict
 import copy
 from functools import total_ordering
+from ipaddress import ip_address
 from operator import itemgetter
 
 import six
@@ -449,6 +450,17 @@ class State(object):
                 }
             }
             dict_update(new_state, ifstate)
+            ipv6_addrs = new_state[Interface.IPV6].get(
+                InterfaceIPv6.ADDRESS, []
+            )
+            if ipv6_addrs:
+                compressed_addresses = [
+                    canonicalize_ipv6_addr(ipv6_addr)
+                    for ipv6_addr in ipv6_addrs
+                ]
+                new_state[Interface.IPV6][
+                    InterfaceIPv6.ADDRESS
+                ] = compressed_addresses
             self._ifaces_state[ifstate[Interface.NAME]] = new_state
 
     def _remove_iface_ipv6_link_local_addr(self):
@@ -558,6 +570,18 @@ def _validate_routes(
                 raise NmstateValueError(
                     'Cannot set IPv4 route when IPv4 is disabled'
                 )
+
+
+def canonicalize_ipv6_addr(ipv6_address_prefix_len):
+    ipv6_address = ipv6_address_prefix_len[InterfaceIPv6.ADDRESS_IP]
+    ipv6_iface = ip_address(six.u(ipv6_address))
+
+    return {
+        InterfaceIPv6.ADDRESS_IP: str(ipv6_iface),
+        InterfaceIPv6.ADDRESS_PREFIX_LENGTH: ipv6_address_prefix_len[
+            InterfaceIPv6.ADDRESS_PREFIX_LENGTH
+        ],
+    }
 
 
 def _get_iface_enable_states(desire_state, current_state):
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,6 +6,7 @@
 #packageD!=0.13.0,<0.14,>=0.12.0
 
 dbus-python
+ipaddress;python_version<"3.3"
 jsonschema
 PyGObject
 PyYAML
--- a/tests/integration/static_ip_address_test.py
+++ b/tests/integration/static_ip_address_test.py
@@ -20,8 +20,11 @@
 import pytest
 
 import libnmstate
+from libnmstate.schema import Interface
 from libnmstate.schema import InterfaceIPv4
 from libnmstate.schema import InterfaceIPv6
+from libnmstate.schema import InterfaceState
+from libnmstate.schema import InterfaceType
 
 from .testlib import assertlib
 from .testlib import statelib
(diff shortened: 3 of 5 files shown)


tests/integration/testlib/statelib.py
```diff
@@ -117,13 +118,30 @@ def normalize(self):
...
         self._sort_interfaces_by_name()
+        self.sanitize_ipv6()
 
     def match(self, other):
         return _state_match(self.state, other.state)
 
     def _sort_interfaces_by_name(self):
         self._state[INTERFACES].sort(key=lambda d: d['name'])
 
+    def sanitize_ipv6(self):
```
EdDev: How about naming this based on what it does specifically?
We already have specific "sanitize" changes like: `_ipv6_skeleton_canonicalization`.

Also note that this probably fits as a private method.

tests/integration/testlib/statelib.py
```diff
@@ -117,13 +118,30 @@ def normalize(self):
...
+        for iface_state in self.state[Interface.KEY]:
+            if Interface.IPV6 in iface_state:
```
EdDev: We can ask and do less lookups:
```
ipv6_state = iface_state.get(Interface.IPV6)
if ipv6_state:
    ...
```

Same for other cases below.

tests/integration/testlib/statelib.py
```diff
@@ -117,13 +118,30 @@ def normalize(self):
...
+                    iface_ipv6_addresses_state = iface_ipv6_state[
+                        InterfaceIP.ADDRESS
+                    ]
+                    compressed_ipv6_addrs = [
+                        compress_ipv6_addr(iface_ipv6_addr)
+                        for iface_ipv6_addr in iface_ipv6_addresses_state
+                    ]
+                    iface_ipv6_state[
+                        InterfaceIP.ADDRESS
+                    ] = compressed_ipv6_addrs
```
EdDev: Can you please check if this will read better?
```
iface_ipv6_state[InterfaceIP.ADDRESS] = [
    compress_ipv6_addr(iface_ipv6_addr)
    for iface_ipv6_addr in iface_ipv6_addresses_state
]
```

maiqueb: could be, at least it makes better use of the 80 char line limit.

I'll update - removing the variable, since we never re-use it.

tests/integration/testlib/statelib.py
maiqueb: using this method as 'public' I manage to use it from the statelib - check [0]. I also need to update the statelib to sanitize the addresses.

Having to kind of duplicate that in the statelib is what I least like ... But I couldn't come up with anything better that fit the existent design.

[0] - https://github.com/nmstate/nmstate/pull/545/files#diff-e631692ae7126c6af4eabc3757dbbde0R138

tests/integration/static_ip_address_test.py
```diff
@@ -539,7 +539,6 @@ def test_add_iface_with_same_static_ipv6_address_to_existing(
     assertlib.assert_state(desired_state)
 
 
-@pytest.mark.xfail(reason='https://bugzilla.redhat.com/1760800', strict=True)
```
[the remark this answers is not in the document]
maiqueb: Fair enough; I tried to separate the tests statelib's changes from the one's in the state module (production code).

Are you OK if I squash the last commit ?

maiqueb:
> Thank you, we need this one!
> 
> Regarding `state, static ipv6: compress ipv6 addrs`, please cover the scenario through at least one test in the `state_test` unit test module.

I thought that since I'm simply using a library to compress the addresses it wouldn't be needed, but, I'm OK with that. 

The more (tests) the better :)

tests/integration/static_ip_address_test.py
```diff
@@ -537,3 +537,40 @@ def test_add_iface_with_same_static_ipv6_address_to_existing(
     libnmstate.apply(desired_state)
 
     assertlib.assert_state(desired_state)
+
+
+@pytest.mark.xfail(reason='https://bugzilla.redhat.com/1760800', strict=True)
+def test_add_iface_with_static_ipv6_expanded_format(eth1_up):
+    ipv6_addr_lead_zeroes = '2001:0db8:85a3:0000:0000:8a2e:0370:7331'
+    desired_state = {
+        INTERFACES: [
```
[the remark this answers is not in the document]
maiqueb: I can compromise and at least move us a bit in the right direction by not adding more keys that will need to be updated :)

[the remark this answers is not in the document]
ffmancera: > We no longer use this constant, but it will be unfair to ask to use the alternative now.
> There are also usages of non constant for the interface keys (`name`, `type`, etc...).
> 
> @ffmancera , I hope you can help to change this throughout the project.

Yes, I've sent a PR #549.

requirements.txt
```diff
@@ -6,6 +6,7 @@
 #packageD!=0.13.0,<0.14,>=0.12.0
 
 dbus-python
+ipaddress;python_version<"3.3"
```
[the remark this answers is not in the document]
cathay4t: @EdDev I am OK with it.

libnmstate/state.py
```diff
@@ -554,6 +561,18 @@ def _validate_routes(
                 )
 
 
+def compress_ipv6_addr(ipv6_address_prefix_len):
```
cathay4t: This function does not actually need the `ADDRESS_PREFIX_LENGTH`.
You may simplify it to just take the ip address string as input parameter.

EdDev:
> I thought that since I'm simply using a library to compress the addresses it wouldn't be needed, but, I'm OK with that.

One is an integration/functional test and the other a unit test. The first checks the libnmstate API while the second checks a unit inside the code.
The unit has also a closer feedback loop.

Regarding the test assert helper... I also dislike it already, but it helps test "everything" for each scenario. Probably we should change the strategy here and use `match` instead with the parameters we are interested in. But if feels like we are loosing something with whatever path we take.

libnmstate/state.py
EdDev: I think that @maiqueb wanted to convert the address dict as a whole in order to simplify the usage:
`ifstate['ipv6'][InterfaceIPv6.ADDRESS] = [compress_ipv6_addr(ipv6_addr) for ipv6_addr in ipv6_addrs]`

I guess if he will use a for-loop at the caller, he can avoid this.
I am ok with both, it reads well IMO.

maiqueb: I think the caller reads better this way.

tyll:
Thank you! It should be called canonicalization instead of compression
because it is the canonical representation as mentioned in
https://tools.ietf.org/html/rfc5952#section-4

Using a different name for this is confusing for me.

Miguel Duarte Barroso <notifications@github.com> schrieb am Di., 22. Okt.
2019, 15:06:

maiqueb:
> Thank you! It should be called canonicalization instead of compression because it is the canonical representation as mentioned in https://tools.ietf.org/html/rfc5952#section-4 Using a different name for this is confusing for me. 

@tyll are you OK with it now ?

Miguel Duarte Barroso <notifications@github.com> schrieb am Di., 22. Okt. 2019, 15:06:

EdDev:
@tyll , I think that @maiqueb addressed your request.
If we missed anything, please let us know and we will fix it.

Merging this PR..

PR #548, opened 2019-10-23
gitignore: ignore htmlcov-py37 directory

ffmancera:
When the unit_py37 test fails it generates the coverage under the
htmlcov-py37 directory. We do not want to track this on git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,8 +9,7 @@ __pycache__/
 dist/
 
 .coverage
-htmlcov-py27/
-htmlcov-py36/
+htmlcov-*/
 
 AUTHORS
 ChangeLog


.gitignore
```diff
@@ -11,6 +11,7 @@ dist/
 .coverage
 htmlcov-py27/
 htmlcov-py36/
+htmlcov-py37/
```
EdDev: Can we just ignore all future ones as well?
Something like this: `htmlcov-*`? (not sure if this works).

ffmancera: Yes, something like htmlcov-*/ should work. I will try it and let you know.

ffmancera: Done!

PR #555, opened 2019-10-25
state, timeout: Increase main timeout to 60s

qinqon:
At kstate we have found that recovering from a reboot 35s is not enough
and 60s is fine.

### DIFF
--- a/libnmstate/netapplier.py
+++ b/libnmstate/netapplier.py
@@ -38,7 +38,7 @@ from libnmstate.error import NmstatePermissionError
 from libnmstate.error import NmstateValueError
 from libnmstate.nm import nmclient
 
-MAINLOOP_TIMEOUT = 35
+MAINLOOP_TIMEOUT = 60
 
 
 def apply(desired_state, verify_change=True, commit=True, rollback_timeout=60):


phoracek:
Please explain when 35 seconds was not enough.

EdDev:
The reasoning here is weak because it does not include the reason it takes so long to activate the profile/interface.
As an example, the 35 figure was updated due to STP on bridges.

Using such a high timeout due to one scenario is also a problem.
If there are specific scenarios in which the user knows it may require a larger timeout, it may be appropriate to let the user control it.
What do you say about adding it as an argument? I am a bit worried about too many arguments, but that is a technicality we can discuss in parallel.

phoracek:
@EdDev argument would be good enough, until more people hit that issue and we will be able to track it down to its source. Maybe we can just configure it via undocumented environment variable while discussing the API?

For kubernetes-nmstate, we have a temporary workaround, although it is not the nicest one https://github.com/nmstate/kubernetes-nmstate/pull/238/files

EdDev:
> Maybe we can just configure it via undocumented environment variable while discussing the API?

We could, but that is also an API discussion.
I suggest you raise it in the upstream mailing list.
Another option is passing a data structure or dict with configuration values.

> For kubernetes-nmstate, we have a temporary workaround

Very sad.

Not investigating the actual reason for such a long delay has the risk of masking bugs.

tyll:
What is the desired state that needs the longer timeout?

tyll:
Also when thinking about the timeout, it should probably depend on the number of interfaces or changes to be done. So before exposing timeout settings it should be clear what kind of timeouts we want to expose.

tyll:
> If there are specific scenarios in which the user knows it may require a larger timeout, it may be appropriate to let the user control it.

Actually when thinking about the design idea of Nmstate it seems to me that the user should not be concerned with knowing about timeouts.

phoracek:
@tyll we set up default bridged network. Then we rebooted and then we tried to apply the same config again (although the configuration was persisted and already running). During this second configuration it timed out. Two interfaces were affected (bridge and its port).

EdDev:
@qinqon , @phoracek , in this form this change has little chance to get in.
The problem which you have detected should be tracked though a BZ or Jira ticket to find out the source problem.
I also think an RFE is reasonable here to allow the client to control this timeout.

@qinqon , are there such tickets opened already? If so, lets close this PR for now.

qinqon:
Didn't have time to open it, I think we can close this and do a proper proporsal with RFE.

PR #559, opened 2019-10-29
nm.device: support reapply on MTU changes

ffmancera:
Support reapply on MTU changes as it is supported by NetworkManager
again.

Ref: https://bugzilla.redhat.com/1702657

### DIFF
--- a/libnmstate/nm/device.py
+++ b/libnmstate/nm/device.py
@@ -23,6 +23,8 @@ from . import active_connection as ac
 from . import connection
 from . import nmclient
 
+from distutils.version import StrictVersion
+
 
 def activate(dev=None, connection_id=None):
     """Activate the given device or remote connection profile."""
@@ -186,7 +188,9 @@ def _modify_callback(src_object, result, user_data):
 
 
 def _requires_activation(dev, connection_profile):
-    if _mtu_changed(dev, connection_profile):
+    if StrictVersion(nmclient.nm_version()) < StrictVersion(
+        '1.18'
+    ) and _mtu_changed(dev, connection_profile):
         logging.debug(
             'Device reapply does not support mtu changes, '
             'fallback to device activation: dev=%s',
--- a/libnmstate/nm/nmclient.py
+++ b/libnmstate/nm/nmclient.py
@@ -50,6 +50,10 @@ def can_disable_ipv6():
     return _can_disable_ipv6
 
 
+def nm_version():
+    return NM.Client.get_version(client())
+
+
 def client(refresh=False):
     global _nmclient
     # refresh is a workaround to get the current state when GMainLoop is not
--- a/tests/integration/ethernet_mtu_test.py
+++ b/tests/integration/ethernet_mtu_test.py
@@ -31,6 +31,7 @@ from libnmstate.error import NmstateVerificationError
 
 from .testlib import assertlib
 from .testlib import statelib
+from .testlib.iproutelib import ip_monitor_assert_stable_link_up
 from .testlib.vlan import vlan_interface
 
 
@@ -155,3 +156,14 @@ def test_set_mtu_on_two_vlans_with_a_shared_base(eth1_up):
         libnmstate.apply(desired_state)
 
         assertlib.assert_state(desired_state)
+
+
+@ip_monitor_assert_stable_link_up('eth1')
+def test_change_mtu_with_stable_link_up(eth1_up):
+    desired_state = statelib.show_only(('eth1',))
+    eth1_desired_state = desired_state[Interface.KEY][0]
+    eth1_desired_state[Interface.MTU] = 1900
+
+    libnmstate.apply(desired_state)
+
+    assertlib.assert_state(desired_state)
--- a/tests/integration/nm/wired_test.py
+++ b/tests/integration/nm/wired_test.py
@@ -19,6 +19,8 @@
 
 import pytest
 
+from distutils.version import StrictVersion
+
 from libnmstate import nm
 from libnmstate import schema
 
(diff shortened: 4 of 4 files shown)


ffmancera:
The CI tests will fail because we need to update the container images. We need to wait for #543.

tests/integration/ethernet_mtu_test.py
```diff
@@ -30,6 +30,7 @@
 from libnmstate.error import NmstateVerificationError
 
 from .testlib import assertlib
+from .testlib.iproutelib import ip_monitor_assert_stable_link_up
```
EdDev: Please sort the imports. I think you need to swap with line 34.

ffmancera:
> Therefore, an easy alternative now could be:
> `if nmclient.nm_version() < 1.18 and _mtu_changed(...) `
> Try to check if you can read the version of NM from its API.

It seems fine to me but then we would need to do the same thing on the integration test, right? Thanks!

EdDev:
> It seems fine to me but then we would need to do the same thing on the integration test, right?

On the test we could use a xfail decorator with the version as the condition. It will also be better to limit the xfail to the monitor failure, for that we need a dedicated exception instead of a plain assert.

EdDev:
There is also a nm specific test that should have failed now:
https://github.com/nmstate/nmstate/blob/5a9f3dd7cd10f67cffcea45a18ebc6dc193668be/tests/integration/nm/wired_test.py#L36

If it passes with the changes, something is wrong.

tests/integration/ethernet_mtu_test.py
ffmancera: Done! Thanks

tests/integration/ethernet_mtu_test.py
```diff
@@ -155,3 +156,14 @@ def test_set_mtu_on_two_vlans_with_a_shared_base(eth1_up):
         libnmstate.apply(desired_state)
 
         assertlib.assert_state(desired_state)
+
+
+@ip_monitor_assert_stable_link_up('eth1')
```
EdDev: I think you need an xfail over this one, as the monitor will pass only with ver >= 1.18.

EdDev:
This is no longer dependent on a different PR, right?

libnmstate/nm/nmclient.py
```diff
@@ -50,6 +50,11 @@ def can_disable_ipv6():
     return _can_disable_ipv6
 
 
+def nm_version():
+    client()
+    return NM.Client.get_version(_nmclient)
```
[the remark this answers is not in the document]
ffmancera: You are right, I forgot to remove that. Thanks!

libnmstate/nm/device.py
```diff
@@ -186,7 +186,9 @@ def _modify_callback(src_object, result, user_data):
 
 
 def _requires_activation(dev, connection_profile):
-    if _mtu_changed(dev, connection_profile):
+    if nmclient.nm_version() < '1.18' and _mtu_changed(
```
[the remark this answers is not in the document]
ffmancera: I propose to use the `packaging` module and do something like this.
```
from packaging import version
...
if version.parse(nmclient.nm_version()) < version.parse('1.18'):
```

What do you think?

tests/integration/ethernet_mtu_test.py
ffmancera: I think there is something wrong with `ip_monitor_assert_stable_link_up` the test passed in the CI.

ffmancera:
> This is no longer dependent on a different PR, right?

Yes, with this approach CI passed.

libnmstate/nm/device.py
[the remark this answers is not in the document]
tyll: https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python contains some guidance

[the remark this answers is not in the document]
ffmancera: > https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python contains some guidance

I've been reading about setuptools and maybe in our case it could be useful because NetworkManager does not follow PEP.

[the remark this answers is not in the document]
EdDev: I do not think that `packaging` is a stdlib and depending on it during runtime may not be a good thing.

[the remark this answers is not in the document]
ffmancera: We have decided to use distutils.version module as it is stdlib.

tests/integration/ethernet_mtu_test.py
ffmancera: There is no need to use xfail here. Information related to this added on commit message.

tests/integration/ethernet_mtu_test.py
```diff
@@ -155,3 +156,14 @@ def test_set_mtu_on_two_vlans_with_a_shared_base(eth1_up):
...
+def test_change_mtu_with_stable_link_up():
+    desired_state = statelib.show_only(('eth1',))
```
EdDev: The ‘eth1_up’ fixture has not been used here, therefore, ‘eth1’ may not even be UP at this point.
Please use the fixture.

EdDev:
> because there is not link DOWN or UNKNOWN events

not —> no

tests/integration/ethernet_mtu_test.py
ffmancera: Oops, fixes. Thanks!

ffmancera:
> Commit message:
> 
> > because there is not link DOWN or UNKNOWN events
> 
> not —> no

Fixed. Thanks!

PR #562, opened 2019-10-29
schema: add OVS link aggregation

phoracek:

### DIFF
--- a/libnmstate/schema.py
+++ b/libnmstate/schema.py
@@ -319,6 +319,15 @@ class OVSBridge(object):
                 MIN_RANGE = 'min'
                 MAX_RANGE = 'max'
 
+        LINK_AGGREGATION_SUBTREE = 'link-aggregation'
+
+        class LinkAggregation:
+            MODE = 'mode'
+            SLAVES_SUBTREE = 'slaves'
+
+            class Slave:
+                NAME = 'name'
+
 
 class Team:
     TYPE = InterfaceType.TEAM
--- a/libnmstate/schemas/operational-state.yaml
+++ b/libnmstate/schemas/operational-state.yaml
@@ -333,6 +333,18 @@ definitions:
                         $ref: "#/definitions/types/bridge-vlan-tag"
                       enable-native:
                         type: boolean
+                  link-aggregation:
+                    type: object
+                    properties:
+                      mode:
+                        type: string
+                      slaves:
+                        type: array
+                        items:
+                          type: object
+                          properties:
+                            name:
+                              type: string
             options:
               type: object
               properties:
--- a/tests/integration/testlib/ovslib.py
+++ b/tests/integration/testlib/ovslib.py
@@ -47,6 +47,17 @@ class Bridge(object):
     def add_system_port(self, name):
         self._add_port(name)
 
+    def add_link_aggregation_port(self, name, slaves, mode=None):
+        self._add_port(name)
+        port = self._get_port(name)
+        port[OVSBridge.Port.LINK_AGGREGATION_SUBTREE] = {
+            OVSBridge.Port.LinkAggregation.SLAVES_SUBTREE: slaves
+        }
+        if mode:
+            port[OVSBridge.Port.LINK_AGGREGATION_SUBTREE][
+                OVSBridge.Port.LinkAggregation.MODE
+            ] = mode
+
     def add_internal_port(self, name, ipv4_state):
         self._add_port(name)
         self._ifaces.append(
@@ -61,6 +72,14 @@ class Bridge(object):
         self._bridge_iface[OVSBridge.CONFIG_SUBTREE].setdefault(
             OVSBridge.PORT_SUBTREE, []
         ).append({OVSBridge.PORT_NAME: name})
+
+    def _get_port(self, name):
+        ports = self._bridge_iface[OVSBridge.CONFIG_SUBTREE].get(
+            OVSBridge.PORT_SUBTREE, []
+        )
+        return next(
+            (port for port in ports if port[OVSBridge.PORT_NAME] == name), None
+        )
 
     @contextmanager
     def create(self):
--- a/tests/lib/schema_validation_test.py
+++ b/tests/lib/schema_validation_test.py
@@ -632,3 +632,28 @@ class TestRouteRules(object):
         with pytest.raises(js.ValidationError) as err:
             libnmstate.validator.validate(default_data)
         assert 'main' in err.value.args[0]
+
+
+class TestOVSBridgeLinkAggregation:
+    def test_valid_link_aggregation_port(self, default_data):
+        link_aggregation_port = {
+            OVSBridge.PORT_NAME: 'bond',
+            OVSBridge.Port.LINK_AGGREGATION_SUBTREE: {
+                OVSBridge.Port.LinkAggregation.MODE: 'bond-mode',
+                OVSBridge.Port.LinkAggregation.SLAVES_SUBTREE: [
+                    {OVSBridge.Port.LinkAggregation.Slave.NAME: 'iface1'},
+                    {OVSBridge.Port.LinkAggregation.Slave.NAME: 'iface2'},
+                ],
+            },
+        }
+        default_data[Interface.KEY].append(
+            {
+                Interface.NAME: 'bridge',
+                Interface.TYPE: InterfaceType.OVS_BRIDGE,
+                OVSBridge.CONFIG_SUBTREE: {
+                    OVSBridge.PORT_SUBTREE: [link_aggregation_port]
+                },
+            }
+        )
+
+        libnmstate.validator.validate(default_data)


phoracek:
@EdDev @cathay4t @tyll could you please take a look?

libnmstate/schemas/operational-state.yaml
```diff
@@ -287,6 +287,23 @@ definitions:
                     type: string
                   access-tag:
                     type: string
+                link-aggregation:
```
EdDev: I do not understand the indentation. Is it supposed to be at the same level as the previous `access-tag` key?

But the main question at this level is: How will we differentiate between a regular port and a link-aggregation one? Please consider that nmstate likes partial editing in general.

libnmstate/schemas/operational-state.yaml
```diff
@@ -287,6 +287,23 @@ definitions:
...
+                  type: object
+                  required:
+                  - mode
+                  - slaves
```
EdDev: If both are required, we cannot do partial editation (now or in the future).
I suggest you leave it up to the implementation to validate this, per its ability.

libnmstate/schemas/operational-state.yaml
```diff
@@ -287,6 +287,23 @@ definitions:
...
+                  properties:
+                    mode:
+                      type: string
+                      enum:
+                      - balance-slb
```
EdDev: Do we want to control this at the schema level? It will make it hard to expend the list in the future.
Adding an interface type is one thing, but limiting a link aggregation mode of an ovs port is something else.

libnmstate/schemas/operational-state.yaml
```diff
@@ -287,6 +287,23 @@ definitions:
...
+                    slaves:
+                      type: array
+                      minItems: 2
+                      items:
+                        type: object
+                        properties:
```
EdDev: Can you please give an example of an additional property? If you can add one now, it will make this much clearer why this layer is needed (instead of a list of iface names).

libnmstate/schemas/operational-state.yaml
phoracek: link-aggretation port has to have two or more interfaces attached. I don't see how that breaks the partial editation, could you elaborate?

libnmstate/schemas/operational-state.yaml
phoracek: There is a **lot** of OVS interface attributes available http://www.openvswitch.org/support/dist-docs/ovs-vswitchd.conf.db.5.html. However, we may just set them in a separate interface entry. We can change it to list of strings to match the linux bonding (although it is scary to limit ourselves).

libnmstate/schemas/operational-state.yaml
phoracek: Ack.

libnmstate/schemas/operational-state.yaml
phoracek: I don't see how that is an issue, but I'm fine making it more free.

libnmstate/schemas/operational-state.yaml
EdDev: You use the existence of the subtree do determine if the user requested a bond or not. This will not work well with partial editation. If the user  wants to change a port vlan for example, he will not mention the subtree.. but then, you will not be able to differentiate that request from just making that port a non bond. I think we need here an ‘enable’ flag.

phoracek: That won't work the other way around though. In OVS database, the only mark that something is a bond is that multiple interfaces are under the same port. Nothing else, no attributes, nothing.

phoracek: Ah, now I get you. So after all we should have kept the `type` of port :)

libnmstate/schemas/operational-state.yaml
EdDev: You may misunderstood me.
I was looking for an argument to leave it like this. It will be hard to change the api from a list of strings to list of objects in the future.

phoracek: I don't have any additional attribute to add now.

libnmstate/schemas/operational-state.yaml
phoracek: Any suggestion how it should look like? @EdDev

libnmstate/schemas/operational-state.yaml
EdDev: I actually remember what I thought at the time of that review.
I thought that we can define the ports as they now and can set link aggregation as a group of other ports. That will work with keeping the slaves as strings.
Makes sense?

libnmstate/schemas/operational-state.yaml
phoracek: We can still check if it is a bond or simple port while applying a partial config. No need for an additional attribute. While generating proxy interfaces, we can check from the system for the current number of ifaces. That is what I did in https://github.com/nmstate/nmstate/pull/514 if I remember correctly.

EdDev: I have an idea: we can declare that for changing a bond to a non bond, it needs to be “absent” first. That will work.
But we now need an absent property.

libnmstate/schemas/operational-state.yaml
phoracek: I am strongly against that (as I raised with you before). The main issue is that the `port` entry we have now has attributes of an OVS `port`, e.g. VLAN.

If I understood you correctly, you want to have `link-aggregation` type linking multiple of these `ports`. That would mean `port` alone is an OVS `port` and `interface`, while `port` under `link-aggregation` is an OVS `interface` only, although it has all attributes of a port. It would make the API way more complicated and hard to read what is defined in a desired state.

libnmstate/schemas/operational-state.yaml
EdDev: > We can still check if it is a bond or simple port while applying a partial config. No need for an additional attribute. While generating proxy interfaces, we can check from the system for the current number of ifaces. That is what I did in #514 if I remember correctly.

The problem is not about what exists currently, but what the client intentions are.
But I think I answered it earlier now.

phoracek: I don't get what you are trying to solve. Changing bond to a non-bond can look like following:

```yaml
port:
- name: bond1
  link-aggregation:
    slaves:
      - name: eth0
      - name: eth1
```

To:

```yaml
port:
- name: eth0
```

You cannot partially change list of slaves. There is no such support in linux bonding nor linux bridge afaik.

phoracek: @EdDev please give me a specific example of a change that you think would not work with the current schema and I will tell you how would it be tackled or why it should not be supported.

libnmstate/schemas/operational-state.yaml
EdDev: I think you got it right. The suggestion is merge port anf interface settings.. which fits all cases except bonding. For bonding, the enslaved ports will ignore the port settings. I think this is a better model.

phoracek: You want to setup interface details? Do it in respective interface `ethernet` or `ovs-internal`. OVS `bridge` `port` list should have `port` specific attributes only. There is this shorthand of "if you defined only name, we implicitly add an interface with the same name" which does not fit the OVS DB schema.

What do you think is a better model again?

libnmstate/schemas/operational-state.yaml
EdDev: Your example should cause a failure and not change a state. If in practice tries to define a contradicting state: on one part eth0 is a port, in parallel it is a bond slave.

phoracek: Ah, you want a partial editing within the port? Oof...

Any idea how to tackle it (other than creating link-aggregation type and making nmstate into a circus)?

libnmstate/schemas/operational-state.yaml
EdDev: I think ovs has interface properties that apply to eth0 and alike as well. Those are relevant for ovs, not for eth0.
I will try to fetch an example later today.

phoracek: Please also try to write an example of the API proposed by you.

libnmstate/schemas/operational-state.yaml
EdDev: > I have an idea: we can declare that for changing a bond to a non bond, it needs to be “absent” first. That will work.
> But we now need an absent property.

tyll: > > I have an idea: we can declare that for changing a bond to a non bond, it needs to be “absent” first. That will work.
> > But we now need an absent property.

IMHO this violates the idea that Nmstate users declare what they want instead of how they get it. Also, where do you want to declare this?

EdDev: This is what we do with routes and rules, so I do not know what this violates.
I will open a discussion on the PR root on this with an example.

EdDev:
Please see below my proposal.

Basic concepts that guided me:
- Nmstate does not want to use two layers of ports and interfaces.
  Ports and interfaces are therefore merged into a single layer: ports.
  For the special case of bonds, a virtual port is defined with a link-aggregation subtree.
- As ports have "weight" on them with potentially complex config, adding and removing them individually would be preferable (like be do with routes and rules).
  The implementation may start with full list-only, but I am not sure it will be simple to implement, just as we saw with the routes case.
- Nmstate allows partial editation whenever possible, we should try and keep this working.
  When trying to hint about some property through an existence of another, this editation may break.

Define a new bond port (showing only the port subtree):
```yaml
port:
  - name: eth0
    lacp-port-id: 12345 # OVS Interface parameter, given as an example
  - name: eth1
    lacp-port-id: 6789 # OVS Interface parameter, given as an example
  - name: bond1
    link-aggregation:
      slaves:
        - eth0
        - eth1 
```
Consider the bond as a virtual port aggregator that groups ports.

Port name lookup algorithm:
- Is it a kernel iface?
- Is it an internal interface?
- Is it a bond port? i.e. defined on the bridge already, or has a link-aggregation subtree (for new ones).

To detach a port from the bridge, including the removal of a bond with its slaves:
```yaml
port:
  - name: bond1
    state: absent
```
This will also remove the slaves implicitly.
We could delay this implementation later. Defining the full-state first and only later the partial is possible, although I find it less convenient.

phoracek:
@EdDev could you please add a northbound port to the example?

phoracek:
@EdDev VLAN in OVS is an attribute of a `port`. What if `eth0` is set with `vlan 10` and `eth1` with `vlan 20`, how are we supposed to merge that into a single bond port?

EdDev:
@phoracek , I added those details.
I'll keep editing it per your requests to make it simple to follow.

phoracek:
@EdDev since the PR I send is just a schema and the schema you proposed is completely different, do you mind taking over this PR creating a new one? I see no benefit in rewriting your comments into a yaml schema. I would review your PR.

I really don't like the idea of explaining someone when you can use which attributes on port and when not and testing the whole corner-case of link-aggregation handling. But whatever API you guys consider nmstate-ish the most, I just want my bond.

EdDev:
> I really don't like the idea of explaining someone when you can use which attributes on port and when not and testing the whole corner-case of link-aggregation handling.

Raising concerns is exactly what the schema design is all about. Explaining users and treating edge cases exists in all options.
The idea is to be specific and point us to the pros/cons of each option.
You are more than welcome to provide the alternative with examples so we can discuss why it is better or not.
As an example, I want to understand what is the alternative to add an interface property that may fit both a slave-port and a regular port.

> But whatever API you guys consider nmstate-ish the most, I just want my bond.

I prefer well over fast, even if well is slow.
Please address the examples and goals to help us make a good long lasting API.

phoracek:
By "someone"  meant users via documentation.

My main concern is that we have a list what looks like a flat list of ports (the current state, simple to understand). What you propose is to add a hierarchy to those ports. I found that really unintuitive. Both our algorithm and users need to crawl to that list and build the image of actual topology themselves (instead of it being obvious from yaml).

I still want to brainstorm a little on my proposal, where you have a list of `ports`. They can be either simple (without aggregation), meaning they represent an `interface` of the same name. Or they can have explicitly set `link-aggregation`, meaning they cover multiple `interfaces`. Can we add `absent` just to that?

Back to my original example in https://github.com/nmstate/nmstate/pull/562#discussion_r351871736. With `absent`, it would be:

```yaml
port:
- name: bond1
  link-aggregation:
    slaves:
      - name: eth0
      - name: eth1
```

Changed into:

```yaml
port:
- name: eth0
- name: bond1
  state: absent
```

EdDev:
Your suggestion is ok, but I am missing one thing:
> As an example, I want to understand what is the alternative to add an interface property that may fit both a slave-port and a regular port.

phoracek:
@EdDev I'd say that would be specific and open to discussion for every single attribute. I don't think we can apply a  generic rule for that. The only rule I'd start with is that you cannot have an interface mentioned in two ports (name and slave).

For your two examples, VLAN would be always only on the top level. LACP port id would be under `link-aggregation`.

Some attributes may need to be described in API on both places, but when `link-aggregation` is in use, they are allowed only on the slave level.

EdDev:
@phoracek please use the lacp example I gave and show it.
We have a bunch of properties already, lets sample some and see how it looks.

In this regard, IMO you are proposing a simple model for the simple case and a complex inconsistent model for the special cases.
I have a strange feeling that you are taking us back to model this the same way ovs does it.. port having interfaces.. at least in their model they do not have duplication.

As the only point left is this interface property, can we delay the decision on it without the need to change api later?

phoracek:
@EdDev the issue is that the current API is hiding interface implicitly in the port. That was OK and simple for simple cases, but is not feasible for more complex cases, such as bonding ;)

```yaml
port:
- name: bond1
  vlan: ... #  OVS generic port property
  link-aggregation:
    mode: balance-slb  # OVS bonding port property
    lacp-system-id: 1000  # OVS bonding port property
    slaves:
      - name: eth0
        lacp-port-id: 1  # OVS bonding interface property
      - name: eth1
        lacp-port-id: 2  # OVS bonding interface property
```

Is there an interface property that would make sense for both standalone port and bonding, while not being configured directly in `ethernet` or `internal-ovs` types?

EdDev:
Ok, lets try this approach for now. I will dig into the interface properties later today and check if something may break.

Please post the schema and all relevant parts as done with the sriov patch: https://github.com/nmstate/nmstate/pull/592/files

phoracek:
@EdDev should be done. I won't be able to address your comments until tomorrow evening. Thanks.

libnmstate/schema.py
```diff
@@ -244,6 +244,16 @@ class OVSBridge(object):
...
 
     PORT_SUBTREE = 'port'
-    PORT_NAME = 'name'
-    PORT_VLAN_MODE = 'vlan-mode'
-    PORT_ACCESS_TAG = 'access-tag'
+
+    class Port(object):
+        NAME = 'name'
+        VLAN_MODE = 'vlan-mode'
+        ACCESS_TAG = 'access-tag'
```
EdDev: This is not a related change and causes a collisions with other PR/s.
Please focus on the added new properties:
```
class Port:
    LINK_AGGREGATION_SUBTREE = 'link-aggregation'
    ...
```

libnmstate/schema.py
```diff
@@ -244,6 +244,16 @@ class OVSBridge(object):
...
+        LINK_AGGREGATION_SUBTREE = 'link-aggregation'
+
+        class LinkAggregation(object):
```
EdDev: Please drop the `(object)` for new classes.

tests/integration/ovs_test.py
```diff
@@ -103,3 +104,16 @@ def test_create_and_remove_ovs_bridge_with_internal_port_and_static_ip():
...
     assertlib.assert_absent(BRIDGE1)
     assertlib.assert_absent(PORT1)
+
+
+@pytest.mark.xfail(
+    raises=NmstateNotImplementedError,
+    reason='OVS link-aggregation is not supported yet',
+    strict=True,
+)
+def test_la_not_implemented():
```
EdDev: Please use `linkaggregation` instead of `la`.

But it does not make sense to test it, and in fact you expect an xfail, which means you are supposed to test the creation of a link-aggregation. Please show that and present how it is supposed to be used.

tests/integration/testlib/ovslib.py
```diff
@@ -47,6 +47,13 @@ def set_options(self, options):
     def add_system_port(self, name):
         self._add_port(name)
 
+    # TODO: Implement adding slaves and mode
+    def add_la_port(self, name):
```
EdDev: How about `add_linkaggregation_port`.

And if you define this helper, it makes sense to use all defined properties in the schema (i.e. the mode and slaves). Otherwise, you can just drop this.

tests/lib/schema_validation_test.py
```diff
@@ -408,3 +409,28 @@ def test_non_interger_route_table(self, default_data):
         with pytest.raises(js.ValidationError) as err:
             libnmstate.validator.validate(default_data)
         assert 'main' in err.value.args[0]
+
+
+class TestOVSBridge(object):
```
EdDev: Drop the `(object)` please.

tests/lib/schema_validation_test.py
```diff
@@ -26,11 +26,12 @@
 from libnmstate.schema import Constants
 from libnmstate.schema import DNS
 from libnmstate.schema import Ethernet
-from libnmstate.schema import LinuxBridge as LB
-from libnmstate.schema import VXLAN
 from libnmstate.schema import Interface
 from libnmstate.schema import InterfaceType
+from libnmstate.schema import LinuxBridge as LB
```
EdDev: This one and VXLAN do not seem to be related to this change.

libnmstate/nm/applier.py
```diff
@@ -390,6 +391,11 @@ def _build_connection_profile(iface_desired_state, base_con_profile=None):
             )
             settings.append(linux_bridge_setting)
     elif iface_type == ovs.BRIDGE_TYPE:
+        for port in iface_desired_state.get(OvsB.CONFIG_SUBTREE, {}).get(
+            OvsB.PORT_SUBTREE, []
+        ):
+            if OvsB.Port.LINK_AGGREGATION_SUBTREE in port:
+                raise NmstateNotImplementedError
```
EdDev: This is not the place we raise the missing implementation.
Missing implementations are usually raised in the setting creation.
For example: https://github.com/nmstate/nmstate/blob/95af61535301ee885d415f5853aa2c06c23972ad/libnmstate/nm/bridge.py#L93

tests/integration/testlib/ovslib.py
```diff
@@ -58,9 +68,11 @@ def add_internal_port(self, name, ipv4_state):
         )
 
     def _add_port(self, name):
+        port = {OVSBridge.PORT_NAME: name}
         self._bridge_iface[OVSBridge.CONFIG_SUBTREE].setdefault(
             OVSBridge.PORT_SUBTREE, []
-        ).append({OVSBridge.PORT_NAME: name})
+        ).append(port)
+        return port
```
EdDev: it is not recommended to do both setting and getting in the same method.
It may be convenient when used, but it causes other problems.
The method changes the state. The caller can fetch the new state or re-calculate `{OVSBridge.PORT_NAME: name}`.

tests/lib/schema_validation_test.py
```diff
@@ -571,3 +571,28 @@ def test_non_interger_route_table(self, default_data):
         with pytest.raises(js.ValidationError) as err:
             libnmstate.validator.validate(default_data)
         assert 'main' in err.value.args[0]
+
+
+class TestOVSBridge:
```
EdDev: Please be specific, like the vlan one is.

PR #563, opened 2019-11-06
netapplier: VLAN MTU change workaround

ffmancera:
If we modify a VLAN interface MTU then we need to reapply the state. If
we do not reapply the state, the rest of the VLAN interfaces based on
the same interface get their MTU restored by default.

Ref: https://bugzilla.redhat.com/1751695

### DIFF
--- a/tests/integration/vlan_test.py
+++ b/tests/integration/vlan_test.py
@@ -37,23 +37,6 @@ from .testlib.vlan import vlan_interface
 VLAN_IFNAME = 'eth1.101'
 VLAN2_IFNAME = 'eth1.102'
 
-TWO_VLANS_STATE = {
-    Interface.KEY: [
-        {
-            Interface.NAME: VLAN_IFNAME,
-            Interface.TYPE: InterfaceType.VLAN,
-            Interface.STATE: InterfaceState.UP,
-            VLAN.CONFIG_SUBTREE: {VLAN.ID: 101, VLAN.BASE_IFACE: 'eth1'},
-        },
-        {
-            Interface.NAME: VLAN2_IFNAME,
-            Interface.TYPE: InterfaceType.VLAN,
-            Interface.STATE: InterfaceState.UP,
-            VLAN.CONFIG_SUBTREE: {VLAN.ID: 102, VLAN.BASE_IFACE: 'eth1'},
-        },
-    ]
-}
-
 
 def test_add_and_remove_vlan(eth1_up):
     with vlan_interface(
@@ -90,9 +73,63 @@ def test_add_and_remove_two_vlans_on_same_iface(eth1_up):
     assert not current_state[Interface.KEY]
 
 
+def test_two_vlans_on_eth1_change_mtu(eth1_up):
+    desired_state = statelib.show_only(('eth1',))
+    eth1_state = desired_state[Interface.KEY][0]
+    vlans_state = create_two_vlans_state()
+    desired_state[Interface.KEY].extend(vlans_state[Interface.KEY])
+    for iface in desired_state[Interface.KEY]:
+        iface[Interface.MTU] = 2000
+    libnmstate.apply(desired_state)
+
+    eth1_102_state = next(
+        ifstate
+        for ifstate in desired_state[Interface.KEY]
+        if ifstate[Interface.NAME] == VLAN2_IFNAME
+    )
+    eth1_state[Interface.MTU] = 2200
+    eth1_102_state[Interface.MTU] = 2200
+    libnmstate.apply(desired_state)
+
+    eth1_vlan_iface_cstate = statelib.show_only((VLAN_IFNAME,))
+    assert eth1_vlan_iface_cstate[Interface.KEY][0][Interface.MTU] == 2000
+
+
+def test_two_vlans_on_eth1_change_base_iface_mtu(eth1_up):
+    desired_state = statelib.show_only(('eth1',))
+    eth1_state = desired_state[Interface.KEY][0]
+    vlans_state = create_two_vlans_state()
+    desired_state[Interface.KEY].extend(vlans_state[Interface.KEY])
+    for iface in desired_state[Interface.KEY]:
+        iface[Interface.MTU] = 2000
+    libnmstate.apply(desired_state)
+
+    eth1_state[Interface.MTU] = 2200
+    libnmstate.apply({Interface.KEY: [eth1_state]})
+    eth1_vlan_iface_cstate = statelib.show_only((VLAN_IFNAME,))
+    assert eth1_vlan_iface_cstate[Interface.KEY][0][Interface.MTU] == 2000
+
+
+def test_two_vlans_on_eth1_change_mtu_rollback(eth1_up):
+    desired_state = statelib.show_only(('eth1',))
+    vlans_state = create_two_vlans_state()
+    desired_state[Interface.KEY].extend(vlans_state[Interface.KEY])
+    for iface in desired_state[Interface.KEY]:
+        iface[Interface.MTU] = 2000
+    libnmstate.apply(desired_state)
+
+    for iface in desired_state[Interface.KEY]:
+        iface[Interface.MTU] = 2200
+    libnmstate.apply(desired_state, commit=False)
+    libnmstate.rollback()
+
+    eth1_vlan_iface_cstate = statelib.show_only((VLAN_IFNAME,))
+    assert eth1_vlan_iface_cstate[Interface.KEY][0][Interface.MTU] == 2000
+
+
 def test_rollback_for_vlans(eth1_up):
     current_state = libnmstate.show()
-    desired_state = TWO_VLANS_STATE
+    desired_state = create_two_vlans_state()
 
     desired_state[Interface.KEY][1]['invalid_key'] = 'foo'
     with pytest.raises(NmstateVerificationError):
@@ -153,7 +190,7 @@ def test_add_new_base_iface_with_vlan():
 
 @contextmanager
 def two_vlans_on_eth1():
-    desired_state = TWO_VLANS_STATE
+    desired_state = create_two_vlans_state()
     libnmstate.apply(desired_state)
     try:
         yield desired_state
@@ -174,3 +211,22 @@ def two_vlans_on_eth1():
                 ]
             }
         )
+
+
+def create_two_vlans_state():
+    return {
+        Interface.KEY: [
+            {
+                Interface.NAME: VLAN_IFNAME,
+                Interface.TYPE: InterfaceType.VLAN,
+                Interface.STATE: InterfaceState.UP,
+                VLAN.CONFIG_SUBTREE: {VLAN.ID: 101, VLAN.BASE_IFACE: 'eth1'},
+            },
+            {
+                Interface.NAME: VLAN2_IFNAME,
+                Interface.TYPE: InterfaceType.VLAN,
+                Interface.STATE: InterfaceState.UP,
+                VLAN.CONFIG_SUBTREE: {VLAN.ID: 102, VLAN.BASE_IFACE: 'eth1'},
+            },
+        ]
+    }


ffmancera:
This is not working properly. The test pass but then I get an error. Probably I am missing something. Any suggestion? Thanks.

ffmancera:
Here is the error from the VLAN test. Thanks.

```
===================================================================== test session starts =====================================================================
platform linux -- Python 3.7.4, pytest-4.6.6, py-1.5.4, pluggy-0.13.0 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /tmp/nmstate, inifile: tox.ini
plugins: cov-2.8.1
collected 7 items                                                                                                                                             

tests/integration/vlan_test.py::test_add_and_remove_vlan PASSED                                                                                         [ 14%]
tests/integration/vlan_test.py::test_vlan_iface_uses_the_mac_of_base_iface PASSED                                                                       [ 28%]
tests/integration/vlan_test.py::test_add_and_remove_two_vlans_on_same_iface PASSED                                                                      [ 42%]
tests/integration/vlan_test.py::tests_two_vlans_on_eth1_change_mtu PASSED                                                                               [ 57%]
tests/integration/vlan_test.py::tests_two_vlans_on_eth1_change_mtu ERROR                                                                                [ 57%]ERROR:root:NM main-loop aborted: Connection deletion failed on con/None: error=g-dbus-error-quark: No such interface “org.freedesktop.NetworkManager.Settings.Connection” on object at path /org/freedesktop/NetworkManager/Settings/12 (19)
  File "/usr/local/bin/pytest", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.7/site-packages/_pytest/config/__init__.py", line 82, in main
    return config.hook.pytest_cmdline_main(config=config)
  File "/usr/local/lib/python3.7/site-packages/pluggy/hooks.py", line 286, in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
  File "/usr/local/lib/python3.7/site-packages/pluggy/manager.py", line 92, in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
  File "/usr/local/lib/python3.7/site-packages/pluggy/manager.py", line 86, in <lambda>
    firstresult=hook.spec.opts.get("firstresult") if hook.spec else False,
  File "/usr/local/lib/python3.7/site-packages/pluggy/callers.py", line 187, in _multicall
    res = hook_impl.function(*args)
  File "/usr/local/lib/python3.7/site-packages/pluggy/callers.py", line 80, in get_result
    raise ex[1].with_traceback(ex[2])
  File "/usr/local/lib/python3.7/site-packages/pluggy/callers.py", line 187, in _multicall
    res = hook_impl.function(*args)
  File "/usr/local/lib/python3.7/site-packages/_pytest/main.py", line 243, in pytest_cmdline_main
    return wrap_session(config, _main)
  File "/usr/local/lib/python3.7/site-packages/_pytest/main.py", line 206, in wrap_session
    session.exitstatus = doit(config, session) or 0
  File "/usr/local/lib/python3.7/site-packages/pluggy/hooks.py", line 286, in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
  File "/usr/local/lib/python3.7/site-packages/pluggy/manager.py", line 92, in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
  File "/usr/local/lib/python3.7/site-packages/pluggy/manager.py", line 86, in <lambda>
    firstresult=hook.spec.opts.get("firstresult") if hook.spec else False,
  File "/usr/local/lib/python3.7/site-packages/pluggy/callers.py", line 203, in _multicall
    gen.send(outcome)
  File "/usr/local/lib/python3.7/site-packages/_pytest/terminal.py", line 662, in pytest_sessionfinish
    outcome.get_result()
  File "/usr/local/lib/python3.7/site-packages/pluggy/callers.py", line 80, in get_result
    raise ex[1].with_traceback(ex[2])
  File "/usr/local/lib/python3.7/site-packages/pluggy/callers.py", line 187, in _multicall
    res = hook_impl.function(*args)
  File "/usr/local/lib/python3.7/site-packages/_pytest/runner.py", line 73, in pytest_sessionfinish
    session._setupstate.teardown_all()
  File "/usr/local/lib/python3.7/site-packages/_pytest/runner.py", line 325, in teardown_all
    self._pop_and_teardown()
  File "/usr/local/lib/python3.7/site-packages/_pytest/runner.py", line 297, in _pop_and_teardown
    self._teardown_with_finalization(colitem)
  File "/usr/local/lib/python3.7/site-packages/_pytest/runner.py", line 315, in _teardown_with_finalization
    self._callfinalizers(colitem)
  File "/usr/local/lib/python3.7/site-packages/_pytest/runner.py", line 312, in _callfinalizers
    six.reraise(*exc)
  File "/usr/lib/python3.7/site-packages/six.py", line 693, in reraise
    raise value
  File "/usr/local/lib/python3.7/site-packages/_pytest/runner.py", line 305, in _callfinalizers
    fin()
  File "/usr/local/lib/python3.7/site-packages/_pytest/fixtures.py", line 859, in finish
    six.reraise(*e)
  File "/usr/lib/python3.7/site-packages/six.py", line 693, in reraise
    raise value
  File "/usr/local/lib/python3.7/site-packages/_pytest/fixtures.py", line 852, in finish
    func()
  File "/usr/local/lib/python3.7/site-packages/_pytest/fixtures.py", line 803, in _teardown_yield_fixture
    next(it)
  File "/tmp/nmstate/tests/integration/conftest.py", line 62, in preserve_old_config
    yield
  File "/usr/lib/python3.7/site-packages/libnmstate/netapplier.py", line 67, in apply
    state.State(desired_state), verify_change, commit, rollback_timeout
  File "/usr/lib/python3.7/site-packages/libnmstate/netapplier.py", line 155, in _apply_ifaces_state
  File "/usr/lib64/python3.7/contextlib.py", line 119, in __exit__
    next(self.gen)
  File "/usr/lib/python3.7/site-packages/libnmstate/netapplier.py", line 246, in _setup_providers
    mainloop.error
libnmstate.error.NmstateLibnmError: Unexpected failure of libnm when running the mainloop: run execution
+ run_exit
+ dump_network_info
+ docker_exec '
      nmcli dev; \
      nmcli con; \
      ip addr; \
      ip route; \
      cat /etc/resolv.conf; \
      head /proc/sys/net/ipv6/conf/*/disable_ipv6; \
    '
+ docker exec -t -i 04620e9c18de911e2a871ff61126786ab14062c53f91fa2ad35b7ec50bc54f5e /bin/bash -c 'cd /tmp/nmstate && 
      nmcli dev; \
      nmcli con; \
      ip addr; \
      ip route; \
      cat /etc/resolv.conf; \
      head /proc/sys/net/ipv6/conf/*/disable_ipv6; \
    '
DEVICE    TYPE      STATE      CONNECTION 
eth1      ethernet  connected  eth1       
eth1.101  vlan      connected  eth1.101   
eth1.102  vlan      connected  eth1.102   
eth0      ethernet  unmanaged  --         
eth2      ethernet  unmanaged  --         
lo        loopback  unmanaged  --         
NAME      UUID                                  TYPE      DEVICE   
eth1      29214628-fd6b-4570-b0f1-fb39e2d8e391  ethernet  eth1     
eth1.101  376271c4-b6f6-4a2c-8bbd-7ac84eb884df  vlan      eth1.101 
eth1.102  f29f4cc2-e924-4682-8560-acec306e71fb  vlan      eth1.102 
ens3      f35c6104-3566-4fe3-87bf-8dfb2dbb661d  ethernet  --       
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
6: eth1.102@eth1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 2200 qdisc noqueue state UP group default qlen 1000
    link/ether 02:42:ac:12:00:02 brd ff:ff:ff:ff:ff:ff
7: eth1.101@eth1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 2000 qdisc noqueue state UP group default qlen 1000
    link/ether 02:42:ac:12:00:02 brd ff:ff:ff:ff:ff:ff
191: eth0@if192: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default 
    link/ether 02:42:ac:11:00:02 brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 172.17.0.2/16 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::42:acff:fe11:2/64 scope link 
       valid_lft forever preferred_lft forever
195: eth1@if196: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 2200 qdisc noqueue state UP group default 
    link/ether 02:42:ac:12:00:02 brd ff:ff:ff:ff:ff:ff link-netnsid 0
197: eth2@if198: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default 
    link/ether 02:42:ac:13:00:02 brd ff:ff:ff:ff:ff:ff link-netnsid 0
default via 172.17.0.1 dev eth0 
172.17.0.0/16 dev eth0 proto kernel scope link src 172.17.0.2 
search redhat.com
nameserver 127.0.0.11
options ndots:0
```

ffmancera:
It is passing all the integration tests except the "test_rollback_for_vlans" from vlan_test.py. I have not been able to find out what is going wrong. Any suggestions about that? Thanks!

EdDev:
Looks like this solution will not work with `commit=False`.
(apply with commit=True and the execute an explicit commit command before the timeout triggers)

We need this supported for the VDSM scenario.

libnmstate/netapplier.py
```diff
@@ -63,9 +63,20 @@ def apply(desired_state, verify_change=True, commit=True, rollback_timeout=60):
     validator.validate_dns(desired_state)
     validator.validate_vxlan(desired_state)
 
+    not_desired_state_vlans = _prepare_vlan_ifaces_for_mtu_change(
```
maiqueb: I'd rename `not_desired_state_vlans`to something more in line w/ reality, like `vlans_with_mtu_changes` or something like that.

libnmstate/netapplier.py
```diff
@@ -63,9 +63,20 @@ def apply(desired_state, verify_change=True, commit=True, rollback_timeout=60):
...
+        desired_state
+    )
     checkpoint = _apply_ifaces_state(
         state.State(desired_state), verify_change, commit, rollback_timeout
     )
+    if not_desired_state_vlans:
+        _apply_ifaces_state(
+            state.State({schema.Interface.KEY: not_desired_state_vlans}),
+            False,
```
maiqueb: this `False` value as is is hard to understand what it stands for. As such, I'd use `verify_change=False` instead. 

Also, why don't we check the state after the transaction commit ?

libnmstate/netapplier.py
```diff
@@ -201,6 +212,42 @@ def _verify_change(desired_state):
     desired_state.verify_dns(current_state)
 
 
+def _prepare_vlan_ifaces_for_mtu_change(desired_state):
+    current_state = state.State(netinfo.show())
+    current_ifaces = list(six.viewvalues(current_state.interfaces))
```
maiqueb: won't it be more efficient to use `six.viewitems` instead ? 

That way, you'd avoid getting the ifstate['name'] in line https://github.com/nmstate/nmstate/pull/563/files#diff-6b531a9cd456f54d73ad106d72174513R226 - if you unpack both ifname, ifstate on the for at line https://github.com/nmstate/nmstate/pull/563/files#diff-6b531a9cd456f54d73ad106d72174513R223. 

just a nit.

tests/integration/vlan_test.py
```diff
@@ -90,6 +90,22 @@ def test_add_and_remove_two_vlans_on_same_iface(eth1_up):
...
 
 
+def test_two_vlans_on_eth1_change_mtu(eth1_up):
+    desired_state = statelib.show_only(('eth1',))
+    desired_state[Interface.KEY].extend(TWO_VLANS_STATE[Interface.KEY])
+    for iface in desired_state[Interface.KEY]:
+        iface[Interface.MTU] = 2000
+    libnmstate.apply(desired_state)
+
+    del desired_state[Interface.KEY][1]
```
maiqueb: could you do something a bit smarter than this, e.g. remove the ifstate based on name ? I'm looking at this and have no clue as to why it's the interface whose index is 1 that needs to be deleted.

tests/integration/vlan_test.py
```diff
@@ -90,6 +90,22 @@ def test_add_and_remove_two_vlans_on_same_iface(eth1_up):
...
+    for iface in desired_state[Interface.KEY]:
+        iface[Interface.MTU] = 2200
+    libnmstate.apply(desired_state)
+
+    current_state = statelib.show_only(('eth1.101',))
```
maiqueb: do you mind renaming `current_state` to something more descriptive - e.g. `eth1_vlan_iface_current_state`, or something like that ?

I think `current_state` applies for `libnmstate.show()` scenarios - e.g. full state.

libnmstate/netapplier.py
```diff
@@ -201,6 +212,42 @@ def _verify_change(desired_state):
...
+            continue
+        ifname = ifstate[schema.Interface.NAME]
+        vlan_dstate = desired_ifaces_states.get(ifname, {})
+        if (
+            vlan_dstate.get(schema.Interface.STATE)
+            == schema.InterfaceState.ABSENT
+        ):
+            continue
+        vlan_ifaces_states.append(
+            {
```
maiqueb: It might improve readability to move this to a method / function. 

Can't say I like all the 'or' stuff you're doing below, but, the only alternative that comes to mind isn't much better - updating a copy of vlan_dstate with the contents of ifstate **that are not** present in vlan_dstate.

Your call.

libnmstate/netapplier.py
ffmancera: Done! Thanks.

libnmstate/netapplier.py
ffmancera: Oops, we are using verify_change variable value now. Thanks!

libnmstate/netapplier.py
ffmancera: I missed that, thanks!

tests/integration/vlan_test.py
ffmancera: I'm recreating the situation raised in the ticket referenced in the commit message. If you take a look to TWO_VLANS_STATE, you would see that I am modifying eth1 and eth1.102 but not eth1.101.

If the workaround works, the MTU of the eth1.101 should be 2000 not 1500. I've also added another test in order to use a rollback scenario.

tests/integration/vlan_test.py
ffmancera: Sure, done! Thanks

libnmstate/netapplier.py
ffmancera: Hm. Not sure about that. Would not it be too much to split that to another function just for a workaround?

maiqueb: Workarounds usually have long lifespans. But that's besides the point. The real question is if the code is more efficient / readable / performs better.

I think it would be more readable the way I'm suggesting. 

But, I don't think it's enough to block - if you think this is better, you can leave it as is.

tests/integration/vlan_test.py
maiqueb: You can also re-create the issue referenced in the ticket removing the ifstate by name, that won't hold you back :)

I think that it's OK if you delete ifstate[Interface.KEY][0], if there's only a single interface. It's easy to reason about that.

When there are multiple, it's trickier, order dependent, and, IMHO, confusing. As you say: I have to go and browse the TWO_VLANS_STATE to know which of the ifaces you're deleting. IMHO, that's not needed. 

I agree that checking TWO_VLANS_STATE for which interface has index 1 won't cost hundreds of millions of dollars over time, but ... Just saying.

Feel free to disregard, but readability is improved the way I'm suggesting.

ffmancera: Oh sorry, I misunderstood you. You are completely right. Thanks!

libnmstate/netapplier.py
ffmancera: Thanks for explaining. Doing it!

libnmstate/netapplier.py
```diff
@@ -63,9 +63,18 @@ def apply(desired_state, verify_change=True, commit=True, rollback_timeout=60):
     validator.validate_dns(desired_state)
     validator.validate_vxlan(desired_state)
 
+    vlans_with_mtu_changes = _prepare_vlan_ifaces_for_mtu_change(desired_state)
     checkpoint = _apply_ifaces_state(
         state.State(desired_state), verify_change, commit, rollback_timeout
     )
+    if vlans_with_mtu_changes:
+        _apply_edited_ifaces_without_checkpoint(
```
tyll: This code needs to be called inside the `_apply_ifaces_state()` function inside the `with nm.checkpoint.CheckPoint()` context. After the the context is left, the checkpoint might be deleted (commited) but the changes to the MTU should happen while the checkpoint is still present.

ffmancera: But `_apply_ifaces_state()` is returning the checkpoint if `commit=False`. If they set `commit=True` it doesn't matter because we are committing the changes after verification phase. Sorry if I am missing something here. Thanks!

tyll: Let's assume that the user requests a state with an invalid MTU for the VLAN and other changes

If `commit == True` all the other changes will be established on the system and the checkpoint will be removed. In a second step, Nmstate will try to apply the invalid MTU, which will fail. The expected behavior would be to return back to the state of the checkpoint. But this is not possible anymore because the checkpoint is already gone. I wonder, how does the workaround actually work? Doesn't `_prepare_vlan_ifaces_for_mtu_change(desired_state)` need to modify the desired state to remove all VLAN MTU changes from there? Otherwise the verification should fail, shouldn't it?
From my understanding of the issue, the special code needs to be added within the `FIXME` markers in the following listing:

```python
def _apply_ifaces_state(
￼    desired_state, verify_change, commit, rollback_timeout
￼):
￼    current_state = state.State(netinfo.show())
￼
￼    _prepare_desired_state_for_apply(current_state, desired_state)
￼
￼    new_interfaces = _list_new_interfaces(desired_state, current_state)
￼
￼    try:
        with nm.checkpoint.CheckPoint(
            autodestroy=commit, timeout=rollback_timeout
        ) as checkpoint:
            with _setup_providers():
# FIXME: This code needs to be modified to re-apply the vlan state twice
                ifaces2add, ifaces_add_configs = _add_interfaces(
                    new_interfaces, desired_state
                )
                state2edit = _create_editable_desired_state(
                    desired_state, current_state, new_interfaces
                )
                ifaces2edit, ifaces_edit_configs = _edit_interfaces(state2edit)  # this should probably first do all the changes without MTU changes, then also call the next function, afterwards another change with the VLAN MTU changes should be done with _edit_interfaces() and 'set_ifaces_admin_state()
                nm.applier.set_ifaces_admin_state(
                    ifaces2add + ifaces2edit,
                )
# FIXME: END
            if not nmclient.can_disable_ipv6():
                _disable_ipv6(desired_state)
            if verify_change:
                _verify_change(desired_state)  # This should verify the original user-provided state
        if not commit:
            return checkpoint
    except nm.checkpoint.NMCheckPointPermissionError:
        raise NmstatePermissionError('Error creating a check point')
    except NmstateError:
        # Assume rollback occured, revert IPv6 stack state.
        # Checkpoint rollback is async, there is a need to wait for it to
        # finish before proceeding with other actions.
        # TODO: https://nmstate.atlassian.net/browse/NMSTATE-103
        time.sleep(5)
        if not nmclient.can_disable_ipv6():
            _disable_ipv6(current_state)
        raise

```

I might be missing something, so correct me if I am wrong.

libnmstate/netapplier.py
```diff
@@ -63,9 +63,18 @@ def apply(desired_state, verify_change=True, commit=True, rollback_timeout=60):
     validator.validate_dns(desired_state)
     validator.validate_vxlan(desired_state)
 
+    vlans_with_mtu_changes = _prepare_vlan_ifaces_for_mtu_change(desired_state)
```
EdDev: Why do we need the complexity of filtering the vlans with changed mtu?
Is there a problem to be consistent and fetch all vlans, no matter if the mtu changed or not?

libnmstate/netapplier.py
EdDev: I agree with @tyll 
@ffmancera do you see a problem with this suggestion?

ffmancera: After @tyll clarification I also agree with him. I will work on that. Thanks for explaining!

libnmstate/netapplier.py
ffmancera: Yes that is what we are doing actually. This variable name is confusing I will change the name to `vlans_ifaces` or something like that. Thanks for point it out!

libnmstate/netapplier.py
ffmancera: Done!

tests/integration/vlan_test.py
ffmancera: Done!

libnmstate/netapplier.py
ffmancera: Done!

libnmstate/netapplier.py
ffmancera: Done! Thanks!

libnmstate/netapplier.py
```diff
@@ -128,6 +125,34 @@ def _apply_ifaces_state(
     validator.validate_interfaces_state(desired_state, current_state)
     validator.validate_routes(desired_state, current_state)
 
+
+def _apply_edited_ifaces_without_checkpoint(
```
tyll: This method does not need the arguments for checkpoint handling like `commit` and `rollback_timeout`

libnmstate/netapplier.py
```diff
@@ -128,6 +125,34 @@ def _apply_ifaces_state(
...
+
+    with _setup_providers():
+        state2edit = _create_editable_desired_state(
+            desired_state, current_state, new_interfaces
+        )
+        ifaces2edit, ifaces_edit_configs = _edit_interfaces(state2edit)
+        nm.applier.set_ifaces_admin_state(
+            ifaces2edit, con_profiles=ifaces_edit_configs
+        )
+        if verify_change:
```
tyll: IMHO the state should be verified only once after all applies happened, so IMHO the `verify_change()` should not be part of this method

libnmstate/netapplier.py
```diff
@@ -128,6 +125,34 @@ def _apply_ifaces_state(
     validator.validate_interfaces_state(desired_state, current_state)
     validator.validate_routes(desired_state, current_state)
 
+
+def _apply_edited_ifaces_without_checkpoint(
+    desired_state, verify_change, commit, rollback_timeout
+):
+    current_state = state.State(netinfo.show())
+
+    _prepare_desired_state_for_apply(current_state, desired_state)
```
tyll: It would be nicer to return the desired state and assign it to show that it will be changed.

libnmstate/netapplier.py
```diff
@@ -150,6 +175,14 @@ def _apply_ifaces_state(
                 _disable_ipv6(desired_state)
             if verify_change:
                 _verify_change(desired_state)
+            if vlan_ifaces_state:
+                _apply_edited_ifaces_without_checkpoint(
+                    state.State({schema.Interface.KEY: vlan_ifaces_state}),
+                    verify_change=False,
```
tyll: Always disable verification should not be needed, if at all it might be ok for the first apply as long as the changes are verified after the second apply to ensure that the final outcome state is as expected.

libnmstate/netapplier.py
```diff
@@ -201,6 +234,47 @@ def _verify_change(desired_state):
...
+        or vlan_cstate[schema.Interface.TYPE],
+        schema.Interface.STATE: vlan_dstate.get(schema.Interface.STATE)
+        or vlan_cstate[schema.Interface.STATE],
+        schema.InterfaceType.VLAN: vlan_dstate.get(schema.InterfaceType.VLAN)
+        or vlan_cstate[schema.InterfaceType.VLAN],
+    }
+
+
+def _prepare_vlan_ifaces_for_mtu_change(desired_state):
+    current_state = state.State(netinfo.show())
```
tyll: AFAICS you could pass the current_state from the caller function to ensure that all functions operate on the same current state.

tests/integration/vlan_test.py
```diff
@@ -90,9 +91,39 @@ def test_add_and_remove_two_vlans_on_same_iface(eth1_up):
...
+    libnmstate.rollback()
+
+    eth1_vlan_iface_cstate = statelib.show_only(('eth1.101',))
+    assert eth1_vlan_iface_cstate[Interface.KEY][0][Interface.MTU] == 1500
+
+
 def test_rollback_for_vlans(eth1_up):
     current_state = libnmstate.show()
-    desired_state = TWO_VLANS_STATE
+    desired_state = copy.deepcopy(TWO_VLANS_STATE)
```
tyll: It would be better to  be better to make the two vlans state a fixture to ensure that tests do not accidentally interfer when someone forgets the deepcopy. This might be a candidate for its own PR to get this merged already.

tests/integration/vlan_test.py
```diff
@@ -90,9 +91,39 @@ def test_add_and_remove_two_vlans_on_same_iface(eth1_up):
...
+    for iface in desired_state[Interface.KEY]:
+        if iface[Interface.NAME] != 'eth1.101':
+            iface[Interface.MTU] = 2200
+    libnmstate.apply(desired_state)
+
+    eth1_vlan_iface_cstate = statelib.show_only(('eth1.101',))
+    assert eth1_vlan_iface_cstate[Interface.KEY][0][Interface.MTU] == 2000
+
+
+def test_two_vlans_on_eth1_change_mtu_using_commit(eth1_up):
```
tyll: The test name seems to imply that commit is used but actually it is not.

libnmstate/netapplier.py
ffmancera: Also, does not need verify_change in the new approach. Done, thanks!

libnmstate/netapplier.py
ffmancera: Yes. You are right. Thanks!

libnmstate/netapplier.py
ffmancera: Done!

libnmstate/netapplier.py
ffmancera: Yes, as I said above we are not using that argument anymore. Thanks!

tests/integration/vlan_test.py
ffmancera: I would do tthat after this PR. Thanks!

tests/integration/vlan_test.py
ffmancera: Yes, I've changed the name to `test_two_vlans_on_eth1_change_mtu_rollback`.

libnmstate/netapplier.py
ffmancera: Sure, done! Thanks!

ffmancera:
> Thank you. It looks better now.
> There are some details especially with the verification that need to be addresed, IMHO the code should work like this (I used pseudo code now, hope this is more clear than my comments in the actual code from yesterday).
> 
> ```
> def _apply_ifaces_state()
>     ...
>     with checkpoint:
>         # apply changes without the MTU changes for VLAN
>         # apply changes to actually adjust the MTU
>         disable_ipv6()
>         verify_change()
> ```
> 
> If this is not working, please explain.

That suggestion works! Thank you very much :)

libnmstate/netapplier.py
```diff
@@ -66,6 +66,7 @@ def apply(desired_state, verify_change=True, commit=True, rollback_timeout=60):
     checkpoint = _apply_ifaces_state(
         state.State(desired_state), verify_change, commit, rollback_timeout
     )
+
```
tyll: Please remove this new line

ffmancera: Oops. Sorry about that.

libnmstate/netapplier.py
tyll: It is not done, yet, the code needs to look like `desired_state = _prepare_desired_state_for_apply(current_state, desired_state)` for this.

ffmancera: Oops, sorry again. I forgot the assign. Thanks!

libnmstate/netapplier.py
ffmancera: Done! It have been fixed. Thanks!

ffmancera:
All changes done. CI is failing for another reason (that test case is also failing in another PR).

ffmancera:
> Please order the functions/methods from top to bottom such that the caller comes before the callee, allowing reading the story from top to bottom, such that the details are always below.
> It makes it easier to review and read the code.

Sorry about that. Done! Thanks for point it out!

libnmstate/netapplier.py
```diff
@@ -167,6 +168,61 @@ def _apply_ifaces_state(
...
+    vlan_ifaces_states = []
+    for ifname, ifstate in current_ifaces:
+        if ifstate[schema.Interface.TYPE] != schema.InterfaceType.VLAN:
+            continue
+        vlan_dstate = desired_ifaces_by_name.get(ifname, {})
+        if (
+            vlan_dstate.get(schema.Interface.STATE)
+            == schema.InterfaceState.ABSENT
+        ):
+            continue
```
EdDev:
```
current_changed_vlan_ifaces = (
(ifname, ifstate) for ifname, ifstate in current_vlan_ifaces
if desired_state.interfaces.get(ifname, {}).get(schema.Interface.STATE) == schema.InterfaceState.UP
)
```
One line here is probably too long and harder to read, so you can use a function for it: `_is_up_iface`.

Then you end up with a list comprehension like this:
```
[
_create_vlan_desired_state_from_vlan_current_state(desired_state.interfaces[ifname], ifstate, ifname)
for ifname, ifstate in current_changed_vlan_ifaces
]
```

But, on the other hand, we do this already in `State._remove_absent_interfaces`. But it looks like it is called from the verify_interface method... which smells wrong (we verify but actually update the state ://).

libnmstate/netapplier.py
```diff
@@ -128,6 +124,38 @@ def _apply_ifaces_state(
     validator.validate_interfaces_state(desired_state, current_state)
     validator.validate_routes(desired_state, current_state)
```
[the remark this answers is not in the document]
ffmancera: Done!

libnmstate/netapplier.py
```diff
@@ -167,6 +168,61 @@ def _apply_ifaces_state(
         raise
 
 
+def _prepare_vlan_ifaces_for_mtu_change(desired_state, current_state):
+    current_ifaces = list(six.viewitems(current_state.interfaces))
+    desired_ifaces = list(six.viewvalues(desired_state.interfaces))
```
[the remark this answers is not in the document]
ffmancera: All done. You are right, we do not need py2 support anymore. Thanks.

libnmstate/netapplier.py
```diff
@@ -167,6 +168,61 @@ def _apply_ifaces_state(
...
+    desired_ifaces_by_name = {
+        ifstate[schema.Interface.NAME]: ifstate for ifstate in desired_ifaces
+    }
```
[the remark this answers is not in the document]
ffmancera: Oops.

libnmstate/netapplier.py
ffmancera: I also added `_is_vlan_iface`. I tried to do it using `interfaces_to_filter=desired_vlan_names` but that would require to filter the ifaces by type too.. Is that fine? Thanks.

tests/integration/vlan_test.py
```diff
@@ -90,9 +91,39 @@ def test_add_and_remove_two_vlans_on_same_iface(eth1_up):
...
+def test_two_vlans_on_eth1_change_mtu(eth1_up):
+    desired_state = statelib.show_only(('eth1',))
+    vlans_state = copy.deepcopy(TWO_VLANS_STATE)
+    desired_state[Interface.KEY].extend(vlans_state[Interface.KEY])
+    for iface in desired_state[Interface.KEY]:
+        iface[Interface.MTU] = 2000
+    libnmstate.apply(desired_state)
+
+    for iface in desired_state[Interface.KEY]:
+        if iface[Interface.NAME] != 'eth1.101':
```
EdDev: You can save this base_iface_state in advance with no trouble to avoid this strange lookup.

tests/integration/vlan_test.py
```diff
@@ -90,9 +91,39 @@ def test_add_and_remove_two_vlans_on_same_iface(eth1_up):
     assert not current_state[Interface.KEY]
 
 
+def test_two_vlans_on_eth1_change_mtu(eth1_up):
+    desired_state = statelib.show_only(('eth1',))
+    vlans_state = copy.deepcopy(TWO_VLANS_STATE)
```
EdDev: Instead of using `deepcopy` it will be nicer to use a creator helper: `create_two_vlans_state` that replaces `TWO_VLANS_STATE`.

libnmstate/netapplier.py
```diff
@@ -167,6 +170,63 @@ def _apply_ifaces_state(
...
+
+
+def _apply_edited_ifaces_without_checkpoint(desired_state):
+    current_state = state.State(netinfo.show())
+
+    desired_state = _prepare_desired_state_for_apply(
+        current_state, desired_state
+    )
+    validator.validate_interfaces_state(desired_state, current_state)
+    validator.validate_routes(desired_state, current_state)
```
EdDev: Is there a reason we do this again?

tests/integration/vlan_test.py
ffmancera: I do not understand this. In this look up I am setting the MTU of all interfaces from the desired state except the "eth1.101" interface.

libnmstate/netapplier.py
ffmancera: Oops, fixed. Thanks

tests/integration/vlan_test.py
ffmancera: Done. Thanks!

tests/integration/vlan_test.py
EdDev: You know in the test who are these "all" except "eth1.101": "eth1" and "eth1.102".
In general, using `for` loops in tests is confusing, it requires interpreting what has been done instead of being explicit. In this case, I got confused and thought you want to change just "eth1" (the base).
I prefer to see what you changed explicitly.

Could you also use the names constants?

ffmancera: Done! Thanks for explaining

tests/integration/vlan_test.py
```diff
@@ -90,9 +73,48 @@ def test_add_and_remove_two_vlans_on_same_iface(eth1_up):
...
+    desired_state[Interface.KEY].extend(vlans_state[Interface.KEY])
+    for iface in desired_state[Interface.KEY]:
+        iface[Interface.MTU] = 2000
+    libnmstate.apply(desired_state)
+
+    eth1_state = next(
+        ifstate
+        for ifstate in desired_state[Interface.KEY]
+        if ifstate[Interface.NAME] == 'eth1'
+    )
```
EdDev: This info was available already:
```
desired_state = statelib.show_only(('eth1',))
eth1_state = desired_state[Interface.KEY][0]
```

ffmancera: Done! Thanks!

ffmancera:
I have included the changes suggested by @cathay4t. If you don't like this new approach then we can go back to the old one at any time.

libnmstate/netapplier.py
```diff
@@ -167,6 +169,44 @@ def _apply_ifaces_state(
...
 
 
+def _prepare_vlan_ifaces_for_mtu_change(desired_state, current_state):
+    current_ifaces = current_state.interfaces
+    desired_ifaces = desired_state.interfaces
+    desired_ifaces_names = desired_ifaces.keys()
+    current_changed_vlan_ifaces = [
+        (ifname, ifstate)
+        for ifname, ifstate in current_ifaces.items()
+        if _is_vlan_iface(ifstate)
```
EdDev: These conditions will always resolve to false, which tells me that there are no tests to cover this logic.
The last two conditions are contradicting.

libnmstate/netapplier.py
```diff
@@ -118,13 +118,15 @@ def _apply_ifaces_state(
...
-    desired_state.merge_dns(current_state)
-    metadata.generate_ifaces_metadata(desired_state, current_state)
-
+    vlan_ifaces_state = _prepare_vlan_ifaces_for_mtu_change(
+        desired_state, current_state
+    )
+    desired_state.merge_interfaces(
+        state.State({schema.Interface.KEY: vlan_ifaces_state})
+    )
+    desired_state = _prepare_desired_state_for_apply(
```
EdDev: Extracting to a function is always good, but now it is not related to this change, therefore belonging to a refactor commit.
But, I also want to state that the name is not helping here, it tells little of what is done (or what it represents). I suggest you extract this to a separate PR for discussion.

libnmstate/netapplier.py
```diff
@@ -118,13 +118,15 @@ def _apply_ifaces_state(
...
 ):
     current_state = state.State(netinfo.show())
-
-    desired_state.sanitize_ethernet(current_state)
-    desired_state.sanitize_dynamic_ip()
-    desired_state.merge_routes(current_state)
-    desired_state.merge_dns(current_state)
-    metadata.generate_ifaces_metadata(desired_state, current_state)
-
+    vlan_ifaces_state = _prepare_vlan_ifaces_for_mtu_change(
```
EdDev: This fits now as a ‘state.State’ method like the others.
‘desired_state.compement_vlan_interfaces(current_state)’
It should identify which current vlan ifaces are not mentioned in the desired state but their base is, and add their name and maybe the state/up (it is enough) to the desired state.
Nothing more should be needed, because later on the flow, a merge will occur anyway.

This new method can now be unit tested easily in the state test module.

libnmstate/netapplier.py
ffmancera: Why? I have debugged it and this is not always false.

libnmstate/netapplier.py
ffmancera: Wouldn't it be better to directly include all VLANs that are not in the desired state?

EdDev: Doing that will touch unnecessary ifaces.. why do it?

libnmstate/netapplier.py
EdDev: If ifname is not in the desired state (cond 3) it cannot be in UP state (cond 2).

ffmancera: I think I am missing something here. Do you mean current state? Why all the UP interfaces must be in the desired state?

EdDev: Please explain me the scenarios in which the last two lines below will not give False.

```
desired_ifaces_names = desired_ifaces.keys()

ifname = ‘eth1’ # it is in the current state

 _is_up_iface(desired_ifaces.get(ifname, {}))

and ifname not in desired_ifaces_names

```

libnmstate/netapplier.py
ffmancera: Not needed anymore. I've deleted it.

libnmstate/netapplier.py
ffmancera: I will do that in a separate PR. Thanks!

libnmstate/netapplier.py
ffmancera: You are right. I've done some changes. Please take a look, the current solution proposed by @cathay4t is now working. I have added another integration test to check what happens if the VLANs are not set in the desired state and it is working.

tests/integration/vlan_test.py
```diff
@@ -90,9 +73,60 @@ def test_add_and_remove_two_vlans_on_same_iface(eth1_up):
...
+def test_two_vlans_on_eth1_change_base_iface_mtu(eth1_up):
+    desired_state = statelib.show_only(('eth1',))
+    eth1_state = desired_state[Interface.KEY][0]
+    vlans_state = create_two_vlans_state()
+    desired_state[Interface.KEY].extend(vlans_state[Interface.KEY])
+    for iface in desired_state[Interface.KEY]:
+        iface[Interface.MTU] = 2000
+    libnmstate.apply(desired_state)
+
+    eth1_state = statelib.show_only(('eth1',))
```
EdDev: Why do you need to read it again? You already had it once.

tests/integration/vlan_test.py
```diff
@@ -90,9 +73,60 @@ def test_add_and_remove_two_vlans_on_same_iface(eth1_up):
...
+    desired_state = statelib.show_only(('eth1',))
+    vlans_state = create_two_vlans_state()
+    desired_state[Interface.KEY].extend(vlans_state[Interface.KEY])
+    for iface in desired_state[Interface.KEY]:
+        iface[Interface.MTU] = 2000
+    libnmstate.apply(desired_state, commit=False)
+    libnmstate.rollback()
+
+    eth1_vlan_iface_cstate = statelib.show_only((VLAN_IFNAME,))
+    assert eth1_vlan_iface_cstate[Interface.KEY][0][Interface.MTU] == 1500
```
EdDev: It is more interesting if it is able to rollback to a non-default value.

tests/integration/vlan_test.py
```diff
@@ -90,9 +73,60 @@ def test_add_and_remove_two_vlans_on_same_iface(eth1_up):
...
+    desired_state[Interface.KEY].extend(vlans_state[Interface.KEY])
+    for iface in desired_state[Interface.KEY]:
+        iface[Interface.MTU] = 2000
+    libnmstate.apply(desired_state)
+
+    eth1_state = statelib.show_only(('eth1',))
+    eth1_state[Interface.KEY][0][Interface.MTU] = 2200
+    libnmstate.apply(eth1_state)
+    eth1_vlan_iface_cstate = statelib.show_only((VLAN_IFNAME,))
+    assert eth1_vlan_iface_cstate[Interface.KEY][0][Interface.MTU] == 2000
```
EdDev: This in practice says that there is no bug at all. The base mtu was changed but it did not touch the vlans above it.

Is it possible that the current copr NM 1.20 version includes the fix?

EdDev:
This is the NM version used in the tests: NetworkManager-1.20.5-1.cab2e4fc96.el8.x86_64

If it includes the fix, then only the tests are needed. The production code should be redundant.

tests/integration/vlan_test.py
ffmancera: Oops, fixed.

tests/integration/vlan_test.py
ffmancera: Done, thanks!

tests/integration/vlan_test.py
ffmancera: Yes as you said the current NM version used includes the fix. So the workaround is not needed.

tests/integration/vlan_test.py
```diff
@@ -90,9 +73,63 @@ def test_add_and_remove_two_vlans_on_same_iface(eth1_up):
...
+    desired_state = statelib.show_only(('eth1',))
+    eth1_state = desired_state[Interface.KEY][0]
+    vlans_state = create_two_vlans_state()
+    desired_state[Interface.KEY].extend(vlans_state[Interface.KEY])
+    for iface in desired_state[Interface.KEY]:
+        iface[Interface.MTU] = 2000
+    libnmstate.apply(desired_state)
+
+    eth1_state[Interface.MTU] = 2200
+    libnmstate.apply(eth1_state)
```
EdDev: CI failed for a good reason:
`libnmstate.apply({Interface.KEY: [eth1_state]}`

EdDev:
@tyll , all comments of your last review have been answered as far as I can see.
I'm merging this change.

PR #565, opened 2019-11-06
tox/black: Output diff for black check

tyll:
Show the necessary diff to make black pass to make it easier to
understand why black failed and to allow contributors to fix the code
without having to run black themselves.

---
This change is [Reviewable](https://reviewable.io/reviews/nmstate/nmstate/565)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -86,6 +86,7 @@ deps =
 commands =
     black \
         --check \
+        --diff \
         {posargs} \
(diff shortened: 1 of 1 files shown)


tyll:
I filed https://github.com/psf/black/issues/1145 to fix the performance issue in black. AFAICS it could just use the cache for `--diff` as well.

PR #566, opened 2019-11-07
run-tests: Show NetworkManager version

tyll:

---
This change is [Reviewable](https://reviewable.io/reviews/nmstate/nmstate/566)

### DIFF
--- a/automation/run-tests.sh
+++ b/automation/run-tests.sh
@@ -283,6 +283,9 @@ docker_exec '
     while ! systemctl is-active NetworkManager; do sleep 1; done
 '
 add_extra_networks
+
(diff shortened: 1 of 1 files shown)


automation/run-tests.sh
```diff
@@ -53,8 +53,9 @@ function add_extra_networks {
     '
 }
 
-function dump_network_info {
+function show_system_info {
     docker_exec '
+      rpm -q NetworkManager;
```
EdDev: Placing this here is a bit strange because we do not need to show this info more than once.
The others are showed before and after the run.

Any chance you will place it outside?

tyll: moved it outside and changed it to also show the OS version

PR #569, opened 2019-11-08
run-tests: add test-file and test-name parameters

ffmancera:
This allow the user to run specific integration test cases.


ffmancera:
We can use --pytest-args='-k add_port_to_existing_bridge' to run an specific test case. This is not needed. Thanks @tyll!

PR #575, opened 2019-11-12
Update spec file

tyll:
This change is [Reviewable](https://reviewable.io/reviews/nmstate/nmstate/575)

### DIFF
--- a/packaging/nmstate.py3.spec
+++ b/packaging/nmstate.py3.spec
@@ -6,7 +6,7 @@ Name:           nmstate
 Version:        @VERSION@
 Release:        @RELEASE@%{?dist}
 Summary:        Declarative network manager API
-License:        LGPLv2.1+
+License:        LGPLv2+
 URL:            https://github.com/%{srcname}/%{srcname}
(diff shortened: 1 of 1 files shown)


tyll:
This needs the CI to not use Fedora 29 anymore because NM is too old there for the libnm dependency update.

EdDev:
On what PR this one is dependent on?

tyll:
> On what PR this one is dependent on?

#543 or #580

packaging/nmstate.py3.spec
```diff
@@ -24,10 +24,12 @@ provider support on the southbound.
 %package -n python3-%{libname}
 Summary:        nmstate Python 3 API library
 Requires:       python3-dbus
-Requires:       NetworkManager-libnm >= 1:1.12
+Requires:       NetworkManager-libnm >= 1:1.20
 # Use Recommends for NetworkManager because only access to NM DBus is required,
 # but NM could be running on a different host
 Recommends:     NetworkManager
```
EdDev: Should we request NM 1.20 (like we do with libnm)?

tyll: NM has a versioned dependency on libnm, therefore libnm >= 1.20 implies that NM also needs to be updated (if it is installed).

PR #576, opened 2019-11-12
Linux bridge port vlan filtering

maiqueb:
Implement vlan filtering for linux bridge ports
    
This patch implements trunk ports and access ports as per the port vlan schema, introduced in [0].

[0] - https://github.com/nmstate/nmstate/pull/440
    
Depends on #591

### DIFF
--- a/examples/linuxbrige_eth1_up_port_vlan.yml
+++ b/examples/linuxbrige_eth1_up_port_vlan.yml
@@ -1,0 +1,23 @@
+---
+interfaces:
+  - name: eth1
+    type: ethernet
+    state: up
+  - name: linux-br0
+    type: linux-bridge
+    state: up
+    bridge:
+      port:
+        - name: eth1
+          stp-hairpin-mode: false
+          stp-path-cost: 100
+          stp-priority: 32
+          vlan:
+            mode: trunk
+            trunk-tags:
+            - id: 101
+            - id-range:
+                min: 500
+                max: 599
+            tag: 100
+            enable-native: true
--- a/libnmstate/appliers/linux_bridge.py
+++ b/libnmstate/appliers/linux_bridge.py
@@ -18,6 +18,15 @@
 #
 
 
+from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceState
+from libnmstate.schema import InterfaceType
+from libnmstate.schema import LinuxBridge as LB
+
+
+BRIDGE_ENABLE_VLAN_FILTERING = "_enable_vlan_filtering"
+
+
 def get_slaves_from_state(state, default=()):
     ports = state.get("bridge", {}).get("port")
     if ports is None:
@@ -29,3 +38,51 @@ def set_bridge_ports_metadata(master_state, slave_state):
     ports = master_state.get("bridge", {}).get("port", [])
     port = next(filter(lambda n: n["name"] == slave_state["name"], ports), {})
     slave_state["_brport_options"] = port
+
+
+def generate_port_vlan_metadata(desired_state, current_state):
+    desired_bridges = _filter_interface_type(
+        desired_state.interfaces, InterfaceType.LINUX_BRIDGE
+    )
+    current_bridges = _filter_interface_type(
+        current_state.interfaces, InterfaceType.LINUX_BRIDGE
+    )
+    desired_port_vlan_status_per_bridge = _get_port_vlan_status(
+        desired_bridges
+    )
+    current_port_vlan_status_per_bridge = _get_port_vlan_status(
+        current_bridges
+    )
+
+    for brname, brstate in desired_bridges.items():
+        desired_port_vlan_status = desired_port_vlan_status_per_bridge.get(
+            brname, {}
+        )
+        current_port_vlan_status = current_port_vlan_status_per_bridge.get(
+            brname, {}
+        )
+        current_port_vlan_status.update(desired_port_vlan_status)
+        enable_vlan_filtering = any(current_port_vlan_status.values())
+        brstate[BRIDGE_ENABLE_VLAN_FILTERING] = enable_vlan_filtering
+
+
+def _get_port_vlan_status(bridges):
+    port_vlan_status = {}
+    for bridge_name, bridge_state in bridges.items():
+        port_vlan_status[bridge_name] = {}
+        bridge_config = bridge_state.get(LB.CONFIG_SUBTREE, {})
+        for port in bridge_config.get(LB.PORT_SUBTREE, []):
+            port_name = port[LB.Port.NAME]
+            bridge_port_vlan_status = port_vlan_status[bridge_name]
+            port_vlan_enabled = LB.Port.VLAN_SUBTREE in port
+            bridge_port_vlan_status[port_name] = port_vlan_enabled
+    return port_vlan_status
+
+
+def _filter_interface_type(interfaces, interface_type):
+    return {
+        if_name: if_state
+        for if_name, if_state in interfaces.items()
+        if if_state.get(Interface.TYPE) == interface_type
+        and if_state.get(Interface.STATE) == InterfaceState.UP
+    }
--- a/libnmstate/nm/applier.py
+++ b/libnmstate/nm/applier.py
@@ -413,14 +413,10 @@ def _build_connection_profile(iface_desired_state, base_con_profile=None):
     if bond_opts:
         settings.append(bond.create_setting(bond_opts, wired_setting))
     elif iface_type == bridge.BRIDGE_TYPE:
-        bridge_options = iface_desired_state.get(bridge.BRIDGE_TYPE, {}).get(
-            LB.OPTIONS_SUBTREE
+        linux_bridge_setting = bridge.create_setting(
+            iface_desired_state, base_profile,
         )
-        if bridge_options:
-            linux_bridge_setting = bridge.create_setting(
-                bridge_options, base_profile
-            )
-            settings.append(linux_bridge_setting)
+        settings.append(linux_bridge_setting)
     elif iface_type == ovs.BRIDGE_TYPE:
         ovs_bridge_state = iface_desired_state.get(OvsB.CONFIG_SUBTREE, {})
         ovs_bridge_options = ovs_bridge_state.get(OvsB.OPTIONS_SUBTREE)
--- a/libnmstate/nm/bridge.py
+++ b/libnmstate/nm/bridge.py
@@ -17,9 +17,9 @@
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 #
 
-from libnmstate.error import NmstateNotImplementedError
 from libnmstate.nm import connection
 from libnmstate.nm import nmclient
+from libnmstate.nm.bridge_port_vlan import PortVlanFilter
 from libnmstate.schema import LinuxBridge as LB
 
 
@@ -27,15 +27,16 @@ BRIDGE_TYPE = "bridge"
 BRIDGE_ENABLE_VLAN_FILTERING_METADATA = "_enable_vlan_filtering"
 
 
-def create_setting(options, base_con_profile, bridge_desired_state=None):
+def create_setting(bridge_state, base_con_profile):
     bridge_setting = _get_current_bridge_setting(base_con_profile)
+    options = bridge_state.get(BRIDGE_TYPE, {}).get(LB.OPTIONS_SUBTREE)
     if not bridge_setting:
         bridge_setting = nmclient.NM.SettingBridge.new()
     if options:
         _set_bridge_properties(bridge_setting, options)
-    if bridge_desired_state:
+    if bridge_state:
         _set_bridge_properties_from_metadata(
-            bridge_setting, bridge_desired_state
+            bridge_setting, bridge_state
         )
 
     return bridge_setting
@@ -100,7 +101,17 @@ def create_port_setting(options, base_con_profile):
         elif key == LB.Port.STP_PATH_COST:
             port_setting.props.path_cost = val
         elif key == LB.Port.VLAN_SUBTREE:
-            raise NmstateNotImplementedError
+            trunk_tags = val.get(LB.Port.Vlan.TRUNK_TAGS)
+            tag = val.get(LB.Port.Vlan.TAG)
+            enable_native_vlan = val.get(LB.Port.Vlan.ENABLE_NATIVE)
+            port_vlan_config = PortVlanFilter()
+            port_vlan_config.create_configuration(
+                trunk_tags, tag, enable_native_vlan
+            )
+            port_setting.clear_vlans()
+
+            for vlan_config in port_vlan_config.to_nm():
+                port_setting.add_vlan(vlan_config)
 
     return port_setting
 
@@ -119,7 +130,9 @@ def get_info(nmdev):
     port_profiles = _get_slave_profiles(nmdev)
     props = bridge_setting.props
     info[LB.CONFIG_SUBTREE] = {
-        LB.PORT_SUBTREE: _get_bridge_ports_info(port_profiles),
+        LB.PORT_SUBTREE: _get_bridge_ports_info(
+            port_profiles, show_vlan_info=bridge_setting.get_vlan_filtering()
+        ),
         LB.OPTIONS_SUBTREE: {
             LB.Options.MAC_AGEING_TIME: props.ageing_time,
             LB.Options.GROUP_FORWARD_MASK: props.group_forward_mask,
@@ -149,11 +162,16 @@ def _get_bridge_setting(nmdev):
     return bridge_setting
 
 
-def _get_bridge_ports_info(port_profiles):
+def _get_bridge_ports_info(port_profiles, show_vlan_info=False):
     ports_info = []
     for p in port_profiles:
         port_info = _get_bridge_port_info(p)
         if port_info:
+            if show_vlan_info:
+                bridge_vlan_config = p.get_setting_bridge_port().props.vlans
+                port_vlan = PortVlanFilter()
+                port_vlan.import_from_bridge_settings(bridge_vlan_config)
+                port_info[LB.Port.VLAN_SUBTREE] = port_vlan.to_dict()
             ports_info.append(port_info)
     return ports_info
 
@@ -162,12 +180,13 @@ def _get_bridge_port_info(port_profile):
     """Report port information."""
 
     port_setting = port_profile.get_setting_bridge_port()
-    return {
+    port_state = {
         LB.Port.NAME: port_profile.get_interface_name(),
         LB.Port.STP_PRIORITY: port_setting.props.priority,
         LB.Port.STP_HAIRPIN_MODE: port_setting.props.hairpin_mode,
         LB.Port.STP_PATH_COST: port_setting.props.path_cost,
     }
+    return port_state
 
 
 def _get_slave_profiles(master_device):
--- a/libnmstate/nm/bridge.py
+++ b/libnmstate/nm/bridge.py
@@ -24,15 +24,19 @@ from libnmstate.schema import LinuxBridge as LB
 
 
 BRIDGE_TYPE = "bridge"
+BRIDGE_ENABLE_VLAN_FILTERING_METADATA = "_enable_vlan_filtering"
 
 
-def create_setting(options, base_con_profile):
+def create_setting(options, base_con_profile, bridge_desired_state=None):
     bridge_setting = _get_current_bridge_setting(base_con_profile)
     if not bridge_setting:
         bridge_setting = nmclient.NM.SettingBridge.new()
-
     if options:
         _set_bridge_properties(bridge_setting, options)
+    if bridge_desired_state:
+        _set_bridge_properties_from_metadata(
+            bridge_setting, bridge_desired_state
+        )
 
     return bridge_setting
 
@@ -70,6 +74,12 @@ def _set_bridge_stp_properties(bridge_setting, bridge_stp):
                 bridge_setting.props.hello_time = stp_val
             elif stp_key == LB.STP.MAX_AGE:
                 bridge_setting.props.max_age = stp_val
+
+
+def _set_bridge_properties_from_metadata(bridge_setting, metadata):
+    for metadata_name, metadata_value in metadata.items():
+        if metadata_name == BRIDGE_ENABLE_VLAN_FILTERING_METADATA:
+            bridge_setting.props.vlan_filtering = metadata_value
 
 
 def create_port_setting(options, base_con_profile):
--- a/libnmstate/nm/bridge_port_vlan.py
+++ b/libnmstate/nm/bridge_port_vlan.py
@@ -1,0 +1,256 @@
+#
+# Copyright (c) 2020 Red Hat, Inc.
+#
+# This file is part of nmstate
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+
+
+from libnmstate.nm import nmclient
+from libnmstate.schema import LinuxBridge as LB
+
+
+class PortVlanFilter(object):
+    Pvid = "pvid"
+    Untagged = "untagged"
+
+    def __init__(self):
+        self._trunk_ports = []
+        self._tag = None
+        self._is_native_vlan = None
+        self._port_type = None
+
+    def create_configuration(self, trunk_tags, tag, is_native_vlan=False):
+        """
+        Fill the PortVlanFilter object with data whose format is tied to the
+        API.
+
+        :param trunk_tags: list of schema.LinuxBridge.Port.Vlan.TrunkTags
+               objects.
+        :param tag: the access tag for access ports, the native vlan ID for
+               trunk ports
+        :param is_native_vlan: boolean attribute indicating if the trunk port
+               has a native vlan.
+        """
+        self._trunk_ports = trunk_tags
+        self._tag = tag
+        self._is_native_vlan = is_native_vlan
+        self._port_type = (
+            LB.Port.Vlan.Mode.TRUNK if trunk_tags else LB.Port.Vlan.Mode.ACCESS
+        )
+
+    @property
+    def trunk_ports(self):
+        return self._trunk_ports
+
+    @property
+    def tag(self):
+        return self._tag
+
+    @property
+    def enable_native_vlan(self):
+        return self._is_native_vlan
+
+    @property
+    def port_type(self):
+        return self._port_type
+
+    def to_nm(self):
+        """
+        Generate a list of NM.BridgeVlan objects from the encapsulated
+        PortVlanFilter data
+        """
+
+        port_vlan_config = []
+        if self.port_type == LB.Port.Vlan.Mode.TRUNK:
+            for trunk_port in self.trunk_ports:
+                port_vlan_config.append(
+                    self._generate_vlan_trunk_port_config(trunk_port)
+                )
+            if self.enable_native_vlan and self.tag:
+                port_vlan_config.append(
+                    self._generate_vlan_access_port_config(self.tag)
+                )
+        elif self.port_type == LB.Port.Vlan.Mode.ACCESS and self.tag:
+            port_vlan_config.append(
+                self._generate_vlan_access_port_config(self.tag)
+            )
+
+        return port_vlan_config
+
+    def to_dict(self):
+        """
+        Get the port vlan filtering configuration in dict format - e.g. in yaml
+        format:
+
+        - name: eth1
+          vlan:
+            type: trunk
+            trunk-tags:
+              - id: 101
+              - id-range:
+                  min: 200
+                  max: 4095
+            tag: 100
+            enable-native: true
+        """
+
+        port_vlan_state = {
+            LB.Port.Vlan.MODE: self.port_type,
+            LB.Port.Vlan.TRUNK_TAGS: self.trunk_ports,
+        }
+        if self.tag:
+            port_vlan_state[LB.Port.Vlan.TAG] = self.tag
+        if self.port_type == LB.Port.Vlan.Mode.TRUNK:
+            port_vlan_state[
+                LB.Port.Vlan.ENABLE_NATIVE
+            ] = self.enable_native_vlan
+        return port_vlan_state
+
+    def import_from_bridge_settings(self, port_vlans):
+        """
+        Instantiates a PortVlanFilter object from a list of NM.BridgeVlan
+        objects.
+        """
+
+        bridge_vlan_config = [
+            self._get_vlan_info(port_vlan_filter_entry)
+            for port_vlan_filter_entry in port_vlans
+        ]
+
+        trunk_tags, tag, enable_native = self._get_vlan_config(
+            bridge_vlan_config
+        )
+
+        self._trunk_ports = trunk_tags
+        self._tag = tag
+        self._is_native_vlan = enable_native
+        self._port_type = (
+            LB.Port.Vlan.Mode.TRUNK if trunk_tags else LB.Port.Vlan.Mode.ACCESS
+        )
+
+    @staticmethod
+    def get_vlan_tag_range(port_vlan):
+        """
+        Extract the vlan tags from the NM.BridgeVlan object.
+        A single NM.BridgeVlan object can have a range of vlan tags, or a
+        single one.
+
+        When a NM.BridgeVlan holds a single tag, the min_range, and max_range
+        returned will have the same vlan tag.
+
+        :return: min_range, max_range
+        """
+
+        port_vlan_tags = port_vlan.to_str().split()
+
+        if "-" in port_vlan_tags[0]:
+            vlan_min, vlan_max = port_vlan_tags[0].split("-")
+            return int(vlan_min), int(vlan_max)
+        else:
+            tag = port_vlan_tags[0]
+            return int(tag), int(tag)
+
+    @staticmethod
+    def _is_access_port(trunk_tags):
+        if (
+            len(trunk_tags) == 1
+            and trunk_tags[0][PortVlanFilter.Pvid]
+            and trunk_tags[0][PortVlanFilter.Untagged]
+        ):
+            return True
+        else:
+            return False
+
+    @staticmethod
+    def _get_vlan_info(port_vlan_info):
+        vlan_min, vlan_max = PortVlanFilter.get_vlan_tag_range(port_vlan_info)
+        pvid = port_vlan_info.is_pvid()
+        untagged = port_vlan_info.is_untagged()
+        if vlan_max != vlan_min:
+            port_data = {
+                LB.Port.Vlan.TrunkTags.ID_RANGE: {
+                    LB.Port.Vlan.TrunkTags.MIN_RANGE: vlan_min,
+                    LB.Port.Vlan.TrunkTags.MAX_RANGE: vlan_max,
+                }
+            }
+        else:
+            port_data = {LB.Port.Vlan.TrunkTags.ID: vlan_min}
+        port_data.update(pvid=pvid, untagged=untagged)
+        return port_data
+
+    @staticmethod
+    def _generate_vlan_trunk_port_config(trunk_port):
+        single_vlan_tag = trunk_port.get(LB.Port.Vlan.TrunkTags.ID)
+        ranged_vlan_tags = trunk_port.get(LB.Port.Vlan.TrunkTags.ID_RANGE)
+        port_vlan = None
+
+        if single_vlan_tag:
+            port_vlan = nmclient.NM.BridgeVlan.new(
+                single_vlan_tag, single_vlan_tag
+            )
+        elif ranged_vlan_tags:
+            min_range, max_range = (
+                ranged_vlan_tags[range_type]
+                for range_type in (
+                    LB.Port.Vlan.TrunkTags.MIN_RANGE,
+                    LB.Port.Vlan.TrunkTags.MAX_RANGE,
+                )
+            )
+            port_vlan = nmclient.NM.BridgeVlan.new(min_range, max_range)
+        port_vlan.set_untagged(False)
+        port_vlan.set_pvid(False)
+        return port_vlan
+
+    @staticmethod
+    def _generate_vlan_access_port_config(vlan_tag):
+        port_vlan = nmclient.NM.BridgeVlan.new(vlan_tag, vlan_tag)
+        port_vlan.set_untagged(True)
+        port_vlan.set_pvid(True)
+        return port_vlan
+
+    @staticmethod
+    def _get_vlan_config(bridge_vlan_config):
+        enable_native = False
+        trunk_tags = []
+        tag = None
+
+        for port_vlan_filter_entry in bridge_vlan_config:
+            single_vlan_tag = port_vlan_filter_entry.get(
+                LB.Port.Vlan.TrunkTags.ID
+            )
+            ranged_vlan_tags = port_vlan_filter_entry.get(
+                LB.Port.Vlan.TrunkTags.ID_RANGE
+            )
+            if PortVlanFilter._is_access_port(bridge_vlan_config):
+                tag = single_vlan_tag
+            else:
+                if port_vlan_filter_entry[PortVlanFilter.Pvid]:
+                    tag = (
+                        single_vlan_tag
+                        or ranged_vlan_tags[LB.Port.Vlan.TrunkTags.MAX_RANGE]
+                    )
+                    enable_native = True
+                else:
+                    PortVlanFilter._sanitize_trunk_port(port_vlan_filter_entry)
+                    trunk_tags.append(port_vlan_filter_entry)
+
+        return trunk_tags, tag, enable_native
+
+    @staticmethod
+    def _sanitize_trunk_port(port):
+        if PortVlanFilter.Pvid and PortVlanFilter.Untagged in port:
+            del port[PortVlanFilter.Pvid]
+            del port[PortVlanFilter.Untagged]
--- a/tests/integration/linux_bridge_test.py
+++ b/tests/integration/linux_bridge_test.py
@@ -24,7 +24,6 @@ import pytest
 import yaml
 
 import libnmstate
-from libnmstate.error import NmstateNotImplementedError
 from libnmstate.error import NmstateVerificationError
 from libnmstate.schema import Interface
 from libnmstate.schema import InterfaceIPv4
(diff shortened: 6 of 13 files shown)


tyll:
> TODO

What is still missing?

maiqueb:
> > TODO
> 
> What is still missing?

Not much, just write proper commit msgs and update the PR description.

libnmstate/nm/bridge.py
```diff
@@ -107,7 +115,13 @@ def get_info(nmdev):
 
     port_profiles = _get_slave_profiles(nmdev)
     info[LB.CONFIG_SUBTREE] = {
-        LB.PORT_SUBTREE: _get_bridge_ports_info(port_profiles),
+        LB.PORT_SUBTREE: _get_bridge_ports_info(
+            port_profiles,
+            show_vlan_info=(
+                is_vlan_filtering_supported
```
ffmancera: This should be `is_vlan_filtering_supported()`, right?

maiqueb: > This should be `is_vlan_filtering_supported()`, right?

yes it should. 

Don't spend time on this, I'm (very) slowly working on this PR, and will be updating it in the following days. Much will have changed.

libnmstate/nm/bridge.py
```diff
@@ -23,15 +23,19 @@
...
 BRIDGE_TYPE = 'bridge'
 
 
-def create_setting(options, base_con_profile):
+def create_setting(options, base_con_profile, enable_vlan_filtering=False):
     bridge_setting = _get_current_bridge_setting(base_con_profile)
     if not bridge_setting:
         bridge_setting = nmclient.NM.SettingBridge.new()
 
+    if is_vlan_filtering_supported():
```
maiqueb: Now that we're using NM-1.20, I don't think this makes sense.

libnmstate/nm/bridge.py
```diff
@@ -107,7 +122,13 @@ def get_info(nmdev):
 
     port_profiles = _get_slave_profiles(nmdev)
     info[LB.CONFIG_SUBTREE] = {
-        LB.PORT_SUBTREE: _get_bridge_ports_info(port_profiles),
+        LB.PORT_SUBTREE: _get_bridge_ports_info(
+            port_profiles,
+            show_vlan_info=(
+                is_vlan_filtering_supported()
```
maiqueb: get. rid. of. this.

libnmstate/nm/bridge.py
```diff
@@ -137,25 +158,34 @@ def _get_bridge_setting(nmdev):
...
+    port_state = {
         LB.PORT_NAME: port_profile.get_interface_name(),
         LB.PORT_STP_PRIORITY: port_setting.props.priority,
         LB.PORT_STP_HAIRPIN_MODE: port_setting.props.hairpin_mode,
         LB.PORT_STP_PATH_COST: port_setting.props.path_cost,
     }
+    if show_vlan_info and is_vlan_filtering_supported():
+        port_state.update(
+            {
+                LB.Port.VLAN_SUBTREE: PortVlanFilter.build_from_port_setting(
```
maiqueb: Rename to `build_from_port_vlans`.

libnmstate/nm/port_vlan.py
```diff
@@ -0,0 +1,238 @@
...
+        }
+        if self.tag:
+            port_vlan_state[LB.Port.Vlan.TAG] = self.tag
+        if self.port_type == LB.Port.Vlan.TRUNK_TYPE:
+            port_vlan_state[
+                LB.Port.Vlan.ENABLE_NATIVE
+            ] = self.enable_native_vlan
+        return port_vlan_state
+
+    @staticmethod
```
tyll: Factory methods should be class methods (`@classmethod`).

libnmstate/nm/port_vlan.py
```diff
@@ -0,0 +1,238 @@
...
+        bridge_vlan_config = [
+            PortVlanFilter._get_vlan_info(port_vlan_filter_entry)
+            for port_vlan_filter_entry in port_vlans
+        ]
+
+        trunk_tags, tag, enable_native = PortVlanFilter._get_vlan_config(
+            bridge_vlan_config
+        )
+
+        return PortVlanFilter(trunk_tags, tag, enable_native)
```
tyll: This should be `return cls(...)`

libnmstate/nm/port_vlan.py
```diff
@@ -0,0 +1,238 @@
...
+
+    @staticmethod
+    def build_from_port_vlans(port_vlans):
+        """
+        Factory method that instantiates a PortVlanFilter object from a
+        list of NM.BridgeVlan objects.
+        """
+
+        bridge_vlan_config = [
+            PortVlanFilter._get_vlan_info(port_vlan_filter_entry)
```
tyll: This should use `cls._get...`

libnmstate/nm/port_vlan.py
```diff
@@ -0,0 +1,238 @@
...
+        A single NM.BridgeVlan object can have a range of vlan tags, or a
+        single one.
+
+        When a NM.BridgeVlan holds a single tag, the min_range, and max_range
+        returned will have the same vlan tag.
+
+        :return: min_range, max_range
+        """
+
+        port_vlan_tags = port_vlan.to_str()
```
tyll: Can't you use split here directly?

libnmstate/nm/port_vlan.py
```diff
@@ -0,0 +1,238 @@
...
+        split_port_vlan_tags = port_vlan_tags.split()
+
+        if split_port_vlan_tags and '-' in split_port_vlan_tags[0]:
```
tyll: Why is `split_port_vlan_tags` part of the condition? In the else block, `split_port_vlan_tags[0]` is called which implies it does not evaluate to False or there will be an IndexError.

libnmstate/nm/port_vlan.py
```diff
@@ -0,0 +1,238 @@
...
+        single one.
+
+        When a NM.BridgeVlan holds a single tag, the min_range, and max_range
+        returned will have the same vlan tag.
+
+        :return: min_range, max_range
+        """
+
+        port_vlan_tags = port_vlan.to_str()
+        split_port_vlan_tags = port_vlan_tags.split()
```
tyll: What is the purpose of the split?

examples/linuxbrige_eth1_up.yml
```diff
@@ -22,3 +22,12 @@ interfaces:
           stp-hairpin-mode: false
           stp-path-cost: 100
           stp-priority: 32
+          vlan:
```
cathay4t: Please create new example file for this.

libnmstate/nm/applier.py
```diff
@@ -381,14 +381,19 @@ def _build_connection_profile(iface_desired_state, base_con_profile=None):
...
         settings.append(bond.create_setting(bond_opts))
     elif iface_type == bridge.BRIDGE_TYPE:
-        bridge_options = iface_desired_state.get(bridge.BRIDGE_TYPE, {}).get(
-            LB.OPTIONS_SUBTREE
+        bridge_state = iface_desired_state.get(bridge.BRIDGE_TYPE, {})
+        bridge_options = bridge_state.get(LB.OPTIONS_SUBTREE)
+        linux_bridge_setting = bridge.create_setting(
+            bridge_options,
+            base_profile,
+            enable_vlan_filtering=any(
```
cathay4t: These code should move to `bridge.create_setting()` where it could determine whether to enable VLAN filtering based on `bridge_options`. The `nm/applier` should not take care such details.

maiqueb:
> If functions in `nm/port_vlan.py` is for LinuxBridge only, please rename to `nm/bridge_port_vlan.py` or merge into `nm/bridge.py`.
> 
> Also need integration tests.

I'm adding a class in
https://github.com/nmstate/nmstate/pull/576/files#diff-41e570df111ba3cf748cf4f22fb0e3c1R378

that adds integration tests for this feature. 

Are you missing a specific test, or had missed that file?

libnmstate/nm/applier.py
maiqueb: I don't think it's possible to decide if vlan filtering should be enabled or not based on the bridge options: it's information that is defined on the bridge ports. How do you suggest we do that ? 

As I see it, I'll have to update the function's signature. Thus, might as well be explicit.

examples/linuxbrige_eth1_up.yml
maiqueb: Done.

libnmstate/nm/port_vlan.py
maiqueb: libnm's BridgeVlan object - [0] - to_str() returns '100 pvid untagged' - e.g. the pvid and untagged information are also part of its output.

by splitting, we remove those 2 properties, which are not relevant.

libnmstate/nm/port_vlan.py
maiqueb: Right.

Since these methods operate based on data that originates from libnm's `port_setting.props.vlans` , I can drop the 'split' condition altogether.

libnmstate/nm/port_vlan.py
maiqueb: Done.

libnmstate/nm/port_vlan.py
maiqueb: Done.

libnmstate/nm/port_vlan.py
maiqueb: Done.

libnmstate/nm/port_vlan.py
maiqueb: Done.

tests/integration/linux_bridge_test.py
```diff
@@ -397,6 +397,111 @@ def test_port_vlan_not_implemented(port0_up):
         pass
 
 
+@pytest.mark.xfail(
+    reason='https://nmstate.atlassian.net/browse/NMSTATE-230', strict=True
+)
+class TestVlanFiltering(object):
```
EdDev: No need to inherit from `object` in py3.

tests/integration/linux_bridge_test.py
```diff
@@ -397,6 +397,111 @@ def test_port_vlan_not_implemented(port0_up):
...
+    access_tag = 300
+    bridge_name = TEST_BRIDGE0
```
EdDev: To access these class variables, the proper way is to use <class-name>.<variable-name>.
It will be easier to just promote these constants to the module level (as constants).

tests/integration/linux_bridge_test.py
```diff
@@ -397,6 +397,111 @@ def test_port_vlan_not_implemented(port0_up):
...
+
+    @pytest.mark.parametrize(
+        'is_native_vlan', [True, False], ids=['native', 'not-native']
+    )
+    def test_trunk_port_config(self, is_native_vlan, port0_up):
+        trunk_tags = self._generate_vlan_id_config(100, 101)
```
EdDev: You declared `_generate_vlan_id_config` as static, but have called it through `self`.
That is technically possible, but very confusing to the reader. Please use `TestVlanFiltering` instead.

tests/integration/linux_bridge_test.py
```diff
@@ -397,6 +397,111 @@ def test_port_vlan_not_implemented(port0_up):
...
+    def _get_port_by_name(linux_bridge, port_name):
+        for port in linux_bridge[LinuxBridge.CONFIG_SUBTREE][
+            LinuxBridge.PORT_SUBTREE
+        ]:
+            if port_name == port[LinuxBridge.PORT_NAME]:
+                return port
+
+    @staticmethod
+    def _generate_vlan_filtering_config(
+        port_type, trunk_tags=None, access_tag=None, native_vlan=None
```
EdDev: `access_tag` in the schema is named `tag`, right? If so, it should be the same here.

tests/integration/linux_bridge_test.py
```diff
@@ -397,6 +397,111 @@ def test_port_vlan_not_implemented(port0_up):
...
+            vlan_access_port_state
+        )
+        with linux_bridge(self.bridge_name, bridge_state) as desired_state:
+            assertlib.assert_state(desired_state)
+
+    @staticmethod
+    def _set_bridge_ports(bridge_state, *new_ports):
+        bridge_state[LinuxBridge.CONFIG_SUBTREE][LinuxBridge.PORT_SUBTREE] = [
+            port[Interface.KEY][0] for port in new_ports
+        ]
```
EdDev: Who uses this?

tests/integration/linux_bridge_test.py
```diff
@@ -397,6 +397,111 @@ def test_port_vlan_not_implemented(port0_up):
...
+
+    @staticmethod
+    def _get_port_by_name(linux_bridge, port_name):
+        for port in linux_bridge[LinuxBridge.CONFIG_SUBTREE][
+            LinuxBridge.PORT_SUBTREE
+        ]:
+            if port_name == port[LinuxBridge.PORT_NAME]:
+                return port
```
EdDev: Who uses this?

libnmstate/nm/port_vlan.py
```diff
@@ -0,0 +1,79 @@
...
+#
+
+
+class PortVlanFilter(object):
+    def __init__(self, trunk_tags, access_tag, enable_native_vlan):
+        self._trunk_tags = trunk_tags
+        self._access_tag = access_tag
+        self._is_native_vlan = enable_native_vlan
+
+    def generate(self):
```
EdDev: Won't it fit the `to_*` pattern you used below? e.g. `to_nm`

libnmstate/nm/port_vlan.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        Generate a list of NM.BridgeVlan objects from the encapsulated data
+        """
+
+        return None
```
EdDev: An empty list may fit better.

libnmstate/nm/applier.py
EdDev: The point is that the `applier` module does not know about such details.
It only knows who to call with what info... the code structure asks to push the details down the stack as much as possible, leaving the upper layers cleaner and generic.

I am also unsure if this check is even correct. Bridge options may be changed without specifying any vlans, but that is not to say that there are no current such ports defined. (unless you handle this as part of the "merge" step).

libnmstate/nm/port_vlan.py
maiqueb: No plans to inherit from PortVlanFilter - for me static makes more sense - that's how it was originally. 

@tyll : are you OK with having this as static ? 

I can go either way.

tests/integration/linux_bridge_test.py
maiqueb: I don't understand the first sentence (... proper way is to use ..). What does '..' do ? 

I sincerely believe these belong in the class, since they're related to a specific feature of linux bridges (all other tests not in this class couldn't care less about them).

EdDev:
@maiqueb , can you continue this PR? Do you prefer someone else take over?

maiqueb:
> @maiqueb , can you continue this PR? Do you prefer someone else take over?

I'll resume this work starting next week.

tests/integration/linux_bridge_test.py
maiqueb: Ditto.

tests/integration/linux_bridge_test.py
maiqueb: No one; vestigial thing copy pasted from the schema tests. 

Gone.

tests/integration/linux_bridge_test.py

tests/integration/linux_bridge_test.py

tests/integration/linux_bridge_test.py
maiqueb: Done

libnmstate/nm/port_vlan.py
maiqueb: Done

libnmstate/nm/port_vlan.py
maiqueb: Done

libnmstate/nm/port_vlan.py
```diff
@@ -0,0 +1,79 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+
+
+class PortVlanFilter(object):
+    def __init__(self, trunk_tags, access_tag, enable_native_vlan):
```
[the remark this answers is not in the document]
maiqueb: Done

libnmstate/nm/applier.py
maiqueb: I possibly am not seeing the forest because of all the trees, but the criteria to enable `vlan_filtering` on the bridge should be if there's any port that has a vlan configuration. Do we agree on that ?

What if my port_list features elements both from the desired state **and** current state ? 
I think that would allow bridge options update (without port vlan config changes) to preserve the old state .

maiqueb: Decided to grab the current vlan_filtering value from the bridge_setting object.

This would preserve the value on a bridge option update.

maiqueb:
> commit `linux bridge, vlan filtering: add PortVlanFilter unit tests`
> 
> I had some non related (sometime contradicting) comments on this.
> Hope I organized it well now:
> 
> * The `nm` subpackage is supposed to expose the NM libnm api to nmstate. On one hand, it needs to represent the NM api, on the other it needs to talk in "nmstate" language.
> * The bridge vlan filtering represents a larger challenge, as the Nmstate API abstracts and limits in a way the on what can be changed, but the NM API does not.
> * The above two API needs need to co-exist nicely without limiting each other.
>   To do this, I think that we can build a data structure class which can serve as the API.
>   A factory can build the data structure from the relevant format (we will need two, building from the desired state and another from the NM config/running).
> 
> Please consider that the NM modules that "talk" to libnm may be re-used in other projects, like ansible system roles. Having this separation, will allow not only reuse, but also changing each without touching the other.

The `port_vlan` module follow the philosophy of the existing `translator` module, which also sits in the `nm` pkg. 

Let me know if that's enough, or if I should move the entire module from there, to the `libnmstate` pkg.

examples/linuxbrige_eth1_up_port_vlan.yml
```diff
@@ -0,0 +1,33 @@
+---
```
tyll: Thank you for an example. Please add this also as a test to https://github.com/nmstate/nmstate/blob/master/tests/integration/examples_test.py to ensure the example works.

maiqueb: Done.

libnmstate/nm/applier.py
ffmancera: In my opinion this should be done inside `create_setting()`.

libnmstate/nm/bridge.py
```diff
@@ -17,20 +17,23 @@
...
 
-def create_setting(options, base_con_profile):
+def create_setting(options, base_con_profile, enable_vlan_filtering=False):
     bridge_setting = _get_current_bridge_setting(base_con_profile)
     if not bridge_setting:
         bridge_setting = nmclient.NM.SettingBridge.new()
+    if not enable_vlan_filtering:
+        enable_vlan_filtering = bridge_setting.props.vlan_filtering
 
+    bridge_setting.props.vlan_filtering = enable_vlan_filtering
```
ffmancera: Okay, my brain have halted thinking on this and sorry if I am missing something but this is the same than: 

```
if enable_vlan_filtering:
    bridge_setting.props.vlan_filtering = enable_vlan_filtering
```

or 

```
bridge_setting.props.vlan_filtering = bridge_setting.props.vlan_filtering or enable_vlan_filtering
```

And I think you should do directly:
```
bridge_setting.props.vlan_filtering = enable_vlan_filtering
```
What do you think? Thanks!

libnmstate/nm/bridge.py
```diff
@@ -139,11 +153,17 @@ def _get_bridge_setting(nmdev):
...
     ports_info = []
     for p in port_profiles:
         port_info = _get_bridge_port_info(p)
         if port_info:
+            if show_vlan_info:
+                port_info[
+                    LB.Port.VLAN_SUBTREE
+                ] = PortVlanFilter.build_from_port_vlans(
+                    p.get_setting_bridge_port().props.vlans
+                ).to_dict()
```
ffmancera: Code format here is quite ugly it would be nice if you could manage to fix it. Please consider this as a nit, not a must :-)

libnmstate/nm/bridge_port_vlan.py
```diff
@@ -0,0 +1,237 @@
+#
+# Copyright (c) 2019 Red Hat, Inc.
```
ffmancera: 2020? Thanks!

tests/lib/nm/linux_bridge_vlan_filtering_test.py
```diff
@@ -0,0 +1,252 @@
+#
+# Copyright (c) 2019 Red Hat, Inc.
```
ffmancera: 2020? Thanks!

tests/integration/linux_bridge_test.py
EdDev: I meant that static class variables should be accessed through the class name and not `self` because the latter is confusing the reader, thinking it is an instance variable.
I do not mind if you leave them here or promote them to the module level, that is up to you.

tests/integration/linux_bridge_test.py
```diff
@@ -397,6 +397,131 @@ def test_port_vlan_not_implemented(port0_up):
...
+    access_tag = 300
+    bridge_name = TEST_BRIDGE0
+
+    @pytest.mark.parametrize(
+        "is_native_vlan", [True, False], ids=["native", "not-native"]
+    )
+    def test_trunk_port_config(self, is_native_vlan, port0_up):
+        trunk_tags = TestVlanFiltering._generate_vlan_id_config(100, 101)
+        trunk_port_state = TestVlanFiltering._generate_vlan_filtering_config(
+            "trunk",
```
EdDev: Seems like this should come from the schema.

tests/integration/linux_bridge_test.py
```diff
@@ -397,6 +397,131 @@ def test_port_vlan_not_implemented(port0_up):
...
+
+        port_name = port0_up[Interface.KEY][0][Interface.NAME]
+        bridge_state = _create_bridge_subtree_config((port_name,))
+        bridge_state[LinuxBridge.PORT_SUBTREE][0].update(trunk_port_state)
+        with linux_bridge(self.bridge_name, bridge_state) as desired_state:
+            assertlib.assert_state(desired_state)
+
+    def test_access_port_config(self, bridge0_with_port0, port0_up):
+        access_port_state = TestVlanFiltering._generate_vlan_filtering_config(
+            "access", tag=self.access_tag
```
EdDev: `access` should probably come from the schema constants.

tests/integration/linux_bridge_test.py
```diff
@@ -397,6 +397,131 @@ def test_port_vlan_not_implemented(port0_up):
...
+            native_vlan=is_native_vlan,
+        )
+
+        port_name = port0_up[Interface.KEY][0][Interface.NAME]
+        bridge_state = _create_bridge_subtree_config((port_name,))
+        bridge_state[LinuxBridge.PORT_SUBTREE][0].update(trunk_port_state)
+        with linux_bridge(self.bridge_name, bridge_state) as desired_state:
+            assertlib.assert_state(desired_state)
+
+    def test_access_port_config(self, bridge0_with_port0, port0_up):
```
EdDev: Why the tests from here on use the `bridge0_with_port0` fixture?
You use the bridge contextmanager in the test body and actually delete it inside when this should have been done by the fixture.

Using the fixture seems ok to me but it checks editation of an existing bridge/port and using the contextmanager is not needed in such cases.

tests/integration/linux_bridge_test.py
```diff
@@ -397,6 +397,131 @@ def test_port_vlan_not_implemented(port0_up):
...
+        access_port_state = TestVlanFiltering._generate_vlan_filtering_config(
+            "access", tag=self.access_tag
+        )
+        port_name = port0_up[Interface.KEY][0][Interface.NAME]
+        bridge_state = _create_bridge_subtree_config((port_name,))
+        bridge_state[LinuxBridge.PORT_SUBTREE][0].update(access_port_state)
+        with linux_bridge(self.bridge_name, bridge_state) as desired_state:
+            assertlib.assert_state(desired_state)
+
+    def test_access_port_update(self, bridge0_with_port0, port0_up):
```
EdDev: Seems identical to the previous but you apply twice the same value (so how do you know it changed?).
BTW, if this is an update check, then the test should only update, not create the initial state.. that should be dine in a fixture.
We do not enforce this pattern much, but it fits here well. Especially when you already have a fixture that provides a non vlanned port.

libnmstate/nm/applier.py
EdDev: 1. As mentioned before, the logic needs to be passed down the stack to the nm.bridge module.
2. The only place that seems to cover all cases is to detect if vlan-filtering is to be enabled or not at the metadata stage, where you have the desired state, the current state and the ability to place metadata on the state that mentions this flag (if vlans exist on the bridge or not).
The other alternative is to query the bridge configuration in nm.bridge*, which may also work. Pick your favorite and lets see the result.

libnmstate/nm/bridge.py
EdDev: This is confusing because the decision is not taken in one place.
Are the ports properties reconfigured from scratch here? I mean, by the time the desired state reaches this place, does it include all ports? If so, the decision is limited to the desired state and it does not matter what existed.
In case the desired state only includes a portion of the ports and all the rest are not touched, then how those are getting removed and is the desired state specify it. In such a case you'll need to pass through all non-removed ports and check if any have a vlan, if all do not, then maybe you need to remove the vlan-filtering.

libnmstate/nm/bridge.py
```diff
@@ -90,7 +93,16 @@ def create_port_setting(options, base_con_profile):
         elif key == LB.Port.STP_PATH_COST:
             port_setting.props.path_cost = val
         elif key == LB.Port.VLAN_SUBTREE:
-            raise NmstateNotImplementedError
+            trunk_tags = val.get(LB.Port.Vlan.TRUNK_TAGS)
+            tag = val.get(LB.Port.Vlan.TAG)
+            enable_native_vlan = val.get(LB.Port.Vlan.ENABLE_NATIVE)
+            port_vlan_config = PortVlanFilter(
+                trunk_tags, tag, enable_native_vlan
+            ).to_nm()
```
EdDev: nit: Can you please first create the the object and only then call its method? I find it hard to follow this way.

libnmstate/nm/bridge.py
EdDev: + 1
It is hard to read

The usage of `PortVlanFilter.build_from_port_vlans` is also strange. The object represent a port but here we have a static method that gets the vlans from the setting. I would expect to instantiate the object and let it import its state from an existing setting.

libnmstate/nm/bridge.py
```diff
@@ -152,12 +172,13 @@ def _get_bridge_port_info(port_profile):
     """Report port information."""
 
     port_setting = port_profile.get_setting_bridge_port()
-    return {
+    port_state = {
```
EdDev: Why change this?

libnmstate/nm/bridge_port_vlan.py
```diff
@@ -18,18 +18,62 @@
...
+
 class PortVlanFilter(object):
-    def __init__(self, trunk_tags, tag, enable_native_vlan):
-        self._trunk_tags = trunk_tags
+    pvid = "pvid"
+    min_range = "vlan-range-min"
+    max_range = "vlan-range-max"
+    untagged = "untagged"
+
+    def __init__(self, trunk_tags, tag, is_native_vlan=False):
```
EdDev: I commented about this earlier, but it may fit here as well:
This object is created from the state data but some of its static methods are implementing the other direction.
This means that this object is not a proxy that can translate in both directions and the static methods do not really belong to it.
If you want the object to be a two way lane, then it needs to input both options (on other objects we used `create` for data from state and `import` for filling it from existing NM setting/profile).
The alternative is to have it represent one direction and another using a function or another class.

tests/integration/linux_bridge_test.py
maiqueb: Done

tests/integration/linux_bridge_test.py
maiqueb: Done

tests/integration/linux_bridge_test.py
maiqueb: Removed the fixture from all tests.

tests/integration/linux_bridge_test.py
maiqueb: I don't get the comment about the non vlaned port.

Never the less, created a new fixture that addresses your comments.

Please resolve if you consider I addressed your concerns.

tests/lib/nm/linux_bridge_vlan_filtering_test.py
maiqueb: Done

libnmstate/nm/bridge_port_vlan.py
maiqueb: Done

libnmstate/nm/bridge.py
maiqueb: Vestigial change that evolution itself cut prematurely.

Got rid of it.

libnmstate/nm/bridge_port_vlan.py
maiqueb: Please point me to references where I can follow the current status quo of the project, since it eludes me.

libnmstate/nm/bridge.py
maiqueb: I can store `p.get_setting_bridge_port().props.vlans` in a variable. Hopefully that'll make it look better.

> The usage of PortVlanFilter.build_from_port_vlans is also strange. The object represent a port but here we have a static method that gets the vlans from the setting. I would expect to instantiate the object and let it import its state from an existing setting.

The advantage there being ? I just don't see it.

You prefer:
```python
a = PortVlanFilter(None, None, None)
a.import(p.get_setting_bridge_port().props.vlans)
port_info[...] = a.to_dict()
```
to 
```python
port_info[...] = PortVlanFilter.build_from_port_vlans(
    p.get_setting_bridge_port().props.vlans
) 
```
? The relevant point (to me) is the intermediate object is useless: It only serves the purpose of translating from NM data to dict and the other way around.

Would renaming build_from_port_vlans to 'import_from_NM' be a compromise between both ?

libnmstate/nm/bridge.py
maiqueb: Stopped exposing a parameter dedicated to this.

libnmstate/nm/applier.py
maiqueb: Implemented it in the metadata section.

libnmstate/nm/port_vlan.py
maiqueb: Stopped using it w/ static / class methods, hence, this no longer applies.

libnmstate/nm/bridge_port_vlan.py
maiqueb: Hope now it's better.

Please re-open if not.

phoracek:
@maiqueb this needs to be rebased now.

@tyll could you please take a look at this? Thanks.

tyll:
> @maiqueb this needs to be rebased now.
> 
> @tyll could you please take a look at this? Thanks.

Travis CI is complaining about the formatting. I will take a look on Monday.

tyll:
@cathay4t you requested changes in the past, can you please verify the changes?

libnmstate/appliers/linux_bridge.py
```diff
@@ -29,3 +38,51 @@ def set_bridge_ports_metadata(master_state, slave_state):
...
+    )
+
+    for brname, brstate in desired_bridges.items():
+        desired_port_vlan_status = desired_port_vlan_status_per_bridge.get(
+            brname, {}
+        )
+        current_port_vlan_status = current_port_vlan_status_per_bridge.get(
+            brname, {}
+        )
+        current_port_vlan_status.update(desired_port_vlan_status)
```
cathay4t: When certain slave interface been removed from this bridge, this might be wrong.

Please check whether slave still valid in this bridge.

libnmstate/nm/applier.py
```diff
@@ -413,14 +413,12 @@ def _build_connection_profile(iface_desired_state, base_con_profile=None):
     if bond_opts:
         settings.append(bond.create_setting(bond_opts, wired_setting))
     elif iface_type == bridge.BRIDGE_TYPE:
-        bridge_options = iface_desired_state.get(bridge.BRIDGE_TYPE, {}).get(
-            LB.OPTIONS_SUBTREE
+        bridge_state = iface_desired_state.get(bridge.BRIDGE_TYPE, {})
+        bridge_options = bridge_state.get(LB.OPTIONS_SUBTREE)
+        linux_bridge_setting = bridge.create_setting(
+            bridge_options, base_profile, iface_desired_state,
```
cathay4t: The `iface_desired_state` argument already contains the information of `bridge_options`.

Passing `iface_desired_state` and `base_profile` to `bridge.create_setting()` is enough.

libnmstate/appliers/linux_bridge.py
```diff
@@ -29,3 +38,51 @@ def set_bridge_ports_metadata(master_state, slave_state):
     ports = master_state.get("bridge", {}).get("port", [])
     port = next(filter(lambda n: n["name"] == slave_state["name"], ports), {})
     slave_state["_brport_options"] = port
+
+
+def generate_port_vlan_metadata(desired_state, current_state):
```
cathay4t: For this PR, I don't think we need to use the metadata.

The `bridge.create_setting()` could determine the set `bridge_setting.props.vlan_filtering` or not
based on `iface_state` passing to it.

This could save code editing on `metadata.py` and `appliers/linux_bridge.py` but limit it inside of `nm/bridge.py`.

libnmstate/nm/bridge_port_vlan.py
```diff
@@ -0,0 +1,258 @@
...
+
+from libnmstate.nm import nmclient
+from libnmstate.schema import LinuxBridge as LB
+
+
+class PortVlanFilter(object):
+    pvid = "pvid"
+    min_range = "vlan-range-min"
+    max_range = "vlan-range-max"
+    untagged = "untagged"
```
cathay4t: constant should be in upper case.

libnmstate/appliers/linux_bridge.py
maiqueb: AFAIK, that wouldn't be enough, since we need information from the current_state. It would true *if* we had exposed the boolean that activates vlan filtering.

That leaves metadata or state_merge as the only options (since that's where we have access to current_state & desired_state.

I chose metadata, since it feels more self-contained.

libnmstate/nm/applier.py
maiqueb: While you're 100% right, this will lead to having to refactor the `linux_bridge_test.py` module.

I'm OK with doing it - just want everyone to be aware this PR will keep on growing,

libnmstate/appliers/linux_bridge.py
cathay4t: At time of `nm.bridge.create_setting()` been called. The state has been merged, hence you don't need to worry about current state.

libnmstate/appliers/linux_bridge.py
maiqueb: The desired bridges are gathered from `_filter_interface_type(...)` which only report bridges whose state is `UP`.

Does that address your concern ?

libnmstate/appliers/linux_bridge.py
cathay4t: I am expecting something like:

```

def create_setting(iface_state, base_con_profile):
    bridge_setting = _get_current_bridge_setting(base_con_profile)
    if not bridge_setting:
        bridge_setting = nmclient.NM.SettingBridge.new()

    bridge_setting.props.vlan_filtering = _have_port_vlan(iface_state)


def _has_port_vlan(iface_state):
    """
    Return True if iface_state contains:
       [LB.CONFIG_SUBTREE][LB.PORT_SUBTREE][LB.Port.VLAN_SUBTREE]
    """
    return any(
        LB.Port.VLAN_SUBTREE in port
        for port in iface_state.get(LB.CONFIG_SUBTREE, {}).get(
            LB.PORT_SUBTREE, []
        )
    )
```

The metadata is for information referring to other interface, not for this case.
This is the comment of `generate_ifaces_metadata()` which explained:
```
    The described desired state for each interface may include references to
    other interfaces. As the provider handles the interface setting in an
    isolated manner, it is sometime necessary to specify a property on an
    interface based on a property from a different interface.
    An example of this, is the bond slaves property, which should mark each
    slave with the setting of it being a slave.
```

maiqueb: Have you considered @EdDev 's comment on https://github.com/nmstate/nmstate/pull/576#discussion_r355195141 ? 

I quote: "I am also unsure if this check is even correct. Bridge options may be changed without specifying any vlans, but that is not to say that there are no current such ports defined. (unless you handle this as part of the "merge" step)."

From your excerpt, if you update the bridge options of - let's say - a bridge that has vlan filtering active, but don't update the port vlan configuration of any of it's ports, it'll disable the bridge's vlan filtering, wouldn't it ?

What you're suggesting is literally what I had in the first place (well, my code was located on the applier, not the bridge module, but anyway... ).

examples/linuxbrige_eth1_up_port_vlan.yml
```diff
@@ -0,0 +1,33 @@
...
+      options:
+        group-forward-mask: 0
+        mac-ageing-time: 300
+        multicast-snooping: true
+        stp:
+          enabled: true
+          forward-delay: 15
+          hello-time: 2
+          max-age: 20
+          priority: 32768
```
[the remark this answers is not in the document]
maiqueb: Done.

libnmstate/nm/bridge_port_vlan.py
maiqueb: Done.

libnmstate/appliers/linux_bridge.py
cathay4t: @EdDev Got it wrong. The vlan filter status can be solely determined by `iface_state`, hence it should not use metadata.

The reason you noticed the `nm.bridge.create_settting()` only get partial data on port section is because the
`state.dict_update()` only do full copy on list.

So the the correct approach should be:
 * Amend the `state.merge_interfaces()` to merge the port settings from current to desire.
 * Passing the merged interface state to `bridge.create_setting()` where it could determine to enable vlan filtering or not.

This is the patch based on your latest commit b71e471:

https://gist.github.com/cathay4t/1dabb3074e1dc003afbd0619254ad12e

I tested it with the yaml files https://paste.centos.org/view/8dad1320

This PR has 100+ comments and out of control. I would suggest you to split this PR into:

 * PR for getting run time status. Reading on-disk config is not runtime status.
   As NM does not provides so, please try to parse `bridge -j vlan show` outputs.

 * PR for create bridge setting and bridge port setting in `libnmstate.nm` folder only.
   I am OK if you don't want to waste time on creating integration test case for `libnmstate.nm` files.
 * PR for amend the `state.py` to merge bridge port settings.
 * PR for full support of linux bridge vlan filtering including integration test cases.
 * PR for adding example of linux bridge vlan filtering.

maiqueb: I've set up a meeting w/ you, where I hope we can discuss most of this.

Please accept or propose another time.

On another note, I have one question about your 'merge_interfaces' proposal;  how can port vlan filtering be disabled on a port that currently has it configured ?

I've always assumed that under this model (where vlan filtering being enabled is not exposed)  , disabling it would be done if the user referred in the desired state a port that currently has vlan filtering, without indicating it's vlan config. 

I'm wondering how can that be done using the merge interfaces, since, afaiu, it's purpose is to (and I quote the comment from the code): "Given the self and other states, complete the self state by merging the missing parts from the current state" .

phoracek: > PR for getting run time status. Reading on-disk config is not runtime status. As NM does not provides so, please try to parse bridge -j vlan show outputs.

We have a bad experience with `bridge -j vlan` from kubernetes-nmstate. It is not always a valid JSON and the output format varies on the OS you run on. If possible, I would advice against using it, at least in production code.

cathay4t: > > PR for getting run time status. Reading on-disk config is not runtime status. As NM does not provides so, please try to parse bridge -j vlan show outputs.
> 
> We have a bad experience with `bridge -j vlan` from kubernetes-nmstate. It is not always a valid JSON and the output format varies on the OS you run on. If possible, I would advice against using it, at least in production code.

Thanks for the info.

cathay4t: > how can port vlan filtering be disabled on a port that currently has it configured ?

Need to set clear rule on how things been merged from current. My propose would be:
 * When desire state does not contain port section, it will copy from current state.
 * When desire state has port section defined as empty list. New configure will remove all  the ports from bridge.
 * When desire state has port section with name only. All information for that port will copy fro current state.
 * When desire state has port section with properties more than name, __no__ information will be copy from current state and let nmstate determine the default value for undefined properties. This means, when user want to change the setting of certain port, it have to define all setting of that port in desire state and expect no merge.

To answer your question, to support that use case, we just ask user to provide full configure
of that port without vlan filter settings. When all the port does not have vlan filter, the bridge will disable vlan filter.

phoracek:
@maiqueb @tyll I believe that in the original RFE we asked to enable filtering by default, allowing trunk on all IDs by default. This is similar behavior to what Open vSwitch bridge does. It would be also more welcoming to users coming from OVS or basic linux bridge without filtering enabled. Being able to use VLANs on a default bridge would be more welcoming than requiring users to setup complex settings of VLAN filterring. Does this PR follow that request? Would it be acceptable?

maiqueb:
> @maiqueb @tyll I believe that in the original RFE we asked to enable filtering by default, allowing trunk on all IDs by default. This is similar behavior to what Open vSwitch bridge does. It would be also more welcoming to users coming from OVS or basic linux bridge without filtering enabled. Being able to use VLANs on a default bridge would be more welcoming than requiring users to setup complex settings of VLAN filterring. Does this PR follow that request? Would it be acceptable?

We can in a final PR adjust the default accordingly - e.g. if no bridge vlan tree is present, add the full range (2-4095) + native vlan set to 1.

Would that work for kubevirt ? 

This would require buy-in from the maintainers: @tyll / @cathay4t / @EdDev / @ffmancera .

phoracek:
> We can in a final PR adjust the default accordingly - e.g. if no bridge vlan tree is present, add the full range (2-4095) + native vlan set to 1.

> Would that work for kubevirt ?

That would be great.

libnmstate/appliers/linux_bridge.py
maiqueb: interface merging PR: https://github.com/nmstate/nmstate/pull/888

maiqueb:
Split this PR into:

add XFAILED integ tests: #882
add interface merging algorithm: #888
add port vlan functionality: #892
activate port vlan functionality in applier/show vlan data: #893
add examples: #895

PR #577, opened 2019-11-12
Prepare run-tests.sh for podman

tyll:
The first commit allows to run multiple containers at once without error messages regarding the docker networks. The other commits also prepare run-tests.sh to be used with podman.

---
This change is [Reviewable](https://reviewable.io/reviews/nmstate/nmstate/577)

### DIFF
--- a/automation/run-tests.sh
+++ b/automation/run-tests.sh
@@ -7,9 +7,6 @@ CONT_EXPORT_DIR="/exported-artifacts"
 
 CONTAINER_WORKSPACE="/workspace/nmstate"
 
-NET0="nmstate-net0"
-NET1="nmstate-net1"
-
 TEST_TYPE_ALL="all"
 TEST_TYPE_FORMAT="format"
 TEST_TYPE_LINT="lint"
@@ -27,8 +24,6 @@ function remove_container {
     [ "$res" -ne 0 ] && echo "*** ERROR: $res"
     docker_exec 'rm -rf $CONTAINER_WORKSPACE/*nmstate*.rpm'
     docker rm $CONTAINER_ID -f
-    docker network rm $NET0
-    docker network rm $NET1
 }
 
 function pyclean {
@@ -42,13 +37,11 @@ function docker_exec {
 }
 
 function add_extra_networks {
-    docker network create $NET0 || true
-    docker network create $NET1 || true
-    docker network connect $NET0 $CONTAINER_ID
-    docker network connect $NET1 $CONTAINER_ID
     docker_exec '
-      ip addr flush eth1 && \
-      ip addr flush eth2
+      ip link add eth1 type veth peer eth1peer && \
+      ip link add eth2 type veth peer eth2peer && \
+      ip link set eth1peer up && \
+      ip link set eth2peer up
     '
 }
 
--- a/automation/run-tests.sh
+++ b/automation/run-tests.sh
@@ -293,9 +293,9 @@ dump_network_info
 
 pyclean
 if [[ "$CI" != "true" ]];then
-    container_exec "cp -rf $CONTAINER_WORKSPACE /tmp/"
+    container_exec "cp -rf $CONTAINER_WORKSPACE /root/nmstate-workspace"
     # Change workspace to keep the original one clean
-    CONTAINER_WORKSPACE="/tmp/nmstate"
+    CONTAINER_WORKSPACE="/root/nmstate-workspace"
 fi
 install_nmstate
 run_tests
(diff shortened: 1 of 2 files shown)


automation/run-tests.sh
```diff
@@ -281,34 +276,34 @@ if [[ "$CI" == "true" ]];then
...
 
-docker_exec '(source /etc/os-release; echo $PRETTY_NAME); rpm -q NetworkManager'
+container_exec '(source /etc/os-release; echo $PRETTY_NAME); rpm -q NetworkManager'
 
 dump_network_info
 
 pyclean
 if [[ "$CI" != "true" ]];then
-    docker_exec "cp -rf $CONTAINER_WORKSPACE /tmp/"
+    container_exec "cp -rf $CONTAINER_WORKSPACE /tmp/"
```
cathay4t: In podman, the `/tmp` is mounted as `no_exec`. Please use another folder.

automation/run-tests.sh
```diff
@@ -18,27 +18,29 @@ TEST_TYPE_INTEG="integ"
 FEDORA_IMAGE_DEV="nmstate/fedora-nmstate-dev"
 CENTOS_IMAGE_DEV="nmstate/centos8-nmstate-dev"
 
+: ${container_cmd:=docker}
```
cathay4t: Please use capital for environment variable

tyll: will change this

automation/run-tests.sh
tyll: Changed it to use `/root/nmstate-workspace`.

tyll:
In pratice, it does not yet work completely with podman because the `preserve_old_config` fixture at https://github.com/nmstate/nmstate/blob/master/tests/integration/conftest.py#L59 does not work with the `tap0` interface that is used in the podman container.

tyll:
@cathay4t I addressed your concerns, can you please re-check?

automation/run-tests.sh
```diff
@@ -21,40 +18,38 @@ TEST_TYPE_INTEG="integ"
...
 function add_extra_networks {
-    docker network create $NET0 || true
-    docker network create $NET1 || true
-    docker network connect $NET0 $CONTAINER_ID
-    docker network connect $NET1 $CONTAINER_ID
-    docker_exec '
+    container_exec '
+      ip link add eth1 type veth peer eth1peer && \
+      ip link add eth2 type veth peer eth2peer && \
       ip addr flush eth1 && \
```
EdDev: These are no longer needed now.

automation/run-tests.sh
```diff
@@ -21,40 +18,38 @@ TEST_TYPE_INTEG="integ"
...
 }
 
 function add_extra_networks {
-    docker network create $NET0 || true
-    docker network create $NET1 || true
-    docker network connect $NET0 $CONTAINER_ID
-    docker network connect $NET1 $CONTAINER_ID
-    docker_exec '
+    container_exec '
+      ip link add eth1 type veth peer eth1peer && \
```
EdDev: Don’t you need to set ‘link up’ on them? At least for the peer ones which we do not suppose to manage using nm.

automation/run-tests.sh
tyll: removed them

automation/run-tests.sh
cathay4t: Yeah. `ip link up` is required.

cathay4t:
I was hoping(PR #441):
 * veth creation could be done by pytest fixture.
 * `run-tests.sh` kicks of pytest/tox commands with preferred arguments.
 * `run-tests-container.sh` setup containers and invoke `run-tests.sh`.

But basically, I am OK with current approach, just a suggestion.

automation/run-tests.sh
tyll: Changed it for the peer. It is passing with NM master in the CI, not sure why it fails for 1.20.

tyll:
> I was hoping(PR #441):
> 
> * veth creation could be done by pytest fixture.

This might also be an option. We can still build upon this PR to do this eventually. It would be nice if pytest would still allow to run it against externally created interfaces, for example when testing it in VMs against virtual interfaces.

> * `run-tests.sh` kicks of pytest/tox commands with preferred arguments.

The preferred arguments should probably be moved from run-tests.sh to tox, so that tox always have the correct arguments. Not sure if there is something special for the CI that needs to be different, though.

> * `run-tests-container.sh` setup containers and invoke `run-tests.sh`.

This would only be needed if special arguments besides the ones from tox are needed AFAICS. However, it would be nice to split the script to do the image setup in an Travis install phase to make the log display more useful (making it show the test output only in the normal case). It can also still be done.

tyll:
@cathay4t can you maybe revisit this? The docker error 129 with cp is driving me nuts and with #588 it should be possible to run podman locally.

tyll:
@cathay4t travis passed on all platforms, can you please approve this PR?

PR #579, opened 2019-11-14
Python2 test cleanup

tyll:
Removing the travis python 2 job will provide more resources for the other jobs.

---
This change is [Reviewable](https://reviewable.io/reviews/nmstate/nmstate/579)


tyll:
Will force-merge because it passed on CentOS 8 with NM 1.20.

PR #582, opened 2019-11-15
stdci: Use CentOS 8

tyll:
Use CentOS 8 for stdci because the Fedora tests are currently
meaningless. Since CentOS 8 only supports the integration tests for now,
run only the integration tests.

---
This change is [Reviewable](https://reviewable.io/reviews/nmstate/nmstate/582)

### DIFF
--- a/.stdci.yml
+++ b/.stdci.yml
@@ -6,4 +6,4 @@ Architectures:
       runtime-requirements:
         host-distro: same
 script:
-    from-file: automation/run-tests.sh
+  from-file: automation/run-tests.sh
--- a/.environment.yaml
+++ b/.environment.yaml
@@ -1,3 +1,0 @@
----
-- name: 'CI'
-  value: 'true'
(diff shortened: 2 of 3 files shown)


tyll:
Seems like stdci does not accept arguments for the command but requires different scripts. Maybe we should just drop stdci for now? It does not provide any benefit at the moment and we are just wasting stdci resources AFAICS.

.stdci.yml
```diff
@@ -6,4 +6,4 @@ Architectures:
       runtime-requirements:
         host-distro: same
 script:
-    from-file: automation/run-tests.sh
+  from-file: automation/run-tests.sh --el8 --test-type integ
```
EdDev: Unfortunately you cannot pass arguments here. You can use env variables (see file in automation).

EdDev:
The advantage we have with stdci:
- We can run on el8/fedora kernel.
- Backup to travis-ci.

If we do not need these options, we can drop it.

tyll:
> All looks good, just one small nit: the 3 commits topic use non consistent caption. Can you please adjust it?

Adjusted it. Forgot to check if stdci works now, let's see after it finished.

tyll:
Since this changes only STDCI and passes on STDCI, I'll ignore the failure on Travis for NM 1.20 because it does not seem to be related but some OVS issue that needs to be fixed. Travis for NM master passed.

PR #586, opened 2019-11-17
nm.device: delete inactive profiles when editting an interface

ffmancera:
When editting an interface that doesn't have an active profile we are
creating a new profile. Now, we are deleting all the inactive profiles
of an interface before creating the new one.

### DIFF
--- a/libnmstate/nm/applier.py
+++ b/libnmstate/nm/applier.py
@@ -54,6 +54,12 @@ def create_new_ifaces(con_profiles):
 
 
 def prepare_new_ifaces_configuration(ifaces_desired_state):
+    # Delete the existing profiles before create the new ones
+    for iface_desired_state in ifaces_desired_state:
+        connection.delete_iface_inactive_connections(
+            iface_desired_state[Interface.NAME]
+        )
+
     return [
         _build_connection_profile(iface_desired_state)
         for iface_desired_state in ifaces_desired_state
@@ -79,7 +85,8 @@ def prepare_edited_ifaces_configuration(ifaces_desired_state):
     con_profiles = []
 
     for iface_desired_state in ifaces_desired_state:
-        nmdev = device.get_device_by_name(iface_desired_state[Interface.NAME])
+        ifname = iface_desired_state[Interface.NAME]
+        nmdev = device.get_device_by_name(ifname)
         cur_con_profile = None
         if nmdev:
             cur_con_profile = connection.ConnectionProfile()
@@ -95,6 +102,7 @@ def prepare_edited_ifaces_configuration(ifaces_desired_state):
             con_profiles.append(cur_con_profile)
         else:
             # Missing connection, attempting to create a new one.
+            connection.delete_iface_inactive_connections(ifname)
             con_profiles.append(new_con_profile)
 
     return con_profiles
--- a/libnmstate/nm/connection.py
+++ b/libnmstate/nm/connection.py
@@ -450,3 +450,16 @@ def get_device_active_connection(nm_device):
     if nm_device:
         active_conn = nm_device.get_active_connection()
     return active_conn
+
+
+def delete_iface_inactive_connections(ifname):
+    for con in list_connections_by_ifname(ifname):
+        con.delete()
+
+
+def list_connections_by_ifname(ifname):
+    return [
+        ConnectionProfile(con)
+        for con in nmclient.NM.Client.get_connections(nmclient.client())
+        if con.get_interface_name() == ifname
+    ]
--- a/tests/integration/profile_test.py
+++ b/tests/integration/profile_test.py
@@ -1,0 +1,143 @@
+#
+# Copyright (c) 2019 Red Hat, Inc.
+#
+# This file is part of nmstate
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+
+from contextlib import contextmanager
+import os
+
+
+import pytest
+
+import libnmstate
+from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceType
+from libnmstate.schema import InterfaceState
+
+from .testlib import cmd as libcmd
+
+
+DUMMY0_IFNAME = 'dummy0'
+
+NMCLI_CON_ADD_DUMMY_CMD = [
+    'nmcli',
+    'con',
+    'add',
+    'type',
+    'dummy',
+    'con-name',
+    'testProfile',
+    'connection.autoconnect',
+    'no',
+    'ifname',
+    DUMMY0_IFNAME,
+]
+
+NMCLI_CON_ADD_ETH_CMD = [
+    'nmcli',
+    'con',
+    'add',
+    'type',
+    'ethernet',
+    'con-name',
+    'testProfile',
+    'connection.autoconnect',
+    'no',
+    'ifname',
+    'eth1',
+]
+
+DUMMY_PROFILE_DIRECTORY = '/etc/NetworkManager/system-connections/'
+
+ETH_PROFILE_DIRECTORY = '/etc/sysconfig/network-scripts/'
+
+
+def test_delete_new_interface_inactive_profiles(dummy_inactive_profile):
+    with dummy_interface(dummy_inactive_profile):
+        profile_exists = _profile_exists(
+            DUMMY_PROFILE_DIRECTORY + 'testProfile.nmconnection'
+        )
+        assert not profile_exists
+
+
+def test_delete_existing_interface_inactive_profiles(eth1_up):
+    with create_inactive_profile(eth1_up[Interface.KEY][0][Interface.NAME]):
+        eth1_up[Interface.KEY][0][Interface.MTU] = 2000
+        libnmstate.apply(eth1_up)
+        profile_exists = _profile_exists(
+            ETH_PROFILE_DIRECTORY + 'ifcfg-testProfile'
+        )
+        assert not profile_exists
+
+
+@contextmanager
+def dummy_interface(ifname):
+    desired_state = {
+        Interface.KEY: [
+            {
+                Interface.NAME: ifname,
+                Interface.TYPE: InterfaceType.DUMMY,
+                Interface.STATE: InterfaceState.UP,
+            }
+        ]
+    }
+    libnmstate.apply(desired_state)
+    try:
+        yield
+    finally:
+        dummy0_dstate = desired_state[Interface.KEY][0]
+        dummy0_dstate[Interface.STATE] = InterfaceState.ABSENT
+        libnmstate.apply(desired_state)
+
+
+@pytest.fixture
+def dummy_inactive_profile():
+    libcmd.exec_cmd(NMCLI_CON_ADD_DUMMY_CMD)
+    profile_exists = _profile_exists(
+        DUMMY_PROFILE_DIRECTORY + 'testProfile.nmconnection'
+    )
+    assert profile_exists
+    try:
+        yield DUMMY0_IFNAME
+    finally:
+        libcmd.exec_cmd(_nmcli_delete_connection('testProfile'))
+
+
+@contextmanager
+def create_inactive_profile(con_name):
+    libcmd.exec_cmd(_nmcli_deactivate_connection(con_name))
+    libcmd.exec_cmd(NMCLI_CON_ADD_ETH_CMD)
+    profile_exists = _profile_exists(
+        ETH_PROFILE_DIRECTORY + 'ifcfg-testProfile'
+    )
+    assert profile_exists
+    try:
+        yield
+    finally:
+        libcmd.exec_cmd(_nmcli_delete_connection('testProfile'))
+
+
+def _nmcli_deactivate_connection(con_name):
+    return ['nmcli', 'con', 'down', con_name]
+
+
+def _nmcli_delete_connection(con_name):
+    return ['nmcli', 'con', 'delete', con_name]
+
+
+def _profile_exists(profile_name):
+    return os.path.isfile(profile_name)


cathay4t:
I don't like the idea of deleting all profiles. Can we just use exists one and edit it?

And please investigate why current code in `nm/applier.py` does not work:
```
device.activate(dev=None, connection_id=ifname)
```

EdDev:
> I don't like the idea of deleting all profiles. Can we just use exists one and edit it?

We do not support having more than one profile per device at the moment. When nmstate takes over a device, having more than one profile may cause unwanted results (them taking over automatically).
We actually already cleanup all device profiles when taking a device down/absent.

> And please investigate why current code in nm/applier.py does not work

What if does not have the same name as we expect?

ffmancera:
> I don't like the idea of deleting all profiles. Can we just use exists one and edit it?
> 
> And please investigate why current code in `nm/applier.py` does not work:
> 
> ```
> device.activate(dev=None, connection_id=ifname)
> ```

If we are editing an existing interface we are not doing `device.activate(dev=None, connection_id=ifname)` instead we are doing 
```
for dev, con_profile in ifaces_to_edit:                                     
    device.modify(dev, con_profile)   
```

And `con_profile` is the already existing active profile or the new active profile.

cathay4t:
OK. I got the root cause, the `nm/connection.py` only check on activate connection
for `import_by_device()`, that's why this PR is required.

```python
    def import_by_device(self, nmdev=None):
        ac = get_device_active_connection(nmdev or self.nmdevice)
        if ac:
            if nmdev:
                self.nmdevice = nmdev
            self.profile = ac.props.connection
```

cathay4t:
Please fix the CI failure.

tests/integration/profile_test.py
```diff
@@ -0,0 +1,67 @@
...
+    'nmcli',
+    'con',
+    'add',
+    'type',
+    'dummy',
+    'con-name',
+    'testProfile',
+    'ifname',
+    'dummy0',
+]
```
cathay4t: The newly create profile might have `connection.autoconnect: True` which will activate the profile once created.

I would suggest use libnmstate to create the dummy interface and use `nmcli c down <dummy0>` to deactivated it.

ffmancera: We decided to use `connection.autoconnect no`.

libnmstate/nm/applier.py
```diff
@@ -95,6 +95,9 @@ def prepare_edited_ifaces_configuration(ifaces_desired_state):
             con_profiles.append(cur_con_profile)
         else:
             # Missing connection, attempting to create a new one.
+            device.delete_device_inactive_profiles(
+                iface_desired_state[Interface.NAME]
```
EdDev: This is first used on line 82 so it merits its own variable and just re-using it everywhere.

libnmstate/nm/device.py
```diff
@@ -324,3 +324,13 @@ def get_device_common_info(dev):
         'type_name': dev.get_type_description(),
         'state': dev.get_state(),
     }
+
+
+def delete_device_inactive_profiles(ifname):
```
EdDev: This function has nothing to do with a device.
If you wanted to delete all device profiles, you can use the existing `device` functions.

I think you need to place this under the `connection` module and name it accordingly.
Something like:
```
for con in list_connections_by_ifname(nmclient.client()):
    con.delete()
```
`list_connections_by_ifname` should return in this case `ConnectionProfile` objects.

libnmstate/nm/device.py
```diff
@@ -324,3 +324,13 @@ def get_device_common_info(dev):
...
+    device_profiles = [
+        connection
+        for connection in nmclient.NM.Client.get_connections(nmclient.client())
+        if connection.get_interface_name() == ifname
+    ]
+    for profile in device_profiles:
+        profile.delete()
```
EdDev: The previous recommendation should fix this, but note that you cannot do this outside of the mainloop... and at this stage the mainloop has not yet been started.

tests/integration/profile_test.py
```diff
@@ -0,0 +1,71 @@
...
+
+import libnmstate
+from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceType
+from libnmstate.schema import InterfaceState
+
+from .testlib import cmd as libcmd
+
+
+NMCLI_CONADD_CMD = [
```
EdDev: spelling

tests/integration/profile_test.py
```diff
@@ -0,0 +1,71 @@
...
+            {
+                Interface.NAME: 'dummy0',
+                Interface.TYPE: InterfaceType.DUMMY,
+                Interface.STATE: InterfaceState.UP,
+            }
+        ]
+    }
+    libnmstate.apply(desired_state)
+    desired_state[Interface.KEY][0][Interface.STATE] = InterfaceState.DOWN
+    libnmstate.apply(desired_state)
```
EdDev: This should be part of the nmstate iface creation in a contextmanager:
```
with dummy_interface(...):
    assert not _profile_exists()
```
You need to assert that the inactive profile is not there anymore after defining the dummy using nmstate, not after setting it `down`.

tests/integration/profile_test.py
```diff
@@ -0,0 +1,71 @@
...
+]
+
+PROFILE_DIRECTORY = '/etc/NetworkManager/system-connections/'
+
+
+def test_delete_inactive_profiles():
+    libcmd.exec_cmd(NMCLI_CONADD_CMD)
+    profile_exists = os.path.isfile(
+        PROFILE_DIRECTORY + 'testProfile.nmconnection'
+    )
```
EdDev: please use a helper function. `_profile_exists()`

tests/integration/profile_test.py
```diff
@@ -0,0 +1,71 @@
...
+    assert profile_exists
+
```
EdDev: This is part of the test setup and not the test.
Please create a fixture that sets this up and returns the name of the interface (to be used in the nmstate iface creation).

libnmstate/nm/applier.py
ffmancera: Done!

libnmstate/nm/device.py
ffmancera: Done, thanks!

tests/integration/profile_test.py
ffmancera: Oops

tests/integration/profile_test.py
ffmancera: Fixed. Thanks!

tests/integration/profile_test.py
ffmancera: Done!

tests/integration/profile_test.py
ffmancera: Done! Thanks!

libnmstate/nm/device.py
ffmancera: Now that should be fixed. Thanks!

EdDev: > list_connections_by_ifname should return in this case ConnectionProfile objects.

You have returned a NM object and not the nmstate one. Therefore the action of deletion is not performed in the mainloop.

libnmstate/nm/connection.py
```diff
@@ -450,3 +450,16 @@ def get_device_active_connection(nm_device):
     if nm_device:
         active_conn = nm_device.get_active_connection()
     return active_conn
+
+
+def list_connections_by_ifname(ifname):
```
EdDev: Nit: Please place it below the caller.

tests/integration/profile_test.py
```diff
@@ -0,0 +1,89 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+
+import os
+
```
EdDev: These two imports are stdlib, no newlines are needed.

tests/integration/profile_test.py
```diff
@@ -0,0 +1,89 @@
...
+
+from contextlib import contextmanager
+
+import libnmstate
+from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceType
+from libnmstate.schema import InterfaceState
+
+import pytest
```
EdDev: This is a 3rd party package, please place below stdlib ones.

tests/integration/profile_test.py
```diff
@@ -0,0 +1,89 @@
...
+        profile_exists = _profile_exists('testProfile.nmconnection')
+        assert not profile_exists
+
+
+def _profile_exists(profile_name):
+    return os.path.isfile(PROFILE_DIRECTORY + profile_name)
+
+
+@contextmanager
+def dummy_interface(ifname):
```
EdDev: Caller before callee please.

tests/integration/profile_test.py
```diff
@@ -0,0 +1,89 @@
...
+        libnmstate.apply(desired_state)
+
+
+@pytest.fixture
+def dummy_prepare():
+    libcmd.exec_cmd(NMCLI_CON_ADD_CMD)
+    profile_exists = _profile_exists('testProfile.nmconnection')
+    assert profile_exists
+
+    return DUMMY0_IFNAME
```
EdDev: You assume here that the test will pass. If it does not, we are left with the profile as a leftover.
There is a need to cleanup here:
try: yield ... finally: libcmd.exec_cmd(...)

tests/integration/profile_test.py
```diff
@@ -0,0 +1,89 @@
...
+                    Interface.NAME: ifname,
+                    Interface.TYPE: InterfaceType.DUMMY,
+                    Interface.STATE: InterfaceState.UP,
+                }
+            ]
+        }
+        libnmstate.apply(desired_state)
+        yield
+    finally:
+        desired_state[Interface.KEY][0][Interface.STATE] = InterfaceState.DOWN
```
EdDev: I think you meant ‘absent’.

tests/integration/profile_test.py
```diff
@@ -0,0 +1,89 @@
...
+        libnmstate.apply(desired_state)
+
+
+@pytest.fixture
+def dummy_prepare():
```
EdDev: How about this name: dummy_inactive_profile

tests/integration/profile_test.py
```diff
@@ -0,0 +1,89 @@
...
+    'connection.autoconnect',
+    'no',
+    'ifname',
+    DUMMY0_IFNAME,
+]
+
+PROFILE_DIRECTORY = '/etc/NetworkManager/system-connections/'
+
+
+def test_delete_inactive_profiles(dummy_prepare):
```
EdDev: I think this test will cover only one part of the production code. The other part is for editing an existing iface which has a non active profile.

This test is about a profile with a non existent iface.
When you add the other test, you’ll need to specify this in the name of the current test name.

libnmstate/nm/device.py
ffmancera: list_connections_by_ifname returns now ConnectionProfile objects. Thanks!

libnmstate/nm/connection.py
ffmancera: Done!

tests/integration/profile_test.py
ffmancera: Fixed

tests/integration/profile_test.py
ffmancera: Done!

tests/integration/profile_test.py
ffmancera: What do you mean here? That is already done. Thanks!

tests/integration/profile_test.py
ffmancera: Sure, fixed.

tests/integration/profile_test.py
ffmancera: Yes, sorry. Thanks for point it out.

tests/integration/profile_test.py
ffmancera: I've create a new test and renamed the existing one. Thanks!

tests/integration/profile_test.py
ffmancera: Done!

tests/integration/profile_test.py
```diff
@@ -0,0 +1,125 @@
...
+            ]
+        }
+        libnmstate.apply(desired_state)
+        profile_exists = _profile_exists('testProfile.nmconnection')
+        assert not profile_exists
+
+
+@contextmanager
+def dummy_interface(ifname):
+    try:
```
EdDev: The `try` should start after the `apply`:
```
...
try
    yield
finally:
    ...
```

tests/integration/profile_test.py
```diff
@@ -0,0 +1,125 @@
...
+        yield
+    finally:
+        dummy0_dstate = desired_state[Interface.KEY][0]
+        dummy0_dstate[Interface.STATE] = InterfaceState.ABSENT
+        libnmstate.apply(desired_state)
+
+
+@pytest.fixture
+def dummy_inactive_profile():
+    try:
```
EdDev: The try is either after the exec_cmd or just before the yield.
(if the command to create fails, it means there is nothing there, so there is nothing to clean)

tests/integration/profile_test.py
EdDev: Now it is, thanks.

tests/integration/profile_test.py
```diff
@@ -0,0 +1,125 @@
...
+
+def test_delete_new_interface_inactive_profiles(dummy_inactive_profile):
+    with dummy_interface(dummy_inactive_profile):
+        profile_exists = _profile_exists('testProfile.nmconnection')
+        assert not profile_exists
+
+
+def test_delete_existing_interface_inactive_profiles(dummy_inactive_profile):
+    with dummy_interface(dummy_inactive_profile):
+        _prepare_profiles_for_editting(DUMMY0_IFNAME)
```
EdDev: I do not understand what you have done here.

tests/integration/profile_test.py
ffmancera: Done! Thanks!

tests/integration/profile_test.py
ffmancera: Done, thanks!

tests/integration/profile_test.py
ffmancera: Take a look to the new test. Thanks!

tests/integration/profile_test.py
```diff
@@ -0,0 +1,142 @@
...
+def test_delete_new_interface_inactive_profiles(dummy_inactive_profile):
+    with dummy_interface(dummy_inactive_profile):
+        profile_exists = _profile_exists(
+            DUMMY_PROFILE_DIRECTORY + 'testProfile.nmconnection'
+        )
+        assert not profile_exists
+
+
+def test_delete_existing_interface_inactive_profiles(eth1_up):
+    _create_inactive_profile('eth1')
```
EdDev: Please use `eth1_up[Interface.KEY][0][Interface.NAME]`

tests/integration/profile_test.py
```diff
@@ -0,0 +1,142 @@
...
+        DUMMY_PROFILE_DIRECTORY + 'testProfile.nmconnection'
+    )
+    assert profile_exists
+    try:
+        yield DUMMY0_IFNAME
+    finally:
+        libcmd.exec_cmd(_nmcli_delete_connection('testProfile'))
+
+
+def _create_inactive_profile(con_name):
```
EdDev: This duplicates the fixture code and is not marked as a contextmanager (and not used as such by the caller).

I do not think the deactivation is a must and after you make it a contextmanager, you can use it from the fixture.

tests/integration/profile_test.py
ffmancera: Done!

EdDev:
@cathay4t , seems your comments have been addressed. I'm merging this PR.
If I missed something, please let me know.

PR #588, opened 2019-11-19
integration tests: Remove preserve_old_config fixture

tyll:
The `preserve_old_config` fixture fails when Nmstate reports a
configuration that it cannot set, for example a gateway together with
DHCPv6 and autoconf. If someone needs to restore a certain network state
after running the tests, they need to use a custom wrapper/script to do
so.

References:
https://lists.fedorahosted.org/archives/list/nmstate-devel@lists.fedorahosted.org/thread/XFMPA5IMYCQCDWUSSNGVDBQA7BKB2N3N/
https://nmstate.atlassian.net/browse/NMSTATE-275

---
This change is [Reviewable](https://reviewable.io/reviews/nmstate/nmstate/588)

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -22,8 +22,6 @@ import subprocess
 
 import pytest
 
-import libnmstate
-
 from .testlib import ifacelib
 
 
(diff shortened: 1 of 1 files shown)


tyll:
Added a new check to warn in case the Nmstate state differs after the test run.

PR #591, opened 2019-11-19
vlan filtering, validator: a trunk port must have at least one vlan tag

maiqueb:
Using the currently defined schema, a trunk port *must* feature
at least 2 trunk tags, otherwise, it is impossible to identify the
correct port type when reading from NetworkManager via libnm.

This patch updates the validator accordingly.

### DIFF
--- a/libnmstate/validator.py
+++ b/libnmstate/validator.py
@@ -250,9 +250,18 @@ def _assert_vxlan_has_missing_attribute(state, *attributes):
 
 def _assert_vlan_filtering_trunk_tags(ports_state):
     for port_state in ports_state:
-        trunk_tags = port_state.get(LB.Port.VLAN_SUBTREE, {}).get(
-            LB.Port.Vlan.TRUNK_TAGS, []
-        )
+        port_vlan_state = port_state.get(LB.Port.VLAN_SUBTREE, {})
+        vlan_type = port_vlan_state.get(LB.Port.Vlan.TYPE)
+        trunk_tags = port_vlan_state.get(LB.Port.Vlan.TRUNK_TAGS, [])
+
+        if vlan_type == LB.Port.Vlan.ACCESS_TYPE:
+            if trunk_tags:
+                raise NmstateValueError('Access port cannot have trunk tags')
+        elif port_vlan_state:
+            if not trunk_tags:
+                raise NmstateValueError(
+                    'A trunk port needs to specify trunk tags'
+                )
         for trunk_tag in trunk_tags:
             _assert_vlan_filtering_trunk_tag(trunk_tag)
 
--- a/tests/lib/validator_test.py
+++ b/tests/lib/validator_test.py
@@ -341,6 +341,64 @@ class TestVxlanValidation(object):
 
 
 class TestVlanFilteringValidation(object):
+    def test_access_port_cant_have_trunks(self):
+        invalid_vlan_config = {
+            LB.Port.Vlan.TYPE: LB.Port.Vlan.ACCESS_TYPE,
+            LB.Port.Vlan.TRUNK_TAGS: [
+                {
+                    LB.Port.Vlan.TrunkTags.ID_RANGE: {
+                        LB.Port.Vlan.TrunkTags.MIN_RANGE: 200,
+                        LB.Port.Vlan.TrunkTags.MAX_RANGE: 299,
+                    }
+                }
+            ],
+        }
+        desired_state = {
+            schema.Interface.KEY: [
+                {
+                    schema.Interface.NAME: 'br0',
+                    schema.Interface.TYPE: LB.TYPE,
+                    schema.Interface.STATE: schema.InterfaceState.UP,
+                    LB.PORT_SUBTREE: [
+                        {
+                            LB.PORT_NAME: 'eth1',
+                            LB.Port.VLAN_SUBTREE: invalid_vlan_config,
+                        }
+                    ],
+                }
+            ]
+        }
+        with pytest.raises(
+            NmstateValueError, match='Access port cannot have trunk tags'
+        ):
+            libnmstate.validator.validate_bridge(desired_state)
+
+    def test_trunk_port_must_have_at_least_1_tag(self):
+        invalid_vlan_config = {
+            LB.Port.Vlan.TYPE: LB.Port.Vlan.TRUNK_TYPE,
+            LB.Port.Vlan.TRUNK_TAGS: [],
+        }
+        desired_state = {
+            schema.Interface.KEY: [
+                {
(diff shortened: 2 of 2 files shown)


libnmstate/validator.py
```diff
@@ -251,9 +251,23 @@ def _assert_vxlan_has_missing_attribute(state, *attributes):
...
+        else:
+            if len(trunk_tags) == 1:
+                if not trunk_tags[0].get(LB.Port.Vlan.TrunkTags.ID_RANGE):
+                    raise NmstateValueError(
+                        'Trunk port must feature multiple vlan tags'
+                    )
+            elif not trunk_tags:
+                raise NmstateValueError(
+                    'Trunk port must feature multiple vlan tags'
+                )
```
ffmancera: Would this be better?
```
try:
    if trunk_tags[0][LB.Port.Vlan.TrunkTags.ID_RANGE]
        continue
except:
    raise NmstateValueError(
        'Trunk port must feature multiple vlan tags'
    )
```

tyll:
CI fails, please take a look

libnmstate/validator.py
tyll: I like the idea to avoid the duplicate exception. The problem is that it also needs to be raised if `trunk_tags[0][LB.Port.Vlan.TrunkTags.ID_RANGE]:` is false.

libnmstate/validator.py
```diff
@@ -251,9 +251,23 @@ def _assert_vxlan_has_missing_attribute(state, *attributes):
...
+        port_vlan_state = port_state.get(LB.Port.VLAN_SUBTREE, {})
+        port_type = port_vlan_state[LB.Port.Vlan.TYPE]
+        trunk_tags = port_vlan_state.get(LB.Port.Vlan.TRUNK_TAGS, [])
+
+        if port_type == LB.Port.Vlan.ACCESS_TYPE:
+            if trunk_tags:
+                raise NmstateValueError('Access port cannot have trunk tags')
+        else:
+            if len(trunk_tags) == 1:
+                if not trunk_tags[0].get(LB.Port.Vlan.TrunkTags.ID_RANGE):
```
tyll: Is it ok if l `len(trunk_tags` is 2 and `trunk_tags[0].get(LB.Port.Vlan.TrunkTags.ID_RANGE)` and `trunk_tags[1].get(LB.Port.Vlan.TrunkTags.ID_RANGE)` are both False or does this need to check all items of `trunk_tags`?

maiqueb: The important thing here is that we have at least 2 trunk tags in the port vlan config.

Hence, 1 tag range is OK; >= 2 single tags are also OK. Unless I'm mistaken, the latter is the scenario you're asking about.

tyll: Can you adjust the error message to say something like `A trunk port needs to specify a vlan tag range or multiple tags` to clarify this?

libnmstate/validator.py
maiqueb: Avoiding that duplication would be the icing on the cake, but I'm not sure how to make the code look simpler.

I welcome any other suggestion. Re-open the conversation if you have one.

libnmstate/validator.py
maiqueb: Sure. Done.

maiqueb:
> > Using the currently defined schema, a trunk port _must_ feature
> > at least 2 trunk tags, otherwise, it is impossible to identify the
> > correct port type when reading from NetworkManager via libnm.
> 
> Sorry for the late question, but why is it impossible? A linux-bridge is supposed to allow such configuration. Is it an NM problem? Can you explain this?

I expected this, the objective of this PR is mostly so we can discuss this.

When I read the port settings, I get an array of BridgeVlan objects - [0]. 

How can I know from those objects if the port is an access / trunk port ? My proposal is to do it based on the number or tags configured on a port: trunks have > 1 tag. 

Let's hear you out; I'm open to any other alternatives.

[0] - https://valadoc.org/libnm/NM.BridgeVlan.html

EdDev:
Maybe I am missing something but here is my attempt to do it:
- If it is pvid and untagged, it may be an access or a native of a port.
— if there is another tagged, it is a native, otherwise it is an access.

Limitations based on the above:
- Do not support setting a trunk mode with no trunks at all.

maiqueb:
> Maybe I am missing something but here is my attempt to do it:
> 
> * If it is pvid and untagged, it may be an access or a native of a port.
>   — if there is another tagged, it is a native, otherwise it is an access.
> 
> Limitations based on the above:
> 
> * Do not support setting a trunk mode with no trunks at all.
> * Do not support pvid without untagged and vice versa.

Fair enough.

I'll update #576 accordingly once I can spare the time.

libnmstate/validator.py
```diff
@@ -250,9 +250,20 @@ def _assert_vxlan_has_missing_attribute(state, *attributes):
 
 def _assert_vlan_filtering_trunk_tags(ports_state):
     for port_state in ports_state:
-        trunk_tags = port_state.get(LB.Port.VLAN_SUBTREE, {}).get(
-            LB.Port.Vlan.TRUNK_TAGS, []
-        )
+        port_vlan_state = port_state.get(LB.Port.VLAN_SUBTREE, {})
+        if not port_vlan_state:
+            return
```
EdDev: Looks like a bug to me.

libnmstate/validator.py
```diff
@@ -250,9 +250,20 @@ def _assert_vxlan_has_missing_attribute(state, *attributes):
...
+        port_type = port_vlan_state[LB.Port.Vlan.TYPE]
```
EdDev: vlan_type?

libnmstate/validator.py
maiqueb: I just wanted to bail out early if there's no port vlan configuration.

Is it better now ?

libnmstate/validator.py
maiqueb: I sincerely don't understand. Please elaborate.

libnmstate/validator.py
EdDev: You can bail out for a specific port, but not for all because of one.

libnmstate/validator.py
EdDev: port_type should probably be vlan_type.

libnmstate/validator.py
maiqueb: touchè.

libnmstate/validator.py
maiqueb: ok, I had it as a 'trunk port' or 'access port', but I'd say yours fits the definition better. 

Let's move forward anyway :)

PR #592, opened 2019-11-20
schema: add SR-IOV basic schema

ffmancera:

### DIFF
--- a/libnmstate/nm/sriov.py
+++ b/libnmstate/nm/sriov.py
@@ -1,0 +1,32 @@
+#
+# Copyright (c) 2019 Red Hat, Inc.
+#
+# This file is part of nmstate
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+
+from libnmstate.error import NmstateNotImplementedError
+from libnmstate.schema import Ethernet
+
+
+def create_setting(iface_state):
+    setting = None
+    sriov_config = iface_state.get(Ethernet.CONFIG_SUBTREE, {}).get(
+        Ethernet.SRIOV_SUBTREE
+    )
+    if sriov_config:
+        raise NmstateNotImplementedError
+
+    return setting
--- a/libnmstate/schema.py
+++ b/libnmstate/schema.py
@@ -198,6 +198,11 @@ class Ethernet(object):
     FULL_DUPLEX = 'full'
     HALF_DUPLEX = 'half'
 
+    SRIOV_SUBTREE = 'sr-iov'
+
+    class SRIOV(object):
+        TOTAL_VFS = 'total-vfs'
+
 
 class VLAN(object):
     TYPE = InterfaceType.VLAN
--- a/libnmstate/schemas/operational-state.yaml
+++ b/libnmstate/schemas/operational-state.yaml
@@ -163,6 +163,12 @@ definitions:
           minimum: 0
         flow-control:
           type: boolean
+        sr-iov:
+          type: object
+          properties:
+            total-vfs:
+              type: integer
+              minimum: 0
   interface-vlan:
     rw:
       properties:
--- a/tests/integration/sriov_test.py
+++ b/tests/integration/sriov_test.py
@@ -1,0 +1,38 @@
+#
+# Copyright (c) 2019 Red Hat, Inc.
+#
+# This file is part of nmstate
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+
+import pytest
+
+import libnmstate
+from libnmstate.error import NmstateNotImplementedError
+from libnmstate.schema import Ethernet
+from libnmstate.schema import Interface
+
+
+SRIOV_CONFIG = {Ethernet.SRIOV_SUBTREE: {Ethernet.SRIOV.TOTAL_VFS: 2}}
+
+
+@pytest.mark.xfail(
+    raises=NmstateNotImplementedError,
+    reason='SR-IOV is not supported yet',
+    strict=True,
+)
+def test_sriov_not_implemented(eth1_up):
+    eth1_up[Interface.KEY][0][Ethernet.CONFIG_SUBTREE] = SRIOV_CONFIG
+    libnmstate.apply(eth1_up)
--- a/tests/lib/schema_validation_test.py
+++ b/tests/lib/schema_validation_test.py
@@ -25,9 +25,11 @@ import jsonschema as js
 import libnmstate
 from libnmstate.schema import Constants
 from libnmstate.schema import DNS
+from libnmstate.schema import Ethernet
 from libnmstate.schema import LinuxBridge as LB
 from libnmstate.schema import VXLAN
 from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceType
 
 INTERFACES = Constants.INTERFACES
 ROUTES = Constants.ROUTES
@@ -167,6 +169,34 @@ class TestIfaceTypeEthernet(object):
         del default_data[INTERFACES][0]['link-speed']
 
         libnmstate.validator.validate(default_data)
+
+    @pytest.mark.parametrize('valid_values', [0, 150, 256])
+    def test_valid_with_sriov_total_vfs(self, default_data, valid_values):
+        default_data[Interface.KEY][0].update(
+            {
+                Interface.TYPE: InterfaceType.ETHERNET,
+                Ethernet.SRIOV_SUBTREE: {
+                    Ethernet.SRIOV.TOTAL_VFS: valid_values
+                },
+            }
+        )
+        libnmstate.validator.validate(default_data)
+
+    @pytest.mark.parametrize('invalid_values', [-50, -1])
+    def test_over_maximum_total_vfs_is_invalid(
+        self, default_data, invalid_values
+    ):
+        default_data[Interface.KEY][0].update(
+            {
+                Interface.TYPE: InterfaceType.ETHERNET,
+                Ethernet.SRIOV_SUBTREE: {
+                    Ethernet.SRIOV.TOTAL_VFS: invalid_values
+                },
+            }
+        )
+
+        with pytest.raises(js.ValidationError, match=str(invalid_values)):
+            libnmstate.validator.validate(default_data)
 
 
 class TestIfaceTypeVxlan(object):
--- a/libnmstate/nm/applier.py
+++ b/libnmstate/nm/applier.py
@@ -34,6 +34,7 @@ from . import ipv4
 from . import ipv6
 from . import nmclient
 from . import ovs
+from . import sriov
 from . import translator
 from . import user
 from . import vlan
@@ -411,6 +412,10 @@ def _build_connection_profile(iface_desired_state, base_con_profile=None):
     vxlan_setting = vxlan.create_setting(iface_desired_state, base_profile)
     if vxlan_setting:
         settings.append(vxlan_setting)
+
+    sriov_setting = sriov.create_setting(iface_desired_state)
+    if sriov_setting:
+        settings.append(sriov_setting)
 
     new_profile = connection.ConnectionProfile()
     new_profile.create(settings)


ffmancera:
This is a proposal for a basic SR-IOV schema with this we will be able to set and get the total number of VFs.

libnmstate/schema.py
```diff
@@ -200,6 +200,14 @@ class Ethernet(object):
     FULL_DUPLEX = 'full'
     HALF_DUPLEX = 'half'
 
+    SR_IOV = 'sr-iov'
+
+
+class SRIOV(object):
+    CONFIG_SUBTREE = 'sr-iov'
```
tyll: You can re-use the `Ethernet.SR_IOV` constant here. Can you please also explain in the commit message why it is a subtree of Ethernet?

tests/lib/schema_validation_test.py
```diff
@@ -170,6 +170,10 @@ def test_valid_without_auto_neg_and_missing_speed(self, default_data):
 
         libnmstate.validator.validate(default_data)
 
+    def test_valid_with_sriov_total_vfs(self, default_data):
+        default_data[Interface.KEY][0].update({'sr-iov': {'total-vfs': 5}})
```
tyll: Please use the constants here for `sr-iov` and `total-vfs`.

tyll:
We probably didn't do it in the past but IMHO it would make sense to add a simple code snippet to Nmstate that will raise an exception that SR-IOV is not yet supported if it is specified in a state to keep the code consistent. Otherwise the schema implies to an outsider that it is supported but Nmstate will silently fail.

libnmstate/schema.py
ffmancera: Done! The explanation added is: `SR-IOV is a subtree of ethernet because it contains the PF information which must be an ethernet interface.`

tests/lib/schema_validation_test.py
ffmancera: Done! Thanks!

ffmancera:
> We probably didn't do it in the past but IMHO it would make sense to add a simple code snippet to Nmstate that will raise an exception that SR-IOV is not yet supported if it is specified in a state to keep the code consistent. Otherwise the schema implies to an outsider that it is supported but Nmstate will silently fail.

I've added a `nm/sriov.py` placeholder which raises a `NmstateNotImplementedError` exception if nmstate finds `SRIOV.CONFIG_SUBTREE` in a desired interface state.

libnmstate/nm/sriov.py
```diff
@@ -0,0 +1,29 @@
...
+from libnmstate.error import NmstateNotImplementedError
+from libnmstate.schema import SRIOV
+
+
+def create_setting(iface_state):
+    setting = iface_state.get(SRIOV.CONFIG_SUBTREE)
+    if setting:
+        raise NmstateNotImplementedError
+
+    return setting
```
tyll: While the code works as expected, using the variable `setting` for both the config and the return value is a little bit misleading because it would have to return some NM object and not a a changed config subtree. Therefore `return None` would be better and maybe even using `config` instead of `setting` for the variable that contains the config. Or just avoid the variables,

```
if iface_state.get(SRIOV.CONFIG_SUBTREE):
    raise NmstateNotImplementedError

return None
```

Ideally you would also add a small integration tests that ensures that `create_setting()` raises the exception when the config is used in  state.

libnmstate/schemas/operational-state.yaml
```diff
@@ -163,6 +163,12 @@ definitions:
           minimum: 0
         flow-control:
           type: boolean
+        sr-iov:
+          type: object
+          properties:
+            total-vfs:
```
maiqueb: I think you should also set the max value of total-vfs to 256.

ffmancera:
> Since you've bound the variable in the schema (min=0) - and I'm requesting you to also set a max - I think you should add some negative testing for those scenarios.
> 
> A question: if I understand correctly, your input for this task is bug [0]. Why are you only defining the max number of virtual functions ? (I'm 100% sure I'm missing some somethings :) )
> 
> [0] - https://bugzilla.redhat.com/show_bug.cgi?id=1695116

That seems fine, thanks! We have decided to implement first the basics and then extend that. You can check sprint planning for this :-)

tests/lib/schema_validation_test.py
```diff
@@ -170,6 +171,34 @@ def test_valid_without_auto_neg_and_missing_speed(self, default_data):
...
+    def test_negative_total_vfs_is_invalid(self, default_data):
+        default_data[Interface.KEY][0].update(
+            {SRIOV.CONFIG_SUBTREE: {SRIOV.TOTAL_VFS: -3}}
+        )
+        libnmstate.validator.validate(default_data)
+
+        with pytest.raises(js.ValidationError) as err:
+            libnmstate.validator.validate(default_data)
+
+        assert '-3 is lower than the minimum of 0' in err.value.args[0]
```
maiqueb: wondering if you can't use the pytest.raises(... **match**) keyword here instead of this (IMHO) ugly assert.

100 % optional, I just think it has potential to look better.

tests/lib/schema_validation_test.py
```diff
@@ -170,6 +171,34 @@ def test_valid_without_auto_neg_and_missing_speed(self, default_data):
...
+    def test_over_maximum_total_vfs_is_invalid(self, default_data):
+        default_data[Interface.KEY][0].update(
+            {SRIOV.CONFIG_SUBTREE: {SRIOV.TOTAL_VFS: 300}}
+        )
+        libnmstate.validator.validate(default_data)
+
+        with pytest.raises(js.ValidationError) as err:
+            libnmstate.validator.validate(default_data)
+
+        assert '300 is greater than the maximum of 256' in err.value.args[0]
```
maiqueb: ditto.

tests/integration/sriov_test.py
```diff
@@ -0,0 +1,50 @@
...
+import libnmstate
+from libnmstate.error import NmstateNotImplementedError
+from libnmstate.schema import Ethernet
+from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceState
+from libnmstate.schema import InterfaceType
+from libnmstate.schema import SRIOV
+
+
+@pytest.mark.xfail(
```
maiqueb: Hm! Neat. Does this throw you a warning resembling this one ? "UserWarning: Network state after test run does not match network state before test run:"

libnmstate/schemas/operational-state.yaml
ffmancera: Done! Thanks!

tests/lib/schema_validation_test.py
ffmancera: It looks much better. Thanks!

tests/lib/schema_validation_test.py
ffmancera: Same here. Please look above

tests/integration/sriov_test.py
ffmancera: No. But I think this warning is not related to your changes. You can check it using `pytest-args='-k <your_test_name>'` when executing run-test.sh

libnmstate/schemas/operational-state.yaml
tyll: > I think you should also set the max value of total-vfs to 256.

Where does this number come from? I was wondering if there was an upper limit as well but it seemed to depend on the actual hardware. Also `256` is a strange limit, if it corresponds to an unsigned byte, the maximum would be `255` since the minimum is `0` - or does `0` indicate that there is no support and there could be `256` different VFs with IDs `0` to `255`?

libnmstate/nm/sriov.py
```diff
@@ -0,0 +1,33 @@
...
+
+from libnmstate.error import NmstateNotImplementedError
+from libnmstate.schema import Ethernet
+from libnmstate.schema import SRIOV
+
+
+def create_setting(iface_state):
+    setting = None
+    ethernet_setting = iface_state.get(Ethernet.CONFIG_SUBTREE)
+    if ethernet_setting:
```
tyll: You can simplify this as

```
setting = None
sriov_config = iface_state.get(Ethernet.CONFIG_SUBTREE, {}).get(SRIOV.CONFIG_SUBTREE)
if sriov_config:
    raise NmstateNotImplementedError

return setting
```
This avoids the multiple ifs. Hope this will please the linter.

tests/lib/schema_validation_test.py
```diff
@@ -170,6 +171,30 @@ def test_valid_without_auto_neg_and_missing_speed(self, default_data):
 
         libnmstate.validator.validate(default_data)
 
+    def test_valid_with_sriov_total_vfs(self, default_data):
```
tyll: Personally, I like testing the edge cases explicitly, which would be -1, 0, 256, 257 and maybe few extra values beside these. Since there are two categories, valid and invalid, it makes in my opinion sense to use `@pytest.parametrize` and one test for valid values and one for invalid values. What do you think?

tests/integration/sriov_test.py
tyll: @maiqueb it is unlikely that this test will expose this because `eth1` is cleaned up a lot of times. I guess for virtual interfaces, it will happen less likely when #593 is merged.

tests/integration/sriov_test.py
```diff
@@ -0,0 +1,50 @@
...
+    raises=NmstateNotImplementedError,
+    reason='SR-IOV is not supported yet',
+    strict=True,
+)
+def test_sriov_not_implemented():
+    desired_state = {
+        Interface.KEY: [
+            {
+                Interface.NAME: 'eth1',
```
tyll: Please make this a constant, maybe `SRIOV_NIC`. Even when it is implemented, this will only work on special hardware and not in the CI.

ffmancera: Yes, in the implementation PR we will need to check if we have SR-IOV support or not.

tests/lib/schema_validation_test.py
ffmancera: That seems better. Will do, thanks!

libnmstate/nm/sriov.py
ffmancera: Yes, that is fine. Thanks!

tests/lib/schema_validation_test.py
ffmancera: Done!

tests/integration/sriov_test.py
maiqueb: tks @tyll .

tests/integration/sriov_test.py
```diff
@@ -0,0 +1,46 @@
...
+    Ethernet.CONFIG_SUBTREE: {SRIOV.CONFIG_SUBTREE: {SRIOV.TOTAL_VFS: 2}},
+}
+
+
+@pytest.mark.xfail(
+    raises=NmstateNotImplementedError,
+    reason='SR-IOV is not supported yet',
+    strict=True,
+)
+def test_sriov_not_implemented():
```
tyll: This should probably use the `eth1_up` fixture since the state uses `eth1`.

tests/integration/sriov_test.py
```diff
@@ -0,0 +1,46 @@
...
+import libnmstate
+from libnmstate.error import NmstateNotImplementedError
+from libnmstate.schema import Ethernet
+from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceState
+from libnmstate.schema import InterfaceType
+from libnmstate.schema import SRIOV
+
+
+SRIOV_NIC = {
```
tyll: Initially I meant to use `SRIOV_NIC = "eth1"`. If you are using it for the full state, it should be something like SRIOV_CONFIG.

tests/integration/sriov_test.py
ffmancera: Done!

tests/integration/sriov_test.py
ffmancera: Done! Thanks!

tyll:
It seems that the integration test is not working as expected (it is not failing): https://travis-ci.com/nmstate/nmstate/jobs/260153026

ffmancera:
> It seems that the integration test is not working as expected (it is not failing): https://travis-ci.com/nmstate/nmstate/jobs/260153026

Yes, just a typo. Fixing it.

tests/lib/schema_validation_test.py
```diff
@@ -170,6 +181,30 @@ def test_valid_without_auto_neg_and_missing_speed(self, default_data):
...
+    ):
+        default_data[Interface.KEY][0].update(
+            {
+                Interface.TYPE: InterfaceType.ETHERNET,
+                SRIOV.CONFIG_SUBTREE: {SRIOV.TOTAL_VFS: invalid_values},
+            }
+        )
+
+        with pytest.raises(js.ValidationError):
+            libnmstate.validator.validate(default_data)
```
maiqueb: shouldn't you verify the error msg as well ? What I had suggested in a previous comment was for you to use the *match* keyword instead, no for your to stop checking that.

I meant something like:
```
with pytest.raises(ValueError, match=r".* 123 .*"):
    myfunc()
```

From [0] I get the understanding this is pytest's way of asserting on py3.

[0] - https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions

ffmancera: That seems fine. Done! Thanks :-)

libnmstate/schemas/operational-state.yaml
```diff
@@ -163,6 +163,13 @@ definitions:
...
+              type: integer
+              minimum: 0
+              maximum: 256
```
maiqueb: Should this allow 256? I think it shouldn't.

tests/lib/schema_validation_test.py
```diff
@@ -170,6 +181,30 @@ def test_valid_without_auto_neg_and_missing_speed(self, default_data):
...
+        self, default_data, invalid_values
+    ):
+        default_data[Interface.KEY][0].update(
+            {
+                Interface.TYPE: InterfaceType.ETHERNET,
+                SRIOV.CONFIG_SUBTREE: {SRIOV.TOTAL_VFS: invalid_values},
+            }
+        )
+
+        with pytest.raises(js.ValidationError, match=str(invalid_values)):
```
maiqueb: not sure it warrants the trouble, but would be better to have it w/ 

```python
invalid_value = 300
max_or_min = 'maximum' if invalid_value > 255 else 'minimum'
comparator = 'greater' if invalid_value > 255 else 'smaller'
with pytest.raises(js.ValidationError, match=f'{invalid_value} is {comparator} than the allowed {max_or_min} of 256'):
    ...
```
Feel free to disregard.

libnmstate/schemas/operational-state.yaml
tyll: The number came from you: https://github.com/nmstate/nmstate/pull/592#discussion_r349574736
Now that you mention it again, I remember that I asked you about it as well:
https://github.com/nmstate/nmstate/pull/592#discussion_r349795251

tests/lib/schema_validation_test.py
tyll: This seems to be testing too many internals of jsonschema that Nmstate does not care about (does it @EdDev ?)

libnmstate/schemas/operational-state.yaml
maiqueb: This is the only reference I could find of the 256 number.
https://www.itprotoday.com/compute-engines/q-what-sr-iov

Even that doc refers it's HW specific, so nmstate shouldn't meddle.

Sorry for misleading you @ffmancera. I think you should remove the 'maximum' .

ffmancera: No problem about that. I am going to remove the `maximum`. Thanks!

tests/lib/schema_validation_test.py
ffmancera: If we are removing the `maximum` limit then we can simplify this a lot. Thanks!

libnmstate/schema.py
```diff
@@ -200,6 +200,13 @@ class Ethernet(object):
     FULL_DUPLEX = 'full'
     HALF_DUPLEX = 'half'
 
+    SR_IOV = 'sr-iov'
+
+    class SRIOV(object):
+        CONFIG_SUBTREE = 'sr-iov'
```
tyll: AFAIU the LinuxBridge example, this constant should be a member of the Ethernet class.

ffmancera: Right, done! Thanks

tests/lib/schema_validation_test.py
```diff
@@ -145,6 +147,15 @@ def test_invalid_type(self, default_data):
...
 
 class TestIfaceTypeEthernet(object):
+
+    parametrize_sriov_valid_total_vfs = pytest.mark.parametrize(
+        'valid_values', [0, 150, 256]
+    )
+
+    parametrize_sriov_invalid_total_vfs = pytest.mark.parametrize(
+        'invalid_values', [-1, 257, 400]
+    )
```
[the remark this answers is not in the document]
ffmancera: Done! Thanks

libnmstate/schema.py
```diff
@@ -200,6 +200,14 @@ class Ethernet(object):
     FULL_DUPLEX = 'full'
     HALF_DUPLEX = 'half'
 
+    SR_IOV = 'sr-iov'
+
+
+class SRIOV(object):
```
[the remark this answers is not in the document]
ffmancera: We have agreed on that. Done, thanks!

ffmancera:
Rebased, thanks!

PR #593, opened 2019-11-20
NM, applier: Explicitly remove any virutal device

almusil:
Virtual devices that are owned by nmstate should be removed.

### DIFF
--- a/libnmstate/nm/applier.py
+++ b/libnmstate/nm/applier.py
@@ -170,17 +170,16 @@ def set_ifaces_admin_state(ifaces_desired_state, con_profiles=()):
                 InterfaceState.ABSENT,
             ):
                 nmdevs = _get_affected_devices(iface_desired_state)
-                for nmdev in nmdevs:
-                    remove_devs_actions[nmdev] = [
+                for affected_nmdev in nmdevs:
+                    remove_devs_actions[affected_nmdev] = [
                         device.deactivate,
                         device.delete,
                     ]
-                    if nmdev.get_device_type() in (
-                        nmclient.NM.DeviceType.OVS_BRIDGE,
-                        nmclient.NM.DeviceType.OVS_PORT,
-                        nmclient.NM.DeviceType.OVS_INTERFACE,
-                    ):
-                        remove_devs_actions[nmdev].append(device.delete_device)
+                if (
+                    nmdev.is_software()
+                    and nmdev.get_device_type() != nmclient.NM.DeviceType.VETH
+                ):
+                    remove_devs_actions[nmdev].append(device.delete_device)
             else:
                 raise NmstateValueError(
                     'Invalid state {} for interface {}'.format(
--- a/libnmstate/nm/device.py
+++ b/libnmstate/nm/device.py
@@ -270,6 +270,12 @@ def _delete_device_callback(src_object, result, user_data):
         if e.matches(
             nmclient.Gio.DBusError.quark(),
             nmclient.Gio.DBusError.UNKNOWN_METHOD,
+        ) or (
+            e.matches(
+                nmclient.NM.DeviceError.quark(),
+                nmclient.NM.DeviceError.NOTSOFTWARE,
+            )
+            and nmdev.is_software()
         ):
             logging.debug(
                 'Device %s has been already deleted: error=%s',
--- a/tests/integration/interface_common_test.py
+++ b/tests/integration/interface_common_test.py
@@ -16,11 +16,53 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 #
+from contextlib import contextmanager
+import time
+
+import pytest
 
 import libnmstate
+from libnmstate.error import NmstateVerificationError
 
+from .testlib import assertlib
+from .testlib import cmd as libcmd
 from .testlib import statelib
 from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceIPv4
+from libnmstate.schema import InterfaceIPv6
+from libnmstate.schema import InterfaceState
+
+DUMMY_INTERFACE = 'dummy_test'
+
+
+@pytest.fixture(scope='function')
+def ip_link_dummy():
+    libcmd.exec_cmd(['ip', 'link', 'add', DUMMY_INTERFACE, 'type', 'dummy'])
+    try:
+        yield
+    finally:
+        libcmd.exec_cmd(['ip', 'link', 'del', DUMMY_INTERFACE])
+
+
+@contextmanager
+def dummy_interface(name):
+    dummy_desired_state = {
+        Interface.KEY: [
+            {
+                Interface.NAME: name,
+                Interface.STATE: InterfaceState.UP,
+                Interface.IPV4: {InterfaceIPv4.ENABLED: False},
+                Interface.IPV6: {InterfaceIPv6.ENABLED: False},
+            }
+        ]
+    }
+    libnmstate.apply(dummy_desired_state)
+    try:
+        yield dummy_desired_state
+    finally:
+        dummy_state = dummy_desired_state[Interface.KEY][0]
+        dummy_state[Interface.STATE] = InterfaceState.ABSENT
+        libnmstate.apply(dummy_desired_state)
 
 
 def test_iface_description_removal(eth1_up):
@@ -34,3 +76,25 @@ def test_iface_description_removal(eth1_up):
     libnmstate.apply(desired_state)
     current_state = statelib.show_only(('eth1',))
     assert Interface.DESCRIPTION not in current_state[Interface.KEY][0]
+
+
+def test_take_over_virtual_interface_then_remove(ip_link_dummy):
+    with dummy_interface(DUMMY_INTERFACE) as dummy_desired_state:
+        assertlib.assert_state_match(dummy_desired_state)
+
+    current_state = statelib.show_only((DUMMY_INTERFACE,))
+    assert len(current_state[Interface.KEY]) == 0
+
+
+def test_take_over_virtual_interface_and_rollback(ip_link_dummy):
+    with dummy_interface(DUMMY_INTERFACE) as dummy_desired_state:
+        assertlib.assert_state_match(dummy_desired_state)
+
+        dummy_desired_state[Interface.KEY][0]['invalid_key'] = 'foo'
+        with pytest.raises(NmstateVerificationError):
+            libnmstate.apply(dummy_desired_state)
+
+        time.sleep(5)
+
+        current_state = statelib.show_only((DUMMY_INTERFACE,))
+        assert len(current_state[Interface.KEY]) == 1


libnmstate/nm/applier.py
```diff
@@ -175,11 +175,7 @@ def set_ifaces_admin_state(ifaces_desired_state, con_profiles=()):
                         device.deactivate,
                         device.delete,
                     ]
-                    if nmdev.get_device_type() in (
-                        nmclient.NM.DeviceType.OVS_BRIDGE,
-                        nmclient.NM.DeviceType.OVS_PORT,
-                        nmclient.NM.DeviceType.OVS_INTERFACE,
-                    ):
+                    if nmdev.is_software():
```
tyll: This seems to be true for Veth-Interfaces as well but removing veths is not appropriate in my opinion.

EdDev: It is appropriate, but it does not work well with the test env.
We should support veth interfaces and with it the ability to remove them.

EdDev: @almusil suggested to do this only for `ABSENT` and not `DOWN`.

almusil: Done

cathay4t:
Please includes the reasons of:
 * Why veth should be treated differently.
 * Why this PR needed. https://github.com/nmstate/nmstate/pull/593#pullrequestreview-319828202

libnmstate/nm/device.py
```diff
@@ -277,6 +277,19 @@ def _delete_device_callback(src_object, result, user_data):
...
             mainloop.execute_next_action()
+        elif (
+            e.matches(
+                nmclient.NM.DeviceError.quark(),
+                nmclient.NM.DeviceError.NOTSOFTWARE,
+            )
+            and nmdev.is_software()
+        ):
+            logging.debug(
+                'Device %s is software ignoring race condition: error=%s',
```
tyll: What exactly is the race here? Can the interface already be removed by NM before the command reaches NM?

almusil: This is probably the case for interfaces that were created by NM in the first place.

tyll: So if this is the race, please explain it in the commit message and the error message. Also to be sure, there should be a check if the device is already deleted instead of assuming that any error message means that it is deleted.

tyll:
This new behavior contradicts this test: https://github.com/nmstate/nmstate/blob/master/tests/integration/iface_admin_state_test.py#L49

libnmstate/nm/applier.py
```diff
@@ -33,7 +33,6 @@
 from . import device
```
tyll: There is a typo in the commit message; it says `virutal` but should say `virtual`.

libnmstate/nm/applier.py
```diff
@@ -170,17 +169,17 @@ def set_ifaces_admin_state(ifaces_desired_state, con_profiles=()):
...
                         device.deactivate,
                         device.delete,
                     ]
-                    if nmdev.get_device_type() in (
-                        nmclient.NM.DeviceType.OVS_BRIDGE,
-                        nmclient.NM.DeviceType.OVS_PORT,
-                        nmclient.NM.DeviceType.OVS_INTERFACE,
-                    ):
-                        remove_devs_actions[nmdev].append(device.delete_device)
+                if (
```
tyll: It would be good to have a test that shows that this is working for example by creating an interface with `ip link`, configuring it with Nmstate and then removing the interface with Nmstate.

libnmstate/nm/device.py
tyll: The handling of the race could be tested by trying to remove the same interface twice.

almusil: I don't know where to find out if the interface was already removed. Can you point me the right direction?

tyll: actually I just saw that a few lines above your code there is a check that is supposed to check if a device was already deleted (it does not look like it explicitly checks this, though).

@thom311  can you tell us how to properly check in the callback of a `NM.Device.delete_async()` call whether the device is deleted in case the call fails? I see that there is a a `is_real()` check: https://lazka.github.io/pgi-docs/#NM-1.0/classes/Device.html#NM.Device.is_real - can it be called in this case?

thom311: > whether the device is deleted in case the call fails?

if the call fails, you don't know whether the deletion succeeded. That depends on why it failed...

Overall, it should be that

 - if a device exists and is-real, then it should be tracked by `NMClient.get_devices()` property.
 - if the device exists and is either real or not real, it should be tracked by `NMClient.get_all_devices()`.
 - after `delete_async()` completes with success, the device should no longer be real and be deleted from `NMClient.get_devices()`.

However, libnm itself does not enforce such consistency. In many aspects, it exposes what is on D-Bus, so theoretically `NMClient.get_devices()` could also track devices that are not real (though, of course, that probably would indicate a bug). You don't have to assume that the consistency is violated, because you have to trust NMClient (and NetworkManager) to not expose you garbage information (otherwise, what would you do anyway to know the truth?).

Note that NMClient also has signals "any-device-added", "any-device-removed" (for `NMClient.get_all_devices()`, and "device-added", "device-removed" (for `NMClient.get_devices()`). And of course, there are also GObject properties "devices" and "all-devices", so you could also subscribe to "notify::devices" and "notify::all-devices" signals.

In general, I think you should not care about unrealized devices (!is_real). Just ignore `NMClient.get_all_devices()`.

The best way to know when a device is gone (or unrealized) is

- listen to `NMClient` "device-removed" signal.
- look for the device in `NMClient.get_devices()`.

tyll: > > whether the device is deleted in case the call fails?
> 
> if the call fails, you don't know whether the deletion succeeded. That depends on why it failed...

[...]

> The best way to know when a device is gone (or unrealized) is
> 
> * listen to `NMClient` "device-removed" signal.
> * look for the device in `NMClient.get_devices()`.

The problem is, that the device might be deleted by NM automatically instead of as a result of the async delete call. So if the call fails but the device is deleted, then 
the error can be ignored. Does NM tell that the call failed because it does not exist? If this is the case, Nmstate could just check for this. If not, then from the options you outlined, it seems that checking if the device is in `NMClient.get_devices()` is the way to go. Listening for signals at that point does not work because the deletion would already have happened. Also if the device is not deleted, this would take forever.

What I do not understand is, why `is_real()' cannot be used. Since it is the callback for a method of a device, we already have the device object, so `nmdevice in NMClient.get_devices()` should return the same result as `nmdevice.is_real()`. What am I missing?

thom311: > Listening for signals at that point does not work because the deletion would already have happened.

It works. You already have a `NMDevice` instance that you know is-real, and that you are about to call `device.delete_async()` on. If you register to `NMClient.device-removed` signal, and you encounter a delete signal for your device, you know it's gone. Of course, you can also check with `any([device is d for d in nmc.get_devices ()])` (which is a bit more expensive, but a failure of `delete_async()` is not the common case).

Theoretically, the error reason will also tell you that the device no longer existed, but that isn't useful here. You are not really interested in success of the `delete_async()` call. You are interested to get rid of the device. You first do something that hopefully will rid it (`delete_async()`) and afterwards you care whether it's still there or not. So, if the call fails with some other failure, but the device is actually gone, what did you gain by caring about the result?

> What I do not understand is, why `is_real()` cannot be used.

is_real() is a property of the NMDevice instance. When an instance gets removed from the cache (`NMClient.get_devices()`), it's not clear that all properties should be cleared. Would it be useful that the device's `ifname` gets set to NULL the moment the instance gets removed from the cache? I don't think so. Maybe we should add a property `nm_object_is_alive()`, which returns TRUE as long as the object is still in the cache. Then you could actually check `not device.is_alive() or not device.is_real()`...

almusil: I have to admit that I totally started to loosing track of what is going on.

libnmstate/nm/applier.py
```diff
@@ -170,17 +169,17 @@ def set_ifaces_admin_state(ifaces_desired_state, con_profiles=()):
...
+                    nmdev.is_software()
+                    and iface_desired_state[Interface.STATE]
+                    == InterfaceState.ABSENT
```
tyll: In todays standup @EdDev @cathay4t and me agreed to add an exception for veth interfaces here for now so that the test `test_removing_a_non_removable_iface` will not fail.

libnmstate/nm/applier.py
almusil: Done

libnmstate/nm/applier.py
almusil: Done

libnmstate/nm/applier.py
almusil: Done, but I don't like the fact that it is not consistent.

libnmstate/nm/device.py
```diff
@@ -277,6 +277,19 @@ def _delete_device_callback(src_object, result, user_data):
                 e,
             )
             mainloop.execute_next_action()
+        elif (
```
EdDev: Why did you placed a distinction between the previous failures and this one? Both results in ignoring the error, no matter if it is software or not.

tests/integration/interface_common_test.py
```diff
@@ -34,3 +39,32 @@ def test_iface_description_removal(eth1_up):
     libnmstate.apply(desired_state)
     current_state = statelib.show_only(('eth1',))
     assert Interface.DESCRIPTION not in current_state[Interface.KEY][0]
+
+
+def test_take_over_virtual_interface_then_remove():
```
tyll: Could you please add a second test that does the same except triggering a rollback instead of a successful apply to check that the device is restored by the rollback?

libnmstate/nm/applier.py
```diff
@@ -170,17 +170,18 @@ def set_ifaces_admin_state(ifaces_desired_state, con_profiles=()):
...
+                    and nmdev.get_device_type() != nmclient.NM.DeviceType.VETH
+                ):
+                    remove_devs_actions[nmdev].append(device.delete_device)
```
tyll: Since the change here is meant to work around certain edge cases, another approach could be to change the success callback for deactivate to remove the interface if it is still there and it should have been removed?

libnmstate/nm/applier.py
```diff
@@ -170,17 +170,18 @@ def set_ifaces_admin_state(ifaces_desired_state, con_profiles=()):
...
-                    if nmdev.get_device_type() in (
-                        nmclient.NM.DeviceType.OVS_BRIDGE,
-                        nmclient.NM.DeviceType.OVS_PORT,
-                        nmclient.NM.DeviceType.OVS_INTERFACE,
-                    ):
-                        remove_devs_actions[nmdev].append(device.delete_device)
+                if (
+                    nmdev.is_software()
+                    and iface_desired_state[Interface.STATE]
+                    == InterfaceState.ABSENT
```
tyll: The original code would remove OVS bridge, port and interface devices also if the state is down, not only when it is absent. Is this an intended change in the logic?

EdDev: Now that the veth has been explicitly checked, there is no need for this condition.

libnmstate/nm/applier.py
EdDev: Good idea. We should try this in a following PR.

tests/integration/interface_common_test.py
EdDev: We already cover these scenarios using bond and bridge rollback.

tyll: Do the bond and bridge rollback tests really test that an interface that was not created by NM is restored on failure?

EdDev: Nope, that is not tested. Good point, thanks.

tests/integration/interface_common_test.py
```diff
@@ -34,3 +39,32 @@ def test_iface_description_removal(eth1_up):
...
+            {
+                Interface.NAME: dummy_name,
+                Interface.STATE: InterfaceState.UP,
+                Interface.IPV4: {InterfaceIPv4.ENABLED: False},
+                Interface.IPV6: {InterfaceIPv6.ENABLED: False},
+            }
+        ]
+    }
+    libnmstate.apply(dummy_desired_state)
+
```
tyll: There seems to be trailing whitespace in this line according to my editor.

libnmstate/nm/device.py
almusil: The previous failure is completely different or am I missing something?

EdDev: They are different failure for the same purpose, therefore they seem to me to fit the same condition and not a new flow.

almusil: Done

libnmstate/nm/applier.py
almusil: Done

tests/integration/interface_common_test.py
almusil: Done, thanks

tests/integration/interface_common_test.py
almusil: Done, thanks

tests/integration/interface_common_test.py
```diff
@@ -16,11 +16,21 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 #
+import pytest
```
EdDev: nit: pytest is a 3rd party package, please place below time with a newline in between.

tests/integration/interface_common_test.py
```diff
@@ -34,3 +44,59 @@ def test_iface_description_removal(eth1_up):
...
+        Interface.KEY: [
+            {
+                Interface.NAME: DUMMY_INTERFACE,
+                Interface.STATE: InterfaceState.UP,
+                Interface.IPV4: {InterfaceIPv4.ENABLED: False},
+                Interface.IPV6: {InterfaceIPv6.ENABLED: False},
+            }
+        ]
+    }
+    libnmstate.apply(dummy_desired_state)
```
EdDev: This and the teardown part needs to be placed in a context manager with a try/finally clause, otherwise we may be left with leftovers from the test.

tests/integration/interface_common_test.py
```diff
@@ -34,3 +44,59 @@ def test_iface_description_removal(eth1_up):
     libnmstate.apply(desired_state)
     current_state = statelib.show_only(('eth1',))
     assert Interface.DESCRIPTION not in current_state[Interface.KEY][0]
+
+
+def test_take_over_virtual_interface_then_remove():
+    assert _add_ip_link_dummy(DUMMY_INTERFACE) == 0
```
EdDev: This needs a contextmanager, for the same reason as mentioned in the next comment below.
It actually fits a fixture as it is part of the test setup (and teardown) and not something that the test itself cares to check.

tests/integration/interface_common_test.py
almusil: Done

tests/integration/interface_common_test.py
almusil: Done, thanks

tests/integration/interface_common_test.py
almusil: Done

tests/integration/interface_common_test.py
```diff
@@ -34,3 +65,30 @@ def test_iface_description_removal(eth1_up):
...
     current_state = statelib.show_only(('eth1',))
     assert Interface.DESCRIPTION not in current_state[Interface.KEY][0]
+
+
+def test_take_over_virtual_interface_then_remove(ip_link_dummy):
+    dummy_desired_state = ip_link_dummy
+    assertlib.assert_state_match(dummy_desired_state)
+
+    dummy_desired_state[Interface.KEY][0][
+        Interface.STATE
```
EdDev: nit: Please use a temporary variable to make this more readable.

tests/integration/interface_common_test.py
```diff
@@ -16,11 +16,42 @@
...
 from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceIPv4
+from libnmstate.schema import InterfaceIPv6
+from libnmstate.schema import InterfaceState
+
+DUMMY_INTERFACE = 'dummy_test'
+
+
+@pytest.fixture(scope='function')
+def ip_link_dummy():
```
EdDev: - The `ip link` fits a fixture as it is a setup/teardown.
- The `apply` fits the tests in a context manager.

almusil: Done

tests/integration/interface_common_test.py
almusil: Done

EdDev:
Thank you @almusil !

PR #595, opened 2019-11-21
Examples: Add combined bond/linuxbridge/vlan example

tyll:
Add a real-life complex example to ensure it is working and to use it
for demos in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/nmstate/nmstate/595)

### DIFF
--- a/examples/bond_linuxbridge_vlan_absent.yml
+++ b/examples/bond_linuxbridge_vlan_absent.yml
@@ -1,0 +1,19 @@
+---
+interfaces:
+  - name: eth1
+    state: down
+  - name: eth2
+    type: ethernet
+    state: down
+  - name: bond0
+    type: bond
+    state: absent
+  - name: br0
+    type: linux-bridge
+    state: absent
+  - name: vlan29
+    type: vlan
+    state: absent
+  - name: br29
+    type: linux-bridge
+    state: absent
--- a/tests/integration/examples_test.py
+++ b/tests/integration/examples_test.py
@@ -64,6 +64,19 @@ def test_add_remove_linux_bridge(eth1_up):
     assertlib.assert_absent('linux-br0')
 
 
+def test_bond_linuxbridge_vlan(eth1_up, eth2_up):
+    with example_state(
+        'bond_linuxbridge_vlan_up.yml',
+        cleanup='bond_linuxbridge_vlan_absent.yml',
+    ) as desired_state:
+        assertlib.assert_state_match(desired_state)
+
+    assertlib.assert_absent('bond0')
+    assertlib.assert_absent('br0')
+    assertlib.assert_absent('br29')
+    assertlib.assert_absent('vlan29')
+
+
 def test_dns_edit(eth1_up):
     with example_state(
         'dns_edit_eth1.yml', cleanup='dns_remove.yml'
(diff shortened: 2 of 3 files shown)


examples/bond_linuxbridge_vlan_absent.yml
```diff
@@ -0,0 +1,20 @@
+---
+interfaces:
+  - name: eth1
+    type: ethernet
+    state: absent
```
EdDev: - There should be no need to specify the type.
- Ethernet types should use `down` as they cannot be removed.

examples/bond_linuxbridge_vlan_absent.yml
```diff
@@ -0,0 +1,20 @@
...
+  - name: eth2
+    type: ethernet
+    state: absent
+  - name: bond0
+    type: bond
+    state: absent
+  - name: bond0-br0
```
EdDev: It is common to start with the device itself and not with its slave.
I would even argue that embedding the port in the name is not appropriate with bridges and bonds.

examples/bond_linuxbridge_vlan_absent.yml
tyll: Changed this.

examples/bond_linuxbridge_vlan_absent.yml
tyll: The order makes it easier to follow the network diagram in the up state.
Not sure what you mean with `port in the name`. The bridge names correspond to the VLAN that they belong to (none, or VLAN 29). I do not see why this would not be appropriate. It makes it easier to comprehend IMHO.

EdDev: It is not easy to comprehend and is not common in places I saw similar setups.
Slaves tend to change so a master name cannot follow that.

tyll: Please propose an alternative naming.
This is about an example derived from a production example and for me it is easy to comprehend (I did not invent it myself). Also as an static example, it does not need to care about possible changes to the slaves because then it would be another example than showing how to create a bridge for vlan 29 on top of a bond for eth1 and eth2.

EdDev: I edited my last response with my suggestion. ‘br0-bond0’ I can understand.

tyll: I changed the bridges to `br0` and `brXY`. To me a top-bottom notation like `br0-bond0` is confusing. Hope this is a working compromise.

tests/integration/examples_test.py
```diff
@@ -64,6 +64,19 @@ def test_add_remove_linux_bridge(eth1_up):
...
 
+def test_bond_linuxbridge_vlan(eth1_up, eth2_up):
+    with example_state(
+        'bond_linuxbridge_vlan_up.yml',
+        cleanup='bond_linuxbridge_vlan_absent.yml',
+    ) as desired_state:
+        assertlib.assert_state_match(desired_state)
+
+    assertlib.assert_absent('bond0')
+    assertlib.assert_absent('bond0-br0')
```
EdDev: Bridge names need to be updated.

tyll: well spotted, changed it!

PR #596, opened 2019-11-22
nm.sriov: SRIOVSetting implementation

ffmancera:
This patch adds basic SR-IOV support in nmstate. It is possible to
retrieve and set the total number of VFs per PF.

### DIFF
--- a/libnmstate/error.py
+++ b/libnmstate/error.py
@@ -101,3 +101,11 @@ class NmstateInternalError(NmstateError):
     """
 
     pass
+
+
+class NmstateNotSupportedError(NmstateError):
+    """
+    A resource like a device does not support the requested feature.
+    """
+
+    pass
--- a/libnmstate/nm/applier.py
+++ b/libnmstate/nm/applier.py
@@ -413,7 +413,7 @@ def _build_connection_profile(iface_desired_state, base_con_profile=None):
     if vxlan_setting:
         settings.append(vxlan_setting)
 
-    sriov_setting = sriov.create_setting(iface_desired_state)
+    sriov_setting = sriov.create_setting(iface_desired_state, base_con_profile)
     if sriov_setting:
         settings.append(sriov_setting)
 
--- a/libnmstate/nm/sriov.py
+++ b/libnmstate/nm/sriov.py
@@ -17,16 +17,63 @@
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 #
 
-from libnmstate.error import NmstateNotImplementedError
+from libnmstate.error import NmstateNotSupportedError
+from libnmstate.nm import connection as nm_connection
+from libnmstate.nm import device
+from libnmstate.nm import nmclient
 from libnmstate.schema import Ethernet
+from libnmstate.schema import Interface
 
 
-def create_setting(iface_state):
-    setting = None
+def create_setting(iface_state, base_con_profile):
+    sriov_setting = None
+    ifname = iface_state[Interface.NAME]
     sriov_config = iface_state.get(Ethernet.CONFIG_SUBTREE, {}).get(
         Ethernet.SRIOV_SUBTREE
     )
     if sriov_config:
-        raise NmstateNotImplementedError
+        if not _has_sriov_capability(ifname):
+            raise NmstateNotSupportedError(
+                f"Interface '{ifname}' does not support SR-IOV"
+            )
 
-    return setting
+        sriov_setting = base_con_profile.get_setting_duplicate(
+            nmclient.NM.SETTING_SRIOV_SETTING_NAME
+        )
+        if not sriov_setting:
+            sriov_setting = nmclient.NM.SettingSriov.new()
+
+        sriov_setting.props.total_vfs = sriov_config[Ethernet.SRIOV.TOTAL_VFS]
+
+    return sriov_setting
+
+
+def _has_sriov_capability(ifname):
+    dev = device.get_device_by_name(ifname)
+    if nmclient.NM.DeviceCapabilities.SRIOV & dev.props.capabilities:
+        return True
+
+    return False
+
+
+def get_info(device):
+    """
+    Provide the current active SR-IOV live configuration for a device
+    """
+    info = {}
+
+    connection = nm_connection.ConnectionProfile()
+    connection.import_by_device(device)
+    if not connection.profile:
+        return info
+
+    sriov_setting = connection.profile.get_setting_by_name(
+        nmclient.NM.SETTING_SRIOV_SETTING_NAME
+    )
+
+    if sriov_setting:
+        info[Ethernet.SRIOV_SUBTREE] = {
+            Ethernet.SRIOV.TOTAL_VFS: sriov_setting.props.total_vfs
+        }
+
+    return info
--- a/tests/integration/sriov_test.py
+++ b/tests/integration/sriov_test.py
@@ -20,19 +20,49 @@
 import pytest
 
 import libnmstate
-from libnmstate.error import NmstateNotImplementedError
+from libnmstate.error import NmstateNotSupportedError
 from libnmstate.schema import Ethernet
 from libnmstate.schema import Interface
 
+from .testlib import assertlib
 
-SRIOV_CONFIG = {Ethernet.SRIOV_SUBTREE: {Ethernet.SRIOV.TOTAL_VFS: 2}}
+
+SRIOV_CONFIG = {Ethernet.SRIOV_SUBTREE: {Ethernet.SRIOV.TOTAL_VFS: 0}}
 
 
 @pytest.mark.xfail(
-    raises=NmstateNotImplementedError,
-    reason='SR-IOV is not supported yet',
-    strict=True,
+    raises=NmstateNotSupportedError,
+    reason='The device does not support SR-IOV.',
 )
-def test_sriov_not_implemented(eth1_up):
+def test_sriov_zero_vfs(sriov_interface):
+    assertlib.assert_state(sriov_interface)
+
+
+@pytest.mark.xfail(
+    raises=NmstateNotSupportedError,
+    reason='The device does not support SR-IOV.',
+)
+def test_sriov_increase_vfs(sriov_interface):
+    eth_config = sriov_interface[Interface.KEY][0][Ethernet.CONFIG_SUBTREE]
+    eth_config[Ethernet.SRIOV_SUBTREE][Ethernet.SRIOV.TOTAL_VFS] = 5
+    libnmstate.apply(sriov_interface)
+    assertlib.assert_state(sriov_interface)
+
+
+@pytest.mark.xfail(
+    raises=NmstateNotSupportedError,
+    reason='The device does not support SR-IOV.',
+)
+def test_sriov_decrease_vfs(sriov_interface):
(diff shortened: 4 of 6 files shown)


ffmancera:
This PR depends on #592. It won't pass CI because of that.

libnmstate/nm/sriov.py
```diff
@@ -0,0 +1,76 @@
...
+from libnmstate.nm import connection as nm_connection
+from libnmstate.nm import nmclient
+from libnmstate.schema import Ethernet
+from libnmstate.schema import SRIOV
+
+
+def create_setting(iface_state, base_con_profile):
+    ethernet_setting = iface_state.get(Ethernet.CONFIG_SUBTREE)
+    if not ethernet_setting:
+        return None
```
tyll: Is None correct here? Why not return `sriov_setting`? Then it could be simplified using `.get()` with `{}` as the second argument, see #592 .

libnmstate/nm/sriov.py
```diff
@@ -0,0 +1,76 @@
...
+    return sriov_setting
+
+
+def get_info(device):
+    """
+    Provide the current active SR-IOV values for a device
+    """
+    info = {}
+
+    connection = nm_connection.ConnectionProfile()
```
tyll: Can we get the SR-IOV settings from the device/interface (IOW the actual values used by the kernel, or is it only available currently from the configuration?)

Maybe @thom311 can help, it might need a RFE from NM.

tyll:
FYI: In theory we could netdevsim devices to test this in a CI: https://developers.redhat.com/blog/2018/10/22/introduction-to-linux-interfaces-for-virtual-networking/#netdevsim - in practice we would need to find a CI that actually provides the kernel module for this. Ubuntu on Travis does not seem to contain it for the current kernel version.

libnmstate/nm/sriov.py
ffmancera: AFAIK, currently libnm only allows us to get the settings from the configuration. If I am wrong, please let me know @thom311. Could be that a problem?

ffmancera:
> FYI: In theory we could netdevsim devices to test this in a CI: https://developers.redhat.com/blog/2018/10/22/introduction-to-linux-interfaces-for-virtual-networking/#netdevsim - in practice we would need to find a CI that actually provides the kernel module for this. Ubuntu on Travis does not seem to contain it for the current kernel version.

That seems fine. I will try to find a CI that allow us to test SR-IOV using netdevsim. Thank you very much!

libnmstate/nm/sriov.py
thom311: no, you cannot... and RFE would be good.

libnmstate/nm/sriov.py
```diff
@@ -0,0 +1,76 @@
...
+#
+
+from libnmstate.nm import connection as nm_connection
+from libnmstate.nm import nmclient
+from libnmstate.schema import Ethernet
+from libnmstate.schema import SRIOV
+
+
+def create_setting(iface_state, base_con_profile):
+    ethernet_setting = iface_state.get(Ethernet.CONFIG_SUBTREE)
```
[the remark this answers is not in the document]
ffmancera: Yes it was confusing. Fixed. Thanks!

libnmstate/nm/sriov.py

ffmancera:
I am preparing the CI tests and searching for a good image that allow us to use netdevsim.

libnmstate/nm/sriov.py
tyll: @ffmancera please file an RFE in Bugzilla for NM, another Bug for Nmstate that depends on the NM RFE. Please also mention it in the code.

libnmstate/nm/sriov.py
```diff
@@ -0,0 +1,73 @@
...
+    if not sriov_setting:
+        sriov_setting = nmclient.NM.SettingSriov.new()
+
+    if sriov_config.get(Ethernet.SRIOV.TOTAL_VFS):
+        sriov_setting.props.total_vfs = sriov_config[Ethernet.SRIOV.TOTAL_VFS]
+
+    return sriov_setting
+
+
+def _get_current_sriov_setting(base_con_profile):
```
tyll: This could probably be a generic function that takes the `base_con_profile` parameter and the constant for the setting. Other code in the nm directory could then re-use this.

ffmancera: This should be another PR right?

libnmstate/nm/sriov.py
ffmancera: Will do. Thanks!

libnmstate/nm/sriov.py
tyll: yes, at least making the other code use the same function. You can already introduce it here and use it for this code IMHO. It might also be useful for the team implementation.

tyll:
> I am preparing the CI tests and searching for a good image that allow us to use netdevsim.

We actually need a host that supports this not just a container image (I guess with the right host kernel, the current images can already support this).

libnmstate/nm/sriov.py
ffmancera: Done thanks!

libnmstate/nm/sriov.py

ffmancera:
This patch have not been tested. I am preparing a machine to test it.

ffmancera:
This patch have been tested with HW that supports SR-IOV. Thanks!

tests/integration/sriov_test.py
```diff
@@ -29,10 +29,10 @@
 
 
 @pytest.mark.xfail(
-    raises=NmstateNotImplementedError,
-    reason='SR-IOV is not supported yet',
+    raises=NotSupportedByDevice,
+    reason='The device does not support SR-IOV.',
     strict=True,
```
tyll: IMHO it should not test eth1 but a dummy interface for the not-supported error and there should be an test that is allowed to fail (not with strict) so it could be tested on bare-metal where eth1 supports SR-IOV.

libnmstate/error.py
```diff
@@ -101,3 +101,11 @@ class NmstateInternalError(NmstateError):
     """
 
     pass
+
+
+class NotSupportedByDevice(Exception):
```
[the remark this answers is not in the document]
ffmancera: I am fine with adding the "Nmstate" prefix but I would keep the "ByDevice" suffix because Nmstate is supporting the features but the device not. What do you think?

libnmstate/nm/sriov.py
```diff
@@ -17,16 +17,62 @@
...
+    )
+    if not sriov_setting:
+        sriov_setting = nmclient.NM.SettingSriov.new()
+
+    sriov_setting.props.total_vfs = sriov_config[Ethernet.SRIOV.TOTAL_VFS]
+
+    return sriov_setting
+
+
+def _has_sriov_capability(ifname):
```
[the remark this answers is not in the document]
ffmancera: Sure, I will add the card. Thanks!

libnmstate/nm/settings.py
```diff
@@ -0,0 +1,27 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+
+
+def get_current_setting(setting_name, base_con_profile):
```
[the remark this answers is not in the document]
ffmancera: Seems fine. Will do.

tests/integration/sriov_test.py
ffmancera: We cannot use dummy interface to arise this error. We need it to be ethernet type. I like the idea of removing the "strict". What do you think? Thanks!

libnmstate/error.py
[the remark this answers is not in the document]
ffmancera: I meant NmstateNotSupportedByDeviceError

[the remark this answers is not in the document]
tyll: To me, `NmstateNotSupportedError` does not imply to me  why it is not supported but leaves this open. When actually raising the exception, it would make sense to add the full reason as an argument, for example:

```
raise NmstateNotSupportedError(f"Interface '{interface}' does not support SR-IOV")
```

Also, if Nmstate would not be supporting a feature in general, it would probably be an `NmstateNotImplementedError`.

ffmancera:
Changes tested in HW with SR-IOV support.

libnmstate/nm/applier.py
```diff
@@ -413,7 +413,7 @@ def _build_connection_profile(iface_desired_state, base_con_profile=None):
     if vxlan_setting:
         settings.append(vxlan_setting)
 
-    sriov_setting = sriov.create_setting(iface_desired_state)
+    sriov_setting = sriov.create_setting(iface_desired_state, base_profile)
```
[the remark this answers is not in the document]
ffmancera: Done! Thanks

libnmstate/error.py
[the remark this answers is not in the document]
ffmancera: That seems fine, thanks!

tests/integration/sriov_test.py
tyll: Currently the test is called `...not_supported`. To keep this name, it should ideally always fail, regardless of the hardware. An alternative to the dummy would be a veth interface or you could mock the NM check to always tell that it is not supported. Not sure if this is too much effort, though.

Whats missing then are tests for the actual features.

ffmancera: Missing tests for the actual features added.

libnmstate/nm/settings.py
```diff
@@ -0,0 +1,27 @@
+#
```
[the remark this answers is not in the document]
ffmancera: Ooops, thank for point it out :-)

libnmstate/error.py
```diff
@@ -101,3 +101,11 @@ class NmstateInternalError(NmstateError):
     """
 
     pass
+
+
+class NmstateNotSupportedError(NmstateError):
+    """
+    The device does not support the required functionality.
```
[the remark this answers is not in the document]
ffmancera: That seems fine, thanks!

libnmstate/nm/sriov.py
```diff
@@ -17,16 +17,64 @@
...
-    setting = None
+def create_setting(iface_state, base_con_profile):
+    ifname = iface_state[Interface.NAME]
     sriov_config = iface_state.get(Ethernet.CONFIG_SUBTREE, {}).get(
         Ethernet.SRIOV_SUBTREE
     )
-    if sriov_config:
-        raise NmstateNotImplementedError
+    if not sriov_config:
+        return sriov_config
```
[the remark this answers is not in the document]

libnmstate/nm/sriov.py
```diff
@@ -17,16 +17,63 @@
...
+    dev = device.get_device_by_name(ifname)
+    if nmclient.NM.DeviceCapabilities.SRIOV & dev.props.capabilities:
+        return True
+
+    return False
+
+
+def get_info(device):
+    """
+    Provide the current active SR-IOV values for a device
```
tyll: The docstring is a little bit misleading since it is reporting the configuration and not the values from the device. On the other hand, we assume that the live configuration is the actual devices settings. What do you think?

ffmancera: I agree. Done, thanks!

PR #597, opened 2019-11-22
linux bridge, port vlan: raise not implemented error

maiqueb:
When the port vlan feature is used, raise NmstateNotImplementedError.

### DIFF
--- a/libnmstate/nm/bridge.py
+++ b/libnmstate/nm/bridge.py
@@ -17,6 +17,7 @@
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 #
 
+from libnmstate.error import NmstateNotImplementedError
 from libnmstate.nm import connection
 from libnmstate.nm import nmclient
 from libnmstate.schema import LinuxBridge as LB
@@ -88,6 +89,8 @@ def create_port_setting(options, base_con_profile):
             port_setting.props.hairpin_mode = val
         elif key == LB.PORT_STP_PATH_COST:
             port_setting.props.path_cost = val
+        elif key == LB.Port.VLAN_SUBTREE:
+            raise NmstateNotImplementedError
 
     return port_setting
 
--- a/tests/integration/linux_bridge_test.py
+++ b/tests/integration/linux_bridge_test.py
@@ -24,6 +24,7 @@ import pytest
 import yaml
 
 import libnmstate
+from libnmstate.error import NmstateNotImplementedError
 from libnmstate.error import NmstateVerificationError
 from libnmstate.schema import Interface
 from libnmstate.schema import InterfaceIPv4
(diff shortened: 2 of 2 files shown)


tests/integration/linux_bridge_test.py
```diff
@@ -375,16 +376,35 @@ def test_activate_empty_bridge_does_not_blocked_by_dhcp():
...
+    bridge_name = TEST_BRIDGE0
+    port_name = port0_up[Interface.KEY][0][Interface.NAME]
+    port_vlan_config = {
+        LinuxBridge.Port.VLAN_SUBTREE: {
+            LinuxBridge.Port.Vlan.TYPE: LinuxBridge.Port.Vlan.ACCESS_TYPE,
+            LinuxBridge.Port.Vlan.TAG: 101,
+        }
+    }
+
+    bridge_state = _create_bridge_subtree_config(
```
tyll: This should probably be named `port_state`.

EdDev: It looks correct to me. The returned state is of the bridge subtree which includes the ports and other parameters. 
The new argument confused things.

tests/integration/linux_bridge_test.py
```diff
@@ -375,16 +376,35 @@ def test_activate_empty_bridge_does_not_blocked_by_dhcp():
...
     add_port_to_bridge(bridge_state, ifname, port_state)
 
 
-def _create_bridge_subtree_config(port_names):
+def _create_bridge_subtree_config(port_names, port_extra_state=None):
     bridge_state = yaml.load(BRIDGE_OPTIONS_YAML, Loader=yaml.SafeLoader)
 
     for port in port_names:
         port_state = yaml.load(BRIDGE_PORT_YAML, Loader=yaml.SafeLoader)
+        port_state.update(port_extra_state or {})
```
EdDev: Does not seem to fit here. The input of ports are plural but you add a single “template” for the extras which does not makes sense much IMO.
I suggest until we make some heavier refactoring here, to just add the needed properties after/outside of this function.

maiqueb: Fair enough .

I'll be adding it manually on the relevant test.

tests/integration/linux_bridge_test.py
```diff
@@ -375,6 +376,23 @@ def test_activate_empty_bridge_does_not_blocked_by_dhcp():
...
+    port_vlan_config = {
+        LinuxBridge.Port.VLAN_SUBTREE: {
+            LinuxBridge.Port.Vlan.TYPE: LinuxBridge.Port.Vlan.ACCESS_TYPE,
+            LinuxBridge.Port.Vlan.TAG: 101,
+        }
+    }
+
+    bridge_state = _create_bridge_subtree_config((port_name,))
+    bridge_state[LinuxBridge.PORT_SUBTREE][0].append(port_vlan_config)
+    with pytest.raises(NmstateNotImplementedError):
```
tyll: The only minor thing I can think of is adding a reference to track the actual implementation, so maybe a link to a Jira card.

libnmstate/nm/bridge.py
```diff
@@ -17,6 +17,8 @@
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 #
 
+
```
tyll: Do we need this extra newline?

tests/integration/linux_bridge_test.py
EdDev: Hmm... you can move this to an xfail decorator and add there this info.

PR #599, opened 2019-11-24
Remove the __import__ statements and six

EdDev:


tyll:
Why do you want to do this change now? In the sprint planning we agreed to not take this change into the current sprint.

tyll:
There is an unexpected error in the CI for NM 1.20/CentOS8: `2019-11-25 06:33:12 nmclient.py:176 ERROR NM main-loop aborted: Connection activation failed on device eth2: error=nm-manager-error-quark: The device 'eth2' has no connections available for activation. (2)`

EdDev:
> Why do you want to do this change now? In the sprint planning we agreed to not take this change into the current sprint.

It annoyed me when working on other tasks. It is a technicality which did not require any brain power.
Unfortunately, it looks like it detected a problem with the `managed` operation.

PR #600, opened 2019-11-26
test ovs-bridge + ovs-interface with same name

qinqon:

### DIFF
--- a/tests/integration/ovs_test.py
+++ b/tests/integration/ovs_test.py
@@ -26,6 +26,7 @@ from libnmstate.schema import InterfaceState
 from libnmstate.schema import InterfaceType
 from libnmstate.schema import OVSBridge
 from libnmstate.error import NmstateLibnmError
+from libnmstate.error import NmstateValueError
 
 from .testlib import assertlib
 from .testlib import nmlib
@@ -122,6 +123,23 @@ def test_create_and_remove_ovs_bridge_with_internal_port_static_ip_and_mac():
 
     assertlib.assert_absent(BRIDGE1)
     assertlib.assert_absent(PORT1)
+
+
+@pytest.mark.xfail(
+    raises=NmstateValueError,
+    reason="https://nmstate.atlassian.net/browse/NMSTATE-286",
+    strict=True,
+)
+def test_create_and_remove_ovs_bridge_with_internal_port_same_name():
+    bridge = Bridge(BRIDGE1)
+    bridge.add_internal_port(
+        BRIDGE1, ipv4_state={InterfaceIPv4.ENABLED: False}
+    )
+
+    with bridge.create() as state:
+        assertlib.assert_state_match(state)
+
+    assertlib.assert_absent(BRIDGE1)
 
 
 def test_vlan_as_ovs_bridge_slave(vlan_on_eth1):


tests/integration/ovs_test.py
```diff
@@ -103,3 +103,17 @@ def test_create_and_remove_ovs_bridge_with_internal_port_and_static_ip():
 
     assertlib.assert_absent(BRIDGE1)
     assertlib.assert_absent(PORT1)
+
+
+@pytest.mark.xfail(
+    raises=(NmstateLibnmError, AssertionError),
+    reason='https://bugzilla.redhat.com/1724901',
```
tyll: This reason does not match this particular case. The problem here is that the interface name is a primary key for Nmstate/needs to be unique. To handle this, Nmstate needs to use a tuple like (name, type) as the primary key or do some special handling for OVS.

tests/integration/ovs_test.py
```diff
@@ -103,3 +103,17 @@ def test_create_and_remove_ovs_bridge_with_internal_port_and_static_ip():
 
     assertlib.assert_absent(BRIDGE1)
     assertlib.assert_absent(PORT1)
+
+
+@pytest.mark.xfail(
+    raises=(NmstateLibnmError, AssertionError),
```
tyll: Does it really raise both exceptions depending on the environment? Since it is not supported at all I would expect only one exception.

libnmstate/validator.py
```diff
@@ -48,6 +48,10 @@ class NmstateRouteWithNoIPInterfaceError(NmstateValueError):
     pass
 
 
+class NmstateDuplicateInterfaceNameError(NmstateValueError):
```
tyll: IMHO this is a very specific error that does not need its own exception. The explanation could be added to the exception message.

libnmstate/validator.py
```diff
@@ -109,6 +113,15 @@ def validate_link_aggregation_state(desired_state, current_state):
                 specified_slaves |= slaves
 
 
+def validate_interface_unique_name(state):
```
tyll: This should be `validate_unique_interface_name`.

libnmstate/validator.py
tyll: Also public exceptions belong in errors.py.

libnmstate/netapplier.py
```diff
@@ -59,6 +59,7 @@ def apply(desired_state, verify_change=True, commit=True, rollback_timeout=60):
     desired_state = copy.deepcopy(desired_state)
     validator.validate(desired_state)
     validator.validate_capabilities(desired_state, netinfo.capabilities())
+    validator.validate_interface_unique_name(desired_state)
```
tyll: IMHO this commit contradicts the other one, since this shows that multiple interfaces with the same name are not supported while the other one indicates that they will be supported eventually.

libnmstate/validator.py
EdDev: Done

libnmstate/validator.py
EdDev: The error name explains what it is, which is better than text.
It inherits from a public error: `NmstateValueError`. Therefore, the caller will catch it if it specify it.

This pattern is already in use and is very helpful to detect the source of the problem without overloading the public error list. If someone will ask, we can promote it to be public easily.

libnmstate/netapplier.py
EdDev: It will help if you suggest an alternative.
The reason it is not in one commit is to preserve the original contribution.

EdDev:
> Please clarify the intentions and fix the exception.

Does this relate to an inline comment or something else?

libnmstate/validator.py
tyll: If something catches the value error and then presents it to the user, it might look like:

`Bad values in desired state: foo0, foo0`

or the user needs to specifically catch for `NmstateDuplicateInterfaceNameError` to change this into a message for the user and for this it becomes part of the API.

EdDev: If you are trying to claim that an exception text is better than a exception type, we are in a disagreement here.
The caller can print the error name + text easily.

If we are interested in exposing these exception classes publicly, it is an easy step and it will not touch the production code. If we need to go the other way around it is harder.

libnmstate/netapplier.py
tyll: Not sure, what you would like me to suggest as an alternative. The easiest one would be not to support multiple interfaces with the same name and drop the first commit. If multiple interfaces are going to be supported, it should probably raise an `NmstateNotImplementedError` for the cases it is going to be supported.

libnmstate/validator.py
tyll: The internal class name does not belong in an error message meant for a user, this is something for developers. Also it would not be obvious for a developer using libnmstate who adds code to except a ValueError that they need to print the class name because it contains the actual information for the user instead of the exception message.

Except for the route errors that you added, all other places in Nmstate specify the problem for a value error in the message, for example:
https://github.com/nmstate/nmstate/blob/master/libnmstate/nm/ovs.py#L69

This should happen in a uniform way. And IMHO the the exception type is meant for the code to allow to run a different code path in case of an exception (for example a FileNotFoundError might trigger a different code path than a PermissionDeniedError) it does not seem to be a good idea to introduce new exceptions.

tyll:
> > Please clarify the intentions and fix the exception.
> 
> Does this relate to an inline comment or something else?

It relates to the inline comments.

libnmstate/validator.py
EdDev: > The internal class name does not belong in an error message meant for a user, this is something for developers. Also it would not be obvious for a developer using libnmstate who adds code to except a ValueError that they need to print the class name because it contains the actual information for the user instead of the exception message.

This is highly opinionated.
Current change uses an existing pattern.
If you want to challenge this pattern, please suggest it though its own PR.

> This should happen in a uniform way. And IMHO the the exception type is meant for the code to allow to run a different code path in case of an exception (for example a FileNotFoundError might trigger a different code path than a PermissionDeniedError) it does not seem to be a good idea to introduce new exceptions.

Again, this is opinionated. There is nothing wrong to raise exceptions that inherit from an API one. The applications that use the lib may do it as well.
If the client of the lib wants to re-use the error text, they need to parse it anyway. Taking the info from the text-only, from the class+text (by serializing the exception using `str` or `logging.except`) is up to the user.

I prefer being specific with a type while you suggest using a more generic type and add text.
In the unit test I could use this exception for the tests and still do not have to expose it through the formal API.

Will adding descriptive text to the specific exception be a good compromise for you?

libnmstate/netapplier.py
EdDev: The original change was contributed to expose the problem through a test. The test fails in an ugly manner (not at the validation but at the verification, this was explained in the solution commit). It was marked as xfail because we wanted to fix it and raise a proper error and earlier with enough info (this was included in the fix commit message was well).

The second commit fixes and removes the xfail with a specific exception at the validation failure.

I want to keep the original contribution intact with just fixes.

EdDev:
The validation of the iface name duplication is extracted to separate PR.
The xfail test that points out that we cannot have an ovs-bridge with the same name as the ovs-interface is left and adjusted to expect the validation failure.

Having the ovs bridge with the same name as one of its internal interfaces is a valid OVS configuration that some users are actually using, therefore, it makes sense to resolve this issue one way or the other. The test just reflects that this is not possible in the current implementation.

EdDev:
This PR now depends on #771

PR #606, opened 2019-11-27
Fix vxlan with non remote

qinqon:


tests/integration/vxlan_test.py
```diff
@@ -22,16 +22,19 @@
 import pytest
 
 import libnmstate
-
```
EdDev: Not looking ;)

tests/integration/vxlan_test.py
```diff
@@ -22,16 +22,19 @@
...
 from libnmstate.error import NmstateVerificationError
 from libnmstate.schema import Interface
 
 from .testlib import assertlib
 from .testlib.bondlib import bond_interface
+from .testlib.cmd import exec_cmd
+from .testlib.cmd import RC_SUCCESS
+from .testlib.cmd import format_exec_cmd_result
```
EdDev: Please sort it. You could also use `from .testlib import cmd`, it will have a nice short context. Up to you.

tests/integration/vxlan_test.py
```diff
@@ -22,16 +22,19 @@
...
 from .testlib.vxlan import VxlanState
 from .testlib.vxlan import vxlan_interfaces
 from .testlib.vxlan import vxlans_up
 from .testlib.vxlan import vxlans_down
+from .testlib.vxlan import vxlans_absent
```
EdDev: We want to keep it sorted. It is already unsorted :( , but we can place it in a better place above already.

tests/integration/vxlan_test.py
```diff
@@ -108,3 +111,21 @@ def test_add_new_bond_iface_with_vxlan(eth1_up):
...
     assertlib.assert_absent(vxlan.name)
     assertlib.assert_absent(bond_name)
+
+
+def test_show_vxlan_with_no_remote(eth1_up):
+    eth_name = eth1_up[Interface.KEY][0][Interface.NAME]
+    vxlan = VxlanState(id=VXLAN1_ID, base_if=eth_name, remote="")
+    add_vxlan_cmd = (
+        f"ip link add {vxlan.name} type vxlan id {vxlan.id}"
+        f" dstport {vxlan.destination_port} dev {eth_name}".split()
```
EdDev: Single quotes please.

libnmstate/nm/vxlan.py
```diff
@@ -59,11 +59,14 @@ def get_info(device):
...
         if device.props.parent:
             base_iface = device.props.parent.get_iface()
+        remote = device.props.group
+        if not remote:
+            remote = ''
         return {
             VXLAN.CONFIG_SUBTREE: {
                 VXLAN.ID: device.props.id,
                 VXLAN.BASE_IFACE: base_iface,
-                VXLAN.REMOTE: device.props.group,
```
maiqueb: don't want to be picky, but why not `VXLAN.REMOTE: device.props.group or '',` . That would allow you to remove all the above.

I think it's more concise. Anyway, not important.

EdDev: My fault @maiqueb , won't block it now that I messed up.

maiqueb: I won't argue with someone that can use my own weight against me :laughing:

PR #612, opened 2019-11-29
Schema constants hierarchy stp

tyll:
This is a follow-up on #608

---
This change is [Reviewable](https://reviewable.io/reviews/nmstate/nmstate/612)

### DIFF
--- a/libnmstate/schema.py
+++ b/libnmstate/schema.py
@@ -167,6 +167,11 @@ DEPRECATED_CONSTANTS = {
     'LinuxBridge.PORT_STP_HAIRPIN_MODE': 'LinuxBridge.Port.STP_HAIRPIN_MODE',
     'LinuxBridge.PORT_STP_PATH_COST': 'LinuxBridge.Port.STP_PATH_COST',
     'LinuxBridge.PORT_STP_PRIORITY': 'LinuxBridge.Port.STP_PRIORITY',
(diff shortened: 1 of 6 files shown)


tyll:
Also added the commits for OVSBridge.Port constants. They do not seem to be used in  libnmstate, therefore no changes were needed there.

libnmstate/schema.py
```diff
@@ -297,3 +297,8 @@ class OVSBridge(object):
     PORT_NAME = 'name'
     PORT_VLAN_MODE = 'vlan-mode'
     PORT_ACCESS_TAG = 'access-tag'
+
+    class Port(object):
```
EdDev: As I remember we discussed that the should use the linux-bridge vlan abstraction which we did a few weeks ago. See line 229. It needs to come with the schema change.
Lets do it in a separate PR.

tyll: I do not recall the discussion and I am not sure what you mean. If you like, you can adjust this PR tomorrow and I can approve it. If the change for LinuxBridge.STP is ok, I can adjust this PR to only contain the respective commits.

EdDev: Yes, the STP looks good, please extract the rest so we can take it. I mean, from this commit forward.
I will try to take over the OVS port/vlan change.

tyll: removed the OVSBridge commits

PR #613, opened 2019-11-30
schema, ovs: Change bridge VLAN schema

EdDev:
Sync the OVS bridge port vlan schema with the one of the Linux bridge.
As the port vlan has never been implemented, this change is backward
safe and requires no production code changes.

### DIFF
--- a/libnmstate/schema.py
+++ b/libnmstate/schema.py
@@ -295,5 +295,22 @@ class OVSBridge(object):
 
     PORT_SUBTREE = 'port'
     PORT_NAME = 'name'
-    PORT_VLAN_MODE = 'vlan-mode'
-    PORT_ACCESS_TAG = 'access-tag'
+
+    class Port:
+        VLAN_SUBTREE = 'vlan'
+
+        class Vlan:
+            TRUNK_TAGS = 'trunk-tags'
+            TAG = 'tag'
+            ENABLE_NATIVE = 'enable-native'
+            MODE = 'mode'
+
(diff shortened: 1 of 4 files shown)


libnmstate/schema.py
```diff
@@ -288,5 +288,20 @@ class OVSBridge(object):
 
     PORT_SUBTREE = 'port'
     PORT_NAME = 'name'
-    PORT_VLAN_MODE = 'vlan-mode'
-    PORT_ACCESS_TAG = 'access-tag'
+
+    class Port:
+        VLAN_SUBTREE = 'vlan'
```
phoracek: Why is the NAME not in here?

libnmstate/schema.py
```diff
@@ -288,5 +288,20 @@ class OVSBridge(object):
...
+
+        class Vlan:
+            TRUNK_TAGS = 'trunk-tags'
+            TAG = 'tag'
+            ENABLE_NATIVE = 'enable-native'
+            TYPE = 'type'
```
phoracek: You set it to MODE in the other PR. Why the difference?

libnmstate/schema.py
EdDev: This is handled by the constants changes PR/s, not this one.
I will move it in a separate one.

libnmstate/schema.py
EdDev: Yea, I missed this and many other changes.
Fixed now.

PR #614, opened 2019-11-30
schema, bridge: Change vlan.type to vlan.mode

EdDev:
Most references to the vlan access/trunk term are using the
name `vlan.mode`.
Therefore, `type` is changed in this context to `mode`.

As the implementation is not yet in, this change is backward safe.

### DIFF
--- a/libnmstate/schema.py
+++ b/libnmstate/schema.py
@@ -227,12 +227,14 @@ class LinuxBridge(metaclass=_DeprecatorType):
         VLAN_SUBTREE = 'vlan'
 
         class Vlan(object):
-            TRUNK_TAGS = 'trunk-tags'
-            TAG = 'tag'
             ENABLE_NATIVE = 'enable-native'
-            TYPE = 'type'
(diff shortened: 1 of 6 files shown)


libnmstate/schema.py
```diff
@@ -227,12 +227,14 @@ class Port(object):
         VLAN_SUBTREE = 'vlan'
 
         class Vlan(object):
-            TRUNK_TAGS = 'trunk-tags'
-            TAG = 'tag'
             ENABLE_NATIVE = 'enable-native'
-            TYPE = 'type'
-            ACCESS_TYPE = 'access'
-            TRUNK_TYPE = 'trunk'
+            MODE = 'type'
```
phoracek: Why don't you change the right-hand-side too?

EdDev: Missing in this PR a few points.
Will treat it in a few hours.
Thanks for the review.

EdDev: Done

PR #616, opened 2019-12-01
schema: remove object inheritance

ffmancera:
The object inheritance was needed in the schema classes in order to
support python2. Now it is not needed anymore.


EdDev:
This now has a merge which needs to be removed. Then rebase over master.

ffmancera:
> This now has a merge which needs to be removed. Then rebase over master.

Done! Thanks :)

PR #622, opened 2019-12-02
nm.connection: report missing NM plugin before device activation

ffmancera:

### DIFF
--- a/libnmstate/nm/connection.py
+++ b/libnmstate/nm/connection.py
@@ -189,6 +189,11 @@ class ConnectionProfile:
                 self._reset_profile()
                 time.sleep(0.1)
                 self.safe_activate_async()
+            elif self._is_ovs_plugin_missing(e):
+                self._mainloop.quit(
+                    "Missing 'NetworkManager-ovs' plugin"
+                    f" to handle device={self.devname}"
+                )
             else:
                 self._mainloop.quit(
                     "Connection activation failed on {} {}: error={}".format(
@@ -231,6 +236,23 @@ class ConnectionProfile:
             and err.domain == "nm-manager-error-quark"
             and err.code == 2
             and "is not available on the device" in err.message
+        )
+
+    def _is_ovs_plugin_missing(self, err):
+        return (
+            isinstance(err, nmclient.GLib.GError)
+            and err.domain == nmclient.NM_MANAGER_ERROR_DOMAIN
+            and (
+                self._con_profile.is_type(
+                    nmclient.NM.SETTING_OVS_INTERFACE_SETTING_NAME
+                )
+                or self._con_profile.is_type(
+                    nmclient.NM.SETTING_OVS_PORT_SETTING_NAME
+                )
+                or self._con_profile.is_type(
+                    nmclient.NM.SETTING_OVS_BRIDGE_SETTING_NAME
+                )
+            )
         )
 
     def _get_activation_metadata(self):
--- a/tests/integration/ovs_test.py
+++ b/tests/integration/ovs_test.py
@@ -17,19 +17,29 @@
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 #
 
+from contextlib import contextmanager
+
 import pytest
 
+import libnmstate
 from libnmstate.schema import Interface
 from libnmstate.schema import InterfaceIPv4
+from libnmstate.schema import InterfaceState
+from libnmstate.schema import InterfaceType
 from libnmstate.schema import OVSBridge
 from libnmstate.error import NmstateLibnmError
 
 from .testlib import assertlib
+from .testlib import cmd as libcmd
 from .testlib.ovslib import Bridge
 
 
 BRIDGE1 = "br1"
 PORT1 = "ovs1"
+
+DNF_REMOVE_NM_OVS_CMD = ("dnf", "remove", "-y", "-q", "NetworkManager-ovs")
+DNF_INSTALL_NM_OVS_CMD = ("dnf", "install", "-y", "-q", "NetworkManager-ovs")
+SYSTEMCTL_RESTART_NM_CMD = ("systemctl", "restart", "NetworkManager")
 
 
 @pytest.mark.xfail(
@@ -103,3 +113,30 @@ def test_create_and_remove_ovs_bridge_with_internal_port_and_static_ip():
 
     assertlib.assert_absent(BRIDGE1)
     assertlib.assert_absent(PORT1)
+
+
+def test_nm_ovs_plugin_missing():
+    with disable_nm_ovs_plugin():
+        with pytest.raises(NmstateLibnmError):
+            libnmstate.apply(
+                {
+                    Interface.KEY: [
+                        {
+                            Interface.NAME: BRIDGE1,
+                            Interface.TYPE: InterfaceType.OVS_BRIDGE,
+                            Interface.STATE: InterfaceState.UP,
+                        }
+                    ]
+                }
+            )
+
+
+@contextmanager
+def disable_nm_ovs_plugin():
+    libcmd.exec_cmd(DNF_REMOVE_NM_OVS_CMD)
+    libcmd.exec_cmd(SYSTEMCTL_RESTART_NM_CMD)
+    try:
+        yield
+    finally:
+        libcmd.exec_cmd(DNF_INSTALL_NM_OVS_CMD)
+        libcmd.exec_cmd(SYSTEMCTL_RESTART_NM_CMD)
--- a/libnmstate/nm/active_connection.py
+++ b/libnmstate/nm/active_connection.py
@@ -26,9 +26,6 @@ from .nmclient import GLib
 from .nmclient import NM
 
 
-NM_MANAGER_ERROR_DOMAIN = "nm-manager-error-quark"
-
-
 class ActivationError(Exception):
     pass
 
@@ -96,7 +93,7 @@ class ActiveConnection:
                 if (
                     isinstance(e, GLib.GError)
                     # pylint: disable=no-member
-                    and e.domain == NM_MANAGER_ERROR_DOMAIN
+                    and e.domain == nmclient.NM_MANAGER_ERROR_DOMAIN
                     and e.code == NM.ManagerError.CONNECTIONNOTACTIVE
                     # pylint: enable=no-member
                 ):
--- a/libnmstate/nm/nmclient.py
+++ b/libnmstate/nm/nmclient.py
@@ -41,6 +41,8 @@ _nmclient = None
 
 _can_disable_ipv6 = hasattr(NM, "SETTING_IP6_CONFIG_METHOD_DISABLED")
 
+NM_MANAGER_ERROR_DOMAIN = "nm-manager-error-quark"
+
 
 def can_disable_ipv6():
     return _can_disable_ipv6
--- a/tests/lib/nm/active_connection_test.py
+++ b/tests/lib/nm/active_connection_test.py
@@ -18,9 +18,8 @@
 #
 
 from libnmstate.nm.nmclient import NM
+from libnmstate.nm.nmclient import NM_MANAGER_ERROR_DOMAIN
 from libnmstate.nm.nmclient import GLib
-
-from libnmstate.nm.active_connection import NM_MANAGER_ERROR_DOMAIN
 
 
 def test_nm_manager_error_domain_str():


ffmancera:
I am testing this patch in different situations. I just pushed it to know if you like this approach. Thanks!

libnmstate/error.py
```diff
@@ -109,3 +109,11 @@ class NmstateNotSupportedError(NmstateError):
     """
 
     pass
+
+
+class NmstateMissingPluginError(NmstateError):
```
tyll: IMHO you can just use `NmstateDependencyError`

libnmstate/nm/connection.py
```diff
@@ -145,6 +147,11 @@ def safe_activate_async(self):
 
         cancellable = self._mainloop.new_cancellable()
 
+        if self.nmdevice and self.nmdevice.get_nm_plugin_is_missing():
```
tyll: What happens when the plugin is missing but the device does not exist? Does this case not happen?

ffmancera: I think this can happen. I am trying to cover this in manual testing but if this happens I don't know how could we fix that case.

ffmancera:
> Do you have an idea how to create a test for this? Can we disable the plugin temporarily via Dbus?

I need to investigate how could we do that. That seems a good idea but I don't if it is possible. Thanks!

tyll:
The original card mentioned a D-Bus error, maybe you can just convert this into a proper message.

libnmstate/nm/connection.py
```diff
@@ -145,6 +147,11 @@ def safe_activate_async(self):
...
+            raise NmstateMissingPluginError(
+                f'Missing NetworkManager plugin for device {self.devname}'
```
tyll: Actually the error message should name the plugin, for example `f"The 'NetworkManager-ovs' plugin is missing but required to manage the device '{self.devname}'"

ffmancera: I let you a comment below.

ffmancera:
If we want to specify the missing plugin then we need to check the interface type, this is the last point where we can do that and even at this point we cannot check this for ovs-bridge types. So I propose the following approach:

We could try to handle the error in the `nm/connection.py` and show up a generic missing plugin error specifying the device, id, etc.. I understand why is better to show up the missing plugin but I am not able to think on a way that could cover all the cases.

Thanks!

tyll:
> If we want to specify the missing plugin then we need to check the interface type, this is the last point where we can do that and even at this point we cannot check this for ovs-bridge types. So I propose the following approach:
> 
> We could try to handle the error in the `nm/connection.py` and show up a generic missing plugin error specifying the device, id, etc.. I understand why is better to show up the missing plugin but I am not able to think on a way that could cover all the cases.
> 
> What do you think @tyll?

I do not follow why it is not possible to determine the necessary plugin. The error message from https://nmstate.atlassian.net/browse/NMSTATE-176 states
`NetworkManager plugin for 'ovs-bridge' unavailable` which contains the interface type.

tyll:
> 1. For missing plugin, it's not debug and should be a error and quit mainloop.

Good catch!

> 3. In stead mentioning the plugin name, how about `Missing plugin to handle {device_type} interface {device_ifname}`

It needs to be clear that it is a NetworkManager plugin. Otherwise users might assume it is an Nmstate plugin. Also it would be most userfriendly to actually tell the users which plugin they need to install. If you do not like to mention the plugin here, I guess the API documentation would be a good place to add a note about plugins that are needed for each interface type.

ffmancera:
It is not possible to catch the error using the error code NM.ManagerError.MISSINGPLUGIN because NM is returning a generic error code (NM.ManagerError.FAILED). Apparently this isn't a bug, if the plugging is missing, it's pretty hard to know that this is the reason why it fails. Thanks @thom311 for explaining.

Any suggestion on this? Thanks!

cathay4t:
> > 1. For missing plugin, it's not debug and should be a error and quit mainloop.
> 
> Good catch!
> 
> > 1. In stead mentioning the plugin name, how about `Missing plugin to handle {device_type} interface {device_ifname}`
> 
> It needs to be clear that it is a NetworkManager plugin. Otherwise users might assume it is an Nmstate plugin. Also it would be most userfriendly to actually tell the users which plugin they need to install. If you do not like to mention the plugin here, I guess the API documentation would be a good place to add a note about plugins that are needed for each interface type.

You raise error on all missgin plugin error, please only mention NetworkManager-OVS when desire interface is OVS type.

cathay4t:
For testing the missing plugin, you may remove the NetworkManager-ovs rpm package and restart daemon, after the test, just install it back and restart the daemon.

thom311:
> It is not possible to catch the error using the error code NM.ManagerError.MISSINGPLUGIN because NM is returning a generic error code (NM.ManagerError.FAILED). Apparently this isn't a bug, if the plugging is missing, it's pretty hard to know that this is the reason why it fails. Thanks @thom311 for explaining.

Changing a `NM.ManagerError.FAILED` error to a more specific error code is usually no problem. Changing error codes is an API change, but probably nobody relies on unspecific "FAILED" errors. On the other hand, see `NM_CAPABILITY_TEAM`. Maybe `NMClient` should expose them as `nm_client_get_capabilities()`, then we could expose there whether the OVS plugin is installed.

ffmancera:
> > > 1. For missing plugin, it's not debug and should be a error and quit mainloop.
> > 
> > 
> > Good catch!
> > > 1. In stead mentioning the plugin name, how about `Missing plugin to handle {device_type} interface {device_ifname}`
> > 
> > 
> > It needs to be clear that it is a NetworkManager plugin. Otherwise users might assume it is an Nmstate plugin. Also it would be most userfriendly to actually tell the users which plugin they need to install. If you do not like to mention the plugin here, I guess the API documentation would be a good place to add a note about plugins that are needed for each interface type.
> 
> You raise error on all missgin plugin error, please only mention NetworkManager-OVS when desire interface is OVS type.

I agree with that but there is no device when this is happening because usually we are creating a new interface when this happens. It is possible to get the device type without having a device? Thanks!

ffmancera:
The proposed solution is working but the integration test is failing because an unknown reason. I am investigating it. Thanks!

tests/integration/ovs_test.py
```diff
@@ -103,3 +117,34 @@ def test_create_and_remove_ovs_bridge_with_internal_port_and_static_ip():
...
+    libcmd.exec_cmd(SYSTEMCTL_RESTART_NM_CMD)
+    ret = [1,0,0]
+    while ret[0]:
+        ret = libcmd.exec_cmd(SYSTEMCTL_IS_ACTIVE_NM_CMD)
+        time.sleep(1)
+    try:
+        yield
+    finally:
+        libcmd.exec_cmd(DNF_INSTALL_NM_OVS_CMD)
+        libcmd.exec_cmd(SYSTEMCTL_RESTART_NM_CMD)
```
tyll: The error log is hard to understand, nevertheless. If this was code is mostly correct, the is-active check should run here as well. Can you check if this fixes it?

ffmancera: Nope, not working..

tests/integration/ovs_test.py
```diff
@@ -103,3 +117,38 @@ def test_create_and_remove_ovs_bridge_with_internal_port_and_static_ip():
...
+
+
+@contextmanager
+def disable_nm_ovs_plugin():
+    libcmd.exec_cmd(DNF_REMOVE_NM_OVS_CMD)
+    libcmd.exec_cmd(SYSTEMCTL_RESTART_NM_CMD)
+    ret = [1,0,0]
+    while ret[0]:
+        ret = libcmd.exec_cmd(SYSTEMCTL_IS_ACTIVE_NM_CMD)
+        time.sleep(1)
```
tyll: I took a deeper look. It seems that NM is actually working already but the dbus code in Nmstate needs to be re-initialized.

This code makes the test work:
`libnmstate.nm.checkpoint._nmdbus_manager = None`

I guess the same is needed after the second restart. Maybe @EdDev has a good idea how to solve this properly. 

Ideally (IMHO) we would create a separate container without the OVS plugin and run the test in there but this would mean a bigger effort.

On the other hand, I guess this discovered a bug that the Nmstate code will fail when a long-running process loaded libnmstate and NM is restarted in-between. So it might be worth fixing this. @EdDev is vdsm a long-running process that loaded libnmstate? What will it do when NetworkManager is restarted while it is running?

It might be worth adding a test with a simple state and restarting NM to show that this breaks NM (however, it will also break all following tests, so we would need to container separation again...).

What are your opinions?

tyll: I took a deeper look. It seems that NM is actually working already but the dbus code in Nmstate needs to be re-initialized.

This code makes the test work:
`libnmstate.nm.checkpoint._nmdbus_manager = None`

I guess the same is needed after the second restart. Maybe @EdDev has a good idea how to solve this properly. 

Ideally (IMHO) we would create a separate container without the OVS plugin and run the test in there but this would mean a bigger effort.

On the other hand, I guess this discovered a bug that the Nmstate code will fail when a long-running process loaded libnmstate and NM is restarted in-between. So it might be worth fixing this. @EdDev is vdsm a long-running process that loaded libnmstate? What will it do when NetworkManager is restarted while it is running?

It might be worth adding a test with a simple state and restarting NM to show that this breaks NM (however, it will also break all following tests, so we would need to container separation again...).

What are your opinions?

tests/integration/ovs_test.py
tyll: I took a deeper look. It seems that NM is actually working already but the dbus code in Nmstate needs to be re-initialized.

tests/integration/ovs_test.py
tyll: I took a deeper look. It seems that NM is actually working already but the dbus code in Nmstate needs to be re-initialized.

This code makes the test work:
`libnmstate.nm.checkpoint._nmdbus_manager = None`

I guess the same is needed after the second restart. Maybe @EdDev has a good idea how to solve this properly. 

Ideally (IMHO) we would create a separate container without the OVS plugin and run the test in there but this would mean a bigger effort.

On the other hand, I guess this discovered a bug that the Nmstate code will fail when a long-running process loaded libnmstate and NM is restarted in-between. So it might be worth fixing this. @EdDev is vdsm a long-running process that loaded libnmstate? What will it do when NetworkManager is restarted while it is running?

It might be worth adding a test with a simple state and restarting NM to show that this breaks NM (however, it will also break all following tests, so we would need to container separation again...).

What are your opinions?

tests/integration/ovs_test.py
```diff
@@ -103,3 +117,38 @@ def test_create_and_remove_ovs_bridge_with_internal_port_and_static_ip():
...
+            }
+        )
+
+
+@contextmanager
+def disable_nm_ovs_plugin():
+    libcmd.exec_cmd(DNF_REMOVE_NM_OVS_CMD)
+    libcmd.exec_cmd(SYSTEMCTL_RESTART_NM_CMD)
+    ret = [1,0,0]
+    while ret[0]:
```
tyll: The while loops need a timeout or similar so that they do not block forever.

tests/integration/ovs_test.py
ffmancera: Wow. Thank you very much for this. Here is my opinion:

1. I think we should use the proposed workaround for this test. 
2. I think that to create a separate container without OVS for this test case is a little overkill.
3. I think it would be nice to have a card for this bug and also have a container to test it.

Thank you again! This helps so much :-)

tests/integration/ovs_test.py
ffmancera: Sure, will do.

tests/integration/ovs_test.py
```diff
@@ -103,3 +117,44 @@ def test_create_and_remove_ovs_bridge_with_internal_port_and_static_ip():
...
+
+
+@contextmanager
+def disable_nm_ovs_plugin():
+    libcmd.exec_cmd(DNF_REMOVE_NM_OVS_CMD)
+    libcmd.exec_cmd(SYSTEMCTL_RESTART_NM_CMD)
+    libnmstate.nm.checkpoint._nmdbus_manager = None
+    ret = [1, 0, 0]
+    timeout = 0
+    while ret[0] or timeout > 20:
```
tyll: The condition for the `timeout` variable should be inversed and the name is not correct (the timeout is IMHO the 20 not the elapsed time). So better use something like: `while ret[0] or elapsed_time < 20`

ffmancera: Yes, you are right. I used `while ret[0] != 0 and elapsed_time < 20`. Thanks!

libnmstate/nm/connection.py
```diff
@@ -202,6 +205,27 @@ def _active_connection_callback(self, src_object, result, user_data):
                 self._reset_profile()
                 time.sleep(0.1)
                 self.safe_activate_async()
+            elif (
+                # pylint: disable=no-member
+                isinstance(e, nmclient.GLib.GError)
```
tyll: Instead of `isinstance` it should catch the exception for the generic exception handling:

```
except nmclient.GLib.GError as e:
....
```

libnmstate/nm/connection.py
```diff
@@ -202,6 +205,27 @@ def _active_connection_callback(self, src_object, result, user_data):
...
+                    )
+                    or self._con_profile.is_type(
+                        nmclient.NM.SETTING_OVS_BRIDGE_SETTING_NAME
+                    )
+                )
+                # pylint: enable=no-member
+            ):
+                self._mainloop.quit(
+                    "Missing 'NetworkManager-ovs' plugin"
+                    " to handle device={}".format(self.devname)
```
tyll: You should be able to use `f" to handle device={self.devname}"` instead of using the `.format()` method.

tests/integration/ovs_test.py
```diff
@@ -17,20 +17,34 @@
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 #
 
+from contextlib import contextmanager
 import pytest
```
tyll: contextlib and time are part of the stdlib but pytest is not, therefore they should be in their own block before pytest.

tests/integration/ovs_test.py
```diff
@@ -103,3 +117,44 @@ def test_create_and_remove_ovs_bridge_with_internal_port_and_static_ip():
 
     assertlib.assert_absent(BRIDGE1)
     assertlib.assert_absent(PORT1)
+
+
+@pytest.mark.xfail(raises=NmstateLibnmError, strict=True)
```
tyll: Since we are expecting the exception, this should use `pytest.raises` in the test instead of expecting the whole test to fail because of an unhanded exception.

tests/integration/ovs_test.py
```diff
@@ -103,3 +117,44 @@ def test_create_and_remove_ovs_bridge_with_internal_port_and_static_ip():
...
+
+
+@contextmanager
+def disable_nm_ovs_plugin():
+    libcmd.exec_cmd(DNF_REMOVE_NM_OVS_CMD)
+    libcmd.exec_cmd(SYSTEMCTL_RESTART_NM_CMD)
+    libnmstate.nm.checkpoint._nmdbus_manager = None
+    ret = [1, 0, 0]
+    elapsed_time = 0
+    while ret[0] != 0 and elapsed_time < 20:
```
tyll: Initially, the hope was that we needed this to get the error fixed but actually we needed to clear the dBus manager. Is this loop still needed? If yes, it would be great to make it more generic as in https://github.com/nmstate/nmstate/blob/ca3450dc93a4dc3af9764c832214c2c250a92535/tests/integration/dynamic_ip_test.py#L688 to avoid repeating it twice here.

tests/integration/ovs_test.py
```diff
@@ -103,3 +117,44 @@ def test_create_and_remove_ovs_bridge_with_internal_port_and_static_ip():
...
+                ]
+            }
+        )
+
+
+@contextmanager
+def disable_nm_ovs_plugin():
+    libcmd.exec_cmd(DNF_REMOVE_NM_OVS_CMD)
+    libcmd.exec_cmd(SYSTEMCTL_RESTART_NM_CMD)
+    libnmstate.nm.checkpoint._nmdbus_manager = None
```
tyll: This should probably have a comment to make it easier to understand why it is happening here in the future.

libnmstate/nm/connection.py
ffmancera: That is not possible. This except needs to catch all the exceptions, this is why I am using `"isinstance(..)`.

libnmstate/nm/connection.py
ffmancera: Done! Thanks

tests/integration/ovs_test.py
ffmancera: Done!

tests/integration/ovs_test.py
ffmancera: This is not needed anymore.

tests/integration/ovs_test.py
ffmancera: Done!

tests/integration/ovs_test.py
ffmancera: Done! Thanks :-)

libnmstate/nm/connection.py
```diff
@@ -25,6 +25,9 @@
 from .active_connection import ActiveConnection
 
 
+NM_MANAGER_ERROR_DOMAIN = "nm-manager-error-quark"
```
EdDev: It also appears in `active_connection.py`, perhaps move it to the `nmclient` module.

libnmstate/nm/connection.py
EdDev: Please hide the condition in a method like it was done with `_is_connection_unavailable`, it will make it clear what you are looking for here (by naming it correctly).

ffmancera: Done! Thanks

libnmstate/nm/connection.py
ffmancera: Done, thanks!

tests/integration/ovs_test.py
```diff
@@ -17,20 +17,33 @@
...
 
+import libnmstate
 from libnmstate.schema import Interface
 from libnmstate.schema import InterfaceIPv4
+from libnmstate.schema import InterfaceState
+from libnmstate.schema import InterfaceType
 from libnmstate.schema import OVSBridge
 from libnmstate.error import NmstateLibnmError
 
+import pytest
```
EdDev: nit: This is a 3rd party lib, please place it after the std libs (after line 21).

tests/integration/ovs_test.py
```diff
@@ -17,20 +17,33 @@
...
 
 
 BRIDGE1 = "br1"
 PORT1 = "ovs1"
 
+DNF_REMOVE_NM_OVS_CMD = ["dnf", "remove", "-y", "-q", "NetworkManager-ovs"]
+
+DNF_INSTALL_NM_OVS_CMD = ["dnf", "install", "-y", "-q", "NetworkManager-ovs"]
+
+SYSTEMCTL_RESTART_NM_CMD = ["systemctl", "restart", "NetworkManager"]
```
EdDev: nit: Using a tuple is "safer" (immutable).

tests/integration/ovs_test.py
EdDev: It is inappropriate to touch a private variable, this should have been handled at each transaction or by a dedicated API of nmstate.
The fact that the tests have a problem is a hint that the user may have the same problem.

I also do not understand why this is a checkpoint only problem. Is it because the rest of the code is using libnm and the checkpoint is the only part that uses dbus directly?

I am ok with opening an issue on this and handling it in a separate PR.

EdDev: https://github.com/nmstate/nmstate/pull/654

tests/integration/ovs_test.py
ffmancera: Done!

tests/integration/ovs_test.py
ffmancera: Done, thanks!

ffmancera:
I am waiting for #654 to be merge and I will update and rebase the PR. Thanks!

ffmancera:
> LGTM. If possible, please change the commit title to:
> 
> `nm.connection: Report missing NM OVS plugin on profile activation`

Done and rebased thanks!

tests/integration/ovs_test.py
```diff
@@ -17,20 +17,32 @@
...
 
 
 BRIDGE1 = "br1"
 PORT1 = "ovs1"
 
+DNF_REMOVE_NM_OVS_CMD = ("dnf", "remove", "-y", "-q", "NetworkManager-ovs")
+
+DNF_INSTALL_NM_OVS_CMD = ("dnf", "install", "-y", "-q", "NetworkManager-ovs")
+
+SYSTEMCTL_RESTART_NM_CMD = ("systemctl", "restart", "NetworkManager")
```
EdDev: nit: The two newlines seem to be redundant.

ffmancera: Done! Thanks!

ffmancera:
> One nit, not a must.
> Thanks.

Done, thanks!

PR #634, opened 2019-12-05
Travis coverage

tyll:
This change is [Reviewable](https://reviewable.io/reviews/nmstate/nmstate/634)

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -2,6 +2,5 @@
 source =
     .
     /workspace/nmstate/
-    /usr/lib/python2.7/site-packages
     /usr/lib/python3.6/site-packages
     /usr/lib/python3.7/site-packages
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,6 +1,10 @@
 [paths]
 source =
     .
-    /workspace/nmstate/
+    /workspace/nmstate/.tox/check-py36/lib/python3.6/site-packages
+    /workspace/nmstate/.tox/check-py37/lib/python3.7/site-packages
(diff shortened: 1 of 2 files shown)


EdDev:
I do not understand the intent in this PR.
- Why path to py3.8 is added? Do we have tests running with py3.8?
- The added path/s for `tox` also seem strange, these should have been managed inside the tox virtenv for the things running with tox. The existing paths worked for the integration tests which we do not run from tox.
- The artifact addition of dmesg logs are unrelated to the coverage, unless I miss something.

tyll:
> I do not understand the intent in this PR.
> 
> * Why path to py3.8 is added? Do we have tests running with py3.8?

Uh, I confused this with the system role. Fedora 32 is going to use Python 3.8. I can remove the commits from this PR for now.

> * The added path/s for `tox` also seem strange, these should have been managed inside the tox virtenv for the things running with tox. The existing paths worked for the integration tests which we do not run from tox.

What do you mean with managing the paths inside the tox virtenv? It is the config file for coverage that we use to fix the paths before uploading the files to coveralls. 

> * The artifact addition of dmesg logs are unrelated to the coverage, unless I miss something.

I can put this into its own PR. When I submitted the PR it looked like NM died (the dbus service was not available) but it was not clear from the log files. I hoped to get some info from dmesg but it seems it did not die, now.

EdDev:
Changes do not affect the production code and therefore we can skip the CI verification on an up-to-date branch (change rebased).

PR #635, opened 2019-12-09
libnmstate: Deprecate positional arguments to be used for keyword arguments

tyll:
Warn when keyword arguments in `libnmstate.apply()`,
`libnmstate.commit()`, `libnmstate.rollback()` and `libnmstate.show()`
are specified as positional arguments. They will become keyword-only
arguments in nmstate-0.3.

### DIFF
--- a/libnmstate/deprecation.py
+++ b/libnmstate/deprecation.py
@@ -1,0 +1,42 @@
+#
+# Copyright (c) 2020 Red Hat, Inc.
+#
+# This file is part of nmstate
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+
+import inspect
+import warnings
+
+
+def _warn_keyword_as_positional(func):
+    signature = inspect.signature(func)
+    parameters = signature.parameters.values()
+    intended_posargs_count = len(
+        [p for p in parameters if p.default == inspect.Parameter.empty]
+    )
+
+    def wrapper(*args, **kwargs):
+        for arg_number in range(intended_posargs_count, len(args)):
+            bad_argname = list(parameters)[arg_number].name
+            warnings.warn(
+                f"Specifying '{bad_argname}' as positional argument is "
+                "deprecated. Please specify it as a keyword argument.",
+                FutureWarning,
+                stacklevel=2,
+            )
+        return func(*args, **kwargs)
+
+    return wrapper
--- a/libnmstate/netapplier.py
+++ b/libnmstate/netapplier.py
@@ -30,6 +30,7 @@ from libnmstate.nm.nmclient import nmclient_context
 from libnmstate import schema
 from libnmstate import state
 from libnmstate import validator
+from libnmstate.deprecation import _warn_keyword_as_positional
 from libnmstate.error import NmstateConflictError
 from libnmstate.error import NmstateError
 from libnmstate.error import NmstateLibnmError
@@ -40,6 +41,7 @@ from libnmstate.nm import nmclient
 MAINLOOP_TIMEOUT = 35
 
 
+@_warn_keyword_as_positional
 @nmclient_context
 def apply(desired_state, verify_change=True, commit=True, rollback_timeout=60):
     """
@@ -71,6 +73,7 @@ def apply(desired_state, verify_change=True, commit=True, rollback_timeout=60):
         return str(checkpoint.dbuspath)
 
 
+@_warn_keyword_as_positional
 def commit(checkpoint=None):
     """
     Commit a checkpoint that was received from `apply()`.
@@ -87,6 +90,7 @@ def commit(checkpoint=None):
         raise NmstateValueError(str(e))
 
 
+@_warn_keyword_as_positional
 def rollback(checkpoint=None):
     """
     Roll back a checkpoint that was received from `apply()`.
--- a/libnmstate/netinfo.py
+++ b/libnmstate/netinfo.py
@@ -20,6 +20,7 @@ from operator import itemgetter
 
 from libnmstate import nm
 from libnmstate import validator
+from libnmstate.deprecation import _warn_keyword_as_positional
 from libnmstate.nm import dns as nm_dns
 from libnmstate.nm.nmclient import nmclient_context
 from libnmstate.schema import Constants
@@ -29,6 +30,7 @@ from libnmstate.schema import Route
 from libnmstate.schema import RouteRule
 
 
+@_warn_keyword_as_positional
 @nmclient_context
 def show(include_status_data=False):
     """
--- a/tests/lib/netapplier_test.py
+++ b/tests/lib/netapplier_test.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2018-2019 Red Hat, Inc.
+# Copyright (c) 2018-2020 Red Hat, Inc.
 #
 # This file is part of nmstate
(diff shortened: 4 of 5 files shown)


CHANGELOG
```diff
@@ -1,4 +1,3 @@
-# Changelog
```
cathay4t: This seems unrelated to this PR.

tyll: Fixed this

tyll:
> This seems break the API:
> 
> Previously, user can do `apply(desired_state, True, True, 60)`, but now all the argument after `desired_state`
> got ignored.

Actually it will not be ignored but fail. The intention is to properly show that the correct way to use this API is by using keyword arguments here. This was not yet supported with Python 2.7.

cathay4t:
> > This seems break the API:
> > Previously, user can do `apply(desired_state, True, True, 60)`, but now all the argument after `desired_state`
> > got ignored.
> 
> Actually it will not be ignored but fail. The intention is to properly show that the correct way to use this API is by using keyword arguments here. This was not yet supported with Python 2.7.

If that fails, it's a break of API, I will not allow it happening in middle of 0.2 branch.

I don't get the reason of this PR.
Why current code doe not works well on future argument changes?

tyll:
> If that fails, it's a break of API, I will not allow it happening in middle of 0.2 branch.

Actually, the arguments are not documented in the API guide, therefore it is not clearly defined:
https://github.com/nmstate/nmstate.github.io/blob/master/devel/api.md#query-network-state

We can also defer this till we branch 0.3 if you prefer this.

> I don't get the reason of this PR.
> Why current code doe not works well on future argument changes?

It came up when we discussed https://github.com/nmstate/nmstate/pull/555#issuecomment-546398045 at a standup meeting, I think. I cannot recall if you were there. @EdDev proposes to add a generic dictionary with options which does not seem so pythonic (there are several examples in the stdlib that allow a lot of keyword arguments). I proposed to enforce keyword arguments in case we drop Python 2 support, so that we are as flexible as if it was a generic dictionary. This would also allow to introduce new positional arguments in case we ever need some in the future (first as non-enforced keyword argument and later as positional argument).

EdDev:
> It came up when we discussed #555 (comment) at a standup meeting, I think. I cannot recall if you were there. @EdDev proposes to add a generic dictionary with options which does not seem so pythonic (there are several examples in the stdlib that allow a lot of keyword arguments). I proposed to enforce keyword arguments in case we drop Python 2 support, so that we are as flexible as if it was a generic dictionary. This would also allow to introduce new positional arguments in case we ever need some in the future (first as non-enforced keyword argument and later as positional argument).

A dict as input is not pythonic? Dicts are the base of python and over it stuff are done.
But I think we did discuss your proposal and I claimed that this approach is programatically improper IMO, no matter what language you choose. An API needs a signature, be it the argument list of a function or the methods of a class/struct.
Using a `*` is not an API, it is a hack. Or more accurately, it is a type of duck-typing.

An API can be a dict that follows a schema (equivalent to json/yaml as input) or perhaps a nicer programmable API in the form of a data structure.
A data structure for example, can have x attributes and then be extended to have more and keeping backward compatibility. If an attribute is removed, it breaks compatibility.
In any case, it is controllable by the API designer.

tyll:
> > It came up when we discussed #555 (comment) at a standup meeting, I think. I cannot recall if you were there. @EdDev proposes to add a generic dictionary with options which does not seem so pythonic (there are several examples in the stdlib that allow a lot of keyword arguments). I proposed to enforce keyword arguments in case we drop Python 2 support, so that we are as flexible as if it was a generic dictionary. This would also allow to introduce new positional arguments in case we ever need some in the future (first as non-enforced keyword argument and later as positional argument).
> 
> A dict as input is not pythonic? Dicts are the base of python and over it stuff are done.

It is not if it is used to simulate function arguments instead of using function arguments. For example using the dict will not allow to use `functools.partial()` to set new defaults to some of the arguments but it 

> But I think we did discuss your proposal and I claimed that this approach is programatically improper IMO, no matter what language you choose. An API needs a signature, be it the argument list of a function or the methods of a class/struct.

AFAIR, you did not like to use "**kwargs". Nevertheless, AFAIU you propose to do something like

```
def foo(args):
    timeout = args.get("timeout", 0)
    other_option = args.get("other_option", "baz")
```

instead of 

```
def foo(*, timeout=0, other_option="baz")
```

and opposed doing something like

```
def (**kwargs):
    timeout = kwargs.get("timeout", 0)
    other_option = kwargs.get("other_option", "baz")
```

> Using a `*` is not an API, it is a hack. Or more accurately, it is a type of duck-typing.

I think you are confusing it with `*args` / `*kwargs` but it is also not duck-typing. Duck-typing is used in python everywhere in Nmstate. Otherwise we would have to use type annotations (which we should probably introduce for the API functions) and check/enforce them (not sure how to enforce it at runtime, though. It can be checked with mypy).

> An API can be a dict that follows a schema (equivalent to json/yaml as input) or perhaps a nicer programmable API in the form of a data structure.
> A data structure for example, can have x attributes and then be extended to have more and keeping backward compatibility. If an attribute is removed, it breaks compatibility.
> In any case, it is controllable by the API designer.

You have these benefits also with the approach in this PR but by using the normal way to specify arguments to a function instead of hiding them in a custom structure.

EdDev:
I consider `*`, `*args` and `**kwargs` as non-api signatures.
They do not declare what is exposed. A user need to dig in the function to understand what is supported.
I do not propose `args`, I propose something meaningful like `config_state`. It is just a syntactic naming thing, but it points to what the API is all about.
This `config_state` can be documented and validated using a schema, like any API.

A client of a library needs to be aware in some form that the api changed, this cannot be done using generic signatures. It can be done using data structures or schemas.

tyll:
> I consider `*`, `*args` and `**kwargs` as non-api signatures.
> They do not declare what is exposed. A user need to dig in the function to understand what is supported.

What does `*` mean in your opinion? It is a complete declaration because it means that there are no more positional arguments after that one, so

```
def foo(a, *, b=0):
    pass
```

cannot be called as `foo(1, 1)` but only as `foo(1, b=1)`. There is no need to look into the function definition.

> I do not propose `args`, I propose something meaningful like `config_state`. It is just a syntactic naming thing, but it points to what the API is all about.
> This `config_state` can be documented and validated using a schema, like any API.

Python already validates the function signature, so there is no need to invent a custom method. Also, can you give a few popular python libraries as an example that use your approach?

> A client of a library needs to be aware in some form that the api changed, this cannot be done using generic signatures. It can be done using data structures or schemas.

Also for example the schema that Nmstate provides does not validate a network state correctly since it does not warn/fail if there are unknown entries. Therefore it mostly does not add benefit IMHO.

tyll:
> With all the above said, enforcing the existing arguments to be called as kwargs is fine by me.
> I suggest that before merging this change, we notify the community of this change which will break backward compatibility.
How would you like to warn the community? Email to our fedorahosted list? @cathay4t would this be ok for you?

EdDev:
> How would you like to warn the community?

I suggest we notify through the two public and internal mailing list.
Then, add a decorator that warns about it, with a full enforcement in nmstate-0.3.
How does that sound?

tyll:
I moved the uncontroversial changes in nmstatectl into its own PR #718

tyll:
> > How would you like to warn the community?
> 
> I suggest we notify through the two public and internal mailing list.

Vdsm is now doing it correctly.

> Then, add a decorator that warns about it, with a full enforcement in nmstate-0.3.

How should the decorator look like? Do you want a generic approach that will warn if any kw-argument is specified as a positional argument or can it be individual decorators for each function?

EdDev:
> Vdsm is now doing it correctly.

Yes. We should check with CNV as well, but I expect anyone who uses this to we notified through warnings or just failing in nmstate-0.3.

> How should the decorator look like? Do you want a generic approach that will warn if any kw-argument is specified as a positional argument or can it be individual decorators for each function?

Whatever works out for you.
I think a generic one is simple enough to implement (it accepts the arguments in an ordered list and will convert them to kwargs + output a warning log).

tyll:
@cathay4t @EdDev updated it to show warnings for now. I guess you do not want to branch nmstate-0.2 for this change so I can update master to properly enforce this, yet?

cathay4t:
The black is complaining `would reformat tests/lib/netapplier_test.py` in CI

tyll:
> The black is complaining `would reformat tests/lib/netapplier_test.py` in CI

thx, fixed it.

Issue #285, opened 2019-02-28
Simplfy and unify the unit test in run-test.sh and tox.ini


cathay4t:
Thank you. I see. If we can assume that nmstate will be installed only for one python version in one running container, it seems that this can be simplified using a shell glob (does not need `PYTHON_SITE_PATH_CMD` anymore):

```
diff --git a/automation/run-tests.sh b/automation/run-tests.sh
index f6a1cf0..8164f12 100755
--- a/automation/run-tests.sh
+++ b/automation/run-tests.sh
@@ -95,8 +95,8 @@ function run_tests {
             --log-date-format='%Y-%m-%d %H:%M:%S' \
             --log-format='%(asctime)s %(filename)s:%(lineno)d %(levelname)s %(message)s' \
             --durations=5 \
-            --cov=\$($PYTHON_SITE_PATH_CMD)/libnmstate \
-            --cov=\$($PYTHON_SITE_PATH_CMD)/nmstatectl \
+            --cov /usr/lib/python*/site-packages/libnmstate \
+            --cov /usr/lib/python*/site-packages/nmstatectl \
             --cov-report=html:htmlcov-integ \
             --cov-report=term \
             tests/integration \
```
And this might also again allow to move this into tox.ini (there are currently only outdated integration test environments that should probably be removed).

_Originally posted by @tyll in https://github.com/nmstate/nmstate/pull/233_

EdDev:
If this is still relevant, please open it through [Jira could](https://nmstate.atlassian.net/secure/RapidBoard.jspa?projectKey=NMSTATE).

tyll:
The `PYTHON_SITE_PATH_CMD` variable is already removed: https://github.com/nmstate/nmstate/blob/master/automation/run-tests.sh#L134

There is also a Jira card open to cleanup tox.ini.

ffmancera:
Could we close this now? Thanks @tyll

PR #640, opened 2019-12-09
Cleanup tox config

tyll:
This also removes the outdated tox env mentioned in #285

---
This change is [Reviewable](https://reviewable.io/reviews/nmstate/nmstate/640)

### DIFF
--- a/automation/run-tests.sh
+++ b/automation/run-tests.sh
@@ -97,7 +97,7 @@ function run_tests {
             echo "Running unit test in $DOCKER_IMAGE container is not " \
                  "support yet"
         else
-            container_exec 'tox -e check-py36'
+            container_exec 'tox -e py36'
         fi
     fi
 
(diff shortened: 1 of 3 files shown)


automation/run-tests.sh
```diff
@@ -97,7 +97,7 @@ function run_tests {
             echo "Running unit test in $DOCKER_IMAGE container is not " \
                  "support yet"
         else
-            container_exec 'tox -e check-py36'
+            container_exec 'tox -e py36'
```
EdDev: I find this name change strange.
`check` points to the run of unit tests, but now it has a generic name of `pyXY` which says nothing compared to the other target names.
If tox interprets the python version from the target name, then isn't it able to do the same with some descriptive prefix as well?

EdDev:
Changes have not touched production code and seems not to collide with anything on master. Taking this in without running the CI on an up-to-date (rebased) branch.

PR #641, opened 2019-12-11
nm.device: support reapply on IPv6 changes

ffmancera:
Support reapply on IPv6 changes as it is supported by NetworkManager
again. In addition, we are supporting the workaround if the
NetworkManager version is lower than 1.20.

### DIFF
--- a/libnmstate/nm/device.py
+++ b/libnmstate/nm/device.py
@@ -197,13 +197,6 @@ def _requires_activation(dev, connection_profile):
             dev.get_iface(),
         )
         return True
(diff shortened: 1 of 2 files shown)


libnmstate/nm/device.py
```diff
@@ -197,13 +197,6 @@ def _requires_activation(dev, connection_profile):
             dev.get_iface(),
         )
         return True
-    if _ipv6_changed(dev, connection_profile):
```
tyll: After this change, nothing calls `_ipv6_changed()` anymore. @ffmancera please file a follow-up PR to remove its definition.

PR #642, opened 2019-12-12
nm.device: drop MTU reapply workaround support

ffmancera:
The minimum NetworkManager version supported is 1.20 so this patch drops
MTU reapply workaround support because it only applies for versions
lower than 1.18.

### DIFF
--- a/libnmstate/nm/device.py
+++ b/libnmstate/nm/device.py
@@ -23,8 +23,6 @@ from . import active_connection as ac
 from . import connection
 from . import nmclient
 
-from distutils.version import StrictVersion
-
 
 def activate(dev=None, connection_id=None):
     """Activate the given device or remote connection profile."""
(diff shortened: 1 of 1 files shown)


libnmstate/nm/device.py
```diff
@@ -187,45 +179,6 @@ def _modify_callback(src_object, result, user_data):
...
-def _mtu_changed(dev, connection_profile):
-    wired_setting = connection_profile.get_setting_wired()
-    configured_mtu = wired_setting.props.mtu if wired_setting else None
-    if configured_mtu:
-        current_mtu = int(dev.get_mtu())
-        return configured_mtu != current_mtu
-    return False
-
-
-def _ipv6_changed(dev, connection_profile):
```
tyll: While it is correct to remove this, it does not belong into this commit, since it is a follow-up on #641

ffmancera: Oops, yes you are right. Thanks!

ffmancera: Fixed!

ffmancera:
Oops I missed it. Sorry

> Uh, actually you seem to have removed too much compared to the previous release or I missed it, the call of `_requires_activation()` needs to be removed as well.

Oops, I missed it. Thanks!

PR #643, opened 2019-12-12
Enable black string normalization

tyll:

---
This change is [Reviewable](https://reviewable.io/reviews/nmstate/nmstate/643)


tyll:
The CI failure does not seem to be related to this change but it fails in a VLAN test that seems to fail often, recently. I wonder if the changes to device deletion or earlier activation now make Nmstate check the outcome state too early.

PR #645, opened 2019-12-12
Travis: Enable codecov

tyll:
This allows to use coveralls for integration test coverage and codecov
for everything else in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,11 +53,12 @@ script:
     - ./automation/run-tests.sh $testflags
 
 after_success:
-    - sudo pip install coveralls
+    - sudo pip install coveralls codecov
     - sudo mv .coverage .coverage.orig || true
     - sudo mv tests/.coverage .coverage.tox || true
     - sudo coverage combine --append .
     - coveralls
+    - codecov
 
 after_failure:
     - ./automation/upload_test_artifacts.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,13 @@
 dist: bionic
 language: generic
 env:
+    global:
+        - use_codecov=true
     matrix:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -57,8 +61,8 @@ after_success:
     - sudo mv .coverage .coverage.orig || true
     - sudo mv tests/.coverage .coverage.tox || true
     - sudo coverage combine --append .
-    - coveralls
-    - codecov
+    - test "${use_coveralls}" == "true" && coveralls || true
+    - test "${use_codecov}" == "true" &&  codecov || true
```
ffmancera: I don't understand why the last `|| true` are needed. Other than that, it looks fine to me.

tyll: Travis aborts if a command returns a non-zero exit code and if one of those environment variables is not set to `true`, the return code will not be zero. In this case, Travis should not consider this an error. the `|| true` makes sure that the return code will always be `0`.

EdDev:
Can someone explain the end result now?
What coverage is seen now with codecov and what is seen with coveralls?
Do we still need that old hack to collect and merge the coverage reports?

tyll:
> Can someone explain the end result now?
> What coverage is seen now with codecov and what is seen with coveralls?

Coveralls will show integration tests on the CentOS 8 images, codecov the remaining coverage. We will see how it looks after it was merged. Currently the web interface seems not to like it that there is no information from the master branch. We can change it, if you like to switch it (coveralls at least shows the coverage for the separate jobs).

> Do we still need that old hack to collect and merge the coverage reports?

AFAIK there is not hack to merge the reports, it happens in the coveralls web service. The command at https://github.com/nmstate/nmstate/blob/master/.travis.yml#L59 sanitizes/normalizes the paths since the coverage files contains the paths from inside the containers - but it is broken currently, see #634.

PR #648, opened 2019-12-13
schema, nm.sr-iov: extend SR-IOV VF configuration

ffmancera:
This PR is going to extend SR-IOV in order to support VF configuration with nmstate.

### DIFF
--- a/libnmstate/nm/sriov.py
+++ b/libnmstate/nm/sriov.py
@@ -25,6 +25,53 @@ from libnmstate.schema import Ethernet
 from libnmstate.schema import Interface
 
 
+SRIOV_NMSTATE_TO_NM_MAP = {
+    Ethernet.SRIOV.VFS.MAC_ADDRESS: (
+        nmclient.NM.SRIOV_VF_ATTRIBUTE_MAC,
+        nmclient.GLib.Variant.new_string,
+    ),
+    Ethernet.SRIOV.VFS.SPOOF_CHECK: (
+        nmclient.NM.SRIOV_VF_ATTRIBUTE_SPOOF_CHECK,
+        nmclient.GLib.Variant.new_boolean,
+    ),
+    Ethernet.SRIOV.VFS.TRUST: (
+        nmclient.NM.SRIOV_VF_ATTRIBUTE_TRUST,
+        nmclient.GLib.Variant.new_boolean,
+    ),
+    Ethernet.SRIOV.VFS.MIN_TX_RATE: (
+        nmclient.NM.SRIOV_VF_ATTRIBUTE_MIN_TX_RATE,
+        nmclient.GLib.Variant.new_uint32,
+    ),
+    Ethernet.SRIOV.VFS.MAX_TX_RATE: (
+        nmclient.NM.SRIOV_VF_ATTRIBUTE_MAX_TX_RATE,
+        nmclient.GLib.Variant.new_uint32,
+    ),
+}
+
+SRIOV_NM_TO_NMSTATE_MAP = {
+    nmclient.NM.SRIOV_VF_ATTRIBUTE_MAC: (
+        Ethernet.SRIOV.VFS.MAC_ADDRESS,
+        nmclient.GLib.Variant.get_string,
+    ),
+    nmclient.NM.SRIOV_VF_ATTRIBUTE_SPOOF_CHECK: (
+        Ethernet.SRIOV.VFS.SPOOF_CHECK,
+        nmclient.GLib.Variant.get_boolean,
+    ),
+    nmclient.NM.SRIOV_VF_ATTRIBUTE_TRUST: (
+        Ethernet.SRIOV.VFS.TRUST,
+        nmclient.GLib.Variant.get_boolean,
+    ),
+    nmclient.NM.SRIOV_VF_ATTRIBUTE_MIN_TX_RATE: (
+        Ethernet.SRIOV.VFS.MIN_TX_RATE,
+        nmclient.GLib.Variant.get_uint32,
+    ),
+    nmclient.NM.SRIOV_VF_ATTRIBUTE_MAX_TX_RATE: (
+        Ethernet.SRIOV.VFS.MAX_TX_RATE,
+        nmclient.GLib.Variant.get_uint32,
+    ),
+}
+
+
 def create_setting(iface_state, base_con_profile):
     sriov_setting = None
     ifname = iface_state[Interface.NAME]
@@ -43,9 +90,52 @@ def create_setting(iface_state, base_con_profile):
         if not sriov_setting:
             sriov_setting = nmclient.NM.SettingSriov.new()
 
+        vfs_config = sriov_config.get(Ethernet.SRIOV.VFS_SUBTREE, [])
+        vf_object_ids = {vf.get_index() for vf in sriov_setting.props.vfs}
+        vf_config_ids = {
+            vf_config[Ethernet.SRIOV.VFS.ID] for vf_config in vfs_config
+        }
+
+        # As the user must do full edit of vfs, nmstate is deleting all the vfs
+        # and then adding all the vfs from the config.
+        for vf_id in _remove_sriov_vfs_in_setting(
+            vfs_config, sriov_setting, vf_object_ids
+        ):
+            sriov_setting.remove_vf_by_index(vf_id)
+
+        for vf_object in _create_sriov_vfs_from_config(
+            vfs_config, sriov_setting, vf_config_ids
+        ):
+            sriov_setting.add_vf(vf_object)
+
         sriov_setting.props.total_vfs = sriov_config[Ethernet.SRIOV.TOTAL_VFS]
 
     return sriov_setting
+
+
+def _create_sriov_vfs_from_config(vfs_config, sriov_setting, vf_ids_to_add):
+    vfs_config_to_add = (
+        vf_config
+        for vf_config in vfs_config
+        if vf_config[Ethernet.SRIOV.VFS.ID] in vf_ids_to_add
+    )
+    for vf_config in vfs_config_to_add:
+        vf_id = vf_config.pop(Ethernet.SRIOV.VFS.ID)
+        vf_object = nmclient.NM.SriovVF.new(vf_id)
+        for key, val in vf_config.items():
+            _set_nm_attribute(vf_object, key, val)
+
+        yield vf_object
+
+
+def _set_nm_attribute(vf_object, key, value):
+    nm_attr, nm_variant = SRIOV_NMSTATE_TO_NM_MAP[key]
+    vf_object.set_attribute(nm_attr, nm_variant(value))
+
+
+def _remove_sriov_vfs_in_setting(vfs_config, sriov_setting, vf_ids_to_remove):
+    for vf_id in vf_ids_to_remove:
+        yield vf_id
 
 
 def _has_sriov_capability(ifname):
@@ -72,8 +162,34 @@ def get_info(device):
     )
 
     if sriov_setting:
-        info[Ethernet.SRIOV_SUBTREE] = {
+        sriov_config = {
             Ethernet.SRIOV.TOTAL_VFS: sriov_setting.props.total_vfs
         }
+        vfs_config = _get_info_sriov_vfs_config(sriov_setting)
+        sriov_config[Ethernet.SRIOV.VFS_SUBTREE] = vfs_config
+
+        info[Ethernet.SRIOV_SUBTREE] = sriov_config
 
     return info
+
+
+def _get_info_sriov_vfs_config(sriov_setting):
+    vfs_config = []
+    vfs_setting = sriov_setting.props.vfs
+    for vf in vfs_setting:
+        vf_config = {}
+        vf_config[Ethernet.SRIOV.VFS.ID] = vf.get_index()
+        for nm_attribute in SRIOV_NM_TO_NMSTATE_MAP.keys():
+            _get_nm_attribute(vf, vf_config, nm_attribute)
+
+        vfs_config.append(vf_config)
+
+    return vfs_config
+
+
+def _get_nm_attribute(vf_object, vf_config, nm_attribute):
+    nmstate_key, nm_variant = SRIOV_NM_TO_NMSTATE_MAP[nm_attribute]
+    if vf_object.get_attribute(nm_attribute) is not None:
+        vf_config[nmstate_key] = nm_variant(
+            vf_object.get_attribute(nm_attribute)
+        )
--- a/libnmstate/schema.py
+++ b/libnmstate/schema.py
@@ -278,6 +278,15 @@ class Ethernet:
 
     class SRIOV:
         TOTAL_VFS = "total-vfs"
+        VFS_SUBTREE = "vfs"
+
+        class VFS:
+            ID = "id"
+            MAC_ADDRESS = "mac-address"
+            SPOOF_CHECK = "spoof-check"
+            TRUST = "trust"
+            MIN_TX_RATE = "min-tx-rate"
+            MAX_TX_RATE = "max-tx-rate"
 
 
 class VLAN:
--- a/libnmstate/schemas/operational-state.yaml
+++ b/libnmstate/schemas/operational-state.yaml
@@ -177,6 +177,28 @@ definitions:
             total-vfs:
               type: integer
               minimum: 0
+            vfs:
+              type: array
+              items:
+                type: object
+                properties:
+                  id:
+                    type: integer
+                    minimum: 0
+                  mac-address:
+                    $ref: "#/definitions/types/mac-address"
+                  spoof-check:
+                    type: boolean
+                  trust:
+                    type: boolean
+                  min-tx-rate:
+                    type: integer
+                    minimum: 0
+                  max-tx-rate:
+                    type: integer
+                    minimum: 0
+                required:
+                  - id
   interface-vlan:
     rw:
       properties:
--- a/tests/integration/sriov_test.py
+++ b/tests/integration/sriov_test.py
@@ -27,7 +27,7 @@ from libnmstate.schema import Interface
 from .testlib import assertlib
 
 
-SRIOV_CONFIG = {Ethernet.SRIOV_SUBTREE: {Ethernet.SRIOV.TOTAL_VFS: 0}}
+SRIOV_CONFIG = {Ethernet.SRIOV_SUBTREE: {Ethernet.SRIOV.TOTAL_VFS: 2}}
 
 
 @pytest.mark.xfail(
@@ -55,14 +55,71 @@ def test_sriov_increase_vfs(sriov_interface):
 )
 def test_sriov_decrease_vfs(sriov_interface):
     eth_config = sriov_interface[Interface.KEY][0][Ethernet.CONFIG_SUBTREE]
-    eth_config[Ethernet.SRIOV_SUBTREE][Ethernet.SRIOV.TOTAL_VFS] = 5
+    eth_config[Ethernet.SRIOV_SUBTREE][Ethernet.SRIOV.TOTAL_VFS] = 0
     libnmstate.apply(sriov_interface)
-    eth_config[Ethernet.SRIOV_SUBTREE][Ethernet.SRIOV.TOTAL_VFS] = 2
+    assertlib.assert_state(sriov_interface)
+
+
+@pytest.mark.xfail(
+    raises=NmstateNotSupportedError,
+    reason="The device does not support SR-IOV.",
+)
+def test_sriov_create_vf_config(sriov_interface):
+    eth_config = sriov_interface[Interface.KEY][0][Ethernet.CONFIG_SUBTREE]
+    eth_config[Ethernet.SRIOV_SUBTREE][Ethernet.SRIOV.VFS_SUBTREE] = [
+        {
+            Ethernet.SRIOV.VFS.ID: 0,
+            Ethernet.SRIOV.VFS.SPOOF_CHECK: True,
+            Ethernet.SRIOV.VFS.MAC_ADDRESS: "00:11:22:33:44:55",
+            Ethernet.SRIOV.VFS.TRUST: False,
+        }
+    ]
+    libnmstate.apply(sriov_interface)
+    assertlib.assert_state(sriov_interface)
+
+
+@pytest.mark.xfail(
+    raises=NmstateNotSupportedError,
+    reason="The device does not support SR-IOV.",
+)
+def test_sriov_edit_vf_config(sriov_iface_vf):
+    eth_config = sriov_iface_vf[Interface.KEY][0][Ethernet.CONFIG_SUBTREE]
+    vf0 = eth_config[Ethernet.SRIOV_SUBTREE][Ethernet.SRIOV.VFS_SUBTREE][0]
+    vf0[Ethernet.SRIOV.VFS.TRUST] = True
+    vf0[Ethernet.SRIOV.VFS.MAC_ADDRESS] = "55:44:33:22:11:00"
+    libnmstate.apply(sriov_interface)
+    assertlib.assert_state(sriov_interface)
+
+
+@pytest.mark.xfail(
+    raises=NmstateNotSupportedError,
+    reason="The device does not support SR-IOV.",
+)
+def test_sriov_remove_vf_config(sriov_iface_vf):
+    eth_config = sriov_iface_vf[Interface.KEY][0][Ethernet.CONFIG_SUBTREE]
+    eth_config[Ethernet.SRIOV_SUBTREE][Ethernet.SRIOV.VFS_SUBTREE] = []
+    libnmstate.apply(sriov_interface)
     assertlib.assert_state(sriov_interface)
 
 
 @pytest.fixture
 def sriov_interface(eth1_up):
     eth1_up[Interface.KEY][0][Ethernet.CONFIG_SUBTREE] = SRIOV_CONFIG
+    libnmstate.apply(eth1_up)
+    yield eth1_up
+
+
+@pytest.fixture
+def sriov_iface_vf(eth1_up):
+    eth1_up[Interface.KEY][0][Ethernet.CONFIG_SUBTREE] = SRIOV_CONFIG
+    eth_config = eth1_up[Interface.KEY][0][Ethernet.CONFIG_SUBTREE]
+    eth_config[Ethernet.SRIOV_SUBTREE][Ethernet.SRIOV.VFS_SUBTREE] = [
+        {
+            Ethernet.SRIOV.VFS.ID: 0,
+            Ethernet.SRIOV.VFS.SPOOF_CHECK: True,
+            Ethernet.SRIOV.VFS.MAC_ADDRESS: "00:11:22:33:44:55",
+            Ethernet.SRIOV.VFS.TRUST: False,
+        }
+    ]
     libnmstate.apply(eth1_up)
     yield eth1_up
--- a/tests/lib/schema_validation_test.py
+++ b/tests/lib/schema_validation_test.py
@@ -40,6 +40,7 @@ ROUTES = Constants.ROUTES
 THE_BRIDGE = "br0"
 VXLAN0 = "vxlan0"
 TEAM0 = "team0"
+SRIOV_MAC = "12:34:56:78:90:ab"
 
 COMMON_DATA = {
     INTERFACES: [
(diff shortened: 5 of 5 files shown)


ffmancera:
I've pushed the schema proposal, I am working on the implementation but it would be nice to discuss the schema here. Thanks!

libnmstate/schemas/operational-state.yaml
```diff
@@ -177,6 +177,41 @@ definitions:
...
+                    type: array
+                    items:
+                      type: object
+                      properties:
+                        id:
+                          type: integer
+                          minimum: 0
+                        priority:
+                          type: integer
+                        proto:
```
cathay4t: Please make it `protocol`.

libnmstate/schema.py
```diff
@@ -278,6 +278,21 @@ class Ethernet:
 
     class SRIOV:
         TOTAL_VFS = "total-vfs"
+        VFS_SUBTREE = "vfs"
+
+        class VFS:
+            ID = "id"
+            MAC = "mac-address"
```
tyll: @EdDev I have a question about the schema. AFAIU we have the JSON schema so that users can auto-generate code from it. But how would they then use the constants from `schema.py`? Especially when they do not match 1:1 like `MAC` and `mac-address`. Also, if users are supposed to auto-generate code from the JSON schema, why does Nmstate not generate the constants from JSON schema as well?

libnmstate/schema.py
```diff
@@ -278,6 +278,21 @@ class Ethernet:
...
+            SPOOF_CHECK = "spoof-check"
+            TRUST = "trust"
+            MIN_TX_RATE = "min-tx-rate"
+            MAX_TX_RATE = "max-tx-rate"
+            VLANS_SUBTREE = "vlans"
+
+            class VLANS:
+                ID = "id"
+                PRIORITY = "priority"
+                PROTO = "proto"
```
tyll: As @cathay4t wrote, `protocol` would be nicer.

ffmancera: Done, thanks!

libnmstate/schemas/operational-state.yaml
ffmancera: Sure, done thanks!

ffmancera:
This patch have not been tested yet, I will try it out on Monday with special HW.

libnmstate/nm/sriov.py
```diff
@@ -43,11 +45,88 @@ def create_setting(iface_state, base_con_profile):
...
+def _create_vf_vlan_setting(vlan_config, vf_object):
+    for vlan in vlan_config:
+        vlan_id = vlan_config[Ethernet.SRIOV.VFS.VLANS.ID]
+        vf_object.add_vlan(vlan_id)
+        if vlan_config.get(Ethernet.SRIOV.VFS.VLANS.PRIORITY):
+            vf_object.set_vlan_qos(
+                vlan_id, vlan_config[Ethernet.SRIOV.VFS.VLANS.PRIORITY]
+            )
+        if vlan_config.get(Ethernet.SRIOV.VFS.VLANS.PROTOCOL):
+            # pylint: disable=no-member
```
cathay4t: Why this line? Is that caused by `nmclient.NM.SriovVFVlanProtocol._1AD`?

tests/lib/schema_validation_test.py
```diff
@@ -231,6 +231,153 @@ def test_over_maximum_total_vfs_is_invalid(
...
+                        {Ethernet.SRIOV.VFS.ID: invalid_values}
+                    ],
+                },
+            }
+        )
+
+        with pytest.raises(js.ValidationError, match=str(invalid_values)):
+            libnmstate.validator.validate(default_data)
+
+    @pytest.mark.parametrize("valid_values", [0, 1, 20])
```
cathay4t: This line is unused.

tests/lib/schema_validation_test.py
```diff
@@ -231,6 +231,153 @@ def test_over_maximum_total_vfs_is_invalid(
...
+                            Ethernet.SRIOV.VFS.VLANS_SUBTREE: [
+                                {Ethernet.SRIOV.VFS.VLANS.ID: invalid_values}
+                            ],
+                        }
+                    ],
+                },
+            }
+        )
+
+        with pytest.raises(js.ValidationError, match=str(invalid_values)):
```
cathay4t: The string of error is not part of API of jsonschema module. There is not need to test on that.

ffmancera:
This patch introduces SR-IOV VF configuration using nmstate. I have test this manually using HW with SR-IOV support and it is working as expected. 

Please, note that currently it is not possible to configure VF VLAN options due to a error in type annotations in get_vlan_ids() function. The fix is already in the upstream [1][2]. Thanks @thom311!

 [1] https://gitlab.freedesktop.org/NetworkManager/NetworkManager/commit/c4a728217d3794627268303475531df5be35a27b 

[2] https://gitlab.freedesktop.org/NetworkManager/NetworkManager/commit/d65aabd5180b9d668610a9deb230d1baecb71a52

tests/lib/schema_validation_test.py
ffmancera: I don't think so. This is testing the valid VF ids.

tests/lib/schema_validation_test.py
ffmancera: This test does not exist anymore. Thanks!

libnmstate/nm/sriov.py
ffmancera: Not needed anymore. Thanks!

libnmstate/schema.py
EdDev: There are tools that generate code from the schema and tools that generate schema from code.
So a possible flow could be something like this (I think this is what kubernetes is doing):
- Define data structs in code.
- Generate from them a schema.
- For clients, generate from the schema code like a REST API.

The binding between the constants and the schema can be automated, but I think we have not gone that step yet. @cathay4t also preferred not to automate this due to his bad experience from other projects I guess.

This by itself could be a task with a research, design and implementation steps.

EdDev:
> Please, note that currently it is not possible to configure VF VLAN options due to a error in type annotations in get_vlan_ids() function. The fix is already in the upstream

Have you referenced this in the tests (xfail) with the BZ?

tests/lib/schema_validation_test.py
```diff
@@ -231,6 +231,73 @@ def test_over_maximum_total_vfs_is_invalid(
         with pytest.raises(js.ValidationError, match=str(invalid_values)):
             libnmstate.validator.validate(default_data)
 
+    @pytest.mark.parametrize("invalid_values", [-50, -1])
```
EdDev: `invalid_values` >> `invalid_value`
I would actually prefer it named per what it represent: `vf_if`.

The test name can also be generalized to represent the invalid ids: `test_invalid_vf_ids`

tests/lib/schema_validation_test.py
```diff
@@ -231,6 +231,73 @@ def test_over_maximum_total_vfs_is_invalid(
...
+                        {Ethernet.SRIOV.VFS.ID: invalid_values}
+                    ],
+                },
+            }
+        )
+
+        with pytest.raises(js.ValidationError, match=str(invalid_values)):
+            libnmstate.validator.validate(default_data)
+
+    @pytest.mark.parametrize("valid_values", [0, 1, 20])
```
EdDev: How about `vf_id` for the argument name and generalize the test name to `test_valid_vf_ids`

libnmstate/schema.py
```diff
@@ -278,6 +278,16 @@ class Ethernet:
...
+        VFS_SUBTREE = "vfs"
+
+        class VFS:
+            ID = "id"
+            MAC = "mac-address"
+            SPOOF_CHECK = "spoof-check"
+            TRUST = "trust"
+            MIN_TX_RATE = "min-tx-rate"
+            MAX_TX_RATE = "max-tx-rate"
+            VLANS_SUBTREE = "vlans"
```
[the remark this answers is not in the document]
ffmancera: Yes you are right. This definition should also not appear.

tests/lib/schema_validation_test.py
ffmancera: Sure, that seems better.

tests/lib/schema_validation_test.py
ffmancera: Seems fine! Thanks

ffmancera:
> > Please, note that currently it is not possible to configure VF VLAN options due to a error in type annotations in get_vlan_ids() function. The fix is already in the upstream
> 
> Have you referenced this in the tests (xfail) with the BZ?

No, because I couldn't implement it so there is no tests about this. Working on the requested changes. Thanks!

libnmstate/nm/sriov.py
```diff
@@ -43,11 +43,79 @@ def create_setting(iface_state, base_con_profile):
...
+
+        sriov_setting.add_vf(vf_object)
+
+
+def _remove_sriov_vfs_in_setting(vfs_config, sriov_setting):
+    vf_object_ids = {vf.get_index() for vf in sriov_setting.props.vfs}
+    vf_config_ids = {
+        vf_config[Ethernet.SRIOV.VFS.ID] for vf_config in vfs_config
+    }
+    vf_ids_to_remove = vf_object_ids.difference(vf_config_ids)
```
EdDev: This is equivalent to `vf_object_ids - vf_config_ids`

libnmstate/nm/sriov.py
```diff
@@ -43,11 +43,79 @@ def create_setting(iface_state, base_con_profile):
...
+            sriov_setting.remove_vf_by_index(vf_id)
+
+        sriov_setting.add_vf(vf_object)
+
+
+def _remove_sriov_vfs_in_setting(vfs_config, sriov_setting):
+    vf_object_ids = {vf.get_index() for vf in sriov_setting.props.vfs}
+    vf_config_ids = {
+        vf_config[Ethernet.SRIOV.VFS.ID] for vf_config in vfs_config
+    }
```
EdDev: This should be common for both adding and removing entries.
The, to one you can pass `vf_object_ids - vf_config_ids` and to the other `vf_config_ids - vf_object_ids`.
The 3rd is editing, which fits sending with `vf_object_ids & vf_config_ids`.

libnmstate/nm/sriov.py
```diff
@@ -72,8 +140,63 @@ def get_info(device):
...
-        info[Ethernet.SRIOV_SUBTREE] = {
-            Ethernet.SRIOV.TOTAL_VFS: sriov_setting.props.total_vfs
-        }
+        sriov_config = {}
+        sriov_config[Ethernet.SRIOV.TOTAL_VFS] = sriov_setting.props.total_vfs
+
+        vfs_config = _create_sriov_vfs_config(sriov_setting)
+        sriov_config[Ethernet.SRIOV.VFS_SUBTREE] = vfs_config
+
+        info[Ethernet.SRIOV_SUBTREE] = sriov_config
```
EdDev: I think we should provide the info from the kernel (sysfs in this case) and open an RFE on NM to provide this data through libnm.

tests/integration/sriov_test.py
```diff
@@ -61,6 +61,47 @@ def test_sriov_decrease_vfs(sriov_interface):
     assertlib.assert_state(sriov_interface)
 
 
+@pytest.mark.xfail(
```
EdDev: These tests are passing on HW with SRIOV?

libnmstate/nm/sriov.py
ffmancera: I don't understand what do you mean with this. I just do this operation once.

libnmstate/nm/sriov.py

tests/integration/sriov_test.py
ffmancera: Yes, when eth1 have SR-IOV support. Then it is passing.

libnmstate/nm/sriov.py
ffmancera: That is okay for `total-vfs` but not for other VF options because it have different formats. I suggest to get `total-vfs` from `/sys/class/net/..` but get the VF options from NM. The RFE is already opened.

ffmancera: Done for `total-vfs`. Thanks!

EdDev: > not for other VF options because it have different formats.
> but get the VF options from NM

I do not understand this point.
If you will not fetch this info from the kernel, then you cannot report the state of all available VF/s that exists.

libnmstate/nm/sriov.py
EdDev: You do something very similar in `_create_sriov_vfs_in_setting`.
I hope you can calculate the ids at the caller and the remove duplicate logic/code.

EdDev:
> commit schema: extend SR-IOV schema:
> 
> - I'm missing validation on required fields (like the vf_id) with their tests.
> 
> You can introduce it in the same commit and the schema on through a separate one.

I could not see if you responded to this one.

libnmstate/nm/sriov.py
ffmancera: > > commit schema: extend SR-IOV schema:
> > 
> > * I'm missing validation on required fields (like the vf_id) with their tests.
> > 
> > You can introduce it in the same commit and the schema on through a separate one.
> 
> I could not see if you responded to this one.

Oops, I didn't see that. Yes, done! Thanks :)

ffmancera:
> > commit schema: extend SR-IOV schema:
> > 
> > * I'm missing validation on required fields (like the vf_id) with their tests.
> > 
> > You can introduce it in the same commit and the schema on through a separate one.
> 
> I could not see if you responded to this one.

Oops, I didn't see that. Done, thanks!

libnmstate/nm/sriov.py
ffmancera: Done! Thanks :-)

libnmstate/nm/sriov.py
ffmancera: We have decided to parse the `ip link show` output in a following PR.

zshi-redhat:
Hi @ffmancera,  this comment maybe too late to be addressed in this PR as it's already in good shape.
Just want to ask if it is possible to add link_state, vlanQoS and mtu configuration support for SR-IOV VFs, perhaps in separate one?

ffmancera:
> Hi @ffmancera, this comment maybe too late to be addressed in this PR as it's already in good shape.
> Just want to ask if it is possible to add link_state, vlanQoS and mtu configuration support for SR-IOV VFs, perhaps in separate one?

Hello! Currently we are supporting link_state and mtu configuration for SR-IOV VFs. When you create a VF (setting the `total-vfs` number in the PF), you could find the VF interface state using nmstate as the other interfaces. e.g:

```
interfaces:
- name: eno1
  type: ethernet
  state: up
  ethernet:
    auto-negotiation: false
    duplex: full
    speed: 10000
    sr-iov:
      total-vfs: 2
      vfs: []
  ipv4:
    dhcp: false
    enabled: false
  ipv6:
    autoconf: false
    dhcp: false
    enabled: false
  mac-address: E4:43:4B:17:F8:48
  mtu: 1500
- name: eno1v0
  type: ethernet
  state: down
  ethernet:
    auto-negotiation: false
    duplex: full
    speed: 10000
  ipv4:
    enabled: false
  ipv6:
    enabled: false
  mac-address: A2:7A:D9:12:D5:78
  mtu: 1500
- name: eno1v1
  type: ethernet
  state: down
  ethernet:
    auto-negotiation: false
    duplex: full
    speed: 10000
  ipv4:
    enabled: false
  ipv6:
    enabled: false
  mac-address: 4A:A3:23:72:5C:83
  mtu: 1500
```

As you can see above `eno1v0` and `eno1v1` are the states of the two SR-IOV VFs from the eno1 PF. Is that okay for you?

About the vlanQoS, currently there is a bug in NetworkManager which does not allow us to configure vlan VFs. The patch is already on upstream [1] [2] so we are going to support it soon. 

[1] https://gitlab.freedesktop.org/NetworkManager/NetworkManager/commit/c4a728217d3794627268303475531df5be35a27b

[2] https://gitlab.freedesktop.org/NetworkManager/NetworkManager/commit/d65aabd5180b9d668610a9deb230d1baecb71a52

libnmstate/nm/sriov.py
```diff
@@ -72,8 +140,65 @@ def get_info(device):
...
+        if vf.get_attribute(nmclient.NM.SRIOV_VF_ATTRIBUTE_TRUST) is not None:
+            vf_config[
+                Ethernet.SRIOV.VFS.TRUST
+            ] = nmclient.GLib.Variant.get_boolean(
+                vf.get_attribute(nmclient.NM.SRIOV_VF_ATTRIBUTE_TRUST)
+            )
+        if (
+            vf.get_attribute(nmclient.NM.SRIOV_VF_ATTRIBUTE_MAX_TX_RATE)
+            is not None
+        ):
```
cathay4t: These duplicated codes be turn into a function:

```
def _fill_vf_atrribute(vf_cfg, vf, key_name, nm_name, get_func):
    nm_val = vf.get_attribute(nm_name)
    if nm_val is not None:
        cf_cfg[key_name] = get_func(nm_val)
```

libnmstate/schema.py
cathay4t: @ffmancera  `MAC = "mac-address"` should be changed to `MAC_ADDRESS` just in case we move to auto generate.

@tyll @EdDev Things might changed a lot in Python, if any of you guys want to do the auto-generate code on schema, we can discuss that in the PR.

libnmstate/nm/sriov.py
```diff
@@ -72,8 +140,65 @@ def get_info(device):
     )
 
     if sriov_setting:
-        info[Ethernet.SRIOV_SUBTREE] = {
-            Ethernet.SRIOV.TOTAL_VFS: sriov_setting.props.total_vfs
-        }
+        sriov_config = {}
+        ifname = device.get_iface()
+        with open(f"/sys/class/net/{ifname}/device/sriov_numvfs", "r") as f:
+            sriov_config[Ethernet.SRIOV.TOTAL_VFS] = int(f.read())
```
cathay4t: If possible, please split this change into another commit or PR.

libnmstate/nm/sriov.py
ffmancera: Yes that is fine but I would need to create 3 different functions as we need to convert the GLib.Variant types to string, int or bool, this is why I am using `nmclient.GLib.Variant.get_uint32` or `nmclient.GLib.Variant.get_boolean`. If you are fine with that I can create them. Also, it would be fine to have a module to handle with GLib.Variant types and use it in other implementations.

libnmstate/schema.py
EdDev: @cathay4t , we use `MAC` also in `Interface`. But I tend to agree that the full name will fit well with a potential future auto-generation.

tests/lib/schema_validation_test.py
EdDev: Why not `vf_if`? Note that currently it is also plural, but the value is singular.
What about the test name?

tests/lib/schema_validation_test.py
EdDev: ping

libnmstate/nm/sriov.py
```diff
@@ -72,8 +140,65 @@ def get_info(device):
...
+
+        vfs_config = _create_sriov_vfs_config(sriov_setting)
+        sriov_config[Ethernet.SRIOV.VFS_SUBTREE] = vfs_config
+
+        info[Ethernet.SRIOV_SUBTREE] = sriov_config
 
     return info
+
+
+def _create_sriov_vfs_config(sriov_setting):
```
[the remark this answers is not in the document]
ffmancera: Done! Thanks :-)

libnmstate/nm/sriov.py
ffmancera: Done! I added the change in another commit

EdDev:
Looks like some of the fixes on commit `schema: extend SR-IOV schema` leaked to the next commit.

ffmancera:
> Looks like some of the fixes on commit `schema: extend SR-IOV schema` leaked to the next commit.

Oops sorry. I am going to fix that.

ffmancera:
> Looks like some of the fixes on commit `schema: extend SR-IOV schema` leaked to the next commit.

I fixed that. All the schema validation tests fixes are now in `schema: extend SR-IOV schema`. Sorry for that. Thanks!

libnmstate/nm/sriov.py
```diff
@@ -75,5 +143,60 @@ def get_info(device):
         info[Ethernet.SRIOV_SUBTREE] = {
             Ethernet.SRIOV.TOTAL_VFS: sriov_setting.props.total_vfs
         }
+        vfs_config = _get_info_sriov_vfs_config(sriov_setting)
+        sriov_config[Ethernet.SRIOV.VFS_SUBTREE] = vfs_config
```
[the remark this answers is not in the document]
EdDev: Looks like this comes from the next commit.

libnmstate/nm/sriov.py
```diff
@@ -43,11 +43,79 @@ def create_setting(iface_state, base_con_profile):
...
 
     return sriov_setting
 
 
+def _create_sriov_vfs_in_setting(vfs_config, sriov_setting, vf_ids_to_add):
+    for vf_config in vfs_config:
+        vf_already_exists = True
+        vf_id = vf_config[Ethernet.SRIOV.VFS.ID]
+        if vf_id in vf_ids_to_add:
+            vf_already_exists = False
```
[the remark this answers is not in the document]
ffmancera: I am doing this because when editting we need to duplicate the existing VF object, then remove it from the list of VFs and add the one that we have duplicated. If we do not do that, we can generate some unstable situations (random fails).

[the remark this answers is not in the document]
ffmancera: But I am going to split this in different functions to make it easier! Thanks!

libnmstate/nm/sriov.py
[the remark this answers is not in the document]
ffmancera: What is wrong with `vfs_config`?

libnmstate/nm/sriov.py

libnmstate/nm/sriov.py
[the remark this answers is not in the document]

libnmstate/nm/sriov.py
```diff
@@ -43,11 +43,79 @@ def create_setting(iface_state, base_con_profile):
...
+
+        vf_object = nmclient.NM.SriovVF.new(vf_config[Ethernet.SRIOV.VFS.ID])
+
+        if vf_config.get(Ethernet.SRIOV.VFS.MAC_ADDRESS, None) is not None:
+            vf_object.set_attribute(
+                nmclient.NM.SRIOV_VF_ATTRIBUTE_MAC,
+                nmclient.GLib.Variant.new_string(
+                    vf_config[Ethernet.SRIOV.VFS.MAC_ADDRESS]
+                ),
+            )
```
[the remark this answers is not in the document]
ffmancera: That helps a lot, thanks!

tests/integration/sriov_test.py
```diff
@@ -61,6 +61,47 @@ def test_sriov_decrease_vfs(sriov_interface):
...
+    raises=NmstateNotSupportedError,
+    reason="The device does not support SR-IOV.",
+)
+def test_sriov_create_vf_config(sriov_interface):
+    eth_config = sriov_interface[Interface.KEY][0][Ethernet.CONFIG_SUBTREE]
+    eth_config[Ethernet.SRIOV_SUBTREE][Ethernet.SRIOV.TOTAL_VFS] = 2
+    libnmstate.apply(sriov_interface)
```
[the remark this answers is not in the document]
ffmancera: Yes, done!

tests/integration/sriov_test.py
```diff
@@ -61,6 +61,47 @@ def test_sriov_decrease_vfs(sriov_interface):
...
+    eth_config[Ethernet.SRIOV_SUBTREE][Ethernet.SRIOV.TOTAL_VFS] = 2
+    eth_config[Ethernet.SRIOV_SUBTREE][Ethernet.SRIOV.VFS_SUBTREE] = [
+        {
+            Ethernet.SRIOV.VFS.ID: 0,
+            Ethernet.SRIOV.VFS.SPOOF_CHECK: True,
+            Ethernet.SRIOV.VFS.MAC_ADDRESS: "00:11:22:33:44:55",
+            Ethernet.SRIOV.VFS.TRUST: False,
+        }
+    ]
+    libnmstate.apply(sriov_interface)
```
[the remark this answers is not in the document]
ffmancera: ACK to the fixture issue and fixed. About the bug, I'm not sure if that situation can happen. I will investigate and discuss it further. Thanks!

libnmstate/nm/sriov.py
```diff
@@ -43,11 +90,67 @@ def create_setting(iface_state, base_con_profile):
...
+        vf_object = nmclient.NM.SriovVF.new(vf_config[Ethernet.SRIOV.VFS.ID])
+        for key, val in vf_config.items():
+            _set_nm_attribute(vf_object, key, val)
+
+        sriov_setting.remove_vf_by_index(vf_id)
+        sriov_setting.add_vf(vf_object)
+
+
+def _set_nm_attribute(vf_object, key, value):
+    if key != Ethernet.SRIOV.VFS.ID:
```
EdDev: Can you pop this out once per `vf_config` at the caller? It looks wasteful to ask this for each attribute.

libnmstate/nm/sriov.py
```diff
@@ -43,11 +90,67 @@ def create_setting(iface_state, base_con_profile):
         if not sriov_setting:
             sriov_setting = nmclient.NM.SettingSriov.new()
 
+        vfs_config = sriov_config.get(Ethernet.SRIOV.VFS_SUBTREE, [])
+        vf_object_ids = {vf.get_index() for vf in sriov_setting.props.vfs}
+        vf_config_ids = {
+            vf_config[Ethernet.SRIOV.VFS.ID] for vf_config in vfs_config
```
EdDev: Seems that there is a hidden assumption that the user/client includes this ID in the desired state.
I think we need to make sure it is included, and if it is not, to raise an error. Seems to fit the validator module.

libnmstate/nm/sriov.py
```diff
@@ -43,11 +90,67 @@ def create_setting(iface_state, base_con_profile):
...
+        vf_config
+        for vf_config in vfs_config
+        if vf_config[Ethernet.SRIOV.VFS.ID] in vf_ids_to_add
+    )
+    for vf_config in vfs_config_to_add:
+        vf_object = nmclient.NM.SriovVF.new(vf_config[Ethernet.SRIOV.VFS.ID])
+        for key, val in vf_config.items():
+            _set_nm_attribute(vf_object, key, val)
+
+        sriov_setting.add_vf(vf_object)
```
EdDev: This is now mutable and we can make it immutable by making this function a generator:
```
yield vf_object
```
Then, doing something like this at the caller:
```
for vf_object in _create_sriov_new_vfs_objects():
    sriov_setting.add_vf(vf_object)
```

libnmstate/nm/sriov.py
```diff
@@ -43,11 +90,67 @@ def create_setting(iface_state, base_con_profile):
         if not sriov_setting:
             sriov_setting = nmclient.NM.SettingSriov.new()
 
+        vfs_config = sriov_config.get(Ethernet.SRIOV.VFS_SUBTREE, [])
+        vf_object_ids = {vf.get_index() for vf in sriov_setting.props.vfs}
+        vf_config_ids = {
+            vf_config[Ethernet.SRIOV.VFS.ID] for vf_config in vfs_config
+        }
+        _edit_sriov_vfs_in_setting(
```
EdDev: I think you can call first remove and then create to do the same.

libnmstate/nm/sriov.py
EdDev: It is enforced in the schema.

libnmstate/nm/sriov.py
```diff
@@ -175,9 +175,11 @@ def get_info(device):
     )
 
     if sriov_setting:
-        info[Ethernet.SRIOV_SUBTREE] = {
-            Ethernet.SRIOV.TOTAL_VFS: sriov_setting.props.total_vfs
-        }
+        sriov_config = {}
+        ifname = device.get_iface()
+        with open(f"/sys/class/net/{ifname}/device/sriov_numvfs", "r") as f:
```
EdDev: `"r"` is the default.

libnmstate/nm/sriov.py
ffmancera: Done! Thanks!

libnmstate/nm/sriov.py
ffmancera: Done in create and remove functions :-)

libnmstate/nm/sriov.py
ffmancera: As we need to remove the in order to edit. First, I am removing all the vf_objects and then creating all the vfs from the config. Thanks!

libnmstate/nm/sriov.py
ffmancera: Yes, thanks! Done!

libnmstate/nm/sriov.py
```diff
@@ -43,11 +90,54 @@ def create_setting(iface_state, base_con_profile):
         if not sriov_setting:
             sriov_setting = nmclient.NM.SettingSriov.new()
 
+        vfs_config = sriov_config.get(Ethernet.SRIOV.VFS_SUBTREE, [])
+        vf_object_ids = {vf.get_index() for vf in sriov_setting.props.vfs}
+        vf_config_ids = {
+            vf_config[Ethernet.SRIOV.VFS.ID] for vf_config in vfs_config
+        }
+
+        # As the user must do full edit of vfs, nmstate is deleting all the vfs
```
EdDev: Please specify in the commit message this limitation: "Currently, only full editation of VF/s is supports. Any VF/s that are emitted from the desired state are therefore implicitly initialized."

libnmstate/nm/sriov.py
```diff
@@ -75,5 +165,31 @@ def get_info(device):
         info[Ethernet.SRIOV_SUBTREE] = {
             Ethernet.SRIOV.TOTAL_VFS: sriov_setting.props.total_vfs
         }
+        vfs_config = _get_info_sriov_vfs_config(sriov_setting)
```
EdDev: This point also needs to be mentioned in the commit message.
It is an exception to report the config and not the running state.

Something like: "Current implementation reports the VF/s config state (from NM) and not the kernel state."

libnmstate/nm/sriov.py
ffmancera: Done, thank you!

libnmstate/nm/sriov.py
```diff
@@ -72,8 +162,36 @@ def get_info(device):
     )
 
     if sriov_setting:
-        info[Ethernet.SRIOV_SUBTREE] = {
-            Ethernet.SRIOV.TOTAL_VFS: sriov_setting.props.total_vfs
-        }
+        sriov_config = {}
+        ifname = device.get_iface()
+        with open(f"/sys/class/net/{ifname}/device/sriov_numvfs") as f:
```
cathay4t: As we promised in API document: https://nmstate.github.io/devel/api.html#error-handling

A `try exception` should be used here to wrap unexpected error  to `NmstateInternalError`

ffmancera: Done, the commit have been moved to #670

EdDev:
@cathay4t , seems all your comments have been addressed.
If I missed something, please let me and @ffmancera know and we will make sure it will be fixed.

PR #649, opened 2019-12-13
Better interface deletion check

tyll:
This should remove two errors we are seeing with the NM 1.22 and master branch for the bond tests:

- `nm-device-error-quark: This device is not a software device or is not realized (5)` 
- `error=nm-client-error-quark: Object is no longer in the client cache (0)`

It appears to be a race condition, sometimes the test just passes, sometimes it warns about using `set_managed` and sometimes one of those errors are triggered.
I ran the new test 100 times as per @cathay4t wishes.

One issue that I found is that the old error handling code was not covered by any tests. I did not manage to figure out how to properly add a test for it as well because of the special mainloop approach in Nmstate. I envision the test to be something like:

Setup:
- create a bond interface
- get the NMDevice for the bond interface
- deactivate the bond interface (variations: Delete the profile, delete the device via NM, delete the device with ip link)
- wait till the interface is gone

Test:
- Try to remove the interface with delete_device
- Check that there is no exception, maybe check the logs, check that the interface is actually gone

### DIFF
--- a/libnmstate/nm/device.py
+++ b/libnmstate/nm/device.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2018-2019 Red Hat, Inc.
+# Copyright (c) 2018-2020 Red Hat, Inc.
 #
 # This file is part of nmstate
 #
@@ -173,62 +173,37 @@ def delete_device(nmdev):
 
 def _safe_delete_device_async(nmdev):
     mainloop = nmclient.mainloop()
-    user_data = mainloop, nmdev
+    user_data = mainloop, nmdev, nmdev.get_iface()
     nmdev.delete_async(
         mainloop.cancellable, _delete_device_callback, user_data
     )
 
 
 def _delete_device_callback(src_object, result, user_data):
-    mainloop, nmdev = user_data
+    mainloop, nmdev, iface = user_data
+    error = None
     try:
-        success = src_object.delete_finish(result)
-    # pylint: disable=catching-non-exception
-    except nmclient.GLib.GError as e:
-        # pylint: enable=catching-non-exception
-        if e.matches(
-            nmclient.Gio.DBusError.quark(),
-            nmclient.Gio.DBusError.UNKNOWN_METHOD,
-        ) or (
-            e.matches(
-                nmclient.NM.DeviceError.quark(),
-                nmclient.NM.DeviceError.NOTSOFTWARE,
-            )
-            and nmdev.is_software()
-        ):
-            logging.debug(
-                "Device %s has been already deleted: error=%s",
-                nmdev.get_iface(),
-                e,
-            )
-            mainloop.execute_next_action()
-        else:
-            mainloop.quit(
-                "Device deletion failed on {}: error={}".format(
-                    nmdev.get_iface(), e
-                )
-            )
-        return
+        src_object.delete_finish(result)
     except Exception as e:
-        if mainloop.is_action_canceled(e):
+        error = e
+        if mainloop.is_action_canceled(error):
             logging.debug(
-                "Device deletion aborted on %s: error=%s", nmdev.get_iface(), e
+                "Device deletion aborted on %s: error=%s", iface, error
             )
-        else:
-            mainloop.quit(
-                "Device deletion failed on {}: error={}".format(
-                    nmdev.get_iface(), e
-                )
-            )
-        return
+            return
 
-    devname = src_object.get_iface()
-    if success:
-        logging.debug("Device deletion succeeded: dev=%s", devname)
+    if not nmdev.is_real():
+        logging.debug("Interface is not real anymore: iface=%s", iface)
+        if error:
+            logging.debug(
+                "Ignored error: %s", error,
+            )
+
         mainloop.execute_next_action()
     else:
         mainloop.quit(
-            "Device deletion failed: dev={}, error=unknown".format(devname)
+            f"Device deletion failed on {iface} ({nmdev.get_path()}): "
+            f"error={error or 'unknown'}"
         )
 
 


tests/integration/bond_test.py
```diff
@@ -99,6 +100,16 @@ def bond99_with_slave(eth2_up):
         yield state
 
 
+def test_remove_existing_bond(setup_remove_bond99):
+    subprocess.check_call(["ip", "link", "add", BOND99, "type", "bond"])
```
EdDev: Could you please use `from .testlib import cmd as libcmd` instead?
This also fits a fixture as it is not part of the test itself (it is a setup of the test).

tests/integration/bond_test.py
```diff
@@ -99,6 +100,16 @@ def bond99_with_slave(eth2_up):
...
+    assertlib.assert_present(BOND99, timeout=1)
+
+    with bond_interface(BOND99, [], create=False):
```
EdDev: This does nothing except creating the state (`create=False` means it is not applied)

tests/integration/testlib/assertlib.py
```diff
@@ -35,13 +36,38 @@ def assert_state(desired_state_data):
     assert desired_state.state == current_state.state
 
 
-def assert_absent(*ifnames):
+def assert_absent(*ifnames, timeout=0):
     """ Assert that a interface is not present in the current state """
 
-    current_state = statelib.show_only(ifnames)
+    now = time.time()
+    while True:
```
EdDev: I would prefer the retry part be separated from the assert.
It can then be shared between asserts.

I can share one example of this from the VDSM project:
This is a decorator that can be used on an assert helper.
```
def retry_assert(func):
    def retry(*args, **kwargs):
        for i in range(4):
            try:
                func(*args, **kwargs)
            except AssertionError:
                time.sleep(1)
            else:
                return
        func(*args, **kwargs)

    return retry
```

libnmstate/nm/device.py
```diff
@@ -173,62 +173,39 @@ def delete_device(nmdev):
...
-        else:
-            mainloop.quit(
-                "Device deletion failed on {}: error={}".format(
-                    nmdev.get_iface(), e
-                )
-            )
-        return
+        src_object.delete_finish(result)
     except Exception as e:
+        error = e
```
cathay4t: You may just use `except Exception as error`.

libnmstate/nm/device.py
```diff
@@ -173,62 +173,39 @@ def delete_device(nmdev):
 
 def _safe_delete_device_async(nmdev):
     mainloop = nmclient.mainloop()
-    user_data = mainloop, nmdev
+    user_data = mainloop, nmdev, nmdev.get_iface()
```
cathay4t: As time of this function been called, `nmdev.get_iface()` could be None.

You may try `assert nmdev.get_iface()` in the repeat calls of bond creation and deletion.

libnmstate/nm/device.py
tyll: Thank you, good catch!

tests/integration/bond_test.py
tyll: What is the benefit of using `cmd` here?

tests/integration/bond_test.py
tyll: AFAIU it removes the bond when the context is left, does it not?

libnmstate/nm/device.py
tyll: The interface is only used for logging, so this does not add extra errors AFAIU. Also AFAIU this code is called before the mainloop is started, therefore it is strange that the interface could be `None`. IMHO the proper solution would be to pass  a static context that contains the information about which state is processed here (the original state, the desired state, the interface that is changed, ...) instead of only using changing NM objects. But this would be a bigger change in Nmstate.

cathay4t: The interface became None as previous mainloop run might enter a partial updated `NM.Device`. Normally, profile deletion will delete the soft interface as well, so then the mainloop of profile deletion quit, the `NM.Device` might be in the middle of deletion.

cathay4t:
@tyll The CI failed on:

```

  File "/usr/lib/python3.6/site-packages/libnmstate/nm/device.py", line 197, in _delete_device_callback

    if error:

UnboundLocalError: local variable 'error' referenced before assignment
```

tyll:
> @tyll The CI failed on:
> 
> ```
> Traceback (most recent call last):
> 
>   File "/usr/lib/python3.6/site-packages/libnmstate/nm/device.py", line 197, in _delete_device_callback
> 
>     if error:
> 
> UnboundLocalError: local variable 'error' referenced before assignment
> ```

Yes, I saw this, too. I do not understand why, the variable is initialized in line 184...

libnmstate/nm/device.py
tyll: Actually, this does not work since it removes `error` from the outer scope, test script:

```python
def test_scope():
    error = "error"
    try:
        raise Exception()
    except Exception as error:
        pass
    print(repr(error))

test_scope()
```

tyll:
> Commit `Integration tests/bond: Test deleting existing bond `: The addition of the retry assertion needs to be explained a bit. A future reader may not understand why it was done.
> 
> Commit `nm.device: Deletion: Assume removed interface as success`: This is a heavier change and may require some discussion. It also touches some work that @cathay4t was involved in about the `nmclient`.
> I would prefer this commit be extracted to its own PR.

I moved the tests to #679, so this is now only the functional change

tests/integration/bond_test.py
EdDev: Right.

libnmstate/nm/device.py
cathay4t: Then use another name instead of overriding `error` variable.

libnmstate/nm/device.py
```diff
@@ -173,62 +173,41 @@ def delete_device(nmdev):
...
+                "Device deletion aborted on %s: error=%s", iface, error
             )
-        else:
-            mainloop.quit(
-                "Device deletion failed on {}: error={}".format(
-                    nmdev.get_iface(), e
-                )
+            return
+
+    mainloop.iterate_all()
```
EdDev: This requires an explanation, probably in the docttext of that function.

tyll: IMHO the commit message fits better because it explains the other code change as well.

tyll: I added it to the commit message.

libnmstate/nm/nmclient.py
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2018-2019 Red Hat, Inc.
+# Copyright (c) 2018-2020 Red Hat, Inc.
```
EdDev: Not needed anymore.

libnmstate/nm/device.py
```diff
@@ -173,62 +173,40 @@ def delete_device(nmdev):
...
-        logging.debug("Device deletion succeeded: dev=%s", devname)
         mainloop.execute_next_action()
     else:
+        if not error:
+            error = "unknown"
+
         mainloop.quit(
-            "Device deletion failed: dev={}, error=unknown".format(devname)
+            f"Device deletion failed on {iface} ({nmdev.get_path()}): "
+            f"error={error}"
```
EdDev: You could do `error or "unknown"`

libnmstate/nm/device.py
tyll: I removed this for now, maybe it is not needed. My debugging session was so long ago, I do not remember all the details.

libnmstate/nm/device.py
tyll: ok

libnmstate/nm/nmclient.py
tyll: hope it is gone now

PR #653, opened 2019-12-19
schema: Fix MAC address pattern

ffmancera:
Infiniband MAC addresses have the following format:

80:00:02:08:FE:80:00:00:00:00:00:00:F4:52:14:03:00:61:AA:91

This patch change the schema "mac-address" pattern in order to match
this format too.

### DIFF
--- a/libnmstate/schemas/operational-state.yaml
+++ b/libnmstate/schemas/operational-state.yaml
@@ -63,7 +63,7 @@ definitions:
         - down
     mac-address:
       type: string
-      pattern: "^([a-fA-F0-9]{2}:){5}[a-fA-F0-9]{2}$"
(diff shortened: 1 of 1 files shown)


cathay4t:
Can you explain why 40 hex chars is enough for IB? (any IB standard specification or linux kernel code confirmed it?)

In linux kernel the `MAX_ADDR_LEN` is 32 char/u8, so I think change the schema to support 64 hex chars is better.

ffmancera:
> Can you explain why 40 hex chars is enough for IB? (any IB standard specification or linux kernel code confirmed it?)
> 
> In linux kernel the `MAX_ADDR_LEN` is 32 char/u8, so I think change the schema to support 64 hex chars is better.

You are right, maximum MAX_ADDR_LEN is 32 char/u8 so.. fixed! Now the schema supports 64 hex chars. Thanks!

PR #658, opened 2019-12-30
nm.team: Team interface basic implementation

ffmancera:

### DIFF
--- a/libnmstate/nm/team.py
+++ b/libnmstate/nm/team.py
@@ -17,14 +17,99 @@
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 #
 
-from libnmstate.error import NmstateNotImplementedError
+import json
+
+from libnmstate.nm import connection as nm_connection
+from libnmstate.nm import nmclient
+from libnmstate.schema import Interface
 from libnmstate.schema import Team
 
 
-def create_setting(iface_state):
-    setting = None
-    team_config = iface_state.get(Team.CONFIG_SUBTREE)
-    if team_config:
-        raise NmstateNotImplementedError
+TEAMD_JSON_DEVICE = "device"
 
-    return setting
+
+def create_setting(iface_state, base_con_profile):
+    team_setting = None
+    team_config = iface_state.get(Team.CONFIG_SUBTREE)
+
+    if not team_config:
+        return None
+
+    if base_con_profile:
+        team_setting = base_con_profile.get_setting_duplicate(
+            nmclient.NM.SETTING_TEAM_SETTING_NAME
+        )
+
+    if not team_setting:
+        team_setting = nmclient.NM.SettingTeam.new()
+
+    teamd_config = _convert_team_config_to_teamd_format(
+        team_config, iface_state[Interface.NAME]
+    )
+
+    team_setting.props.config = json.dumps(teamd_config)
+
+    return team_setting
+
+
+def _convert_team_config_to_teamd_format(team_config, ifname):
+    team_config[TEAMD_JSON_DEVICE] = ifname
+
+    team_ports = team_config.get(Team.PORT_SUBTREE, ())
+    team_ports_formatted = {
+        port[Team.Port.NAME]: _dict_key_filter(port, Team.Port.NAME)
+        for port in team_ports
+    }
+    team_config[Team.PORT_SUBTREE] = team_ports_formatted
+
+    return team_config
+
+
+def _dict_key_filter(dict_to_filter, key):
+    return dict(filter(lambda elem: elem[0] == key, dict_to_filter.items()))
+
+
+def get_info(device):
+    """
+    Provide the current active teamd values for an interface.
+    Ref: https://bugzilla.redhat.com/1792232
+    """
+    info = {}
+
+    connection = nm_connection.ConnectionProfile()
+    connection.import_by_device(device)
+    if not connection.profile:
+        return info
+
+    team_setting = connection.profile.get_setting_by_name(
+        nmclient.NM.SETTING_TEAM_SETTING_NAME
+    )
+
+    if team_setting:
+        teamd_json = team_setting.get_config()
+        if teamd_json:
+            teamd_config = json.loads(teamd_json)
+            team_config = _convert_teamd_config_to_nmstate_config(teamd_config)
+            info[Team.CONFIG_SUBTREE] = team_config
+
+    return info
+
+
+def _convert_teamd_config_to_nmstate_config(team_config):
+    team_config.pop("device")
+    port_config = team_config.get(Team.PORT_SUBTREE)
+
+    if port_config:
+        team_port = _teamd_port_to_nmstate_port(port_config)
+        team_config[Team.PORT_SUBTREE] = team_port
+
+    return team_config
+
+
+def _teamd_port_to_nmstate_port(port_config):
+    port_list = []
+    for name, port in port_config.items():
+        port.update({Team.Port.NAME: name})
+        port_list.append(port)
+
+    return port_list
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -25,6 +25,7 @@ import pytest
 
 import libnmstate
 
+from .testlib import cmdlib
 from .testlib import ifacelib
 
 
@@ -98,3 +99,10 @@ def _get_osname():
             if line.startswith("PRETTY_NAME="):
                 return line.split("=", maxsplit=1)[1].strip().strip('"')
     return ""
+
+
+@pytest.fixture(scope="session", autouse=True)
+def download_nm_team_plugin():
+    cmdlib.exec_cmd(
+        ("dnf", "install", "-y", "--downloadonly", "NetworkManager-team",)
+    )
--- a/tests/integration/team_test.py
+++ b/tests/integration/team_test.py
@@ -17,34 +17,115 @@
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 #
 
+from contextlib import contextmanager
+import os
+
 import pytest
 
 import libnmstate
-from libnmstate.error import NmstateNotImplementedError
+from libnmstate.error import NmstateLibnmError
 from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceState
 from libnmstate.schema import InterfaceType
 from libnmstate.schema import Team
 
+from .testlib import assertlib
+from .testlib import cmdlib
+
+
+DNF_INSTALL_NM_TEAM_PLUGIN_CMD = (
+    "dnf",
+    "install",
+    "-y",
+    "--cacheonly",
+    "NetworkManager-team",
+)
+
+DNF_REMOVE_NM_TEAM_PLUGIN_CMD = (
+    "dnf",
+    "remove",
+    "-y",
+    "-q",
+    "NetworkManager-team",
+)
+
+SYSTEMCTL_RESTART_NM_CMD = ("systemctl", "restart", "NetworkManager")
 
 TEAM0 = "team0"
 
-TEAM0_STATE = {
-    Interface.KEY: [
-        {
-            Interface.NAME: TEAM0,
-            Interface.TYPE: InterfaceType.TEAM,
-            Team.CONFIG_SUBTREE: {
-                Team.PORT_SUBTREE: [{Team.Port.NAME: "eth1"}]
-            },
-        }
-    ]
-}
 
-
-@pytest.mark.xfail(
-    raises=NmstateNotImplementedError,
-    reason="Team interface is not supported yet",
-    strict=True,
+pytestmark = pytest.mark.skipif(
+    os.environ.get("CI") == "true",
+    reason="Team kmod not available in Travis CI",
 )
-def test_sriov_not_implemented():
-    libnmstate.apply(TEAM0_STATE)
+
+
+def test_create_team_iface():
+    with team_interface(TEAM0) as team_state:
+        assertlib.assert_state(team_state)
+
+
+def test_edit_team_iface():
+    with team_interface(TEAM0) as team_state:
+        team_state[Interface.KEY][0][Team.CONFIG_SUBTREE] = {
+            Team.RUNNER_SUBTREE: {
+                Team.Runner.NAME: Team.Runner.RunnerMode.LOAD_BALANCE
+            },
+            Team.PORT_SUBTREE: [{Team.Port.NAME: "eth1"}],
+        }
+        libnmstate.apply(team_state)
+        assertlib.assert_state(team_state)
+
+
+def test_nm_team_plugin_missing():
+    with disable_nm_team_plugin():
+        with pytest.raises(NmstateLibnmError):
+            libnmstate.apply(
+                {
+                    Interface.KEY: [
+                        {
+                            Interface.NAME: TEAM0,
+                            Interface.TYPE: InterfaceType.TEAM,
+                            Interface.STATE: InterfaceState.UP,
+                        }
+                    ]
+                }
+            )
+
+
+@contextmanager
+def team_interface(ifname):
+    desired_state = {
+        Interface.KEY: [
+            {
+                Interface.NAME: ifname,
+                Interface.TYPE: InterfaceType.TEAM,
+                Interface.STATE: InterfaceState.UP,
+            }
+        ]
+    }
+    libnmstate.apply(desired_state)
+    try:
+        yield desired_state
+    finally:
+        libnmstate.apply(
+            {
+                Interface.KEY: [
+                    {
+                        Interface.NAME: ifname,
+                        Interface.STATE: InterfaceState.ABSENT,
+                    }
+                ]
+            }
+        )
+
+
+@contextmanager
+def disable_nm_team_plugin():
+    cmdlib.exec_cmd(DNF_REMOVE_NM_TEAM_PLUGIN_CMD)
+    cmdlib.exec_cmd(SYSTEMCTL_RESTART_NM_CMD)
+    try:
+        yield
+    finally:
+        cmdlib.exec_cmd(DNF_INSTALL_NM_TEAM_PLUGIN_CMD)
+        cmdlib.exec_cmd(SYSTEMCTL_RESTART_NM_CMD)
--- a/automation/run-tests.sh
+++ b/automation/run-tests.sh
@@ -283,7 +283,7 @@ if [[ "$CI" == "true" ]];then
 fi
 
 mkdir -p $EXPORT_DIR
-CONTAINER_ID="$(${CONTAINER_CMD} run --privileged -d -v /sys/fs/cgroup:/sys/fs/cgroup:ro -v $PROJECT_PATH:$CONTAINER_WORKSPACE -v $EXPORT_DIR:$CONT_EXPORT_DIR $CONTAINER_IMAGE)"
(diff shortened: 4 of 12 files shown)


ffmancera:
This is failing because we need to update the docker image in docker.io

ffmancera:
The CI is failing because it depends on #667.

libnmstate/nm/team.py
```diff
@@ -17,14 +17,81 @@
...
 
 
-def create_setting(iface_state):
-    setting = None
+def create_setting(iface_state, base_con_profile):
+    team_setting = None
     team_config = iface_state.get(Team.CONFIG_SUBTREE)
-    if team_config:
-        raise NmstateNotImplementedError
+    if base_con_profile:
```
EdDev: This means that you can never create a new team, you can just edit an existing one.
Or do I miss something here?

libnmstate/nm/team.py
```diff
@@ -17,14 +17,81 @@
...
+        if team_config:
+            team_setting = base_con_profile.get_setting_duplicate(
+                nmclient.NM.SETTING_TEAM_SETTING_NAME
+            )
+
+            if not team_setting:
+                team_setting = nmclient.NM.SettingTeam.new()
+
+            team_config = _convert_team_config_to_setting_format(
```
EdDev: The function name describes  a conversion between config to setting, but the result is assigned to config.
I'm confused.

libnmstate/nm/team.py
```diff
@@ -17,14 +17,81 @@
...
+                team_config, iface_state.get(Interface.NAME)
```
EdDev: I do not think the `name` value can be absent at this stage.

libnmstate/nm/team.py
```diff
@@ -17,14 +17,81 @@
...
+            )
+
+            team_setting.props.config = json.dumps(team_config)
+
+    return team_setting
+
+
+def _convert_team_config_to_setting_format(team_config, ifname):
+    team_config["device"] = ifname
```
EdDev: Please give `"device"` a constant and explain where it came from. (By explain, I mean give it a good name)

libnmstate/nm/team.py
```diff
@@ -17,14 +17,81 @@
...
+
+    if team_config.get(Team.PORT_SUBTREE):
```
EdDev: You can fetch it once into a variable so you will not need to fetch it twice.

libnmstate/nm/team.py
```diff
@@ -17,14 +17,81 @@
...
+        team_port = {
+            port.pop(Team.Port.NAME): port
```
EdDev: Ohh my... it looks like `d[i++] = [++i]`

It is neat, but the reader may raise :)
How about something less neat like this:
```
port[Team.Port.NAME]: _dict_key_filter(port, Team.Port.NAME)
```

If you can avoid a mutable operation, prefer it over the immutable solution.

I won't fight over this, so if you or others have strong feeling to keep this neat, let me know.

libnmstate/nm/team.py
```diff
@@ -17,14 +17,81 @@
...
+        team_json = team_setting.get_config()
+        if team_json:
+            team_config = json.loads(team_json)
+            team_config = _convert_team_config_to_config_format(team_config)
+            info[Team.CONFIG_SUBTREE] = team_config
+
+    return info
+
+
+def _convert_team_config_to_config_format(team_config):
```
EdDev: "config to config format"?

libnmstate/nm/team.py
ffmancera: Oops. I missed to amend some changes into the commit. Fixed :-)

libnmstate/nm/team.py
ffmancera: Yes, it is a little bit confusing. I renamed it to `_convert_team_config_to_teamd_format`.

libnmstate/nm/team.py
ffmancera: Yes, you are right. Fixed!

libnmstate/nm/team.py
ffmancera: I added `TEAMD_JSON_DEVICE` constant. Thanks!

libnmstate/nm/team.py
ffmancera: Done, thanks!

libnmstate/nm/team.py
ffmancera: I think you are right. That is better. Done and thanks!

libnmstate/nm/team.py
ffmancera: Oops. This name is confusing again. I renamed it to `_convert_teamd_config_to_nmstate_config`. Thanks!

tests/integration/team_test.py
```diff
@@ -17,34 +17,88 @@
...
 from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceState
 from libnmstate.schema import InterfaceType
 from libnmstate.schema import Team
 
+from .testlib import assertlib
+from .testlib import cmd as libcmd
+
+
+DNF_REMOVE_NM_TEAM_CMD = ("dnf", "remove", "-y", "-q", "NetworkManager-team")
```
cathay4t: When I run the test in my VM, since nmstate will lose default gateway(old long story), the `dnf` command will fail to
install the package back.

Can you try to move the so file to /tmp and move it back after done?

```
mv `rpm -ql NetworkManager-team |egrep so$` /tmp
```

tests/integration/team_test.py
```diff
@@ -17,34 +17,88 @@
...
 
-TEAM0_STATE = {
-    Interface.KEY: [
-        {
-            Interface.NAME: TEAM0,
-            Interface.TYPE: InterfaceType.TEAM,
-            Team.CONFIG_SUBTREE: {
-                Team.PORT_SUBTREE: [{Team.Port.NAME: "eth1"}]
+
+@pytest.mark.xfail(raises=NmstateLibnmError, reason="Team kmod not installed")
```
cathay4t: This is not xfail, you should ask use to modprobe team.

And also do modprobe in `automation/run-tests.sh` like `function modprobe_ovs` did.

tyll: There is no team kernel module for the Travis Ubuntu kernel, so it cannot be loaded there.

ffmancera: As it is not possible to do have team kernel module for the Travis Ubuntu kernel, I am going to add a xskip with a condition on $CI. Does that work for you? Thanks

tests/integration/team_test.py
ffmancera: This is not working. If I do that the test is not failing.

tests/integration/team_test.py
ffmancera: Done!

tests/integration/team_test.py
ffmancera: Done, thanks!

libnmstate/nm/team.py
```diff
@@ -17,14 +17,94 @@
...
+    return team_config
+
+
+def _dict_key_filter(dict_to_filter, key):
+    return dict(filter(lambda elem: elem[0] == key, dict_to_filter.items()))
+
+
+def get_info(device):
+    """
+    Provide the current active teamd values for an interface.
```
EdDev: Please ref a ticket here for the missing kernel state.

libnmstate/nm/team.py
```diff
@@ -17,14 +17,94 @@
...
+
+    return info
+
+
+def _convert_teamd_config_to_nmstate_config(team_config):
+    team_config.pop("device")
+
+    if team_config.get(Team.PORT_SUBTREE):
+        team_port = [
+            port.update({Team.Port.NAME: name})
```
EdDev: This seems wrong. Update returns None.
Please check why the tests did not detected this.

libnmstate/nm/team.py
```diff
@@ -17,14 +17,94 @@
...
+            team_config = _convert_teamd_config_to_nmstate_config(teamd_config)
+            info[Team.CONFIG_SUBTREE] = team_config
+
+    return info
+
+
+def _convert_teamd_config_to_nmstate_config(team_config):
+    team_config.pop("device")
+
+    if team_config.get(Team.PORT_SUBTREE):
```
EdDev: Please extract it to a variable as you do this fetch twice.

tests/integration/team_test.py
```diff
@@ -17,34 +17,106 @@
...
+    with team_interface(TEAM0) as team_state:
+        team_state[Team.CONFIG_SUBTREE] = {
+            Team.RUNNER_SUBTREE: {
+                Team.Runner.NAME: Team.Runner.RunnerMode.LOAD_BALANCE
             },
+            Team.PORT_SUBTREE: [{Team.Port.NAME: "eth1"}],
         }
-    ]
-}
+        libnmstate.apply(team_state)
```
EdDev: What about asserting that the new config is in?

tests/integration/team_test.py
```diff
@@ -17,34 +17,106 @@
...
+                Interface.TYPE: InterfaceType.TEAM,
+                Interface.STATE: InterfaceState.UP,
+            }
+        ]
+    }
+    libnmstate.apply(desired_state)
+    try:
+        yield desired_state
+    finally:
+        team_dstate = desired_state[Interface.KEY][0]
```
EdDev: The user of the fixture may have done changes to the desired_state, it is safer to build the absent command directly with the original name (ifname).

tests/integration/team_test.py
```diff
@@ -17,34 +17,106 @@
...
+def disable_nm_team_plugin():
+    rpm_team_plugin = libcmd.exec_cmd(RPM_NM_TEAM_PLUGIN_PATH)[1]
+    plugin_path = re.search(r".*so", rpm_team_plugin).group()
+    disable_team_plugin = (
+        "mv",
+        f"{plugin_path}",
+        "/tmp/",
+    )
+    libcmd.exec_cmd(disable_team_plugin)
+    libcmd.exec_cmd(SYSTEMCTL_RESTART_NM_CMD)
```
EdDev: If this fails, no restoration occurs (finally).

tests/integration/team_test.py
```diff
@@ -17,34 +17,106 @@
...
+    try:
+        yield desired_state
+    finally:
+        team_dstate = desired_state[Interface.KEY][0]
+        team_dstate[Interface.STATE] = InterfaceState.ABSENT
+        libnmstate.apply(desired_state)
+
+
+@contextmanager
+def disable_nm_team_plugin():
```
EdDev: This context test is destructive, I would just either uninstall the rpm if possible or allow running it only in a container.

ffmancera: This change was requested by @cathay4t, I prefer just to uninstall the rpm.. how can we check that we are in a container? And is @cathay4t okay with that?

tests/integration/team_test.py
```diff
@@ -17,34 +17,106 @@
...
+    rpm_team_plugin = libcmd.exec_cmd(RPM_NM_TEAM_PLUGIN_PATH)[1]
+    plugin_path = re.search(r".*so", rpm_team_plugin).group()
+    disable_team_plugin = (
```
tyll: In the future (not necessarily here, since this might need to be rewritten) please use `os.rename()`

tests/integration/team_test.py
```diff
@@ -17,34 +17,106 @@
...
+        "mv",
+        f"{plugin_path}",
+        "/tmp/",
+    )
+    libcmd.exec_cmd(disable_team_plugin)
+    libcmd.exec_cmd(SYSTEMCTL_RESTART_NM_CMD)
+    try:
+        yield
+    finally:
+        ls_tmp = libcmd.exec_cmd(("ls", "/tmp/"))[1]
```
tyll: Use `os.listdir()`

tests/integration/team_test.py
```diff
@@ -17,34 +17,106 @@
...
+        plugin_filename = re.search(r".*so", ls_tmp).group()
```
tyll: Actually, just use `glob.glob("/tmp/*.so")`, no need for listdir.

tests/integration/team_test.py
```diff
@@ -17,34 +17,106 @@
...
+        enable_team_plugin = ("mv", f"/tmp/{plugin_filename}", plugin_path)
```
tyll: This would be `os.rename()` again.

tests/integration/team_test.py
tyll: Gris' problem was that this breaks network connectivity and then he cannot re-install the RPM. Another workaround would be to run `dnf install --downloadonly NetworkManager-team` before breaking NM and installing the rpm again with `dnf install --cacheonly NetworkManager-team`.

tests/integration/team_test.py
tyll: Actually it would be better to copy the plugin to a temp file created with `tempfile.NamedTemporaryFile()` so that you do not need to do any globing that might fail on non CI systems.

tests/integration/team_test.py
ffmancera: I like this workaround. Does this work for you @cathay4t? Thanks!

cathay4t: `cacheonly` works for me.

libnmstate/nm/team.py
```diff
@@ -17,14 +17,94 @@
...
+
+    if base_con_profile:
+        team_setting = base_con_profile.get_setting_duplicate(
+            nmclient.NM.SETTING_TEAM_SETTING_NAME
+        )
+
+    if not team_setting:
+        team_setting = nmclient.NM.SettingTeam.new()
+
+    if team_config is not None:
```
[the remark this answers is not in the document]
ffmancera: Yep you are right, done!

libnmstate/nm/team.py
```diff
@@ -17,14 +17,94 @@
...
+
+        team_setting.props.config = json.dumps(teamd_config)
+
+    return team_setting
+
+
+def _convert_team_config_to_teamd_format(team_config, ifname):
+    team_config[TEAMD_JSON_DEVICE] = ifname
+
+    team_ports = team_config.get(Team.PORT_SUBTREE, None)
```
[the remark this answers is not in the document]
ffmancera: Done, thanks!

libnmstate/nm/team.py
ffmancera: Sure, done!

libnmstate/nm/team.py
ffmancera: Whoops, tests didn't detect this because there was a typo and the apply ignored the team subtree. Now both are fixed.

libnmstate/nm/team.py
ffmancera: Done, thanks!

tests/integration/team_test.py
ffmancera: I fixed a typo and also added the asserting. Thanks :)

tests/integration/team_test.py
ffmancera: Done, thanks!

tests/integration/team_test.py
ffmancera: I change the approach, I think you are going to like this most.

tests/integration/team_test.py
ffmancera: Done, we are now doing the `cacheonly` stuff. Thanks!

tests/integration/team_test.py
ffmancera: Not needed anymore.

tests/integration/team_test.py
ffmancera: Not needed anymore.

tests/integration/team_test.py
ffmancera: Not needed anymore.

tests/integration/team_test.py
ffmancera: Not needed anymore.

tests/integration/team_test.py
```diff
@@ -17,34 +17,123 @@
...
+from .testlib import cmd as libcmd
+
+
+DNF_DOWNLOAD_NM_TEAM_PLUGIN_CMD = (
+    "dnf",
+    "install",
+    "-y",
+    "--downloadonly",
+    "NetworkManager-team",
+)
```
cathay4t: Can we move this to `conftest.py` and make it run before all the test cases where network still works?

ffmancera: I am fine with moving it to conftest.py. I will also send a PR to fix the OVS missing plugin test.

ffmancera: Done, thanks! :-)

ffmancera:
Rebased! :-)

tests/integration/conftest.py
```diff
@@ -98,3 +99,10 @@ def _get_osname():
             if line.startswith("PRETTY_NAME="):
                 return line.split("=", maxsplit=1)[1].strip().strip('"')
     return ""
+
+
+def download_NM_plugin(plugin):
```
EdDev: Please use lowercase chars, even for `nm`.
But based on what the function does, I do not understand how it related to `NM` at all.
It just installs a package.
Therefore, it may fit something like: `install_package`.

tests/integration/conftest.py
```diff
@@ -98,3 +99,10 @@ def _get_osname():
...
+    cmdlib.exec_cmd(("dnf", "install", "-y", "--downloadonly", f"{plugin}",))
+
+
+download_NM_plugin("NetworkManager-team")
```
EdDev: If you meant to call it before the tests, it fits as a session fixture.

tests/integration/conftest.py
ffmancera: As it is fixture now, I have renamed it to `download_nm_team_plugin()`. Thanks!

tests/integration/conftest.py
ffmancera: Done! Thanks!

tests/integration/team_test.py
```diff
@@ -17,34 +17,117 @@
...
-TEAM0_STATE = {
-    Interface.KEY: [
-        {
-            Interface.NAME: TEAM0,
-            Interface.TYPE: InterfaceType.TEAM,
-            Team.CONFIG_SUBTREE: {
-                Team.PORT_SUBTREE: [{Team.Port.NAME: "eth1"}]
+
+@pytest.mark.skipif(
+    os.environ.get("CI") == "true",
```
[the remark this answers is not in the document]
ffmancera: I prefer to use the CI check. I am working in a better solution for this but I prefer to keep this for now. I have did the skipif for the whole class/module. Thanks!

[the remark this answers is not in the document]
tyll: No need to change this now but please consider to actually identify the CI with `CI=travis` for Travis and check for this. In the `run-tests.sh` nothing indicates that the CI variable is only set on Travis. It was also set on STDCI, when we still used it.

libnmstate/nm/team.py
```diff
@@ -17,14 +17,98 @@
...
+    return info
+
+
+def _convert_teamd_config_to_nmstate_config(team_config):
+    team_config.pop("device")
+    port_config = team_config.get(Team.PORT_SUBTREE)
+
+    if port_config:
+        team_port = [
+            port for port in _teamd_port_to_nmstate_port(port_config.items())
```
tyll: It seems to me that this could just be `list( _teamd_port_to_nmstate_port(port_config.items()))`. It also seems to me that it would be more natural to call the `.items()` method inside `_teamd_port_to_nmstate_port` and it could also just return a list if you convert the generator to a list right away. Are there some subtle requirements for the current method that I am missing?

EdDev: +1
`list(_teamd_port_to_nmstate_port(port_config))` is nicer.
Returning a list makes it a bit uglier IMO, but I do not really mind much.

ffmancera: Done, the function is returning a list now. Thanks!

PR #659, opened 2019-12-30
tests: add unit tests definitions for python 3.8

ffmancera:
This patch introduces python 3.8 support for unit tests.


ffmancera:
> Thanks!
> Don’t we need to run this on CI?

Yes, but we need to do that in another commit. If we do this here, it is going to fail as the docker container is not going to be updated. I am having the same issue in #658 where I need to update the docker container in order to test team interfaces :-)

ffmancera:
Rebased!

PR #660, opened 2019-12-31
metadata: Ignore existing unmanaged and unknown type ports

EdDev:
When editing bridges, ports which are not managed or have an unknown
type should not be consumed by nmstate.

The issue has been observed when tap devices have been connected to the
bridge and its editation caused nmstate to implicitly define them.


EdDev:
TODO: Add unit and integration tests.

PR #662, opened 2019-12-31
tests, metadata: Add Linux Bridge metadata unit tests

EdDev:


tests/lib/metadata_linux_bridge_test.py
```diff
@@ -0,0 +1,366 @@
+#
+# Copyright (c) 2019 Red Hat, Inc.
```
ffmancera: Everything looks fine to me. Just a nit: Should we use `Copyright (c) 2019 Red Hat, Inc.`?

EdDev: The header is identical to the others. Did I miss something?

tests/lib/metadata_linux_bridge_test.py
```diff
@@ -0,0 +1,366 @@
...
+
+from libnmstate import metadata
+from libnmstate import state
+from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceState
+from libnmstate.schema import InterfaceType
+from libnmstate.schema import LinuxBridge
+
+
+InterfaceType.UNKNOWN
```
tyll: What is the purpose of this line?

tests/lib/metadata_linux_bridge_test.py
```diff
@@ -0,0 +1,365 @@
+#
+# Copyright (c) 2019 Red Hat, Inc.
```
tyll: This needs to be 2020

tests/lib/metadata_linux_bridge_test.py
```diff
@@ -0,0 +1,365 @@
...
+            }
+        )
+        expected_cstate = state.State(current_state.state)
+
+        metadata.generate_ifaces_metadata(desired_state, current_state)
+
+        assert state.State(desired_state.state) == expected_dstate
+        assert current_state == expected_cstate
+
+    def test_editing_option(self):
```
tyll: What is the option that is edited here?

tests/lib/metadata_linux_bridge_test.py
```diff
@@ -0,0 +1,365 @@
...
+                    ),
+                ]
+            }
+        )
+        expected_cstate = state.State(current_state.state)
+
+        metadata.generate_ifaces_metadata(desired_state, current_state)
+
+        assert state.State(desired_state.state) == expected_dstate
+        assert current_state == expected_cstate
```
tyll: What is tested here? That the metadata function does not change the current state?

tests/lib/metadata_linux_bridge_test.py
```diff
@@ -0,0 +1,365 @@
...
+
+BRIDGE0 = "br0"
+BRIDGE1 = "br1"
+IFACE0 = "eth0"
+IFACE1 = "eth1"
+
+
+class TestLinuxBridgeMetadata:
+    def test_creation_with_new_ports(self):
+        desired_state = state.State(
```
tyll: IMHO the tests are very hard to understand since there are only subtle differences between the original and expected state that are hidden in the complex structures that create the states.

tests/lib/metadata_linux_bridge_test.py
```diff
@@ -0,0 +1,365 @@
...
+                        IFACE1,
+                        InterfaceType.UNKNOWN,
+                        extra=_create_metadata_state(BRIDGE0, IFACE1),
+                    ),
+                ]
+            }
+        )
+        expected_cstate = state.State(current_state.state)
+
+        metadata.generate_ifaces_metadata(desired_state, current_state)
```
tyll: This test shows that `metadata.generate_ifaces_metadata()` is badly designed, since it seems to modify its arguments instead of returning the changed data. It should probably be a method of `state.State()` anyhow. Just a note, though.

tests/lib/metadata_linux_bridge_test.py
EdDev: The commit is from December, the year is good.

tests/lib/metadata_linux_bridge_test.py
EdDev: These are unit tests, therefore there is a need to understand how metadata is behaving.
Making tests and code simpler is always welcome, so if you have a better idea, please share.

tests/lib/metadata_linux_bridge_test.py
EdDev: It is mutable indeed. Refactoring this to an immutable version is welcome indeed.
I do not see how it is related to `State`.

tests/lib/metadata_linux_bridge_test.py
EdDev: Yes.
If we will make the metadata generator immutable, then we will not need it.

tests/lib/metadata_linux_bridge_test.py
EdDev: I need to understand what this one is... looks suspicious indeed.
It was copied from the OVS bridge.

tests/lib/metadata_linux_bridge_test.py
tyll: Ugh, this needs to be automated. GitHub writes `Commits on Jan 6, 2020` at the top of the commits tab but then the commit says its from 7 days ago...

tests/lib/metadata_linux_bridge_test.py
EdDev: I changed the test name.

tests/lib/metadata_linux_bridge_test.py
tyll: It might be nicer to create the desired stated and the expected desired state from a common state so that it is more clear that the difference is the added metadata.

tests/lib/metadata_linux_bridge_test.py
tyll: The metadata is a property of the state - or actually a workaround for passing around a dictionary instead of an object with properties or actually the items of the metadata should be properties of a InterfaceState object, I guess.

tests/lib/metadata_linux_bridge_test.py
EdDev: You will just move the hard reading to the editation of the expected state. Editing an item in the list is not trivial.
In this case, it should be easy to see that the difference is with the `_create_iface_state`, where the expected is adding the metadata.

PR #663, opened 2020-01-01
metadata: Ignore existing unmanaged and unknown type ports

EdDev:
When editing bridges, ports which are not managed or have an unknown
type should not be consumed by nmstate.

The issue has been observed when tap devices have been connected to the
bridge and its editation caused nmstate to implicitly define them.

Note: Backport from master.


ffmancera:
This look good but I am missing the tests here. I think we should port them too.

EdDev:
Includes now the fix introduced in #668 .

EdDev:
> This looks good but I am missing the tests here. I think we should port them too.

I'll backport #662 after it is merged.

PR #664, opened 2020-01-02
nm, wired: Read mac-address from sysfs when missing from NM

EdDev:
In some device types (e.g. ovs-interface type) NM does not provide the mac
address from the kernel.
For such cases, use sysfs as a fallback method to read the data.

### DIFF
--- a/libnmstate/nm/wired.py
+++ b/libnmstate/nm/wired.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2018-2019 Red Hat, Inc.
+# Copyright (c) 2018-2020 Red Hat, Inc.
 #
 # This file is part of nmstate
 #
@@ -129,6 +129,9 @@ def get_info(device):
         pass
 
     mac = device.get_hw_address()
+    if not mac:
+        mac = _get_mac_address_from_sysfs(iface)
+
     # A device may not have a MAC or it may not yet be "realized" (zeroed mac).
     if mac and mac != ZEROED_MAC:
         info[Interface.MAC] = mac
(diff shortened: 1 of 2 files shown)


libnmstate/nm/wired.py
```diff
@@ -141,6 +143,17 @@ def get_info(device):
     return info
 
 
+def _get_mac_address_from_sysfs(ifname):
+    mac = None
+    sysfs_path = f'/sys/class/net/{ifname}/address'
+    try:
+        with open(sysfs_path) as f:
+            mac = f.read().rstrip('\n')
```
ffmancera: Black is complaining about the usage of `'`.

libnmstate/nm/wired.py
```diff
@@ -130,6 +130,8 @@ def get_info(device):
 
     mac = device.get_hw_address()
     # A device may not have a MAC or it may not yet be "realized" (zeroed mac).
+    if not mac:
```
tyll: This needs to be inserted before the comment.

tyll: This needs an integration test to get the MAC address from a OVS interface.

libnmstate/nm/wired.py
```diff
@@ -141,6 +143,17 @@ def get_info(device):
     return info
 
 
+def _get_mac_address_from_sysfs(ifname):
```
tyll: IMHO there should be a bug for NetworkManager to always provide the MAC and the bug should be referenced here in a docstring.

EdDev: Done

PR #665, opened 2020-01-02
version: add libnmstate.__version__

ffmancera:
This patch introduces version.py which is being used to get the current
version defined in VERSION constant. The idea is to use this module in
the future for version utilities.


ffmancera:
> The information is persisted in https://github.com/nmstate/nmstate/blob/master/VERSION .
> Can you just read from it?

This is what I thought first, but we are not installing `https://github.com/nmstate/nmstate/blob/master/VERSION` so there it is not possible to read from it when libnmstate is installed.

EdDev:
We either need to include that file or add to setup.py a function that embeds the version into libnmstate.__init__ .
The second option sounds better to me.

cathay4t:
This link could be a good HOWTO:

https://packaging.python.org/guides/single-sourcing-package-version/

tyll:
> I prefer to depend on `setup.py` generated VERSION file.

What do you mean with generated here?

tyll:
> This link could be a good HOWTO:
> 
> https://packaging.python.org/guides/single-sourcing-package-version/

My current preference would be option 4 from this guide. I do not like the current approach of generating a python file. I hope this is not what @cathay4t meant with his comment.

ffmancera:
> > This link could be a good HOWTO:
> > https://packaging.python.org/guides/single-sourcing-package-version/
> 
> My current preference would be option 4 from this guide. I do not like the current approach of generating a python file. I hope this is not what @cathay4t meant with his comment.

In order to do this, we would need to put the VERSION file under libnmstate/

tyll:
> > > This link could be a good HOWTO:
> > > https://packaging.python.org/guides/single-sourcing-package-version/
> > 
> > 
> > My current preference would be option 4 from this guide. I do not like the current approach of generating a python file. I hope this is not what @cathay4t meant with his comment.
> 
> In order to do this, we would need to put the VERSION file under libnmstate/

Works for me.

libnmstate/__init__.py
```diff
@@ -36,3 +41,9 @@
     "schema",
     "PrettyState",
 ]
+
+
+def __version__():
```
tyll: iThis should be a property called `__version__` not a method....

libnmstate/__init__.py
```diff
@@ -36,3 +41,9 @@
...
+    with open(os.path.join(ROOT_DIR, "VERSION")) as f:
+        version = f.read().strip()
```
tyll: change this to `__version__ = ...` and move out of the function. Also there should be a test for this code that at least accesses the property without verifying its value. Verifying the value would, be good, too.

libnmstate/__init__.py
ffmancera: Done!

libnmstate/__init__.py
```diff
@@ -36,3 +41,12 @@
     "schema",
     "PrettyState",
 ]
+
+
+def get_version():
```
tyll: This function should be prefixed with `_` since it is not meant to be part of the API.

tests/lib/version.py
```diff
@@ -0,0 +1,24 @@
...
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+
+import libnmstate
+
+
+def test_get_version():
+    assert libnmstate.__version__
```
tyll: Please test it can be parsed by `StrictVersion` and that it is a string.

libnmstate/__init__.py
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2018-2019 Red Hat, Inc.
```
tyll: Please update the commit message, it is not a method anymore.

tests/lib/version.py
```diff
@@ -0,0 +1,24 @@
...
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+
+import libnmstate
+
+
+def test_get_version():
```
tyll: This does not directly test `get_version()` but the `__version__` attribute.

tests/lib/version.py
```diff
@@ -0,0 +1,24 @@
+#
```
tyll: Please name the file `version_test.py`, otherwise pytest does not discover it. It needs to have `test` in its filename.

libnmstate/__init__.py
EdDev: I do not think we need to expose this, especially if we indent to use `__version__` instead.
@tyll , what do you think?

ffmancera: Done!

tests/lib/version.py
ffmancera: Right now if StrictVersion is able to parse it, is because the version is a string. Done, thanks!

libnmstate/__init__.py
ffmancera: Done, thanks!

tests/lib/version.py
ffmancera: Changed to `test_version()`

tests/lib/version.py
ffmancera: Oops. Done! Thanks :-)

libnmstate/__init__.py
ffmancera: Yes, you are right. Done!

PR #667, opened 2020-01-03
container: install NetworkManager-team plugin

ffmancera:

### DIFF
--- a/packaging/Dockerfile.centos8-nmstate-dev
+++ b/packaging/Dockerfile.centos8-nmstate-dev
@@ -17,6 +17,7 @@ RUN dnf copr enable nmstate/ovs-el8 -y && \
 RUN dnf -y install --setopt=install_weak_deps=False \
                    NetworkManager \
                    NetworkManager-ovs \
+                   NetworkManager-team \
                    NetworkManager-config-server \
                    openvswitch2.11 \
                    systemd-udev \
--- a/packaging/Dockerfile.fedora-nmstate-dev
+++ b/packaging/Dockerfile.fedora-nmstate-dev
@@ -11,6 +11,7 @@ RUN bash ./docker_enable_systemd.sh && rm ./docker_enable_systemd.sh
 RUN dnf -y install --setopt=install_weak_deps=False \
                    NetworkManager \
                    NetworkManager-ovs \
+                   NetworkManager-team \
                    NetworkManager-config-server \
                    dhcp-client \
                    openvswitch \


packaging/Dockerfile.centos8-nmstate-dev
```diff
@@ -17,6 +17,7 @@ RUN dnf copr enable nmstate/ovs-el8 -y && \
 RUN dnf -y install --setopt=install_weak_deps=False \
                    NetworkManager \
                    NetworkManager-ovs \
+		   NetworkManager-team \
```
EdDev: Please fix the indentation.

packaging/Dockerfile.fedora-nmstate-dev
```diff
@@ -11,6 +11,7 @@ RUN bash ./docker_enable_systemd.sh && rm ./docker_enable_systemd.sh
 RUN dnf -y install --setopt=install_weak_deps=False \
                    NetworkManager \
                    NetworkManager-ovs \
+		   NetworkManager-team \
```
EdDev: Please fix the indentation.

tyll:
If the plugin is needed now, is there also a proper error message in case it is missing? Also it should be added as `Suggests` to the spec I guess.

ffmancera:
> If the plugin is needed now, is there also a proper error message in case it is missing? Also it should be added as `Suggests` to the spec I guess.

Yes, I am adding that in the implementation PR. This is just to update the container (needed for CI in the implementation PR)

packaging/Dockerfile.centos8-nmstate-dev
ffmancera: Oops! Done!

packaging/Dockerfile.fedora-nmstate-dev
ffmancera: Oops! Done!

tyll:
> > If the plugin is needed now, is there also a proper error message in case it is missing? Also it should be added as `Suggests` to the spec I guess.
> 
> Yes, I am adding that in the implementation PR. This is just to update the container (needed for CI in the implementation PR)

If you change the container specification (Dockerfiles) in the implementation PR, the CI should rebuild the container images before running the tests. Therefore this change should not require a separate PR. Am I missing something?

ffmancera:
> > If the plugin is needed now, is there also a proper error message in case it is missing? Also it should be added as `Suggests` to the spec I guess.
> 
> Yes, I am adding that in the implementation PR. This is just to update the container (needed for CI in the implementation PR)

AFAIK, CI builds the container from the docker.io image, so when the PR triggers the CI, it will fetch the container from docker.io.

tyll:
> AFAIK, CI builds the container from the docker.io image, so when the PR triggers the CI, it will fetch the container from docker.io.

Please take a look at https://github.com/nmstate/nmstate/blob/master/automation/run-tests.sh#L195 it will rebuild the image locally if necessary.

ffmancera:
This is not needed. Done in #658

PR #668, opened 2020-01-05
state: Fix bridge type interface state parsing

EdDev:
A bridge type interface state includes the `bridge` subconfig only if it
is in an UP state.
When a bridge exists on the system but has no NM profile, it will appear
with state DOWN and will miss the bridge subconfig tree.

When parsing the current bridge interfaces, ignore the ones that their
state is not UP.


EdDev:
> The date comments need to be updated.
Done

PR #670, opened 2020-01-06
nm.sriov: get total-vfs property from sysfs

ffmancera:

### DIFF
--- a/libnmstate/nm/sriov.py
+++ b/libnmstate/nm/sriov.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2019 Red Hat, Inc.
+# Copyright (c) 2019-2020 Red Hat, Inc.
 #
 # This file is part of nmstate
 #
@@ -148,13 +148,24 @@ def _has_sriov_capability(ifname):
 
 def get_info(device):
     """
-    Provide the current active SR-IOV live configuration for a device
+    Provide the current active SR-IOV total-vfs runtime value and the live
+    configuration for each VF for a device.
     """
     info = {}
+    sriov_config = {}
+
+    ifname = device.get_iface()
+    numvf_path = f"/sys/class/net/{ifname}/device/sriov_numvfs"
+    try:
+        with open(numvf_path) as f:
+            sriov_config[Ethernet.SRIOV.TOTAL_VFS] = int(f.read())
+    except FileNotFoundError:
+        return info
 
     connection = nm_connection.ConnectionProfile()
     connection.import_by_device(device)
     if not connection.profile:
+        info[Ethernet.SRIOV_SUBTREE] = sriov_config
         return info
 
     sriov_setting = connection.profile.get_setting_by_name(
@@ -162,9 +173,6 @@ def get_info(device):
     )
 
     if sriov_setting:
-        sriov_config = {
-            Ethernet.SRIOV.TOTAL_VFS: sriov_setting.props.total_vfs
-        }
         vfs_config = _get_info_sriov_vfs_config(sriov_setting)
         sriov_config[Ethernet.SRIOV.VFS_SUBTREE] = vfs_config
 


ffmancera:
This is needed for #648

libnmstate/nm/sriov.py
```diff
@@ -18,6 +18,7 @@
 #
```
tyll: The commit message is missing a justification for this change. Also there should be a bug for NetworkManager to request the missing feature.

libnmstate/nm/sriov.py
```diff
@@ -67,13 +68,15 @@ def get_info(device):
...
-        }
+    ifname = device.get_iface()
+    try:
+        with open(f"/sys/class/net/{ifname}/device/sriov_numvfs") as f:
+            info[Ethernet.SRIOV_SUBTREE] = {
+                Ethernet.SRIOV.TOTAL_VFS: int(f.read())
+            }
+    except FileNotFoundError:
+        raise NmstateInternalError(
+            "Could not read from /sys/class/net/{ifname}/device/sriov_numvfs"
```
tyll: This is missing the `f` specifier for the format string. This mistake indicates that there should be a test to check the error condition.
Also, please compute the path for the path only once and then use it both in the `open()` command and the error message to ensure that they are always the same.

libnmstate/nm/sriov.py
```diff
@@ -67,13 +68,15 @@ def get_info(device):
     if not connection.profile:
```
tyll: The docstring says this gets the live configuration but with this change it is more like the runtime state IMHO. Also given that it checks for the path in the sysfs, it is not clear why this functions checks for the presence of a connection profile since it does not matter anymore, does it?

libnmstate/nm/sriov.py
tyll: The date needs to be adjusted to say `2019-2020`.

libnmstate/nm/sriov.py
ffmancera: It does not matter anymore, I will remove it. Thank you!

libnmstate/nm/sriov.py
ffmancera: Oops. I will add the test for this and build the path once.

libnmstate/nm/sriov.py
ffmancera: Yes, there is already an existing BZ for NM. I will add it as a reference in the commit message among the justification for this change. Thanks!

libnmstate/nm/sriov.py
ffmancera: Fixed the `f ` specifier. About the test, please see comments below

libnmstate/nm/sriov.py
ffmancera: Comments below!

ffmancera:
In this approach I am keeping the `sriov_setting` for 2 reasons:
- If we do not check if `sriov_setting` then this will fail for all the ethernet interfaces that don't support SR-IOV.
- This is going to be needed for the SR-IOV extension PR #648 as we cannot get the other parameters from runtime easily.

About the test, I am not able to test the error case when reading the file because is an unexpected situation.

tyll:
> In this approach I am keeping the `sriov_setting` for 2 reasons:
> 
> * If we do not check if `sriov_setting` then this will fail for all the ethernet interfaces that don't support SR-IOV.

Maybe it would be better to rely on the presence of the sysfs path instead of on the NM setting and then treat the missing path as an indicator that SR-IOV is not supported instead of treating it as an error.

> * This is going to be needed for the SR-IOV extension PR #648 as we cannot get the other parameters from runtime easily.

I did not look there, but it could be added there, again, if it is needed. Maybe this would be overkill but it does not look completely right to keep it here.

libnmstate/nm/sriov.py
```diff
@@ -72,8 +73,14 @@ def get_info(device):
     )
 
     if sriov_setting:
```
EdDev: The case where a user changes this value on the kernel directly (e.g. though sysfs) is valid. The report should show this, but if the user never configured a profile for the device, this solution will not report it.

tyll:
The implications for kubernetes-nmstate need to be evaluated before merging this.

libnmstate/nm/sriov.py
ffmancera: Fixed! Thanks. Please see the comment below.

ffmancera:
This new approach allow us to get the `total-vf` value from the runtime configuration instead from the live configuration of NM. I am preparing another patch (different PR) to get all the VFs runtime configuration.

This patch have been tested using SR-IOV HW and it is working fine. Thanks!

libnmstate/nm/sriov.py
```diff
@@ -148,23 +151,34 @@ def _has_sriov_capability(ifname):
...
     info = {}
+    sriov_config = {}
+
+    ifname = device.get_iface()
+    numvf_path = f"/sys/class/net/{ifname}/device/sriov_numvfs"
+    if os.path.exists(numvf_path):
+        try:
+            with open(numvf_path) as f:
+                sriov_config[Ethernet.SRIOV.TOTAL_VFS] = int(f.read())
+        except FileNotFoundError:
```
tyll: Instead of checking if the path exists, return info here in the except block. It is odd to catch this specific exception here if he code first checks if the file exists and does not treat it as an error.

ffmancera: Done! Thanks :-)

PR #671, opened 2020-01-07
tests: ovs interface set mac test

EdDev:
Now it is possible to set the mac address of an ovs-interface.

Depends on #664

### DIFF
--- a/tests/integration/testlib/ovslib.py
+++ b/tests/integration/testlib/ovslib.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2019 Red Hat, Inc.
+# Copyright (c) 2019-2020 Red Hat, Inc.
 #
 # This file is part of nmstate
 #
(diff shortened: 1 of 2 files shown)


tests/integration/testlib/ovslib.py
```diff
@@ -58,15 +58,18 @@ def add_link_aggregation_port(self, name, slaves, mode=None):
                 OVSBridge.Port.LinkAggregation.MODE
             ] = mode
```
tyll: The year needs to be updated in the header.

tests/integration/testlib/ovslib.py
```diff
@@ -58,15 +58,18 @@ def add_link_aggregation_port(self, name, slaves, mode=None):
...
 
-    def add_internal_port(self, name, ipv4_state):
+    def add_internal_port(self, name, mac=None, ipv4_state=None):
```
tyll: Please consider adding a `*, ` before `mac` to enforce the keyword arguments and make future changes easier. This change is changing the signature already.

tests/integration/testlib/ovslib.py
EdDev: Done

tests/integration/testlib/ovslib.py
EdDev: Done

PR #673, opened 2020-01-07
Prevent against container name squatting and update automation README

tyll:


.travis.yml
```diff
@@ -5,38 +5,38 @@ env:
     global:
         - use_codecov=true
     matrix:
-        - CONTAINER_IMAGE=nmstate/centos8-nmstate-dev
+        - CONTAINER_IMAGE=docker.io/nmstate/centos8-nmstate-dev
```
EdDev: Can you use a global variable to avoid the need to re-specify this all over id we change this in the future?

packaging/Dockerfile.centos8-nmstate-dev
```diff
@@ -2,7 +2,7 @@
 # Fedora official repository
 # (https://hub.docker.com/r/fedora/systemd-systemd/).
 # It enables systemd to be operational.
-FROM centos:8
+FROM docker.io/library/centos:8
```
EdDev: I do not think this is a common practice.
When running the build, you can specify the repository AFAIK.

.travis.yml
tyll: Currently we have 5 times the CentOS image and 7 times the Fedora image which is fairly even and it might be possible that either of these images might be used more in the future (we might move unit tests to CentOS or more integration tests to Fedora or remove the Python 3.7 unit test completely). I am not sure that we will certainly save something.

packaging/Dockerfile.centos8-nmstate-dev
tyll: When there was only docker and a single registry, it was not yet a problem. Therefore it is unlikely that it is a common practice. What is the problem with this?

.travis.yml
EdDev: I meant only the change portions, i.e. `docker.io`.

packaging/Dockerfile.centos8-nmstate-dev
EdDev: I see no specific problem, but I would like to check what the k8s projects are doing (or plan to) as well.

EdDev: Seems like some use the full path and some do not.
Seems ok to proceed.

tyll:
This needs to be  merged in smaller steps to ensure we test all cases in the CI.

packaging/Dockerfile.centos8-nmstate-dev
tyll: This change is now split in #682 to simplify this PR.

PR #679, opened 2020-01-08
Better deletion tests

tyll:
Split from #649 as requested by @EdDev

### DIFF
--- a/tests/integration/bond_test.py
+++ b/tests/integration/bond_test.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2018-2019 Red Hat, Inc.
+# Copyright (c) 2018-2020 Red Hat, Inc.
 #
 # This file is part of nmstate
 #
@@ -33,6 +33,7 @@ from libnmstate.schema import InterfaceIPv4
 from libnmstate.schema import InterfaceIPv6
 
 from .testlib import assertlib
+from .testlib import cmdlib
 from .testlib import statelib
 from .testlib.assertlib import assert_mac_address
 from .testlib.bondlib import bond_interface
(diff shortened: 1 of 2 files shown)


tests/integration/testlib/assertlib.py
```diff
@@ -35,13 +36,38 @@ def assert_state(desired_state_data):
     assert desired_state.state == current_state.state
 
 
-def assert_absent(*ifnames):
+def assert_absent(*ifnames, timeout=0):
     """ Assert that a interface is not present in the current state """
 
-    current_state = statelib.show_only(ifnames)
+    now = time.time()
+    while True:
```
EdDev: I think I suggested a decorator in the original PR.
Can we do that?

(it is already duplicated here in two asserts and probably other asserts will need it in the future)

tests/integration/bond_test.py
```diff
@@ -99,6 +100,16 @@ def bond99_with_slave(eth2_up):
         yield state
 
 
+def test_remove_existing_bond(setup_remove_bond99):
+    subprocess.check_call(["ip", "link", "add", BOND99, "type", "bond"])
```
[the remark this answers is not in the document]
tyll: changed it.

tests/integration/bond_test.py
```diff
@@ -99,6 +100,19 @@ def bond99_with_slave(eth2_up):
...
 
 
+def test_remove_existing_bond(setup_remove_bond99):
+    assert (
+        cmdlib.exec_cmd(["ip", "link", "add", BOND99, "type", "bond"])[0]
+        == cmdlib.RC_SUCCESS
+    )
+    assertlib.assert_present(BOND99, timeout=1)
+
+    with bond_interface(BOND99, [], create=False):
```
tyll: It might be a bad idea to use the contextmanager here because it hides verification errors. But it might also be that it does not matter in this specific test case. However, it is for me a surprising behavior and makes it harder to understand the tests.

PR #681, opened 2020-01-08
Rename DOCKER_IMAGE to CONTAINER_IMAGE and cleanup run-tests.sh README

tyll:


tyll:
These are non-controversial changes from #673.

PR #685, opened 2020-01-09
tests: set a mtu lower than 1280

ffmancera:
This bug have been already fixed. This test checks that nmstate can set
a mtu lower than 1280 when IPv6 is not enabled.

Ref: https://bugzilla.redhat.com/1753128

### DIFF
--- a/tests/integration/ethernet_mtu_test.py
+++ b/tests/integration/ethernet_mtu_test.py
@@ -135,6 +135,12 @@ def test_decrease_to_lower_than_min_ipv6_iface_mtu(eth1_with_ipv6):
     assertlib.assert_state(original_state)
 
 
+def test_mtu_without_ipv6(eth1_up):
+    eth1_up[Interface.KEY][0][Interface.MTU] = 576
+    libnmstate.apply(eth1_up)
+    assertlib.assert_state(eth1_up)
+
+
 @pytest.mark.xfail(reason="https://bugzilla.redhat.com/1751079", strict=True)
 def test_set_mtu_on_two_vlans_with_a_shared_base(eth1_up):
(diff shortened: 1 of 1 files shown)


tests/integration/ethernet_mtu_test.py
```diff
@@ -135,6 +135,12 @@ def test_decrease_to_lower_than_min_ipv6_iface_mtu(eth1_with_ipv6):
     assertlib.assert_state(original_state)
 
 
+def test_decrease_mtu_to_lower_than_1280(eth1_up):
```
tyll: Since IPv6 being disabled is important here it should be mentioned in the test name, maybe `test_mtu_without_ipv6` and then parametrize it with 1279, 926 and 576 to test the edge cases and a middle value.

Even better would be to introduce a constant for 1280 such as `MINIMAL_IPV6_MTU` and then use `MINIMAL_IPV6_MTU - 1`.

ffmancera: Done! Thanks!

ffmancera:
Rebased!

tests/integration/ethernet_mtu_test.py
```diff
@@ -135,6 +138,13 @@ def test_decrease_to_lower_than_min_ipv6_iface_mtu(eth1_with_ipv6):
     assertlib.assert_state(original_state)
 
 
+@pytest.mark.parametrize("mtu", [MINIMAL_IPV6_MTU - 1, 926, 576])
```
EdDev: What are these `926, 576` mtu values represent?
These are not unit tests, why do we need to test these?

ffmancera: After discuss this on IRC. We are only testing 576. Thanks!

PR #689, opened 2020-01-10
tests: set a vlan as an ovs-bridge slave

ffmancera:


ffmancera:
Rebased! Thanks!

PR #694, opened 2020-01-17
nm.sriov: get SR-IOV VFs configuration from runtime

ffmancera:
This is a workaround to get the SR-IOV VFs configuration from runtime.
We are currently parsing these values from the output of "ip link show
<device>" command.

Ref: https://bugzilla.redhat.com/1777520

### DIFF
--- a/libnmstate/nm/sriov.py
+++ b/libnmstate/nm/sriov.py
@@ -17,8 +17,10 @@
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 #
 
+import re
+import subprocess
+
 from libnmstate.error import NmstateNotSupportedError
-from libnmstate.nm import connection as nm_connection
 from libnmstate.nm import device
 from libnmstate.nm import nmclient
 from libnmstate.schema import Ethernet
@@ -48,27 +50,14 @@ SRIOV_NMSTATE_TO_NM_MAP = {
     ),
 }
 
-SRIOV_NM_TO_NMSTATE_MAP = {
-    nmclient.NM.SRIOV_VF_ATTRIBUTE_MAC: (
-        Ethernet.SRIOV.VFS.MAC_ADDRESS,
-        nmclient.GLib.Variant.get_string,
+SRIOV_NMSTATE_TO_REGEX = {
+    Ethernet.SRIOV.VFS.MAC_ADDRESS: re.compile(
+        r"[a-fA-F0-9:]{17}|[a-fA-F0-9]{12}"
     ),
-    nmclient.NM.SRIOV_VF_ATTRIBUTE_SPOOF_CHECK: (
-        Ethernet.SRIOV.VFS.SPOOF_CHECK,
-        nmclient.GLib.Variant.get_boolean,
-    ),
-    nmclient.NM.SRIOV_VF_ATTRIBUTE_TRUST: (
-        Ethernet.SRIOV.VFS.TRUST,
-        nmclient.GLib.Variant.get_boolean,
-    ),
-    nmclient.NM.SRIOV_VF_ATTRIBUTE_MIN_TX_RATE: (
-        Ethernet.SRIOV.VFS.MIN_TX_RATE,
-        nmclient.GLib.Variant.get_uint32,
-    ),
-    nmclient.NM.SRIOV_VF_ATTRIBUTE_MAX_TX_RATE: (
-        Ethernet.SRIOV.VFS.MAX_TX_RATE,
-        nmclient.GLib.Variant.get_uint32,
-    ),
+    Ethernet.SRIOV.VFS.SPOOF_CHECK: re.compile(r"checking (on|off)"),
+    Ethernet.SRIOV.VFS.TRUST: re.compile(r"trust (on|off)"),
+    Ethernet.SRIOV.VFS.MIN_TX_RATE: re.compile(r"min_tx_rate ([0-9]+)"),
+    Ethernet.SRIOV.VFS.MAX_TX_RATE: re.compile(r"max_tx_rate ([0-9]+)"),
 }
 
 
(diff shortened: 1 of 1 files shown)


ffmancera:
This patch have been tested using HW with SR-IOV support.

libnmstate/nm/sriov.py
```diff
@@ -162,42 +140,69 @@ def get_info(device):
...
 
-def _get_info_sriov_vfs_config(sriov_setting):
-    vfs_config = []
-    vfs_setting = sriov_setting.props.vfs
-    for vf in vfs_setting:
-        vf_config = {}
-        vf_config[Ethernet.SRIOV.VFS.ID] = vf.get_index()
-        for nm_attribute in SRIOV_NM_TO_NMSTATE_MAP.keys():
-            _get_nm_attribute(vf, vf_config, nm_attribute)
+def _get_sriov_vfs_configuration(ifname):
```
cathay4t: IMHO, in nmstate, the `configuration` means on-disk profile/connection. So please use `_get_sriov_vfs_info`.

libnmstate/nm/sriov.py
```diff
@@ -162,42 +140,69 @@ def get_info(device):
...
     return vfs_config
 
 
-def _get_nm_attribute(vf_object, vf_config, nm_attribute):
-    nmstate_key, nm_variant = SRIOV_NM_TO_NMSTATE_MAP[nm_attribute]
-    if vf_object.get_attribute(nm_attribute) is not None:
-        vf_config[nmstate_key] = nm_variant(
-            vf_object.get_attribute(nm_attribute)
-        )
+def _create_vf_config(vfs):
```
cathay4t: This function does not create any configure, it just extract/convert ip link output to nmstate info. So maybe `_parse_ip_link_output_for_vfs`

libnmstate/nm/sriov.py
```diff
@@ -162,42 +140,69 @@ def get_info(device):
...
+        vf_config[Ethernet.SRIOV.VFS.SPOOF_CHECK] = spoof
+        trust = trust_expr.search(vf).group(0).split()[1]
+        if trust == "on":
+            trust = True
+        else:
+            trust = False
+        vf_config[Ethernet.SRIOV.VFS.TRUST] = trust
+        min_tx = min_tx_expr.search(vf)
+        if min_tx:
+            min_tx = min_tx.group(0).split()[1]
```
cathay4t: You can just change regex group like `max_tx_rate ([0-9]*)` instead of doing `split()`.

Since you do `split()[1]`, you are expecting there is a number after it, the regex should be `max_tx_rate ([0-9]+)`.

libnmstate/nm/sriov.py
```diff
@@ -162,42 +140,69 @@ def get_info(device):
...
-    if vf_object.get_attribute(nm_attribute) is not None:
-        vf_config[nmstate_key] = nm_variant(
-            vf_object.get_attribute(nm_attribute)
-        )
+def _create_vf_config(vfs):
+    mac_expr = re.compile(r"[a-fA-F0-9:]{17}|[a-fA-F0-9]{12}")
+    min_tx_expr = re.compile(r"min_tx_rate [0-9]*")
+    max_tx_expr = re.compile(r"max_tx_rate [0-9]*")
+    spoof_expr = re.compile(r"checking (on|off)")
+    trust_expr = re.compile(r"trust (on|off)")
```
cathay4t: Doing so, you have a lot duplicate code and don't grow well when more options come.

You may create a function taking three arguments:

iplink_output, regex, key_name

In that function, you can do converting `on` to True or converting string to int.

libnmstate/nm/sriov.py

libnmstate/nm/sriov.py
ffmancera: Done, thank you!

libnmstate/nm/sriov.py
ffmancera: Done and then using `.group(1)` in order to match it.

libnmstate/nm/sriov.py
ffmancera: Done, I hope you like the new approach. Thanks!

ffmancera:
I have tested the updated patch in HW with SR-IOV support.

ffmancera:
I've added some comments to clarify why we are ignoring the first two lines of `ip link show <device>` output.

libnmstate/nm/sriov.py
```diff
@@ -148,8 +135,7 @@ def _has_sriov_capability(ifname):
 
 def get_info(device):
     """
-    Provide the current active SR-IOV total-vfs runtime value and the live
-    configuration for each VF for a device.
+    Provide the current active SR-IOV runtime values
     """
     info = {}
```
EdDev: nit: I think this one is redundant now.
You can just create it on return (e.g. `return {Ethernet.SRIOV_SUBTREE: sriov_config}`)

libnmstate/nm/sriov.py
```diff
@@ -162,42 +148,66 @@ def get_info(device):
     except FileNotFoundError:
         return info
 
-    connection = nm_connection.ConnectionProfile()
-    connection.import_by_device(device)
-    if not connection.profile:
-        info[Ethernet.SRIOV_SUBTREE] = sriov_config
-        return info
+    if sriov_config[Ethernet.SRIOV.TOTAL_VFS]:
```
EdDev: `sriov_config` is no longer a "config", it is now a sriov running_info or something similar.

libnmstate/nm/sriov.py
```diff
@@ -162,42 +148,66 @@ def get_info(device):
...
 
 
-def _get_nm_attribute(vf_object, vf_config, nm_attribute):
-    nmstate_key, nm_variant = SRIOV_NM_TO_NMSTATE_MAP[nm_attribute]
-    if vf_object.get_attribute(nm_attribute) is not None:
-        vf_config[nmstate_key] = nm_variant(
-            vf_object.get_attribute(nm_attribute)
-        )
+def _parse_ip_link_output_for_vfs(vfs):
+    vf_id = 0
```
EdDev: Can you use an enumerator instead in the for loop?

libnmstate/nm/sriov.py
```diff
@@ -162,42 +148,66 @@ def get_info(device):
...
+        }
+        vf_config.update(_parse_ip_link_output_options_for_vf(vf))
+        vf_id += 1
+        yield vf_config
+
+
+def _parse_ip_link_output_options_for_vf(vf):
+    vf_options = {}
+    for option, expr in SRIOV_NMSTATE_TO_REGEX.items():
+        option_expr = re.compile(expr)
```
EdDev: You can store the compiled expression in the map and not recompile it each time.

libnmstate/nm/sriov.py
```diff
@@ -162,42 +148,66 @@ def get_info(device):
...
-        vf_config[nmstate_key] = nm_variant(
-            vf_object.get_attribute(nm_attribute)
-        )
+def _parse_ip_link_output_for_vfs(vfs):
+    vf_id = 0
+    for vf in vfs:
+        vf_config = {
+            Ethernet.SRIOV.VFS.ID: vf_id,
+        }
+        vf_config.update(_parse_ip_link_output_options_for_vf(vf))
```
EdDev: Reversing the order and avoiding the update makes more sense to me, but it is not a must.

libnmstate/nm/sriov.py
```diff
@@ -162,42 +148,66 @@ def get_info(device):
...
-        vf_config = {}
-        vf_config[Ethernet.SRIOV.VFS.ID] = vf.get_index()
-        for nm_attribute in SRIOV_NM_TO_NMSTATE_MAP.keys():
-            _get_nm_attribute(vf, vf_config, nm_attribute)
+def _get_sriov_vfs_info(ifname):
+    """
+    This is a workaround to get the VFs configuration from runtime.
+    Ref: https://bugzilla.redhat.com/1777520
+    """
+    proc = subprocess.run(
```
EdDev: Just some notes for the future:
- If we will need it again (I am guessing we would), it should be moved to a module in order to share it.
- We should consider changing the test module `cmd` to use the new `run` if it fits.

libnmstate/nm/sriov.py

libnmstate/nm/sriov.py

libnmstate/nm/sriov.py

libnmstate/nm/sriov.py

libnmstate/nm/sriov.py

PR #695, opened 2020-01-17
nm.ipv6: drop workaround for missing support to disable IPv6

ffmancera:
Since NM 1.20,  IPv6 can be properly disabled with NetworkManager.


libnmstate/netapplier.py
```diff
@@ -163,8 +162,7 @@ def _apply_ifaces_state(
         # finish before proceeding with other actions.
         # TODO: https://nmstate.atlassian.net/browse/NMSTATE-103
         time.sleep(5)
-        if not nmclient.can_disable_ipv6():
-            _disable_ipv6(current_state)
+        _disable_ipv6(current_state)
```
cathay4t: Please also remove the use of `_disable_ipv6()` which is not required on NM 1.20+

ffmancera: Oops. You are right, done! Thanks :-)

libnmstate/netapplier.py
```diff
@@ -160,13 +157,7 @@ def _apply_ifaces_state(
     except nm.checkpoint.NMCheckPointCreationError:
         raise NmstateConflictError("Error creating a check point")
     except NmstateError:
-        # Assume rollback occured, revert IPv6 stack state.
-        # Checkpoint rollback is async, there is a need to wait for it to
-        # finish before proceeding with other actions.
-        # TODO: https://nmstate.atlassian.net/browse/NMSTATE-103
         time.sleep(5)
```
EdDev: You left the "sleep" part, therefore the follwing is still relevant.
```
# Assume rollback occurred,
# Checkpoint rollback is async, there is a need to wait for it to
# finish before proceeding with other actions.
```

ffmancera: Oops. Yes, you are right. Thanks!

ffmancera:
Fixed and rebased! Thanks

ffmancera:
Rebased! :-)

PR #699, opened 2020-01-21
Integration tests: Rename cmd to cmdlib

tyll:
The `testlib.cmd` module is imported under a custom name because `cmd`
is too generic. Rename it to `cmdlib` to follow other sub libs in
`testlib`.

### DIFF
--- a/tests/integration/testlib/iprule.py
+++ b/tests/integration/testlib/iprule.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2019 Red Hat, Inc.
+# Copyright (c) 2019-2020 Red Hat, Inc.
 #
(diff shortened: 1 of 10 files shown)


tests/integration/testlib/iprule.py
```diff
@@ -21,7 +21,7 @@
 import logging
 
 from libnmstate import iplib
-from . import cmd as libcmd
+from . import cmd as cmdlib
```
ffmancera: That seems wrong, it should be `from . import cmdlib`. Thanks! :-)

tyll: well spotted, thank you. fixed.

PR #702, opened 2020-01-26
tests, examplelib: Do not use verify on cleanup

EdDev:


tests/integration/testlib/examplelib.py
```diff
@@ -41,7 +41,7 @@ def example_state(initial, cleanup=None):
         yield desired_state
     finally:
         if cleanup:
-            libnmstate.apply(load_example(cleanup))
+            libnmstate.apply(load_example(cleanup), verify_change=False)
```
tyll: IMHO it is wrong for the examples not to verify that the cleanup example works as well. It documents to our users how interfaces can be removed and therefore it should be expected to always work. Why do you want to disable it?

EdDev: The test is responsible to assert if the action succeeded or not.

The action of cleanup should always be done, even if the end result is not as expected because we *must* cleanup at the test end. If there is a bug at the creation, we want the cleanup to happen and do its best.
So I prefer something over nothing.

tyll: So how about

```
try:
    libnmstate.apply(load_example(cleanup))
except:
    libnmstate.apply(load_example(cleanup), verify_change=False)
    raise
```

This way the test should fail if the cleanup works (it does , doesn't it?) and it will still try to cleanup with libnmstate a little bit more. It would probably be better to do a proper cleanup without using libnmstate by just removing all profiles/interfaces, see https://lists.fedorahosted.org/archives/list/nmstate-devel@lists.fedorahosted.org/thread/ZHTJTFZZR6YV3FLMX2AUYQA75GH5SSDZ/

EdDev: As I mentioned before, the cleanup should occur in the best way possible and if it did not do the job well, it should fail on an assertion.
Your suggestion may work, it is a bit wasteful IMO, but if it works, I do not mind.

Regarding cleaning up NM profiles directly... it is possible, but it sounds to me a cleanup of a cleanup. If we do it right in a test, there should be no leftovers.
I also prefer not to depend on NM being the backend, at least the tests should now depend on that.. as much as we possible can (there are always exceptions).

ffmancera:
Why is this failing now and not before?

tests/integration/testlib/examplelib.py
ffmancera: IMHO, @tyll 's proposed solution looks good to me. I think it is good to know that we have known issues on the cleanup verification. About the profile, I think @EdDev is right.

EdDev:
I disagree with the notion that the lib needs to catch cleanup errors. What we need from the helper lib in the context is to clean after itself as well as possible. If the test wants to check if the context cleanup worked, it can assert for that.
If the test wants to explicitly check a cleanup, it should call the cleanup directly and not through the context.

But as this is specific to the examples  tests and these are a bit different from the other libs, I will not fight over this. Please to not use this pattern in other test helper libs.

tyll:
> I disagree with the notion that the lib needs to catch cleanup errors. What we need from the helper lib in the context is to clean after itself as well as possible. If the test wants to check if the context cleanup worked, it can assert for that.

The tests assert that the cleanup actually worked:
https://github.com/nmstate/nmstate/blob/master/tests/integration/examples_test.py#L74

However, if `example_state()` hides cleanup failures, then the assert does not assert the application of the cleanup yaml with verify enabled but only without verify enabled. Therefore my proposed solution is also not really correct. A better solution would be a different cleanup solution maybe as a fixture. However, we decided that cleaning up after the tests is not a priority currently and there are easy workarounds to cleanup individual systems or to just use a volatile container to make the errors irrelevant. Is there an important issue with the cleanup?

> If the test wants to explicitly check a cleanup, it should call the cleanup directly and not through the context.

IMHO it is valid to not hide cleanup errors and AFAIK the way to assert that no exception is raised is to do nothing, isn't it?

> But as this is specific to the examples tests and these are a bit different from the other libs, I will not fight over this. Please to not use this pattern in other test helper libs.

How do you propose this? AFAICS you want it to be like

```
with example_state("eth1.yml", cleanup="eth1_absent.yml):
    assert eth1 is there
    # do something to invoke cleanup
    assert eth1 is gone
````

Is this correct? I guess the contextmanager could return an object to do something like

```
with example_state("eth1.yml", cleanup="eth1_absent.yml) as context:
    assert eth1 is there
    context.cleanup()
    assert eth1 is gone
```

It might need a better name than `context`.

**Comment by EdDev** [comment]
> The tests assert that the cleanup actually worked
> However, if example_state() hides cleanup failures, then the assert does not assert the application of the cleanup yaml with verify enabled but only without verify enabled.

As stated before, the contextmanager is there to cleanup and not to detect errors in the cleanup as you suggests here. If you want to explicitly check a removal yaml, do it explicitly.

> Therefore my proposed solution is also not really correct.

I do not see the problem now. If the execution of the cleanup yaml fails, it reverts back and we do it again and then raise. We will see the problem in the log.

> A better solution would be a different cleanup solution maybe as a fixture. However, we decided that cleaning up after the tests is not a priority currently and there are easy workarounds to cleanup individual systems or to just use a volatile container to make the errors irrelevant.

IMO this is out of this PR scope. A test, should cleanup itself as best as it can and anything beyond that is another layer which can be discussed in a dedicated thread.

> Is there an important issue with the cleanup?

During investigation of failing tests and leftovers, I noticed this issue and tried to addressed it.
The verification may fail because actions are still in progress for example (we are seeing such async problems from time to time, sometime due to the cache and sometime because the actual did not reached the expected state).

**Comment by EdDev** [comment]
> How do you propose this? AFAICS you want it to be like
> 
> ```
> with example_state("eth1.yml", cleanup="eth1_absent.yml):
>     assert eth1 is there
>     # do something to invoke cleanup
>     assert eth1 is gone
> ````

The last assert needs to be at the same indentation as the `with`.

> Is this correct? I guess the contextmanager could return an object to do something like
> 
> ```
> with example_state("eth1.yml", cleanup="eth1_absent.yml) as context:
>     assert eth1 is there
>     context.cleanup()
>     assert eth1 is gone
> ```

This was not my intention.

File: tests/integration/testlib/examplelib.py [hunk 2]
```diff
@@ -41,7 +41,11 @@ def example_state(initial, cleanup=None):
         yield desired_state
     finally:
         if cleanup:
-            libnmstate.apply(load_example(cleanup))
+            try:
+                libnmstate.apply(load_example(cleanup), verify_change=True)
+            except libnmstate.error.NmstateVerificationError:
```
tyll: Why only for the verification error? Shouldn't it try to cleanup in case of any error?

EdDev: The `verify_change=False` is only useful if all is good except of the verification step (like partial cleanup vs all-or-nothing cleanup).
If there is a code error or anything unexpected, it should blow up as we have nothing better to do.

tyll: There could still be a coding error in the verification step. Nevertheless, the more I think about this, the cleanup as part as the tear down belongs IMHO in a fixture or decorator instead of in a context inside the test. This change makes it hard to understand its intentions.

EdDev: A fixture usually uses the contextmanager. When and where you call it is a topic by itself.
Please re-suggest a solution that will allow us to move forward with cleaning up the example tests.

tests/integration/testlib/examplelib.py
tyll: > Regarding cleaning up NM profiles directly... it is possible, but it sounds to me a cleanup of a cleanup. If we do it right in a test, there should be no leftovers.

Please read https://lists.fedorahosted.org/archives/list/nmstate-devel@lists.fedorahosted.org/thread/ZHTJTFZZR6YV3FLMX2AUYQA75GH5SSDZ/ - there is a problem with using checkpoints during cleanup therefore it would not be a second cleanup/teardown but the only one. However, for the examples we also want to test that the examples to remove something also work and probably in other tests, too. But these tests should probably still have a teardown cleanup phase.

> I also prefer not to depend on NM being the backend, at least the tests should now depend on that.. as much as we possible can (there are always exceptions).

Once we have another backend, it could be added to the cleanup tool so it does the appropriate cleanup depending on the tested backend.

tests/integration/testlib/examplelib.py
```diff
@@ -1,5 +1,5 @@
 #
```
tyll: Proposal for the commit message:

```
tests, examplelib: Retry cleanup without verify

In case the cleanup example fails in the verification step, perform a second cleanup without verification to cleanup as much as possible.

Raise the cleanup exception to expose the error in the test since the cleanup examples should work as-is with verification.
```

tests/integration/testlib/examplelib.py
tyll: I can live with your current solution if you update the commit message. I provided a suggestion since it did not seem to be accurate or correct English.

tyll:
> > How do you propose this? AFAICS you want it to be like
> > ```
> > with example_state("eth1.yml", cleanup="eth1_absent.yml):
> >     assert eth1 is there
> >     # do something to invoke cleanup
> >     assert eth1 is gone
> > ```
> 
> The last assert needs to be at the same indentation as the `with`.

If the last assert is at the same indentation, then the cleanup in the context manager must not ignore verification errors. Otherwise the assert would assert the result from the wrong command.

My intent is to understand what you suggest for the cleanest solution given that you do not really like the current change.

EdDev:
> Proposal for the commit message

Done, thanks.

> If the last assert is at the same indentation, then the cleanup in the context manager must not ignore verification errors. Otherwise the assert would assert the result from the wrong command.

If you consider the contextmanager main purpose is to perform the fist setup and consider the teardown as a cleanup step only (and not an intended action to be tested), then whether the cleanup worked or not is left to the test to re-check if it is interested.
Your reasoning that the cleanup needs to be done by a fixture, outside of the test, is valid. We are not enforcing well enough the separation between fixtures and tests and perhaps that is the confusing part.
Nevertheless, this change was minimal and tried to attack only the leftover problem. A more overall change is needed to introduce the separation of fixtures & tests which will probably fit much better than the current step.
I guess the current change is somewhere in between, trying to balance between the two needs.

PR #705, opened 2020-01-26
Packaging: Use LC_TIME for RPM date

tyll:
Use `LC_TIME` to generate the RPM date instead of only setting `LANG` to
ensure that user changes to `LC_TIME` do not interfere.


ffmancera:
Why is this failing now and not before?

tyll:
> Do we have something that is testing this? Can you mention how this was tested to work?

The CI is using this, so it would fail building the RPM when it would not work (see the e-mail from Vlad that led to this PR).

> Why is this failing now and not before?

It fails when someone runs the script and set `LC_TIME` to a non-working value.

tyll:
DCO check looks fine, I will ignore that no status was reported.

PR #706, opened 2020-01-26
nm, ovs: Learn the ovs bridge ports and ifaces from NM

EdDev:
OVS ports and interfaces have been interpreted by parsing the current
state instead of fetching it from the NM device.
Now that the `get_slaves` function is available (from NM 1.14), the
parsing and the hard-coded port name prefix assumptions are no longer
needed.

Note: The hard-coded prefix of the ovs ports names limited changes to
their name format in the future. The current change allows future
changes to the port names, keeping backward compatibility.


ffmancera:
Hm. CI is complaining. I am restarting the job..

PR #707, opened 2020-01-26
netapplier: Wait before the verification step

EdDev:
CI tests have frequently failed on the `test_dhcp4_with_static_ipv6`
test, missing the IPv6 address.

This change assumes that such problems occur due to the kernel or NM
client cache not updating fast enough, therefore, a small sleep period
is added before the verification step.

### DIFF
--- a/libnmstate/netapplier.py
+++ b/libnmstate/netapplier.py
@@ -149,6 +149,9 @@ def _apply_ifaces_state(
                     con_profiles=ifaces_add_configs + ifaces_edit_configs,
                 )
             if verify_change:
+                # Some actions are not getting updated fast enough in the
+                # kernel/cache. Wait a bit before the verification step.
+                time.sleep(2)
                 _verify_change(desired_state)
         if not commit:
             return checkpoint


cathay4t:
I don't like the fix here as it might hide root cause of ci failure. I will investigate on this to support my standing.

EdDev:
> I don't like the fix here as it might hide root cause of ci failure. I will investigate on this to support my standing.

I totally agree, but the CI fails frequently now with different reasons.
One of these was this one and only on ci.
Can we confirm this by running it multiple times and take it temporary with a FIXME + ticket in the comment?

tyll:
> > I don't like the fix here as it might hide root cause of ci failure. I will investigate on this to support my standing.
> 
> I totally agree, but the CI fails frequently now with different reasons.
> One of these was this one and only on ci.
> Can we confirm this by running it multiple times and take it temporary with a FIXME + ticket in the comment?

I am fine with this temporary fix. Ideally Nmstate would retry until the timeout expires to check if everything is fine earlier. Also it would be great to get statistics about how often and for which tests this occurs.

EdDev:
> I am fine with this temporary fix. Ideally Nmstate would retry until the timeout expires to check if everything is fine earlier. Also it would be great to get statistics about how often and for which tests this occurs.

@cathay4t wanted to investigate this further when he gets back.
I guess he can suggest a retry verification mechanism with configurable thresholds to investigate this.

At the moment, I just want to calm down the CI.

EdDev:
CI failing on a known VLAN issue which is supposed to be solved in NM 1.22.6.

PR #708, opened 2020-01-27
Dynamic: Use dhcp-iaid introduced in NM 1.22

almusil:
Use dhcp-iaid=mac for IPv4 and IPv6 dynamic address.

### DIFF
--- a/libnmstate/nm/ipv6.py
+++ b/libnmstate/nm/ipv6.py
@@ -119,6 +119,7 @@ def create_setting(config, base_con_profile):
         nmclient.NM.SettingIP6ConfigAddrGenMode.EUI64
     )
     setting_ip.props.dhcp_duid = "ll"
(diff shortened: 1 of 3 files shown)


almusil:
This cannot be merged before we move to NM 1.22

EdDev:
Depends on #711

EdDev:
> This would also mean that we cannot update Nmstate in Fedora 31 anymore unless we ask NM to also update there.

nmstate-0.1 is released on fc31, so we do not intend to have 0.2 there.
And I do not think we are interested in back-porting this (we cannot).

tyll:
> > This would also mean that we cannot update Nmstate in Fedora 31 anymore unless we ask NM to also update there.
> 
> nmstate-0.1 is released on fc31, so we do not intend to have 0.2 there.

Ok, thank you for the clarification.

PR #714, opened 2020-01-28
stdci: Remove oVirt STDCI

EdDev:
oVirt STDCI has been optional for some time now and we rarely look at
its results.
Therefore, there is no point in keeping them running.

### DIFF
--- a/.stdci.yml
+++ b/.stdci.yml
@@ -1,9 +1,0 @@
----
-Architectures:
-  - x86_64:
-      Distributions:
-        - fc30
-      runtime-requirements:
-        host-distro: same
-script:
-  from-file: automation/run-tests.sh
--- a/automation/README.md
+++ b/automation/README.md
@@ -19,14 +19,6 @@ It may be used both locally and through CI.
 
   It also handles the cleanup of the container and nets (stop,rm).
 
-- run-tests.mounts: Includes mounts to be used by the oVirt CI (STDCI) worker.
-
-- run-tests.packages: Includes the packages needed by the oVirt CI (STDCI)
-  worker.
-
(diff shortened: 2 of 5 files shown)


EdDev:
@tyll , the results clutter the total sum at the pr list. It also wastes ovirt ci resources.
For now, we chose travis-ci, lets stick to it and drop this.

If you do not mind, please approve.

tyll:
> If you do not mind, please approve.

I am fine with removing it. STDCI still reported a status here. It seems that you caught all files but can you check for sure? Does nmstate needs to be removed somewhere else, too? I am happy to approve it once the status is gone from this PR.

EdDev:
The status will stay there because it needs to be removed from the oVirt CI.
I will ask for this, but it does not need to block this PR, it will just give an "empty" ack.

tyll:
The problem with the empty ack is that it makes it look like it is passing
the tests but it is actually not testing anything. If you guarantee it will
be removed by the end of the week I can live with this but I do not like
having a misleading status for a long time there. I am probably the one
that will fail for it at some point.

EdDev:
> The problem with the empty ack is that it makes it look like it is passing
> the tests but it is actually not testing anything.

I see no problem with this, as it counts for nothing and has no weight to allow a merge or not.
Yes, it is not nice, but it does no merits blocking this change on this techincality.
For sure it is no worse then when we had this red and ignored it.

> If you guarantee it will be removed by the end of the week I can live with this but I do not like
> having a misleading status for a long time there. I am probably the one
> that will fail for it at some point.

You should agree (or not) based on the existing state, not based on promises.

EdDev:
I have deactivated the webhook that notifies oVirt stdci.

PR #716, opened 2020-01-30
GitHub: Add stale workflow

tyll:

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,12 @@
+---
+# Configuration for probot-stale - https://github.com/probot/stale
+daysUntilStale: 14
+daysUntilClose: 14
+staleLabel: stale
+markComment: >
+  Thank you for your contribution! There was not activity in this pull request
+  recently. To avoid pull requests to pile up, an automated process marked this
+  pull request as stale. It will close this pull request if no further activity
+  occurs. The current policy is available at:
+  https://github.com/nmstate/nmstate/blob/master/.github/stale.yml
+only: pulls


.github/workflows/stale.yml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/stale@v1
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          stale-pr-message: >
+            Thank you for your contribution! To avoid pull requests to pile up,
+            an automated process marked this pull request as stale.
+            It will close this pull request if no further activity occurs.
```
ffmancera: IMO, It would be nice to warn the contributor about the number of days of inactivity allowed. Thanks!

tyll: How important is this for you? I thought about this, too. The problem I identified is that changes to the numbers will cause inconsistencies. If state PRs were already reported, their comment might be out of sync (say we lower the number to 7 instead of 14 days. The comment will still say 14 but the workflow will close it after 7). Also we might forget to update the number in the message since there is no indication that we can use a template for this number. Also contributors could re-open/re-file a new PR anytime anyhow (maybe we should add a message about this somehow, but it would belong in a close statement. What are you comments?

ffmancera: It is not so important for me. Maybe we can document the number in CONTRIBUTING.md, this way if we change the number we are also going to change it in the documentation. Is that fins with you? If not, you can merge this anyway, is not that important for me :-)

.github/workflows/stale.yml
```diff
@@ -0,0 +1,23 @@
...
+          stale-pr-label: 'stale'
+          days-before-state: 14
+          day-bore-close: 14
```
EdDev: Typo?

.github/workflows/stale.yml
```diff
@@ -0,0 +1,23 @@
+---
+name: Mark stale issues and pull requests
```
EdDev: Are we doing this for issues? It does not feel right if we do.

.github/workflows/stale.yml
EdDev: I would set a number for the visitor to be fair. Or at least a link to where this number is specified.

Regarding updates of the number, perhaps you can comment near the number that it needs to be updated in another place.
You can even set a test that makes sure the numbers are in sync.

EdDev:
Aren't you using this app? https://probot.github.io/apps/stale/

tyll:
> Aren't you using this app? https://github.com/marketplace/stale
> This is the official GitHub stale app.

It seemed to me that it did not support PRs but now I found more docs and it also seems to support PRs. What makes it more official, the action is also from Github AFAICS. I do not care about the solution. I will address your concerns once we have an agreement on the solution.

.github/workflows/stale.yml
tyll: Adding a link to the config file works for me.

.github/workflows/stale.yml
tyll: Currently not, because it does not support both. However, it is not a bad idea in general.

EdDev:
> It seemed to me that it did not support PRs but now I found more docs and it also seems to support PRs. What makes it more official, the action is also from Github AFAICS. I do not care about the solution. I will address your concerns once we have an agreement on the solution.

On the marketplace the app is marked as “verified” and it does not require the “how” configuration.. it is simpler as far as I can see.
Are they both maintained by GitHub?

tyll:
> > It seemed to me that it did not support PRs but now I found more docs and it also seems to support PRs. What makes it more official, the action is also from Github AFAICS. I do not care about the solution. I will address your concerns once we have an agreement on the solution.
> 
> On the marketplace the app is marked as “verified” and it does not require the “how” configuration.. it is simpler as far as I can see.
> Are they both maintained by GitHub?

The stale action is also marked as "verified creator" - whatever this actually means: https://github.com/marketplace/actions/close-stale-issues
The initial commit to the repo behind is is done from GitHub staff. The action has the "how" configuration because it runs on GitHub infrastructure as an Action and the other thing runs on external infrastructure (external to the GitHub git forge) that is also owned by GitHub the organization. If you prefer the bot (external infra) over the action, it is fine by me, too.

EdDev:
> The stale action is also marked as "verified creator" - whatever this actually means: https://github.com/marketplace/actions/close-stale-issues
The initial commit to the repo behind is is done from GitHub staff.

Ack

> If you prefer the bot (external infra) over the action, it is fine by me, too.

I prefer to have lower maintenance burden and from a first look the App looked simpler to me.
But I do not have any special preference beyond that.
Updating the commit message with the intent/reasoning will suffice for me, anything else is nice-to-have and should not block this addition.

tyll:
New commit message:

GitHub: Add stale probot configuration.

To ensure that old/untouched pull requests do not pile up, configure the
probot stale bot to mark pull requests without activity in the last 14
days as stale and close them after 14 more days if there is no further
activity. The pull requests are marked with the `stale` label.

There are two possible solutions for this problem, the stale probot,
that is used here and the GitHub stale action. The stale probot seems to
unmark pull requests once there is activity while the GitHub stale
action requires the label to be removed manually. Since the latter means
more work, use the probot solution.

tyll:
@ffmancera please re-visit this

tyll:
I ignore the Travis error because this change does not touch any code and the error in Travis is unrelated.

PR #717, opened 2020-01-30
nm, ovs: Properly delete ovs-port profiles

EdDev:
When an interface that is connected to an OVS bridge is removed, there
is a need to remove also the ovs-port profile which implicitly exists.

This change removes the ovs-port profile toghether with the enslaved
interface.

This PR replaces #511 .

### DIFF
--- a/libnmstate/nm/applier.py
+++ b/libnmstate/nm/applier.py
@@ -274,6 +274,10 @@ def _get_affected_devices(iface_state):
             devs += bridge.get_slaves(nmdev)
         elif iface_type == bond.BOND_TYPE:
             devs += bond.get_slaves(nmdev)
+
+        ovs_port_dev = ovs.get_port_by_slave(nmdev)
+        if ovs_port_dev:
+            devs.append(ovs_port_dev)
     return devs
 
 
--- a/libnmstate/nm/applier.py
+++ b/libnmstate/nm/applier.py
@@ -266,9 +266,7 @@ def _get_affected_devices(iface_state):
         iface_type = iface_state[Interface.TYPE]
         if iface_type == ovs.BRIDGE_TYPE:
             port_slaves = ovs.get_slaves(nmdev)
-            iface_slaves = [
-                iface for port in port_slaves for iface in ovs.get_slaves(port)
-            ]
+            iface_slaves = _get_ovs_interfaces(iface_state)
             devs += port_slaves + iface_slaves
         elif iface_type == LB.TYPE:
             devs += bridge.get_slaves(nmdev)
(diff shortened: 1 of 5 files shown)


tests/integration/testlib/ovslib.py
```diff
@@ -100,6 +109,32 @@ def create(self):
...
+    """
+    Given an interface name, return the OVS port device/profile name.
+    This assumes NM as the backend.
+    """
+    nmdev = device.get_device_by_name(iface_name)
+    ovs_port_dev = ovs.get_port_by_slave(nmdev)
+    return ovs_port_dev.get_iface()
+
+
+def list_profiles_names(iface_name):
```
tyll: This helper does not seem to be specific to OVS, so it should be somewhere else. In other tests we use something with "show only" in its name to get the profiles. Despite its name, this is actually returning device names as it seems. Why is this special? Shouldn't it just use the other method?

EdDev: > This helper does not seem to be specific to OVS, so it should be somewhere else.
> In other tests we use something with "show only" in its name to get the profiles. Despite its name, this is actually returning device names as it seems. Why is this special? Shouldn't it just use the other method?

`show*` returns existing devices while this one returns defined profiles (including inactive ones).
It does not really matter what it returns, it could just become an assert helper.
I was mainly concern about not exposing `nm` on the tests directly.

I will create an `nmlib` module and push it there.

EdDev:
There still seem to be a problem left.
When deleting the OVS bridge, the port device is also removed, but not the profile.
We need to support the deletion of an inactive profile for such cases.

libnmstate/nm/applier.py
```diff
@@ -266,17 +266,38 @@ def _get_affected_devices(iface_state):
...
     return devs
 
 
+def _get_ovs_interfaces(iface_state):
+    """
+    Report a list of all interfaces that are connected to the OVS bridge.
+    This is a workaround solution for https://bugzilla.redhat.com/1797478.
+    """
+    ifaces = [
+        p[Interface.NAME]
```
ffmancera: Why `p`? I think the variable name should be more precise..

tests/integration/ovs_test.py
```diff
@@ -31,6 +31,8 @@
 
 from .testlib import assertlib
 from .testlib import cmdlib
+from .testlib.nmlib import get_ovs_port_by_slave
+from .testlib.nmlib import list_profiles_by_iface_name
```
ffmancera: I think it would be better to do `from .testlib import nmlib`. Not a must, just a nit. :-)

libnmstate/nm/applier.py
EdDev: It enclosed on a list comprehension and very (very) close to what it iterates over.
Do you find it hard to read and understand?

tests/integration/ovs_test.py
EdDev: Done

libnmstate/nm/applier.py

EdDev:
The only test that did not ran yet is the py37 unit test.
This is passing locally and the py36 and py38 are passing, therefore, merging as is to save time.

PR #718, opened 2020-01-30
Nmstatectl: Use keyword arguments

tyll:
Use keyword arguments for API functions in case they will become
keyword-only in the future.


tyll:
FYI: Forgot to add a change to the test file initially.

PR #719, opened 2020-01-31
integ: download plugin packages before running the tests

ffmancera:
This patch is going to avoid connectivity issues during the package
installation when running tests.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2018-2019 Red Hat, Inc.
+# Copyright (c) 2018-2020 Red Hat, Inc.
 #
 # This file is part of nmstate
 #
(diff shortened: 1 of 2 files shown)


tests/integration/conftest.py
```diff
@@ -102,7 +108,6 @@ def _get_osname():
 
 
 @pytest.fixture(scope="session", autouse=True)
-def download_nm_team_plugin():
-    cmdlib.exec_cmd(
-        ("dnf", "install", "-y", "--downloadonly", "NetworkManager-team",)
-    )
+def download_packages():
+    for package in PACKAGE_LIST:
+        cmdlib.exec_cmd(("dnf", "install", "-y", "--downloadonly", package,))
```
tyll: No need for a `for` loop, dnf can download multiple packages at once, which is faster. Using
```
cmdlib.exec_cmd(("dnf", "install", "-y", "--downloadonly",) + PACKAGE_LIST)
```
should work. Actually since the package list is only used here, you can probably just write the packages here and drop the constant. I do not see the value of the constant here.

ffmancera: Done, thanks!

PR #723, opened 2020-02-03
nm, applier, ovs: Hash the OVS port device name

EdDev:
nm, applier, ovs: Hash the OVS port device name

The OVS port device name is limited by NM to the kernel interface name
length (15 chars) even though there is no real underline interface. [1]

As a workaround, this change hashes the profile name into a max of 15 hex
characters and uses it as the profile interface/device name.
The interface name is creaded as part of the OVS port (proxy) interface
state and passed on through the setup flow by adding it as metadata to
the state.

[1] https://bugzilla.redhat.com/1788432

Depends on #717 
An alternative to #678

### DIFF
--- a/libnmstate/nm/applier.py
+++ b/libnmstate/nm/applier.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2018-2019 Red Hat, Inc.
+# Copyright (c) 2018-2020 Red Hat, Inc.
 #
 # This file is part of nmstate
 #
@@ -17,6 +17,8 @@
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 #
 
+import base64
+import hashlib
 import itertools
 
 from libnmstate.error import NmstateValueError
@@ -43,11 +45,15 @@ from . import vxlan
 from . import wired
 
 
+MAXIMUM_INTERFACE_LENGTH = 15
+
 MASTER_METADATA = "_master"
 MASTER_TYPE_METADATA = "_master_type"
 MASTER_IFACE_TYPES = ovs.BRIDGE_TYPE, bond.BOND_TYPE, LB.TYPE
 
 BRPORT_OPTIONS_METADATA = "_brport_options"
+
+IFACE_NAME_METADATA = "_iface_name"
 
 
 def create_new_ifaces(con_profiles):
@@ -246,6 +252,10 @@ def _get_new_ifaces(con_profiles):
         ifname = con_profile.devname
         nmdev = device.get_device_by_name(ifname)
         if not nmdev:
+            # When the profile id is different from the iface name, use the
+            # profile id.
+            if ifname != con_profile.con_id:
+                ifname = con_profile.con_id
             ifaces_without_device.add(ifname)
     return ifaces_without_device
 
@@ -325,21 +335,38 @@ def prepare_proxy_ifaces_desired_state(ifaces_desired_state):
         new_ifaces_desired_state.append(port_iface_desired_state)
         # The "visible" slave/interface needs to point to the port profile
         iface_desired_state[MASTER_METADATA] = port_iface_desired_state[
-            Interface.NAME
+            IFACE_NAME_METADATA
         ]
         iface_desired_state[MASTER_TYPE_METADATA] = ovs.PORT_TYPE
     return new_ifaces_desired_state
 
 
 def _create_ovs_port_iface_desired_state(iface_desired_state, port_options):
+    port_name = ovs.PORT_PROFILE_PREFIX + iface_desired_state[Interface.NAME]
     return {
-        "name": ovs.PORT_PROFILE_PREFIX + iface_desired_state[Interface.NAME],
-        "type": ovs.PORT_TYPE,
-        "state": iface_desired_state[Interface.STATE],
+        Interface.NAME: port_name,
+        Interface.TYPE: ovs.PORT_TYPE,
+        Interface.STATE: iface_desired_state[Interface.STATE],
+        OvsB.OPTIONS_SUBTREE: port_options,
         MASTER_METADATA: iface_desired_state[MASTER_METADATA],
         MASTER_TYPE_METADATA: iface_desired_state[MASTER_TYPE_METADATA],
-        "options": port_options,
+        IFACE_NAME_METADATA: _generate_hash_iface_name(port_name),
     }
+
+
+def _generate_hash_iface_name(name):
+    """
+    Given a name, generate a hash string that may be used as an interface name.
+
+    The OVS port does not have an actual interface in the kernel, but
+    its length is still limited by NM connection.interface-name.
+    https://bugzilla.redhat.com/1788432
+    As a workaround, hash the full name and use it as an alternative device
+    name (assigned as a metadata). The profile/connection name format has
+    an OVS PORT prefix
+    """
+    name_ = base64.urlsafe_b64encode(hashlib.sha256(name.encode()).digest())
+    return name_[:MAXIMUM_INTERFACE_LENGTH].decode()
 
 
 def _build_connection_profile(iface_desired_state, base_con_profile=None):
@@ -362,9 +389,13 @@ def _build_connection_profile(iface_desired_state, base_con_profile=None):
     if base_profile:
         con_setting.import_by_profile(base_con_profile)
     else:
+        iface_name = (
+            iface_desired_state.get(IFACE_NAME_METADATA)
+            or iface_desired_state[Interface.NAME]
+        )
         con_setting.create(
             con_name=iface_desired_state[Interface.NAME],
-            iface_name=iface_desired_state[Interface.NAME],
+            iface_name=iface_name,
             iface_type=iface_type,
         )
     master = iface_desired_state.get(MASTER_METADATA)
--- a/libnmstate/nm/connection.py
+++ b/libnmstate/nm/connection.py
@@ -125,7 +125,8 @@ class ConnectionProfile:
 
     @property
     def con_id(self):
-        return self._con_id
+        con_id = self._con_profile.get_id() if self._con_profile else None
+        return self._con_id or con_id
 
     @con_id.setter
     def con_id(self, connection_id):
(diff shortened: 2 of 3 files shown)


libnmstate/nm/applier.py
```diff
@@ -304,23 +338,39 @@ def prepare_proxy_ifaces_desired_state(ifaces_desired_state):
...
+    Given a name, generate a hash string that may be used as an interface name.
+
+    The OVS port does not have an actual interface in the kernel, but
+    its length is still limited by NM connection.interface-name.
+    https://bugzilla.redhat.com/1788432
+    As a workaround, hash the full name and use it as an alternative device
+    name (assigned as a metadata). The profile/connection name format has
+    an OVS PORT prefix
+    """
+    return f"{hashlib.md5(name.encode()).hexdigest():.15}"
```
tyll: Could you please use sha256 here? Even though using md5 is not a problem here it requires people to evaluate if it is ok or not. Using sha256 should not cause any trouble. Also you could use base64 instead of hex to reduce the probability of a collisions slightly more.

EdDev: sha256 for a name hashing? That sounds an overkill. Anyway, I do not mind, it cannot do anything wrong except questioning why the coder exaggerated.

I am not sure I understand what you suggests about `base64`.

tyll: Currently the hash value is encoded in hexadecimal, therefore 15 characters mean its 15 * 4 = 60 bits, with base64 it would be 15 * 6 = 90 bits.
Also I just noticed that using a format string seems to be overkill, you could just slice it with '[:15]` and the 15 should probably be a constant like `MAX_INTERFACE_NAME_LENGTH` to clarify why it is 15.

EdDev: You have 4 comments on a single line and I am not sure if I understand them all.
It can help if you present (through code) the alternative and avoid misunderstandings.

For example, I do not understand how you represent a char of 6 bit using an alphanumeric representation. A hex seems the simplest and easiest.

tyll: Proposal:

```
MAXIMUM_INTERFACE_LENGTH=15
return base64.b64encode(hashlib.sha256(name.encode()).digest(), altchars=b"-_")[:MAXIMUM_INTERFACE_LENGTH]
```

EdDev: Done.

libnmstate/nm/applier.py
```diff
@@ -270,13 +280,41 @@ def _get_affected_devices(iface_state):
...
 
 
+def _get_ovs_bridge_port_devices_from_state(iface_state):
+    """
+    Report a list of all ovs ports and interfaces that are connected to the
+    OVS bridge.
+    This is a workaround solution for https://bugzilla.redhat.com/1797478.
+    """
+    ifaces = [
+        p[Interface.NAME]
```
ffmancera: Why `p`? I think the variable name should be more precise..

tests/integration/ovs_test.py
```diff
@@ -144,6 +144,22 @@ def test_vlan_as_ovs_bridge_slave(vlan_on_eth1):
         assertlib.assert_state_match(state)
 
 
+@pytest.mark.xfail(
+    raises=(NmstateLibnmError, AssertionError),
+    reason="https://bugzilla.redhat.com/1724901",
+)
+def test_ovs_interface_with_max_length_name():
+    bridge = Bridge(BRIDGE1)
+    ovs_interface_name = "ovs123456789012"
```
ffmancera: Have you also tried with more than 15 chars?

tests/integration/ovs_test.py
```diff
@@ -144,6 +144,22 @@ def test_vlan_as_ovs_bridge_slave(vlan_on_eth1):
         assertlib.assert_state_match(state)
 
 
+@pytest.mark.xfail(
+    raises=(NmstateLibnmError, AssertionError),
+    reason="https://bugzilla.redhat.com/1724901",
+)
```
ffmancera: This is not needed right? The test should pass.. Am I missing something? Thanks!

EdDev: https://bugzilla.redhat.com/1724901 is about OVS instability in general, it waits for NM to fix their other problems

tests/integration/ovs_test.py
EdDev: No, that should fail on the internal interface.

tests/integration/ovs_test.py
ffmancera: Oh, sorry. Looks good then

tests/integration/ovs_test.py
ffmancera: Nvm, thank you for the explanation :-)

libnmstate/nm/applier.py
```diff
@@ -246,6 +252,10 @@ def _get_new_ifaces(con_profiles):
         ifname = con_profile.devname
         nmdev = device.get_device_by_name(ifname)
         if not nmdev:
+            # When the profile id is different from the iface name, use the
```
tyll: AFAIU, you could just always use `con_profile.con_id` instead. But then it is also strange that the method is named liked it will return interfaces but it can also return connection IDs.

libnmstate/nm/applier.py
```diff
@@ -325,23 +335,40 @@ def prepare_proxy_ifaces_desired_state(ifaces_desired_state):
...
         iface_desired_state[MASTER_METADATA] = port_iface_desired_state[
-            Interface.NAME
+            IFACE_NAME_METADATA
         ]
         iface_desired_state[MASTER_TYPE_METADATA] = ovs.PORT_TYPE
     return new_ifaces_desired_state
 
 
 def _create_ovs_port_iface_desired_state(iface_desired_state, port_options):
+    port_name = ovs.PORT_PROFILE_PREFIX + iface_desired_state[Interface.NAME]
```
tyll: This seems to combine a lot of changes to use the proper constants with the actual change of adding the metadata. Would be nicer to separate it. The constants change could be easily reviewed and merged.

libnmstate/nm/connection.py
```diff
@@ -125,7 +125,8 @@ def nmdevice(self, dev):
 
     @property
     def con_id(self):
-        return self._con_id
+        con_id = self._con_profile.get_id() if self._con_profile else None
```
tyll: This looks like a workaround and it appears to me that `self._con_id` should be properly initialized when `self._con_profile` is set. Or is there a use case where `self._con_id` needs to be different than `self._con_profile.get_id()`?

libnmstate/nm/applier.py
```diff
@@ -325,23 +335,40 @@ def prepare_proxy_ifaces_desired_state(ifaces_desired_state):
...
-        "type": ovs.PORT_TYPE,
-        "state": iface_desired_state[Interface.STATE],
+        Interface.NAME: port_name,
+        Interface.TYPE: ovs.PORT_TYPE,
+        Interface.STATE: iface_desired_state[Interface.STATE],
+        OvsB.OPTIONS_SUBTREE: port_options,
         MASTER_METADATA: iface_desired_state[MASTER_METADATA],
         MASTER_TYPE_METADATA: iface_desired_state[MASTER_TYPE_METADATA],
-        "options": port_options,
+        IFACE_NAME_METADATA: _generate_hash_iface_name(port_name),
```
tyll: This constant name does not really fit IMHO. How about `NM_IFACE_NAME_OVERRIDE`? The current name does not make it clear to me that it is used to override the default name used by NM in its profiles.

tests/integration/ovs_test.py
```diff
@@ -143,6 +143,22 @@ def test_vlan_as_ovs_bridge_slave(vlan_on_eth1):
...
+def test_ovs_interface_with_max_length_name():
+    bridge = Bridge(BRIDGE1)
+    ovs_interface_name = "ovs123456789012"
```
tyll: Do we need a special test for this or could be just use this name in an existing test?

libnmstate/nm/connection.py
EdDev: There are multiple places where the profile is set (create, imports, explicitly setting it).
I see no problem in adding a simple logic when someone asks for this info: Either fetch the explicitly set value or read it from the profile itself.
Consider that a user of this object may ask about this ID even before the actual profile is created.

tests/integration/ovs_test.py
EdDev: There is only one such test and it already does several things. (`test_create_and_remove_ovs_bridge_with_internal_port_static_ip_and_mac`)

I do not mind using this long name there of you prefer, the only disadvantage is that it will overload it.

EdDev:
> I am worried about the code changes might impact non-OVS interfaces/connections/profiles.

We need to trust the tests, I see no other way.

> It could be better if we add some interface type check, just a nit, not must.

If there are bugs in the logic, we will find them faster when the solution is generic.
This can also be resolved by keeping the id and name identical using the hash (on the expense of debugging).
Ideally NM will remove the name length limitation and we can remove this mess out.

cathay4t:
> > I am worried about the code changes might impact non-OVS interfaces/connections/profiles.
> 
> We need to trust the tests, I see no other way.
> 
> > It could be better if we add some interface type check, just a nit, not must.
> 
> If there are bugs in the logic, we will find them faster when the solution is generic.
> This can also be resolved by keeping the id and name identical using the hash (on the expense of debugging).
> Ideally NM will remove the name length limitation and we can remove this mess out.

Sounds good to me.

libnmstate/nm/applier.py
tyll: I mean to use this code:

```python
if not nmdev:
    ifaces_without_device.add(con_profile.con_id)
```

AFAIU, it has the identical result as the current code. To me the current code does not make this more clear.

tyll: Line 252 can then be removed.

tyll: I guess this function should be named `_get_profiles_for_new_interfaces()`.

PR #725, opened 2020-02-03
nm.nmclient: add timeout per action

ffmancera:
When doing a lot of operations in a single transaction the timeout
expires and the transaction fails. Adding a timeout per action instead
of per transaction solves this issue.

### DIFF
--- a/automation/run-tests.sh
+++ b/automation/run-tests.sh
@@ -13,9 +13,19 @@ TEST_TYPE_LINT="lint"
 TEST_TYPE_UNIT_PY36="unit_py36"
 TEST_TYPE_UNIT_PY38="unit_py38"
 TEST_TYPE_INTEG="integ"
+TEST_TYPE_INTEG_SLOW="integ_slow"
 
 FEDORA_IMAGE_DEV="nmstate/fedora-nmstate-dev"
 CENTOS_IMAGE_DEV="nmstate/centos8-nmstate-dev"
+
+PYTEST_OPTIONS="--verbose --verbose \
+        --log-level=DEBUG \
+        --log-date-format='%Y-%m-%d %H:%M:%S' \
+        --log-format='%(asctime)s %(filename)s:%(lineno)d %(levelname)s %(message)s' \
+        --durations=5 \
+        --cov /usr/lib/python*/site-packages/libnmstate \
+        --cov /usr/lib/python*/site-packages/nmstatectl \
+        --cov-report=term"
 
 : ${CONTAINER_CMD:=docker}
 
@@ -109,15 +119,20 @@ function run_tests {
         container_exec "
           cd $CONTAINER_WORKSPACE &&
           pytest \
-            --verbose --verbose \
-            --log-level=DEBUG \
-            --log-date-format='%Y-%m-%d %H:%M:%S' \
-            --log-format='%(asctime)s %(filename)s:%(lineno)d %(levelname)s %(message)s' \
-            --durations=5 \
-            --cov /usr/lib/python*/site-packages/libnmstate \
-            --cov /usr/lib/python*/site-packages/nmstatectl \
+            $PYTEST_OPTIONS \
             --cov-report=html:htmlcov-integ \
-            --cov-report=term \
+            tests/integration \
+            ${nmstate_pytest_extra_args}"
+    fi
+
+    if [ $TEST_TYPE == $TEST_TYPE_ALL ] || \
+       [ $TEST_TYPE == $TEST_TYPE_INTEG_SLOW ];then
+        container_exec "
+          cd $CONTAINER_WORKSPACE &&
+          pytest \
+            $PYTEST_OPTIONS \
+            --cov-report=html:htmlcov-integ_slow \
+            -m slow --runslow \
             tests/integration \
             ${nmstate_pytest_extra_args}"
     fi
@@ -238,6 +253,7 @@ while true; do
         echo "     * $TEST_TYPE_FORMAT"
         echo "     * $TEST_TYPE_LINT"
         echo "     * $TEST_TYPE_INTEG"
+        echo "     * $TEST_TYPE_INTEG_SLOW"
         echo "     * $TEST_TYPE_UNIT_PY36"
         echo "     * $TEST_TYPE_UNIT_PY38"
         echo -n "--customize allows to specify a command to customize the "
--- a/libnmstate/nm/checkpoint.py
+++ b/libnmstate/nm/checkpoint.py
@@ -21,6 +21,8 @@ import logging
 
 import dbus
 
+from gi.repository import GLib
+
 from libnmstate.nm.nmclient import nmclient_context
 from libnmstate.nm.nmclient import client
 
@@ -123,6 +125,27 @@ class CheckPoint:
             if e.get_dbus_name() == NM_PERMISSION_DENIED:
                 raise NMCheckPointPermissionError(str(e))
             raise NMCheckPointCreationError(str(e))
+
+        GLib.timeout_add(
+            self._timeout * 0.5, self._refresh_checkpoint_timeout, None,
+        )
+
+    def _refresh_checkpoint_timeout(self, data):
+        cancellable, cb, cb_data = (None, None, None)
+        nm_client = client()
+
+        # Activation finished, _delete_client() invoked, which then invoke
+        # _nmclient.delete_client(). Then it set self._client as None. The main
+        # context got event from timeout, then invoke
+        # _refresh_checkpoint_timeout(). So that time, the _nmclient is not
+        # None while _nmclient.client is None.
+        if nm_client:
+            nm_client.checkpoint_adjust_rollback_timeout(
+                self._dbuspath, self._timeout, cancellable, cb, cb_data
+            )
+            return GLib.SOURCE_CONTINUE
+        else:
+            return GLib.SOURCE_REMOVE
 
     def destroy(self):
         try:
--- a/tests/integration/timeout_test.py
+++ b/tests/integration/timeout_test.py
@@ -1,0 +1,58 @@
+#
+# Copyright (c) 2020 Red Hat, Inc.
+#
+# This file is part of nmstate
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+
+import time
+
+import pytest
+
+import libnmstate
+from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceType
+from libnmstate.schema import InterfaceState
+from libnmstate.schema import LinuxBridge
+from libnmstate.schema import VLAN
+
+
+@pytest.mark.slow
+def test_lot_of_vlans_with_bridges(eth1_up):
+    interfaces = []
+    for i in range(100, 400):
+        interfaces.extend(
+            [
+                {
+                    Interface.NAME: "vlan." + str(i),
+                    Interface.TYPE: InterfaceType.VLAN,
+                    Interface.STATE: InterfaceState.UP,
+                    VLAN.CONFIG_SUBTREE: {VLAN.BASE_IFACE: "eth1", VLAN.ID: i},
+                },
+                {
+                    Interface.NAME: "linux-br" + str(i),
+                    Interface.TYPE: InterfaceType.LINUX_BRIDGE,
+                    Interface.STATE: InterfaceState.UP,
+                    LinuxBridge.CONFIG_SUBTREE: {
+                        LinuxBridge.PORT_SUBTREE: [
+                            {LinuxBridge.Port.NAME: "vlan." + str(i)}
+                        ]
+                    },
+                },
+            ]
+        )
+    checkpoint = libnmstate.apply({Interface.KEY: interfaces}, commit=False)
+    libnmstate.rollback(checkpoint)
+    time.sleep(5)
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,8 @@ env:
           use_coveralls=false
         - CONTAINER_IMAGE=nmstate/fedora-nmstate-dev
           testflags="--test-type integ --pytest-args='-x'"
+        - CONTAINER_IMAGE=nmstate/centos8-nmstate-dev
+          testflags="--test-type integ_slow --pytest-args='-x'"
         - CONTAINER_IMAGE=nmstate/centos8-nmstate-dev-nm-1.22-git
           testflags="--test-type integ --pytest-args='-x'"
         - CONTAINER_IMAGE=nmstate/centos8-nmstate-dev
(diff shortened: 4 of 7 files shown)


ffmancera:
This fixes the transaction timeout issue. But we also need to take care of the checkpoint timeout. I would need some suggestions about where should we place `checkpoint_adjust_rollback_timeout()` in order to refresh the checkpoint timeout.

I have been trying multiple things but didn't find a solution. Thanks!

tyll:
> This fixes the transaction timeout issue. But we also need to take care of the checkpoint timeout. I would need some suggestions about where should we place `checkpoint_adjust_rollback_timeout()` in order to refresh the checkpoint timeout.

These are my notes from the meeting on 30 January:

> Checkpoint timeout interferes with peer-action timeout:
> 
> Define that the timeout for Checkpoint means the time after Nmstate finished doing changes to allow the user to revert/commit after Nmstate returned
> 
> Use a timer specific for the checkpoint that will increase the checkpoint timeout regularly if there are still actions pending

So when the checkpoint is created, a new timer needs to be created as well that then does the refreshing in its callback (and creates a new timer).

ffmancera:
> > This fixes the transaction timeout issue. But we also need to take care of the checkpoint timeout. I would need some suggestions about where should we place `checkpoint_adjust_rollback_timeout()` in order to refresh the checkpoint timeout.
> 
> These are my notes from the meeting on 30 January:
> 
> > Checkpoint timeout interferes with peer-action timeout:
> > Define that the timeout for Checkpoint means the time after Nmstate finished doing changes to allow the user to revert/commit after Nmstate returned
> > Use a timer specific for the checkpoint that will increase the checkpoint timeout regularly if there are still actions pending
> 
> So when the checkpoint is created, a new timer needs to be created as well that then does the refreshing in its callback (and creates a new timer).

In addition, in that callback we need to check if the mainloop action queue is empty or not, right? Thanks.

tyll:
> > So when the checkpoint is created, a new timer needs to be created as well that then does the refreshing in its callback (and creates a new timer).
> 
> In addition, in that callback we need to check if the mainloop action queue is empty or not, right? Thanks.

So, this seems to be a complicated question. Actually the condition should be whether we still expect a callback to an action or signal. Not sure if this is easy to implement. AFAIU, even if the mainloop action queue is empty, Nmstate might still wait for an action to call the callback or signal a state change.

It might also be possible to not have a special condition for this. Since once Nmstate considers everything to be done (or if there was a fatal error), it will quit the mainloop, it would also mean that the timer callback will not be callled anymore. So when exiting the mainloop, the timer should probably also be canceled (to avoid any problems with some references since Nmstate is going to remove the NMClient object at some point and it will also wait for the timer to expire AFAIU). At the same time when the mainloop is terminated, the checkpoint timeout needs to be expanded one last time. Otherwise it might be that the checkpoint expires shortly after the mainloop terminated but before Nmstate is done with verification.

ffmancera:
A few comments about the current approach:
* This patch introduces a timeout per action instead of transaction
* Introduce a timer which is refreshing checkpoint timeout

I have not been able to do to the last checkpoint timeout refreshing because I've not been able to get the checkpoint timeout from the mainloop context. Do you have any suggestions for this?

I've tested this with the following code with different ranges (1000, 2000, 3000..):
```
def test_lot_of_ifaces():
    interfaces = []
    for i in range(1000):
        interfaces.append(
            {
                Interface.NAME: "dummy"+str(i),
                Interface.TYPE: InterfaceType.DUMMY,
            }
        )
    checkpoint = libnmstate.apply({Interface.KEY: interfaces}, commit=False)
    libnmstate.rollback(checkpoint)
```

tyll:
> A few comments about the current approach:
> 
> * This patch introduces a timeout per action instead of transaction
> * Introduce a timer which is refreshing checkpoint timeout
> 
> I have not been able to do to the last checkpoint timeout refreshing because I've not been able to get the checkpoint timeout from the mainloop context. Do you have any suggestions for this?

Since we are missing an object like NMApplierContext that could contain this reference, the only hack that I see directly would be to change `_setup_providers()` to store a reference to the checkpoint in the mainloop because that is the only thing that is accessible there IIRC. Maybe @EdDev has a better solution.

Also I just remembered that the checkpoint code might use its own main loop. Not sure if this is a problem with the approach or if it actually makes it unnecessary to update the timeout at the end of the netapplier mainloop.

> I've tested this with the following code with different ranges (1000, 2000, 3000..):
> 
> ```
> def test_lot_of_ifaces():
>     interfaces = []
>     for i in range(1000):
>         interfaces.append(
>             {
>                 Interface.NAME: "dummy"+str(i),
>                 Interface.TYPE: InterfaceType.DUMMY,
>                 Interface.STATE: InterfaceState.UP,
>             }
>         )
>     checkpoint = libnmstate.apply({Interface.KEY: interfaces}, commit=False)
>     libnmstate.rollback(checkpoint)
> ```
> 
> I've not included this test into the patch because it would consume a lot of time when running the tests.

How long does it take? It would be nice to add a test that is skipped by default but could be easily activated by manual testing or in a separate CI job. Ideally it would also use the scenario with bridges and vlans and log the duration for the run.

libnmstate/nm/checkpoint.py
```diff
@@ -124,6 +126,23 @@ def create(self):
...
+        GLib.timeout_add(
+            timeout * 0.5,
+            self._refresh_checkpoint_timeout,
+            (self._dbuspath, timeout),
+        )
+
+    def _refresh_checkpoint_timeout(self, data):
+        dbuspath, timeout = data
+        nm_client = client()
+        if nm_client:
```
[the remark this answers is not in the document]
ffmancera: After investigating I've discovered that sometimes after finishing all the actions and quitting the mainloop, `nm_client` is `None`. I am not sure but I think this happens because the GLib timeout does not are about the NM mainloop.

So, right now, I am not doing nothing. If `nm_client` is `None` it means mainloop quit and therefore I am not refreshing the timeout (because I can't do it indeeed).

I am not 100% sure about this. Thanks!

libnmstate/nm/checkpoint.py
```diff
@@ -124,6 +126,23 @@ def create(self):
...
+            nm_client.checkpoint_adjust_rollback_timeout(
+                dbuspath, timeout, None, None, None
```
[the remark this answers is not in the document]
ffmancera: Sure, will do.

libnmstate/nm/checkpoint.py
```diff
@@ -124,6 +126,23 @@ def create(self):
                 raise NMCheckPointPermissionError(str(e))
             raise NMCheckPointCreationError(str(e))
 
+        GLib.timeout_add(
```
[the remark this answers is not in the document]
ffmancera: Oh, you are right. Thanks

libnmstate/nm/checkpoint.py
```diff
@@ -124,6 +126,23 @@ def create(self):
...
+            timeout * 0.5,
+            self._refresh_checkpoint_timeout,
```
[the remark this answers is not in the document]
ffmancera: What return value? I do not undestand this, thanks.

ffmancera:
> > A few comments about the current approach:
> > 
> > * This patch introduces a timeout per action instead of transaction
> > * Introduce a timer which is refreshing checkpoint timeout
> > 
> > I have not been able to do to the last checkpoint timeout refreshing because I've not been able to get the checkpoint timeout from the mainloop context. Do you have any suggestions for this?
> 
> Since we are missing an object like NMApplierContext that could contain this reference, the only hack that I see directly would be to change `_setup_providers()` to store a reference to the checkpoint in the mainloop because that is the only thing that is accessible there IIRC. Maybe @EdDev has a better solution.
> 

I propose to open a new issue about this and fix it in another PR. This issue is not completely related to this PR. And it needs further investigation. Would you mind to move this to another issue? Thanks :-)

> Also I just remembered that the checkpoint code might use its own main loop. Not sure if this is a problem with the approach or if it actually makes it unnecessary to update the timeout at the end of the netapplier mainloop.
> 
> > I've tested this with the following code with different ranges (1000, 2000, 3000..):
> > ```
> > def test_lot_of_ifaces():
> >     interfaces = []
> >     for i in range(1000):
> >         interfaces.append(
> >             {
> >                 Interface.NAME: "dummy"+str(i),
> >                 Interface.TYPE: InterfaceType.DUMMY,
> >                 Interface.STATE: InterfaceState.UP,
> >             }
> >         )
> >     checkpoint = libnmstate.apply({Interface.KEY: interfaces}, commit=False)
> >     libnmstate.rollback(checkpoint)
> > ```
> > 
> > 
> > I've not included this test into the patch because it would consume a lot of time when running the tests.
> 
> How long does it take? It would be nice to add a test that is skipped by default but could be easily activated by manual testing or in a separate CI job. Ideally it would also use the scenario with bridges and vlans and log the duration for the run.

In my system it takes like 3~4 minutes for 1000 dummy interfaces. As you wish, I could add an optional test or something like this.

libnmstate/nm/checkpoint.py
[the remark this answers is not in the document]
tyll: Oh, this comment was meant for `nm_client.checkpoint_adjust_rollback_timeout`, not sure how it got here.

libnmstate/nm/checkpoint.py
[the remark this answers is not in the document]
tyll: @cathay4t can you help? When is `nm_client` supposed to be `None`?

libnmstate/nm/checkpoint.py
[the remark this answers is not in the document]
ffmancera: Done! Thanks

libnmstate/nm/checkpoint.py
```diff
@@ -124,6 +126,23 @@ def create(self):
...
+            (self._dbuspath, timeout),
+        )
+
+    def _refresh_checkpoint_timeout(self, data):
+        dbuspath, timeout = data
```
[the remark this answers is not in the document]
ffmancera: Done, thanks!

libnmstate/nm/checkpoint.py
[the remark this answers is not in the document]
ffmancera: Done, thanks!

libnmstate/nm/checkpoint.py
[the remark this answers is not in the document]
cathay4t: It's a long story:

 * Activation finished, `_delete_client()` invoked, which then invoke `_nmclient.delete_client()`.
 * In `_nmclient.delete_client()` it set `self._client` as None.  iterating the main context to wait NM.Client clean up.
 * The main context got event from timeout, then invoke `_refresh_checkpoint_timeout()`. So that time, the `_nmclient` is not None while `_nmclient.client` is None.

That's why.

You code has many problems:
 * Never do `GLib.timeout_destroy()`.
 * Created a lot lot timeout event by recursively calling `GLib.timeout_add()`.
 * Need correct handling above problem.

ffmancera:
@cathay4t I think the current approach fixes the problem that you have described. When the nmclient is `None` the callback returns  `False` and according to `GLib.timeout_add()` documentation, the timeout event should be destroyed automatically.

libnmstate/nm/checkpoint.py
```diff
@@ -124,6 +126,21 @@ def create(self):
...
+    def _refresh_checkpoint_timeout(self, data):
+        cancellable, cb, cb_data = (None, None, None)
+        nm_client = client()
+        if nm_client:
+            nm_client.checkpoint_adjust_rollback_timeout(
+                self._dbuspath, self._timeout, cancellable, cb, cb_data
+            )
+            return True
+        else:
+            return False
```
cathay4t: nit: Can we use G_SOURCE_CONTINUE and G_SOURCE_REMOVE ?

libnmstate/nm/checkpoint.py
```diff
@@ -124,6 +126,21 @@ def create(self):
...
+
+    def _refresh_checkpoint_timeout(self, data):
+        cancellable, cb, cb_data = (None, None, None)
+        nm_client = client()
+        if nm_client:
+            nm_client.checkpoint_adjust_rollback_timeout(
+                self._dbuspath, self._timeout, cancellable, cb, cb_data
+            )
+            return True
+        else:
```
cathay4t: Please add a comment here explaining why NM.Client might be None here.

libnmstate/nm/checkpoint.py
```diff
@@ -124,6 +126,21 @@ def create(self):
                 raise NMCheckPointPermissionError(str(e))
             raise NMCheckPointCreationError(str(e))
 
+        GLib.timeout_add(
+            self._timeout * 0.5, self._refresh_checkpoint_timeout, None,
+        )
```
cathay4t: This basicly keep refreshing the checkpoint timeout tile exception happens or checkpoint destroyed.

How about we just set the checkpoint timeout as 0 and depend on `CheckPoint.__del__()` to destroy this checkpoint?

Note: `__del__` is different from `__exit__`. The later is for `with` context. The `__dell__` is invoked when class destroyed( including context exit).

libnmstate/nm/checkpoint.py
cathay4t: Should be `GLib.SOURCE_CONTINUE` in python

libnmstate/nm/checkpoint.py
ffmancera: If we do this and for some unexpected reason nmstate crash then the checkpoint is going to stay there forever. We discussed this and we decided to avoid this situation, right? @tyll @EdDev 

Thanks!

tyll: Yes, an unlimited checkpoint would be bad.

libnmstate/nm/checkpoint.py
ffmancera: I've adapted your comment about this. Thanks

libnmstate/nm/checkpoint.py
ffmancera: Done, thanks!

ffmancera:
Integration test added but it is skipped by default.

tests/integration/timeout_test.py
```diff
@@ -0,0 +1,55 @@
...
+
+import libnmstate
+from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceType
+from libnmstate.schema import InterfaceState
+from libnmstate.schema import LinuxBridge
+from libnmstate.schema import VLAN
+
+
+@pytest.mark.skip(reason="This test would take a lot of time.")
```
cathay4t: Please use the similar way like https://github.com/nmstate/nmstate/pull/776 did for pytest slow marker.

cathay4t: Please also use https://docs.travis-ci.com/user/customizing-the-build#Fast-Finishing which allow  travis to mark test as passed as soon as mandatory test passed.

ffmancera: Done, thanks!

automation/run-tests.sh
```diff
@@ -13,10 +13,22 @@ TEST_TYPE_LINT="lint"
 TEST_TYPE_UNIT_PY36="unit_py36"
 TEST_TYPE_UNIT_PY38="unit_py38"
 TEST_TYPE_INTEG="integ"
+TEST_TYPE_INTEG_SLOW="integ_slow"
 
 FEDORA_IMAGE_DEV="nmstate/fedora-nmstate-dev"
 CENTOS_IMAGE_DEV="nmstate/centos8-nmstate-dev"
 
+PYTEST_OPTIONS="
+    --verbose --verbose
```
cathay4t: This does not work as expected.

https://travis-ci.com/nmstate/nmstate/jobs/286905321#L895

Looks like we need a ending `\`

ffmancera: Fixed! Thanks!

tests/integration/timeout_test.py
```diff
@@ -0,0 +1,58 @@
...
+                    Interface.STATE: InterfaceState.UP,
+                    LinuxBridge.CONFIG_SUBTREE: {
+                        LinuxBridge.PORT_SUBTREE: [
+                            {LinuxBridge.Port.NAME: "vlan." + str(i)}
+                        ]
+                    },
+                },
+            ]
+        )
+    checkpoint = libnmstate.apply({Interface.KEY: interfaces}, commit=False)
```
cathay4t: Can you add a small `ip link|grep vlan` check to make sure we got 300 VLANs created?

ffmancera: We are supposed to have 300 VLANs and 1 Linux Bridge per VLAN. If not, this should have failed during the verification. I would prefer to do not add this additional step. Thanks!

cathay4t:
@EdDev This PR use the `integration slow` code from PR #776 .  Can you take a look also? Thank you.

ffmancera:
Rebased! Thanks!

PR #739, opened 2020-02-04
netapplier, mainloop: Delay the exit from the mainloop run

EdDev:
Convert the existing delay before the verification step with a delay in
the mainloop graceful termination.


cathay4t:
The CI failure is a bug of NM  which is fixed in NM 1.22.6.

PR #741, opened 2020-02-04
nm, connection: Fix commit callback error logging

EdDev:
In cases where the commit operation has not included a NM device, and a
failure occurs in the callback function, the mainloop quit logging
causes a secondary error.

### DIFF
--- a/libnmstate/nm/connection.py
+++ b/libnmstate/nm/connection.py
@@ -387,20 +387,25 @@ class ConnectionProfile:
     @staticmethod
     def _commit_changes_callback(src_object, result, user_data):
         mainloop, nmdev = user_data
+        devname = src_object.get_interface_name()
         try:
             success = src_object.commit_changes_finish(result)
         except Exception as e:
             if mainloop.is_action_canceled(e):
                 logging.debug("Connection update aborted: error=%s", e)
             else:
+                if nmdev:
+                    devname = nmdev.props.interface
+                    devstate = nmdev.props.state
+                else:
+                    devstate = "absent"
                 mainloop.quit(
                     "Connection update failed: error={}, dev={}/{}".format(
-                        e, nmdev.props.interface, nmdev.props.state
+                        e, devname, devstate
                     )
                 )
             return
 
-        devname = src_object.get_interface_name()
         if success:
             logging.debug("Connection update succeeded: dev=%s", devname)
             mainloop.execute_next_action()


libnmstate/nm/connection.py
```diff
@@ -387,20 +387,26 @@ def _delete_connection_callback(self, src_object, result, user_data):
...
         except Exception as e:
             if mainloop.is_action_canceled(e):
                 logging.debug("Connection update aborted: error=%s", e)
             else:
+                if nmdev:
+                    iface_name = nmdev.props.interface
+                    iface_state = nmdev.props.state
+                else:
+                    iface_name = devname
+                    iface_state = "absent"
```
cathay4t: Why `absent`?

libnmstate/nm/connection.py
```diff
@@ -387,20 +387,26 @@ def _delete_connection_callback(self, src_object, result, user_data):
...
         mainloop, nmdev = user_data
+        devname = src_object.get_interface_name()
         try:
             success = src_object.commit_changes_finish(result)
         except Exception as e:
             if mainloop.is_action_canceled(e):
                 logging.debug("Connection update aborted: error=%s", e)
             else:
+                if nmdev:
+                    iface_name = nmdev.props.interface
```
tyll: iface_name and devname should be the same, shouldn't they? It is confusing to have multiple variables with different names for the same name.
So maybe remove iface_name and renamse iface_state to devstate.

libnmstate/nm/connection.py
```diff
@@ -387,20 +387,26 @@ def _delete_connection_callback(self, src_object, result, user_data):
     @staticmethod
     def _commit_changes_callback(src_object, result, user_data):
         mainloop, nmdev = user_data
```
tyll: Proposal for the commit message:

Handle `nmdev` being `None` by getting the device name from `src_object` for error logging. This fixes the following traceback:

```
  File "/usr/lib/python3.6/site-packages/libnmstate/nm/connection.py", line 398, in _commit_changes_callback
    e, nmdev.props.interface, nmdev.props.state
AttributeError: 'NoneType' object has no attribute 'props'
```

libnmstate/nm/connection.py
EdDev: It is not there... the device.
Any other suggestions?

libnmstate/nm/connection.py
EdDev: Done

libnmstate/nm/connection.py
EdDev: Too much details for a commit message IMO, but if you prefer it, I will not argue.
Thanks!

libnmstate/nm/connection.py
```diff
@@ -387,20 +387,25 @@ def _delete_connection_callback(self, src_object, result, user_data):
...
+        devname = src_object.get_interface_name()
         try:
             success = src_object.commit_changes_finish(result)
         except Exception as e:
             if mainloop.is_action_canceled(e):
                 logging.debug("Connection update aborted: error=%s", e)
             else:
+                if nmdev:
+                    devname = nmdev.props.interface
```
tyll: Does this need to be assigned again here? Is the first assignment not enough?

EdDev: To be consistent, I think we need to take the state and name from the same source.
Taking the name from the device and the name from the profile seems wrong to me.

ffmancera: In my opinion @EdDev is right. If we are taking the state from the device then we should also take the name from the device.

libnmstate/nm/connection.py
ffmancera: I like the details :-)

PR #742, opened 2020-02-04
CHANGELOG: update change log for 0.2.2 release

ffmancera:

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,20 @@
 # Changelog
-## [0.2.1] - 2019-01-13
+## [0.2.2] - 2020-02-04
+### Breaking changes
+ - NetworkManager 1.22 is required now.
+
+### New Features
+ - Team interface support.
+
+### Bug Fixes
+ - Fix file descriptor leak.
+ - Get SR-IOV VFs configuration from kernel instead of reading the NetworkManager configuration.
+ - Fix DHCPv4 with static IPv6.
+ - Wait a short time before verification to allow the network state to stabilize.
+ - Properly delete ovs-port profiles.
+ - Use dhcp-iaid='mac' on dynamic IPv6 configuration.
+
+## [0.2.1] - 2020-01-13
 ### Breaking Changes
  - Deprecations in `libnmstate.schema`:
     - 'LinuxBridge.GROUP_FORWARD_MASK' -> 'LinuxBridge.Options.GROUP_FORWARD_MASK'


CHANGELOG
```diff
@@ -1,4 +1,15 @@
 # Changelog
+## [0.2.2] - 2019-02-03
+### New Features
+ - Team interface support
+
+### Bug Fixes
+ - Fix memory leak of NM.Client
+ - Get SR-IOV VFs configuration from runtime
+ - Fix DHCPv4 with static IPv6
+ - Use LC_TIME for RPM date
```
tyll: This change does not matter for users.

ffmancera: Removed. Thanks!

CHANGELOG
```diff
@@ -1,4 +1,14 @@
 # Changelog
+## [0.2.2] - 2019-02-03
```
tyll: It is 2020 already and it is also not 3 February anymore. 0.2.1 also used the wrong year, please update it as well. :-)

tyll: It should probably contain a "Breaking Change" note that NM 1.22 is now required.

CHANGELOG
```diff
@@ -1,4 +1,14 @@
...
+### New Features
+ - Team interface support
+
+### Bug Fixes
+ - Fix memory leak of NM.Client
```
tyll: I think this was a file descriptor leak, how about `Fix file descriptor leak` - the NM.Client part is an internal detail

CHANGELOG
```diff
@@ -1,4 +1,14 @@
...
+ - Get SR-IOV VFs configuration from runtime
```
tyll: `from kernel instead of reading the NetworkManager configuration`

CHANGELOG
```diff
@@ -1,4 +1,14 @@
...
+ - Fix DHCPv4 with static IPv6
+ - Learn the OVS Bridge ports and interfaces from NetworkManager
```
tyll: I am not sure what this really means and I guess users will not know, too.

CHANGELOG
```diff
@@ -1,4 +1,14 @@
...
+
```
tyll: Missing is something like:
- Wait a short time before verification to allow the network state to stabilize

CHANGELOG
ffmancera: Maybe that is something not important to users, it just means that now we are using libnm functions to get all bridge ports..

CHANGELOG
ffmancera: Ooops, sorry. It seems my mind is not in 2020 yet :-)

CHANGELOG
ffmancera: Sure, done thanks!

CHANGELOG

CHANGELOG
ffmancera: Added. Thank you for point it out :)

PR #744, opened 2020-02-04
VERSION: update to 0.2.3

ffmancera:


tyll:
This CI already showed that it builds correctly. No need to wait for the full results IMHO.

PR #748, opened 2020-02-05
automation, travis-ci: Separate integration tests

EdDev:
In order to parallelize the integration test run, separate the nm
subpackage tests from the rest and dedicate for them a CI job.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,10 @@ env:
           testflags="--test-type integ --pytest-args='-x'"
           use_codecov=true
           use_coveralls=false
+        - CONTAINER_IMAGE=nmstate/centos8-nmstate-dev
+          testflags="--test-type integ_nm --pytest-args='-x'"
+          use_codecov=true
+          use_coveralls=false
         - CONTAINER_IMAGE=nmstate/fedora-nmstate-dev
           testflags="--test-type integ --pytest-args='-x'"
         - CONTAINER_IMAGE=nmstate/centos8-nmstate-dev
--- a/automation/run-tests.sh
+++ b/automation/run-tests.sh
@@ -14,6 +14,7 @@ TEST_TYPE_UNIT_PY36="unit_py36"
 TEST_TYPE_UNIT_PY37="unit_py37"
 TEST_TYPE_UNIT_PY38="unit_py38"
 TEST_TYPE_INTEG="integ"
+TEST_TYPE_INTEG_NM="integ_nm"
 
 FEDORA_IMAGE_DEV="nmstate/fedora-nmstate-dev"
 CENTOS_IMAGE_DEV="nmstate/centos8-nmstate-dev"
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -9,6 +9,10 @@ env:
           testflags="--test-type integ --pytest-args='-x'"
           use_codecov=true
           use_coveralls=false
+        - CONTAINER_IMAGE=nmstate/centos8-nmstate-dev
+          testflags="--test-type integ_nm --pytest-args='-x'"
+          use_codecov=true
```
tyll: Codecov does not handle multiple uploads from one job that good IIRC. Something is strange with the diff in this PR, it shows the timeout change as diff:
https://codecov.io/gh/nmstate/nmstate/pull/748/diff

automation/run-tests.sh
```diff
@@ -14,6 +14,7 @@ TEST_TYPE_UNIT_PY36="unit_py36"
 TEST_TYPE_UNIT_PY37="unit_py37"
 TEST_TYPE_UNIT_PY38="unit_py38"
 TEST_TYPE_INTEG="integ"
+TEST_TYPE_INTEG_NM="integ_nm"
```
tyll: IMHO it would be nicer to have one target to run all tests locally like `integ` and introduce two new targets that will be called from `integ`.

.travis.yml
```diff
@@ -9,6 +9,10 @@ env:
           testflags="--test-type integ --pytest-args='-x'"
           use_codecov=true
           use_coveralls=false
+        - CONTAINER_IMAGE=nmstate/centos8-nmstate-dev
```
tyll: How about running two containers in the same job instead of adding another job?

EdDev: You assume multiple CPU/s to carry the task which may not be a valid assumption.
It will also mess up the logs, I have bad experience from the oVirt project with this.

.travis.yml
EdDev: I expected this may be a problem. I do not have a solution for this.

automation/run-tests.sh
EdDev: Fine by me.

.travis.yml
tyll: According to Travis, there are two cores available. Also needing more cores assumes that the tests are CPU bound but there is also a lot of sleeping happening which makes the CPU idle and would allow to run something else in parallel.

EdDev: Ack for the CPU/s, but AFAIK the logging of running two containers in parallel has no known solution.

tyll: We could use pytest-xdist to have one pytest command run on both containers: https://pypi.org/project/pytest-xdist/

EdDev: This will require a lot more investigation and work.
At the moment I do not know how it will fit exactly.

EdDev:
Until the coverage report is solved, I do not think we can go forward with this separation.

I am closing this PR for the time being.

PR #750, opened 2020-02-05
tests, integ: skip SR-IOV tests in CI

ffmancera:
SR-IOV tests xfail are consuming a lot of time in Travis CI.

### DIFF
--- a/tests/integration/sriov_test.py
+++ b/tests/integration/sriov_test.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2019 Red Hat, Inc.
+# Copyright (c) 2019-2020 Red Hat, Inc.
 #
 # This file is part of nmstate
 #
@@ -17,6 +17,8 @@
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 #
 
+import os
+
 import pytest
 
 import libnmstate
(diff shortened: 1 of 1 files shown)


ffmancera:
> nit: please update the file date `2019-2020`.

Done, thanks!

tests/integration/sriov_test.py
```diff
@@ -30,6 +32,11 @@
 SRIOV_CONFIG = {Ethernet.SRIOV_SUBTREE: {Ethernet.SRIOV.TOTAL_VFS: 2}}
 
 
+pytestmark = pytest.mark.skipif(
+    os.environ.get("CI") == "true", reason="CI devices do not support SR-IOV",
```
tyll: How about using `not os.path.exists("/sys/class/net/eth1/device/sriov_numvfs")` as a condition? Ideally `eth1` would be a constant here, though.

ffmancera: The thing is that we do not expect to have eth1 with SR-IOV support in CI soon. So I prefer to use the existing CI environment variable. If you are not fine with this I can change it, it is not a must for me :-)

tyll: By actually checking the support on eth1, it would also skip the tests when running locally on a system without SR-IOV support. Actually, it would also allow to remove all the xfails. Also the CI variable has already too many implications (for example it also controls whether the working copy is copied in the container and whether to rebuild the container images), therefore I am not convinced that using it here, too, is a good idea.

PR #756, opened 2020-02-06
schema: modify MAC address minimum limitation

ffmancera:
In some virtual interfaces like 'tun' or 'ip-tunnel' the MAC address
could be less than six octets. In order to fix that, we have modify the
minimum limitation to one octet.

If an user sets an invalid MAC address then NetworkManager is
reporting it clearly.

```
2020-02-06 14:50:20,585 root         ERROR    NM main-loop aborted:
		Connection update failed: error=nm-connection-error-quark:
		802-3-ethernet.cloned-mac-address: '12:BC:B9:2B' is not a valid MAC
		address (7), dev=eth1/<enum NM_DEVICE_STATE_ACTIVATED of type NM.DeviceState>
```

Ref: https://bugzilla.redhat.com/1796838

### DIFF
--- a/libnmstate/schemas/operational-state.yaml
+++ b/libnmstate/schemas/operational-state.yaml
@@ -63,7 +63,7 @@ definitions:
         - down
     mac-address:
       type: string
-      pattern: "^([a-fA-F0-9]{2}:){5,31}[a-fA-F0-9]{2}$"
+      pattern: "^([a-fA-F0-9]{2}:){3,31}[a-fA-F0-9]{2}$"
     bridge-vlan-tag:
       type: integer
       minimum: 0
--- a/tests/lib/schema_validation_test.py
+++ b/tests/lib/schema_validation_test.py
@@ -178,6 +178,34 @@ class TestIfaceCommon:
             libnmstate.validator.validate(default_data)
 
 
+class TestIfaceMacAddress:
+    @pytest.mark.parametrize(
+        "mac_address",
+        [
+            "00:11:22:33",
+            "00:11:22:33:44:55",
+            "80:00:02:08:fe:80:00:00:00:00:00:00:f4:52:14:03:00:8d:52:11",
+        ],
+    )
+    def test_valid_mac_address(self, default_data, mac_address):
+        default_data[INTERFACES][0][Interface.MAC] = mac_address
+        libnmstate.validator.validate(default_data)
+
+    @pytest.mark.parametrize(
+        "mac_address",
+        [
+            "00:11:22",
+            "00:xx:xx:yy:100",
+            "90:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00"
+            "00:00:00:00:00:00:00:00:00:00:00",
+        ],
+    )
+    def test_invalid_mac_address(self, default_data, mac_address):
+        default_data[INTERFACES][0][Interface.MAC] = mac_address
+        with pytest.raises(js.ValidationError, match=str(mac_address)):
+            libnmstate.validator.validate(default_data)
+
+
 class TestIfaceTypeEthernet:
     def test_valid_ethernet_with_auto_neg(self, default_data):
         default_data[INTERFACES][0].update(


tyll:
> * I expected to see a minimum of 4 octets. Are there interfaces with less?

When I asked my colleagues, I got the answer that there is no lower limit in the Linux kernel. IMHO the question should be: Is it guaranteed that all interfaces will have at least a 4 byte MAC address?

> * Unit tests should show that 4 octets work and what does not work (like 3 octets).

Good catch that the unit tests should be adjusted.

EdDev:
> When I asked my colleagues, I got the answer that there is no lower limit in the Linux kernel. IMHO the question should be: Is it guaranteed that all interfaces will have at least a 4 byte MAC address?

Never noticed less than 6 until now.
I actually think it is wrong to use anything less than 6 based on the standard.

I suggest you either leave it 4 or take it out completely (no format validation).
A following patch should enforce 6 octets for ethernet related devices.

tyll:
> > When I asked my colleagues, I got the answer that there is no lower limit in the Linux kernel. IMHO the question should be: Is it guaranteed that all interfaces will have at least a 4 byte MAC address?
> 
> Never noticed less than 6 until now.
> I actually think it is wrong to use anything less than 6 based on the standard.

You mean the ethernet standard? It only applies to ethernet interfaces, I guess, but the Linux kernel supports all kind of interfaces.

> I suggest you either leave it 4 or take it out completely (no format validation).

It is still valid to check for the maximum size (it is a Linux limitation) and if it is specified in the hex/colon notation.

> A following patch should enforce 6 octets for ethernet related devices.

Maybe. Why does this require special handling in the schema given that NM already checks this?

ffmancera:
> > When I asked my colleagues, I got the answer that there is no lower limit in the Linux kernel. IMHO the question should be: Is it guaranteed that all interfaces will have at least a 4 byte MAC address?
> 
> Never noticed less than 6 until now.
> I actually think it is wrong to use anything less than 6 based on the standard.
> 
> I suggest you either leave it 4 or take it out completely (no format validation).
> A following patch should enforce 6 octets for ethernet related devices.

In my opinion we should not take it out completely. I am fine with reducing it to 4 octets instead of 1 octet as we do not have an use case for less than 3 octets. Yes, we can prepare another patch to enforce 6 octets for ethernet devices, I think that is a good idea.

Thanks!

EdDev:
> You mean the ethernet standard? It only applies to ethernet interfaces, I guess, but the Linux kernel supports all kind of interfaces.

Bluetooth is another. What other interface you know that has a mac with less than 6 octets?

It is questionable why the tunnel interfaces use 4 octets, nothing defines that because these are not L2 protocols. What is that 4 octets represent there?

> > I suggest you either leave it 4 or take it out completely (no format validation).
> 
> It is still valid to check for the maximum size (it is a Linux limitation) and if it is specified in the hex/colon notation.

ack

> > A following patch should enforce 6 octets for ethernet related devices.
> 
> Maybe. Why does this require special handling in the schema given that NM already checks this?

I suggest doing this in the validation step, not in the schema.
If NM is providing a reasonable error that the user/client can interpret, then I'm all in favor of letting NM do it.

libnmstate/schemas/operational-state.yaml
```diff
@@ -63,7 +63,6 @@ definitions:
         - down
     mac-address:
       type: string
-      pattern: "^([a-fA-F0-9]{2}:){5,31}[a-fA-F0-9]{2}$"
```
EdDev: @tyll pointed out correctly that it makes sense to keep the maximum length and the format validated.
If there is no explicit reason, I would use a minimum of 4 octets.

tests/lib/schema_validation_test.py
```diff
@@ -178,6 +178,30 @@ def test_invalid_type(self, default_data):
             libnmstate.validator.validate(default_data)
 
 
+class TestIfaceMacAddress:
+    def test_valid_ethernet_mac_address(self, default_data):
```
EdDev: Can you also add invalid cases? Min, max and wrong characters should suffice. Strange that we did not have these already.

EdDev: BTW, you could use here parametrized https://docs.pytest.org/en/latest/parametrize.html

One test for valid entries and another for invalid ones.

ffmancera: Done, thanks!

libnmstate/schemas/operational-state.yaml
ffmancera: Done, thank you!

tests/lib/schema_validation_test.py
```diff
@@ -178,6 +178,34 @@ def test_invalid_type(self, default_data):
...
+    @pytest.mark.parametrize(
+        "mac_address",
+        [
+            "00:11:22:33",
+            "00:11:22:33:44:55",
+            "80:00:02:08:fe:80:00:00:00:00:00:00:f4:52:14:03:00:8d:52:11",
+        ],
+    )
+    def test_valid_mac_address(self, default_data, mac_address):
+        default_data[INTERFACES][0].update({Interface.MAC: mac_address})
```
EdDev: How about a direct approach: `default_data[INTERFACES][0][Interface.MAC] = mac_address`

tests/lib/schema_validation_test.py
```diff
@@ -178,6 +178,34 @@ def test_invalid_type(self, default_data):
...
+        "mac_address",
+        [
+            "00:11:22",
+            "00:xx:xx:yy:100",
+            "90:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00"
+            "00:00:00:00:00:00:00:00:00:00:00",
+        ],
+    )
+    def test_invalid_mac_address(self, default_data, mac_address):
+        default_data[INTERFACES][0].update({Interface.MAC: mac_address})
```
EdDev: Same here.

tests/lib/schema_validation_test.py
ffmancera: Done, thanks!

tests/lib/schema_validation_test.py
ffmancera: Done

PR #757, opened 2020-02-07
 tox, tavis: remove python 3.7 unit tests

ffmancera:
As run-tests.sh is now using Fedora 32, it requires to install python 3.7
to run python 3.7 unit tests.

```
GLOB sdist-make: /root/nmstate-workspace/setup.py
py37 create: /root/nmstate-workspace/.tox/py37
SKIPPED: InterpreterNotFound: python3.7
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,6 @@ env:
         - CONTAINER_IMAGE=nmstate/centos8-nmstate-dev
           testflags="--test-type unit_py36"
         - CONTAINER_IMAGE=nmstate/fedora-nmstate-dev
-          testflags="--test-type unit_py37"
-        - CONTAINER_IMAGE=nmstate/fedora-nmstate-dev
           testflags="--test-type unit_py38"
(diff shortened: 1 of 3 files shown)


tyll:
@EdDev we do not need to support Fedora 31 anymore for VDSM, correct? Then we can just remove the Python 3.7 unit tests from the CI instead of fixing it for F32.

EdDev:
> we do not need to support Fedora 31 anymore for VDSM, correct?

Correct. We also agreed not to support 0.2 on Fedora 31, that is why we moved to rawhide.

> Then we can just remove the Python 3.7 unit tests from the CI instead of fixing it for F32.

I’m good with dropping it.

ffmancera:
> Please change it to remove the py37 test from tox and travis.yml

Done, thanks!

EdDev:
Rebase/merge exception: No code changed, travis-ci and tox config files were not touched on master.

Issue #759, opened 2020-02-09
Test `test_nm_ovs_plugin_missing` fails all following OVS tests


EdDev:
The `tests/integration/ovs_test.py::test_nm_ovs_plugin_missing` test is breaking all OVS following tests because it is not able to re-install the `NetworkManager-ovs` package.

See errors from the logs:
```
DEBUG    root:cmdlib.py:40 dnf install -y --cacheonly NetworkManager-ovs (cwd None)
DEBUG    root:cmdlib.py:53 FAILED: <err> = b'error: rpmdb: damaged header #178 retrieved -- skipping.\nerror: rpmdb: damaged header #179 retrieved -- skipping.\nerror: rpmdbNextIterator: skipping h#     179 blob size(1752749994): BAD, 8 + 16 * il(1886415219) + dl(1634877554)\nerror: rpmdbNextIterator: skipping h#     178 blob size(7736): BAD, 8 + 16 * il(71) + dl(6592)\nError opening file for checksum: /var/cache/dnf/copr:copr.fedorainfracloud.org:networkmanager:NetworkManager-1.22-25c098209a2a8596/packages/NetworkManager-ovs-1.22.6-2.el8.x86_64.rpm\nPackage "NetworkManager-ovs-1:1.22.6-2.el8.x86_64" from repository "copr:copr.fedorainfracloud.org:networkmanager:NetworkManager-1.22" has incorrect checksum\nError: Some packages have invalid cache, but cannot be downloaded due to "--cacheonly" option\n'; <rc> = 1
```

In addition, the test needs to assure that the plugin is installed before proceeding, making it clear what is the problem (currently it is hidden by the ovs xfail markings).

tyll:
Looks to me like it is the same problem as #740

tyll:
IMHO a better solution for the plugin tests would be to run them in seperate containers where the plugin is missing and then the container is discarded after the test ran.

ffmancera:
#767 should fix this. Thanks!

PR #760, opened 2020-02-09
tests, integ: Do not run test_nm_ovs_plugin_missing

EdDev:
When running test_nm_ovs_plugin_missing test, the NM ovs plugin is
removed. Unfortunatly it fails to be re-installed when the test ends.

Until this issue is fixed, the test is marked not to run in the suite.

Ref: https://github.com/nmstate/nmstate/issues/759

### DIFF
--- a/tests/integration/ovs_test.py
+++ b/tests/integration/ovs_test.py
@@ -163,6 +163,9 @@ def test_ovs_interface_with_max_length_name():
     assertlib.assert_absent(ovs_interface_name)
 
 
+@pytest.mark.xfail(
+    reason="https://github.com/nmstate/nmstate/issues/759", run=False
+)
 def test_nm_ovs_plugin_missing():
     with disable_nm_ovs_plugin():
         with pytest.raises(NmstateLibnmError):


tests/integration/ovs_test.py
```diff
@@ -163,6 +163,9 @@ def test_ovs_interface_with_max_length_name():
     assertlib.assert_absent(ovs_interface_name)
 
 
+@pytest.mark.xfail(
```
ffmancera: This can be a `skip`, just a nit. Thanks!

EdDev: This is a bug.
`skip` means we accept the current state (due to the env usually) and there is no plan to "fix" it.
`xfail` means we intend to fix it and remove it.

The difference is very thin, I like to think of it as: `xfail` means that we intend to treat it.

ffmancera: Oh, thanks for explaining. LGTM

EdDev:
Rebase/merge exception: DCO job seems to be stuck but the signature is present.

PR #761, opened 2020-02-09
Remove xfail from all nm ovs tests

EdDev:
tests, integ, nm: Remove xfail from all ovs tests

The OVS tests have been marked with xfail because there were instability
issues when running them.

In order to proceed with the development of OVS and detect new issues,
the xfail marking is removed.

Depends on #760

### DIFF
--- a/tests/integration/nm/ovs_test.py
+++ b/tests/integration/nm/ovs_test.py
@@ -27,7 +27,6 @@ from libnmstate.schema import Interface
 from libnmstate.schema import OVSBridge as OB
 
 from .testlib import mainloop
-from .testlib import MainloopTestError
 
 
 BRIDGE0 = "brtest0"
@@ -54,10 +53,6 @@ def bridge_default_config():
     }
 
 
-@pytest.mark.xfail(
-    raises=(MainloopTestError, AssertionError),
-    reason="https://bugzilla.redhat.com/1724901",
-)
 def test_create_and_remove_minimum_config_bridge(
     bridge_minimum_config, bridge_default_config
 ):
@@ -71,10 +66,6 @@ def test_create_and_remove_minimum_config_bridge(
     assert not _get_bridge_current_state()
 
 
-@pytest.mark.xfail(
-    raises=(MainloopTestError, AssertionError),
-    reason="https://bugzilla.redhat.com/1724901",
-)
 def test_bridge_with_system_port(eth1_up, bridge_default_config):
     bridge_desired_state = bridge_default_config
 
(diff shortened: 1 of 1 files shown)


tests/integration/ovs_test.py
```diff
@@ -163,6 +163,9 @@ def test_ovs_interface_with_max_length_name():
     assertlib.assert_absent(ovs_interface_name)
 
 
+@pytest.mark.xfail(
```
ffmancera: This should be a `skip` instead of a `xfail`. Thanks!

ffmancera: Sorry wrong PR

tyll:
Just one thing, if the issues from https://bugzilla.redhat.com/show_bug.cgi?id=1724901 are now fixed, then the bug should be updated.

EdDev:
> Just one thing, if the issues from https://bugzilla.redhat.com/show_bug.cgi?id=1724901 are now fixed, then the bug should be updated.

The BZ mentioned is too generic, I think the specific issues that are not included in the latest NM release (1.22.6) and are currently known are these: https://bugzilla.redhat.com/1782701 and https://bugzilla.redhat.com/1788432.
The problem with the current xfail marking is that they mask other problems and as we now focus back on OVS, I prefer the problem to surface up so we can feedback if other issues exists.

tyll:
> > Just one thing, if the issues from https://bugzilla.redhat.com/show_bug.cgi?id=1724901 are now fixed, then the bug should be updated.
> 
> The BZ mentioned is too generic, I think the specific issues that are not included in the latest NM release (1.22.6) and are currently known are these: https://bugzilla.redhat.com/1782701 and https://bugzilla.redhat.com/1788432.
> The problem with the current xfail marking is that they mask other problems and as we now focus back on OVS, I prefer the problem to surface up so we can feedback if other issues exists.

The referenced bugzilla (1724901) is against Nmstate and not NetworkManager. What do you propose to do with it?
The second one against NM is worked-around in Nmstate by hashing the interface names - isn't it?

EdDev:
> The referenced bugzilla (1724901) is against Nmstate and not NetworkManager. What do you propose to do with it?

Link it to the fixes that came from NM and check if we still have other instability problems or not.
If we do, open specific issues, otherwise consider it closed after we remove all ovs xfails (from the integration tests).

> The second one against NM is worked-around in Nmstate by hashing the interface names - isn't it?

Right, it is not a priority for now although it could have simplify our code considerably.
It may also create challenges with the OVS bond now, unsure yet.

PR #762, opened 2020-02-09
nm, ovs: Remove translate {port, bridge} options step

EdDev:
The OVS {port, bridge} options translation step has no added value at the moment and
can be dropped.

Depends on #761, #760

### DIFF
--- a/libnmstate/nm/applier.py
+++ b/libnmstate/nm/applier.py
@@ -311,9 +311,8 @@ def prepare_proxy_ifaces_desired_state(ifaces_desired_state):
         port_opts_metadata = iface_desired_state.get(BRPORT_OPTIONS_METADATA)
         if port_opts_metadata is None:
             continue
-        port_options = ovs.translate_port_options(port_opts_metadata)
         port_iface_desired_state = _create_ovs_port_iface_desired_state(
-            iface_desired_state, port_options
+            iface_desired_state, port_opts_metadata
         )
         new_ifaces_desired_state.append(port_iface_desired_state)
         # The "visible" slave/interface needs to point to the port profile
--- a/libnmstate/nm/applier.py
+++ b/libnmstate/nm/applier.py
@@ -407,7 +407,8 @@ def _build_connection_profile(iface_desired_state, base_con_profile=None):
             )
             settings.append(linux_bridge_setting)
     elif iface_type == ovs.BRIDGE_TYPE:
-        ovs_bridge_options = ovs.translate_bridge_options(iface_desired_state)
+        ovs_bridge_state = iface_desired_state.get(OvsB.CONFIG_SUBTREE, {})
+        ovs_bridge_options = ovs_bridge_state.get(OvsB.OPTIONS_SUBTREE)
         if ovs_bridge_options:
             settings.append(ovs.create_bridge_setting(ovs_bridge_options))
     elif iface_type == ovs.PORT_TYPE:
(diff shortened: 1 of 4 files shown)


tests/lib/nm/ovs_test.py
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2018-2019 Red Hat, Inc.
+# Copyright (c) 2018-2020 Red Hat, Inc.
```
tyll: This seems to be unnecessary.

libnmstate/nm/applier.py
```diff
@@ -328,9 +328,8 @@ def prepare_proxy_ifaces_desired_state(ifaces_desired_state):
         port_opts_metadata = iface_desired_state.get(BRPORT_OPTIONS_METADATA)
```
tyll: Please explain in the commit why it does not add any value. Why was it there in the first place? What did it actually do?

libnmstate/nm/applier.py
```diff
@@ -424,7 +424,8 @@ def _build_connection_profile(iface_desired_state, base_con_profile=None):
             )
             settings.append(linux_bridge_setting)
     elif iface_type == ovs.BRIDGE_TYPE:
-        ovs_bridge_options = ovs.translate_bridge_options(iface_desired_state)
```
tyll: Please explain the reasons for the change in the commit message, see other comment.

libnmstate/nm/applier.py
EdDev: What did I miss? I explained it was a `step` that does not seem to be needed.
If it would have given something interested and was worth changing, then I guess it would have merit an explanation of what it currently does (but now it does nothing at all).

Will changing `step` to `abstraction level` help clarify this?

tyll: What is missing is why you came to the conclusion that it does not add value. Also it says `at the moment`, so will it be needed later? To me it is not clear what the removed code did and why it is not needed anymore, therefore it would be nice to explain it.

EdDev: Sound to me like you are asking for a recursive "why".
Why? Because it does nothing now.
Why? Because the it "translate" a dict to a dict with same key to same value.
why? Because in the past we thought it makes sense to separate between the nmstate state and nm provider state using a translation step.
why? nm api could be independent from nmstate schema.
why? nm could be then easily re-used by other projects.
...
I can continue on for many more.

> Also it says at the moment, so will it be needed later?

Yes, it may come back, I cannot say. I do not see it happening anytime soon.
It just expresses that *now* it has no value, I hoped it is enough reasoning to merit its removal.

tyll: > Sound to me like you are asking for a recursive "why".

> Why? Because the it "translate" a dict to a dict with same key to same value.
> why? Because in the past we thought it makes sense to separate between the nmstate state and nm provider state using a translation step.
> why? nm api could be independent from nmstate schema.
> why? nm could be then easily re-used by other projects.

Write this in the commit message and the change would be properly motivated and easy to understand.

EdDev: Asking this for an idle code that does nothing is redundant IMO, especially when the base reasoning has been given.
Adding more reasoning per your request.

tests/lib/nm/ovs_test.py
EdDev: Done

libnmstate/nm/applier.py
tyll: Thank you for explaining this. To me "does not add value" is not that clear but 'it "translate" a dict to a dict with same key to same value.' is.

EdDev: Updated

PR #763, opened 2020-02-09
nm, ovs: Support link-aggregation port

EdDev:
Link aggregation is implemented in the nm.ovs module, supporting both
setting and reporting the state.

The state (report) is interpreted from the NM configuration.

As part of the change, existing bond related properties (updelay and
downdelay) have been relocated to the link-aggregation subtree.

The lacp property has been integrated into the `mode` property.
Specifically, lacp is set to `active` when the lag mode is "lacp" or
"balance-tcp" and to "off" when the lag mode is "active-backup" or
"balance-slb".

The following OVS port state may be used to configure a bond:
```
name: ovs-bond1
link-aggregation:
  mode: balance-slb
  slaves:
    - name: eth1
    - name: eth2
```

Depends on #760, #761, #762

### DIFF
--- a/libnmstate/nm/ovs.py
+++ b/libnmstate/nm/ovs.py
@@ -31,6 +31,11 @@ PORT_PROFILE_PREFIX = "ovs-port-"
 CAPABILITY = "openvswitch"
 
 
+class LacpValue:
+    ACTIVE = "active"
+    OFF = "off"
+
+
 def has_ovs_capability():
     try:
         nmclient.NM.DeviceType.OVS_BRIDGE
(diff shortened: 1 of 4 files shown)


libnmstate/schema.py
```diff
@@ -351,6 +351,16 @@ class LinkAggregation:
             class Slave:
                 NAME = "name"
 
+            class Options:
+                DOWN_DELAY = "bond-downdelay"
```
tyll: Why do the options need the `bond-` prefix? Since they are already in the LinkAggregation subtree it should be already clear what they belong to.

EdDev: These correspond 1:1 with the OVS property names.
"Our" property name does not include it.

It will be simpler to keep 1:1 naming for simplicity at this stage although it is not a must.

tyll:
> The following OVS port state may be used to configure a bond:
> 
> ```
> name: ovs-bond1
> link-aggregation:
>   mode: balance-slb
>   slaves:
>     - name: eth1
>     - name: eth2
> ```

Could you please add this into an example yaml?

EdDev:
> Could you please add this into an example yaml?

We are not there yet. This is about `libnmstate.nm`, not the `nmstate` core.
It will come in in the following PR/s, like @phoracek did.

libnmstate/nm/ovs.py
```diff
@@ -150,15 +170,39 @@ def _get_bridge_port_info(port_profile, devices_info):
...
 
     return port_info
 
 
+def _get_lag_info(port_name, port_setting, port_slave_names):
+    port_info = {}
+
+    lacp = port_setting.props.lacp
+    mode = port_setting.props.bond_mode
+    if not mode and lacp:
```
tyll: Why is `lacp` a boolean here. The previous code makes it look like it would be `off` or `active`.

libnmstate/nm/ovs.py
```diff
@@ -57,19 +62,34 @@ def create_bridge_setting(options_state):
 
 def create_port_setting(port_state):
     port_setting = nmclient.NM.SettingOvsPort.new()
+
+    lag_state = port_state.get(OB.Port.LINK_AGGREGATION_SUBTREE)
+    if lag_state:
+        mode = lag_state.get(OB.Port.LinkAggregation.MODE)
+        if mode == OB.Port.LinkAggregation.Mode.LACP:
```
tyll: It seems to me that tests for the different modes are missing/conditionals are missing.

EdDev: Done

PR #765, opened 2020-02-10
metadata: Use helper to determine iface state

EdDev:
Use a function helper to determine if the interface state destine its
removal.

Depends on #764

### DIFF
--- a/libnmstate/metadata.py
+++ b/libnmstate/metadata.py
@@ -141,8 +141,7 @@ def _generate_link_master_metadata(
         (ifname, ifstate)
         for ifname, ifstate in ifaces_desired_state.items()
         if ifstate.get(Interface.TYPE) == master_type
-        and ifstate.get(Interface.STATE)
-        not in (InterfaceState.DOWN, InterfaceState.ABSENT)
+        and _iface_to_remove(ifstate)
     ]
     for master_name, master_state in desired_masters:
         desired_slaves = get_slaves_func(master_state)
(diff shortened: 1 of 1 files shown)


libnmstate/metadata.py
```diff
@@ -198,6 +194,13 @@ def _generate_link_master_metadata(
                         )
 
 
+def _iface_to_remove(iface_state):
```
tyll: Is this the correct name? The profiles are going to be removed and the interface only if it is a virtual interface (but not veth).

EdDev: We do not distinct at the moment (but we should) between down and absent. We look at this as if it is absent because it removes the profile, but you are right that it is problematic.

The other alternative is to call it `_is_iface_state_down_or_absent`.

Maybe it is not worth the effort and we should not do this at all.

tyll: The other name fits. I'll leave it up to you whether to close or rename.

libnmstate/metadata.py
```diff
@@ -198,6 +194,13 @@ def _generate_link_master_metadata(
...
+    return iface_state.get(Interface.STATE) not in (
```
tyll: Do you know why this uses get? The state is mandatory, isn't it? It needs to be set explicitly or it will always be merged in from the current state, wouldn't it?

EdDev: The state is not a must, it will use the current.

libnmstate/metadata.py
EdDev: If we cannot name it, it is not worth the change.
Closing.

PR #767, opened 2020-02-10
tests, integ: fix missing plugin tests

ffmancera:

### DIFF
--- a/tests/integration/ovs_test.py
+++ b/tests/integration/ovs_test.py
@@ -17,8 +17,6 @@
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 #
 
-from contextlib import contextmanager
-
 import pytest
 
 import libnmstate
@@ -30,8 +28,8 @@ from libnmstate.schema import OVSBridge
 from libnmstate.error import NmstateLibnmError
 
 from .testlib import assertlib
-from .testlib import cmdlib
 from .testlib import nmlib
+from .testlib.nmplugin import disable_nm_plugin
 from .testlib.ovslib import Bridge
 from .testlib.vlan import vlan_interface
 
@@ -39,20 +37,6 @@ from .testlib.vlan import vlan_interface
 BRIDGE1 = "br1"
 PORT1 = "ovs1"
 VLAN_IFNAME = "eth101"
-
-DNF_REMOVE_NM_OVS_CMD = (
-    "rpm",
-    "-e",
-    "NetworkManager-ovs",
-)
-DNF_INSTALL_NM_OVS_CMD = (
-    "dnf",
-    "install",
-    "-y",
-    "--cacheonly",
-    "NetworkManager-ovs",
-)
-SYSTEMCTL_RESTART_NM_CMD = ("systemctl", "restart", "NetworkManager")
 
 
 @pytest.fixture
@@ -163,11 +147,8 @@ def test_ovs_interface_with_max_length_name():
     assertlib.assert_absent(ovs_interface_name)
 
 
-@pytest.mark.xfail(
-    reason="https://github.com/nmstate/nmstate/issues/759", run=False
-)
 def test_nm_ovs_plugin_missing():
-    with disable_nm_ovs_plugin():
+    with disable_nm_plugin("ovs"):
         with pytest.raises(NmstateLibnmError):
             libnmstate.apply(
                 {
@@ -200,17 +181,6 @@ def test_ovs_remove_port(bridge_with_ports):
             }
         )
         assert not nmlib.list_profiles_by_iface_name(nm_port_profile_name)
-
-
-@contextmanager
-def disable_nm_ovs_plugin():
-    cmdlib.exec_cmd(DNF_REMOVE_NM_OVS_CMD)
-    cmdlib.exec_cmd(SYSTEMCTL_RESTART_NM_CMD)
-    try:
-        yield
-    finally:
-        cmdlib.exec_cmd(DNF_INSTALL_NM_OVS_CMD)
-        cmdlib.exec_cmd(SYSTEMCTL_RESTART_NM_CMD)
 
 
 @pytest.fixture
--- a/tests/integration/testlib/cmdlib.py
+++ b/tests/integration/testlib/cmdlib.py
@@ -25,7 +25,7 @@ RC_SUCCESS = 0
 RC_FAIL2 = 2
 
 
-def exec_cmd(cmd, env=None, stdin=None):
+def exec_cmd(cmd, env=None, stdin=None, check=False):
     """
     Execute cmd in an external process, collect its output and returncode
 
@@ -39,16 +39,18 @@ def exec_cmd(cmd, env=None, stdin=None):
     """
     logging.debug(command_log_line(cmd))
 
-    p = subprocess.Popen(
+    p = subprocess.run(
         cmd,
         close_fds=True,
-        stdin=subprocess.PIPE,
         stdout=subprocess.PIPE,
         stderr=subprocess.PIPE,
         env=env,
+        check=check,
+        input=stdin,
     )
 
-    out, err = p.communicate(stdin)
+    out = p.stdout
+    err = p.stderr
 
     logging.debug(_retcode_log_line(p.returncode, err=err))
 
--- a/tests/integration/testlib/nmplugin.py
+++ b/tests/integration/testlib/nmplugin.py
@@ -1,0 +1,61 @@
+#
+# Copyright (c) 2020 Red Hat, Inc.
+#
+# This file is part of nmstate
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+
+from contextlib import contextmanager
+from subprocess import CalledProcessError
+
+from . import cmdlib
+
+
+@contextmanager
+def disable_nm_plugin(plugin):
+    try:
+        _, rpm_output, _ = cmdlib.exec_cmd(
+            ("rpm", "-ql", f"NetworkManager-{plugin}"), check=True
+        )
+    except CalledProcessError:
+        yield
+    else:
+        lib_path = [l for l in rpm_output.split() if l.endswith(".so")][0]
+        with mount_devnull_to_path(lib_path):
+            yield
+
+
+@contextmanager
+def mount_devnull_to_path(lib_path):
+    try:
+        cmdlib.exec_cmd(("mount", "--bind", "/dev/null", lib_path), check=True)
+        with nm_service_restart():
+            try:
+                yield
+            finally:
+                cmdlib.exec_cmd(("umount", lib_path), check=True)
+    except CalledProcessError:
+        cmdlib.exec_cmd(("umount", lib_path), check=True)
+        raise
+
+
+@contextmanager
+def nm_service_restart():
+    systemctl_restart_nm_cmd = ("systemctl", "restart", "NetworkManager")
+    cmdlib.exec_cmd(systemctl_restart_nm_cmd, check=True)
+    try:
+        yield
+    finally:
+        cmdlib.exec_cmd(systemctl_restart_nm_cmd, check=True)
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -25,7 +25,6 @@ import pytest
 
 import libnmstate
 
-from .testlib import cmdlib
 from .testlib import ifacelib
 
 
@@ -99,17 +98,3 @@ def _get_osname():
             if line.startswith("PRETTY_NAME="):
                 return line.split("=", maxsplit=1)[1].strip().strip('"')
     return ""
-
-
-@pytest.fixture(scope="session", autouse=True)
-def download_packages():
-    cmdlib.exec_cmd(
-        (
-            "dnf",
-            "install",
-            "-y",
-            "--downloadonly",
-            "NetworkManager-ovs",
-            "NetworkManager-team",
-        )
-    )
--- a/tests/integration/team_test.py
+++ b/tests/integration/team_test.py
@@ -30,24 +30,8 @@ from libnmstate.schema import InterfaceType
 from libnmstate.schema import Team
 
 from .testlib import assertlib
-from .testlib import cmdlib
+from .testlib.nmplugin import disable_nm_plugin
 
-
-DNF_INSTALL_NM_TEAM_PLUGIN_CMD = (
-    "dnf",
-    "install",
-    "-y",
-    "--cacheonly",
-    "NetworkManager-team",
-)
-
-DNF_REMOVE_NM_TEAM_PLUGIN_CMD = (
-    "rpm",
-    "-e",
-    "NetworkManager-team",
-)
-
-SYSTEMCTL_RESTART_NM_CMD = ("systemctl", "restart", "NetworkManager")
 
 TEAM0 = "team0"
 
@@ -78,7 +62,7 @@ def test_edit_team_iface():
 
 
 def test_nm_team_plugin_missing():
-    with disable_nm_team_plugin():
+    with disable_nm_plugin("team"):
         with pytest.raises(NmstateLibnmError):
             libnmstate.apply(
                 {
@@ -118,14 +102,3 @@ def team_interface(ifname):
                 ]
             }
         )
-
-
-@contextmanager
-def disable_nm_team_plugin():
-    cmdlib.exec_cmd(DNF_REMOVE_NM_TEAM_PLUGIN_CMD)
-    cmdlib.exec_cmd(SYSTEMCTL_RESTART_NM_CMD)
-    try:
-        yield
-    finally:
-        cmdlib.exec_cmd(DNF_INSTALL_NM_TEAM_PLUGIN_CMD)
-        cmdlib.exec_cmd(SYSTEMCTL_RESTART_NM_CMD)


ffmancera:
I've ran the tests like 100 times and none of the are failing. This approach allow us to disable the plugin without uninstalling the NM plugin package.

tests/integration/ovs_test.py
```diff
@@ -204,12 +193,18 @@ def test_ovs_remove_port(bridge_with_ports):
 
 @contextmanager
 def disable_nm_ovs_plugin():
-    cmdlib.exec_cmd(DNF_REMOVE_NM_OVS_CMD)
-    cmdlib.exec_cmd(SYSTEMCTL_RESTART_NM_CMD)
+    tmp_path = "/tmp/libnm-device-plugin-ovs.so"
```
tyll: I would prefer not to use a hard-coded path in `/tmp` but to either create a proper temporary directory with `tempfile.mkdtemp` or to rename the file inside the actual directory so it will not move across file systems. You can then use `os.rename` to move the file.

tests/integration/ovs_test.py
```diff
@@ -204,12 +193,18 @@ def test_ovs_remove_port(bridge_with_ports):
 
 @contextmanager
 def disable_nm_ovs_plugin():
```
tyll: Since the two plugin helpers are identical except for the plugin RPM, it would be nice to have one context manager that gets the RPM name as argument to reduce the duplication IMHO. Then you also do not need the constants for the commands but can use them directly in the function.

tests/integration/ovs_test.py
ffmancera: I cannot use the os.rename(). If I use it then I am getting an error because of a symlink. I will reproduce it and paste here the error. I am fine with creating a temporary directory.

tests/integration/ovs_test.py
ffmancera: It looks good. Where should I put this new context manager? Thanks!

tests/integration/ovs_test.py
tyll: You cannot use `os.rename` to move the file from /usr/lib64/NetworkManager/... to /tmp but you could use it to rename the plugin from `libnm-device-plugin-team.so` to `libnm-device-plugin-team.nmstate_disabled` or similar.

tests/integration/ovs_test.py
tyll: How about `nmstate/tests/integration/testlib/plugin.py`? Needs to be created. IMHO it could also be directly in `__init__.py` but it might be that @EdDev has arguments against it.

EdDev:
Will mount bind work?
‘ mount -o bind /dev/null /path/to/file’

ffmancera:
> Will mount bind work?
> ‘ mount -o bind /dev/null /path/to/file’

Do you mean instead of renaming it? I can try but it seems quite ugly to me.

EdDev:
> > Will mount bind work?
> > ‘ mount -o bind /dev/null /path/to/file’
> 
> Do you mean instead of renaming it? I can try but it seems quite ugly to me.

Yes, it is masking the file so it cannot be loaded by NM.
Seems to work fine:
Executing: `mount --bind /dev/null /usr/lib64/NetworkManager/1.22.6-2.el8/libnm-device-plugin-ovs.so`

shows this in the log:

`plugin: skip invalid file /usr/lib64/NetworkManager/1.22.6-2.el8/libnm-device-plugin-ovs.so: not a regular file`

`umount  /dev/null /usr/lib64/NetworkManager/1.22.6-2.el8/libnm-device-plugin-ovs.so` should revert everything back.

No need for special temp files, renaming or anything fancy.

tests/integration/ovs_test.py
ffmancera: I did the @EdDev approach. What do you think? Thanks!

tests/integration/ovs_test.py
ffmancera: Done, thanks!

ffmancera:
> > > Will mount bind work?
> > > ‘ mount -o bind /dev/null /path/to/file’
> > 
> > 
> > Do you mean instead of renaming it? I can try but it seems quite ugly to me.
> 
> Yes, it is masking the file so it cannot be loaded by NM.
> Why do you think it is ugly? (at least why uglier than moving or renaming?)
> 
> Seems to work fine:
> Executing: `mount --bind /dev/null /usr/lib64/NetworkManager/1.22.6-2.el8/libnm-device-plugin-ovs.so`
> 
> shows this in the log:
> 
> `plugin: skip invalid file /usr/lib64/NetworkManager/1.22.6-2.el8/libnm-device-plugin-ovs.so: not a regular file`
> 
> `umount /dev/null /usr/lib64/NetworkManager/1.22.6-2.el8/libnm-device-plugin-ovs.so` should revert everything back.
> 
> No need for special temp files, renaming or anything fancy.

Done, thanks! :-)

tests/integration/testlib/plugin.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+@contextmanager
+def disable_nm_plugin(plugin):
+    ret, rpm_output, _ = cmdlib.exec_cmd(
+        ("rpm", "-ql", f"NetworkManager-{plugin}")
+    )
+    if not ret:
+        lib_path = re.search(r"[A-Za-z0-9-/.]*.so", rpm_output).group()
+        cmdlib.exec_cmd(("mount", "--bind", "/dev/null", lib_path))
```
EdDev: This one and its counterpart need to have their own contextmanager.
If the next commands fail, the mount is not reverted.

tests/integration/testlib/plugin.py
```diff
@@ -0,0 +1,46 @@
...
+        cmdlib.exec_cmd(SYSTEMCTL_RESTART_NM_CMD)
+        plugin_disable = True
+    else:
+        plugin_disable = False
+    try:
+        yield
+    finally:
+        if plugin_disable:
+            cmdlib.exec_cmd(("umount", "/dev/null", lib_path))
```
EdDev: I do not think you need to specify "/dev/null". When I tried it, it actually gave an error.

tests/integration/testlib/plugin.py
```diff
@@ -0,0 +1,46 @@
...
+    ret, rpm_output, _ = cmdlib.exec_cmd(
+        ("rpm", "-ql", f"NetworkManager-{plugin}")
+    )
+    if not ret:
+        lib_path = re.search(r"[A-Za-z0-9-/.]*.so", rpm_output).group()
+        cmdlib.exec_cmd(("mount", "--bind", "/dev/null", lib_path))
+        cmdlib.exec_cmd(SYSTEMCTL_RESTART_NM_CMD)
+        plugin_disable = True
+    else:
+        plugin_disable = False
```
EdDev: Can't you just exit earlier if ret evaluates to True? I see no need for the flags in that case.
```
if ret:
    return
```

tests/integration/testlib/plugin.py
```diff
@@ -0,0 +1,46 @@
...
+
+@contextmanager
+def disable_nm_plugin(plugin):
+    ret, rpm_output, _ = cmdlib.exec_cmd(
+        ("rpm", "-ql", f"NetworkManager-{plugin}")
+    )
+    if not ret:
+        lib_path = re.search(r"[A-Za-z0-9-/.]*.so", rpm_output).group()
+        cmdlib.exec_cmd(("mount", "--bind", "/dev/null", lib_path))
+        cmdlib.exec_cmd(SYSTEMCTL_RESTART_NM_CMD)
```
EdDev: Using a contextmanager to manage the NM service will also be nice.

tests/integration/testlib/plugin.py
ffmancera: Oh you are right. Thanks!

tests/integration/testlib/plugin.py
ffmancera: That is not possible. If we do this and the user does not have the plugin installed the test is going to fail with the following error:
```
    def test_nm_ovs_plugin_missing():
>       with disable_nm_plugin("ovs"):

tests/integration/ovs_test.py:151: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <contextlib._GeneratorContextManager object at 0x7ff54c4e9a60>

    def __enter__(self):
        # do not keep args and kwds alive unnecessarily
        # they are only needed for recreation, which is not possible anymore
        del self.args, self.kwds, self.func
        try:
            return next(self.gen)
        except StopIteration:
>           raise RuntimeError("generator didn't yield") from None
E           RuntimeError: generator didn't yield
```

tests/integration/testlib/plugin.py
ffmancera: Done, thanks!

tests/integration/testlib/plugin.py
ffmancera: Done!

tests/integration/testlib/plugin.py
ffmancera: Done, thanks!

tests/integration/testlib/nmplugin.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+@contextmanager
+def mount_devnull_to_path(lib_path):
+    cmdlib.exec_cmd(("mount", "--bind", "/dev/null", lib_path))
+    with nm_service_restart():
+        try:
+            yield
+        finally:
+            cmdlib.exec_cmd(("umount", lib_path))
```
EdDev: You need another try/except (not finally), over the `with` in case the service fails the first restart.

ffmancera: Done, thanks!

tests/integration/testlib/nmplugin.py
```diff
@@ -0,0 +1,61 @@
...
+def mount_devnull_to_path(lib_path):
+    cmdlib.exec_cmd(("mount", "--bind", "/dev/null", lib_path))
+    try:
+        with nm_service_restart():
+            try:
+                yield
+            finally:
+                cmdlib.exec_cmd(("umount", lib_path))
+    except Exception:
+        cmdlib.exec_cmd(SYSTEMCTL_RESTART_NM_CMD)
```
EdDev: @tyll noticed this and I missed... you need to umount, not to restart the service.

ffmancera: Oh you are right. Sorry.

tests/integration/testlib/nmplugin.py
```diff
@@ -0,0 +1,61 @@
...
+                yield
+            finally:
+                cmdlib.exec_cmd(("umount", lib_path))
+    except Exception:
+        cmdlib.exec_cmd(("umount", lib_path))
+
+
+@contextmanager
+def nm_service_restart():
+    cmdlib.exec_cmd(SYSTEMCTL_RESTART_NM_CMD)
```
tyll: It is inconsistent to use a constant here but not for the unmount. Also IMHO it does not need to be a constant that is defined at the beginning but it could be defined in the function to make it easier to read.

tests/integration/testlib/nmplugin.py
```diff
@@ -0,0 +1,61 @@
...
+
+SYSTEMCTL_RESTART_NM_CMD = ("systemctl", "restart", "NetworkManager")
+
+
+@contextmanager
+def disable_nm_plugin(plugin):
+    ret, rpm_output, _ = cmdlib.exec_cmd(
+        ("rpm", "-ql", f"NetworkManager-{plugin}")
+    )
+    if not ret:
```
tyll: IMHO the condition would be more clear if it is `if ret == 0`.

tests/integration/testlib/nmplugin.py
```diff
@@ -0,0 +1,61 @@
...
+        lib_path = re.search(r"[A-Za-z0-9-/.]*.so", rpm_output).group()
```
tyll: You can use something like `[l for l in rpm_output.split() if l.endswith(b".so")][0]` to avoid the regular expression which  is harder to understand IMHO. (Meant to mention this before but I could not find an old comment about this).

tests/integration/testlib/nmplugin.py
```diff
@@ -0,0 +1,61 @@
...
+        with mount_devnull_to_path(lib_path):
+            yield
+    else:
+        yield
+
+
+@contextmanager
+def mount_devnull_to_path(lib_path):
+    cmdlib.exec_cmd(("mount", "--bind", "/dev/null", lib_path))
```
tyll: This does not check the return code to check if this actually worked. The same holds true for the other commands.

tests/integration/testlib/nmplugin.py
```diff
@@ -0,0 +1,61 @@
...
+    try:
+        with nm_service_restart():
+            try:
+                yield
+            finally:
+                cmdlib.exec_cmd(("umount", lib_path))
+    except Exception:
```
tyll: What kind of exceptions are expected here? in case it is not possible to execute commands (which is the only thing that `nm_service_restart()` does, it should also not be possible to recover from this to execute just another command. Therefore this might be a little bit of overkill. But maybe I just looked too long at this code.

ffmancera: I think you are right. This try: except: is a little bit overkill but I am not completely against it so I am not going to fight for it.

tests/integration/testlib/nmplugin.py
tyll: The returnode should be checked here.

tests/integration/testlib/nmplugin.py
tyll: Actually, when nm_service_restart would raise an exception if the exception handling becomes more meaningful.

tests/integration/testlib/nmplugin.py
ffmancera: Oh! Thanks for this one. I am glad to drop the regexp :-)

tests/integration/testlib/nmplugin.py
ffmancera: Done, thanks!

tests/integration/testlib/nmplugin.py
ffmancera: Done, thanks!

tests/integration/testlib/nmplugin.py
ffmancera: Done, thanks!

tests/integration/testlib/nmplugin.py
ffmancera: Done, I've implemented check_returncode parameter in exec_cmd.

tests/integration/testlib/nmplugin.py
ffmancera: Done, thanks!

tests/integration/testlib/nmplugin.py
ffmancera: Done, thanks!

tests/integration/testlib/cmdlib.py
```diff
@@ -49,10 +49,13 @@ def exec_cmd(cmd, env=None, stdin=None):
     )
 
     out, err = p.communicate(stdin)
+    returncode = p.returncode
```
tyll: Please call `Popen()` instead with `check=check_returncode` and IMHO rename `check_returncode` to `check`.

tests/integration/testlib/nmplugin.py
```diff
@@ -0,0 +1,64 @@
...
+    try:
+        cmdlib.exec_cmd(
+            ("mount", "--bind", "/dev/null", lib_path), check_returncode=True
+        )
+        with nm_service_restart():
+            try:
+                yield
+            finally:
+                cmdlib.exec_cmd(("umount", lib_path), check_returncode=True)
+    except CalledProcessError:
```
tyll: Please change this back to `Exception` and try to run umount instead of raising.

tests/integration/testlib/nmplugin.py
```diff
@@ -0,0 +1,64 @@
...
+@contextmanager
+def nm_service_restart():
+    systemctl_restart_nm_cmd = ("systemctl", "restart", "NetworkManager")
+    try:
+        cmdlib.exec_cmd(systemctl_restart_nm_cmd, check_returncode=True)
+        try:
+            yield
+        finally:
+            cmdlib.exec_cmd(systemctl_restart_nm_cmd, check_returncode=True)
+    except CalledProcessError:
```
tyll: This does not make sense, raise is the default action if there is no except, just remove this try/except pair.

tests/integration/testlib/nmplugin.py
tyll: Probably it should umount and then raise.

tests/integration/testlib/nmplugin.py
```diff
@@ -0,0 +1,64 @@
...
+from contextlib import contextmanager
+from subprocess import CalledProcessError
+
+from . import cmdlib
+
+
+@contextmanager
+def disable_nm_plugin(plugin):
+    try:
+        ret, rpm_output, _ = cmdlib.exec_cmd(
```
tyll: Please change `ret` to `_`.

tests/integration/testlib/nmplugin.py
```diff
@@ -0,0 +1,64 @@
...
+            ("rpm", "-ql", f"NetworkManager-{plugin}"), check_returncode=True
+        )
```
tyll: The except CalledProcessError should move here to guard the rpm -ql.

tests/integration/testlib/cmdlib.py
ffmancera: Reading the subprocess documentation I am not able to see the `check` parameter in the `Popen()` constructor.

tests/integration/testlib/nmplugin.py
ffmancera: Oops.Sorry

tests/integration/testlib/nmplugin.py
ffmancera: Sure, thanks

tests/integration/testlib/cmdlib.py
tyll: Are you maybe looking into Python 2.7 docs? I see it in 3.6 and later: https://docs.python.org/3.6/library/subprocess.html

tests/integration/testlib/nmplugin.py
ffmancera: Hmm not sure. If this command fails it means the plugin is not installed. In that case, we should proceed with the test directly without unmounting or restarting NM. What do you think? Thanks!

tests/integration/testlib/cmdlib.py
ffmancera: Sorry, I am looking into `Popen()` and I am not able to find `check` here 
```
class subprocess.Popen(args, bufsize=-1, executable=None, stdin=None, stdout=None, stderr=None, preexec_fn=None, close_fds=True, shell=False, cwd=None, env=None, universal_newlines=False, startupinfo=None, creationflags=0, restore_signals=True, start_new_session=False, pass_fds=(), *, encoding=None, errors=None)
```

I will try it anyway, thanks!

tyll:
Hm, now I do not see it anymore, too.

tyll:
Ah, sorry, I was looking at run which might actually be more appropriate
than Popen.

tests/integration/testlib/cmdlib.py
```diff
@@ -49,10 +49,13 @@ def exec_cmd(cmd, env=None, stdin=None):
...
+    if check_returncode and returncode:
+        raise subprocess.CalledProcessError(returncode, cmd, out)
```
EdDev: This seems odd to me.
I would expect a simple change here: On failure, raise and let the caller handle it.
The caller can use a helper function that handles the exception, but having branching here does not seem right.

`CalledProcessError` is supposed to be raised by subprocess, you raising it yourself is a smell IMO.

tests/integration/testlib/cmdlib.py
ffmancera: Fixed. We are using now subprocess.run(). Thanks!

tests/integration/testlib/nmplugin.py
ffmancera: Done, thanks!

tests/integration/testlib/nmplugin.py
ffmancera: Done, thanks!

tests/integration/testlib/nmplugin.py
ffmancera: Done, thanks!

tests/integration/testlib/nmplugin.py
ffmancera: I fixed this, thanks!

tests/integration/testlib/cmdlib.py
ffmancera: Yes, you are right. Fixed, thanks!

tests/integration/testlib/cmdlib.py
```diff
@@ -25,7 +25,7 @@
 RC_FAIL2 = 2
 
 
-def exec_cmd(cmd, env=None, stdin=None):
+def exec_cmd(cmd, env=None, stdin=None, check=False):
```
tyll: The commit message needs to be adjusted.

ffmancera: Opps, done. Thanks!

PR #771, opened 2020-02-11
validator: Validate duplicate interface names

EdDev:
The current interface state is using the name property as an unique
key. Passing in duplicate interfaces names causes a failure only at the
verification step (if it is executed), making it hard to understand the
problem.

This change adds a validation for duplicate interface names, raising an
error early on.

### DIFF
--- a/libnmstate/validator.py
+++ b/libnmstate/validator.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2018-2019 Red Hat, Inc.
+# Copyright (c) 2018-2020 Red Hat, Inc.
 #
 # This file is part of nmstate
 #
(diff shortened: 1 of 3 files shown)


EdDev:
Extracted from #600 and does not include the xfail test that shows the no support of ovs-bridge/ovs-interface with the same name.

libnmstate/validator.py
```diff
@@ -109,6 +113,15 @@ def validate_link_aggregation_state(desired_state, current_state):
                 specified_slaves |= slaves
 
 
+def validate_unique_interface_name(state):
+    ifaces_names = [
+        ifstate[schema.Interface.NAME]
+        for ifstate in state.get(schema.Interface.KEY, [])
+    ]
+    if len(ifaces_names) != len(set(ifaces_names)):
+        raise NmstateDuplicateInterfaceNameError(str(sorted(ifaces_names)))
```
tyll: The exception should use a user-friendly message instead of just listing interfaces.

EdDev: Done

PR #772, opened 2020-02-11
nmclient: Use exceptions in mainloop.run and introduce NmstateTimeoutError

EdDev:
Instead of using return values, use exceptions to reflect the reason of
the mainloop error.

In addition, raise NmstateTimeoutError on mainloop timeout

From the existing errors reported by the mainloop it is hard to
distinguish between a timeout error and an action failure.

This change introduces a dedicated timeout exception that may be used to
identify it by callers.

The integration tests may now be more specific when xfail-ing a test.

### DIFF
--- a/libnmstate/error.py
+++ b/libnmstate/error.py
@@ -109,3 +109,11 @@ class NmstateNotSupportedError(NmstateError):
     """
 
     pass
+
+
+class NmstateTimeoutError(NmstateLibnmError):
+    """
+    The transaction execution timed out.
+    """
+
+    pass
--- a/libnmstate/netapplier.py
+++ b/libnmstate/netapplier.py
@@ -203,14 +203,11 @@ def _verify_change(desired_state):
 def _setup_providers():
     mainloop = nmclient.mainloop()
     yield
-    success = mainloop.run(timeout=MAINLOOP_TIMEOUT)
-    if not success:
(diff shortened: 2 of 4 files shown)


libnmstate/error.py
```diff
@@ -109,3 +109,11 @@ class NmstateNotSupportedError(NmstateError):
     """
 
     pass
+
+
+class NmstateTimeoutError(NmstateLibnmError):
```
cathay4t: The `NmstateLibnmError` is `Exception for unexpected libnm failure.`

But NmstateTimeoutError is expected failure.

EdDev:
This PR enables the nmstate mainloop object to raise exceptions on the run failure (instead of returning error codes).
At the first step, this allows us to raise explicit errors for the timeout.
In following PR/s, we can let the individual action/callback that runs in the mainloop to decide what error to raise, it will then pass it on outside the mainloop run (probably through the mainloop object somehow) so the run() will raise it instead of the current `NmstateLibnmError` general one.

Then, we can fine tune our tests for expected errors and not mask others.

libnmstate/error.py
EdDev: I am not sure I follow your point here.
An unexpected timeout occurred and instead of raising the general exception, we raise a specific one.

PR #783, opened 2020-02-12
CHANGELOG: update change log for 0.2.4 release

ffmancera:

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,12 @@
 # Changelog
+## [0.2.4] - 2020-02-12
+### Bug Fixes
+ - Reduce minimal required MAC address length to 4 octets
+ - Warn early when the same interface name is mentioned multiple times
+
+### New Features
+ - Raise NmstateTimeoutError on mainloop timeout
+
 ## [0.2.3] - 2020-02-05
 ### Bug Fixes
  - Reduce the delay before the verification step to 0.5 second.


CHANGELOG
```diff
@@ -1,4 +1,14 @@
 # Changelog
+## [0.2.4] - 2020-02-12
+### Bug Fixes
+ - Reduce minimal required MAC address length to 4 octets
+ - Raise NmstateTimeoutError on mainloop timeout
+ - Validate duplicate interface names
```
tyll: IMHO this should be something like "Warn early when the same interface name is mentioned multiple times" to focus on the user-visible behavior change.

CHANGELOG
```diff
@@ -1,4 +1,14 @@
...
+ - Support ovs-bridge and ovs-interface with the same name
+
+### New features
+ - Support link-aggregation port in OVS
```
cathay4t: We only have NM implementation, not public interface support.

CHANGELOG
```diff
@@ -1,4 +1,14 @@
 # Changelog
+## [0.2.4] - 2020-02-12
+### Bug Fixes
+ - Reduce minimal required MAC address length to 4 octets
+ - Raise NmstateTimeoutError on mainloop timeout
+ - Validate duplicate interface names
+ - Support ovs-bridge and ovs-interface with the same name
```
tyll: This is not supported, we only have a test that shows that it is not supported. This line can be dropped IMHO.

CHANGELOG
tyll: This is not finished, yet, only the internals are done.

CHANGELOG
```diff
@@ -1,4 +1,14 @@
 # Changelog
+## [0.2.4] - 2020-02-12
+### Bug Fixes
+ - Reduce minimal required MAC address length to 4 octets
+ - Raise NmstateTimeoutError on mainloop timeout
```
tyll: This is more a feature than a bugfix, since the exception is new.

CHANGELOG
ffmancera: Thanks, removed.

CHANGELOG

CHANGELOG

CHANGELOG

PR #792, opened 2020-02-14
validator: report missing NM plugin during validation

ffmancera:
Now it is possible to get the plugin capabilities from the NM client.
Nmstate is reporting the missing NetworkManager plugin during the
validation instead of during profile activation.

Ref: https://bugzilla.redhat.com/1785160

### DIFF
--- a/libnmstate/nm/team.py
+++ b/libnmstate/nm/team.py
@@ -25,7 +25,13 @@ from libnmstate.schema import Interface
 from libnmstate.schema import Team
 
 
+CAPABILITY = "team"
 TEAMD_JSON_DEVICE = "device"
+
+
+def has_team_capability():
+    nm_client = nmclient.client()
+    return nmclient.NM.Capability.TEAM in nm_client.get_capabilities()
 
 
 def create_setting(iface_state, base_con_profile):
--- a/tests/integration/team_test.py
+++ b/tests/integration/team_test.py
@@ -23,7 +23,7 @@ import os
 import pytest
 
 import libnmstate
-from libnmstate.error import NmstateLibnmError
+from libnmstate.error import NmstateDependencyError
 from libnmstate.schema import Interface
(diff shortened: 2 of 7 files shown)


libnmstate/nm/team.py
```diff
@@ -18,16 +18,26 @@
 #
 
 import json
+import logging
```
tyll: This does not seem to be needed

ffmancera: Whoops, leftover from debugging. Sorry

ffmancera: Fixed. Thanks!

libnmstate/nm/team.py
```diff
@@ -25,9 +25,18 @@
 from libnmstate.schema import Team
 
 
+CAPABILITY = "team"
 TEAMD_JSON_DEVICE = "device"
 
 
+def has_team_capability():
+    nm_client = nmclient.client()
+    if nmclient.NM.Capability.TEAM in nm_client.get_capabilities():
```
tyll: you can simplify this by doing `return nmclient.NM.Capability.TEAM in nm_client.get_capabilities()`

tests/integration/team_test.py
```diff
@@ -23,7 +23,7 @@
 import pytest
 
 import libnmstate
-from libnmstate.error import NmstateLibnmError
+from libnmstate.error import NmstateDependencyError
```
tyll: Would be good to mention the new exception type in the commit message and later in the changelog

libnmstate/nm/team.py
ffmancera: Thanks! Done :-)

tests/integration/team_test.py
ffmancera: Done, thanks!

PR #794, opened 2020-02-14
nm.team: Check if "device" exists before delete it

ffmancera:
If a team interface with options is set using nmcli then "device" is not
found in the team JSON configuration because is optional.

### DIFF
--- a/libnmstate/nm/team.py
+++ b/libnmstate/nm/team.py
@@ -102,7 +102,7 @@ def get_info(device):
 
 
 def _convert_teamd_config_to_nmstate_config(team_config):
-    team_config.pop("device")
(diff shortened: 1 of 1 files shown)


libnmstate/nm/team.py
```diff
@@ -102,7 +102,9 @@ def get_info(device):
 
 
 def _convert_teamd_config_to_nmstate_config(team_config):
-    team_config.pop("device")
+    if team_config.get(TEAMD_JSON_DEVICE):
+        team_config.pop(TEAMD_JSON_DEVICE)
```
cathay4t: Please use pop(key, None) to ignore missing key

ffmancera: Done, thanks! :-)

PR #796, opened 2020-02-16
libnmstate: Do not remove slaves implicitly with master

EdDev:
When a master interface is removed, nmstate should not implicitly remove
its slaves. If such an action is desired, the user needs to explicitly
specify it.

This change scope includes the following interface types:
- Bond
- Linux Bridge
- OVS Bridge
- Team

Note: OVS internal interfaces (ovs-interface type) are an exception
here, as such interfaces may not exist without an OVS bridge. Therefore,
these are expected to get removed together with their bridge.

This is also fixing the issue reported in https://nmstate.atlassian.net/browse/NMSTATE-272.

### DIFF
--- a/libnmstate/state.py
+++ b/libnmstate/state.py
@@ -29,6 +29,10 @@ from operator import itemgetter
 
 from libnmstate import iplib
 from libnmstate import metadata
+from libnmstate.appliers import bond
+from libnmstate.appliers import linux_bridge
+from libnmstate.appliers import ovs_bridge
+from libnmstate.appliers import team
 from libnmstate.error import NmstateNotImplementedError
 from libnmstate.error import NmstateValueError
 from libnmstate.error import NmstateVerificationError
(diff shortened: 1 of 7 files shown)


libnmstate/state.py
```diff
@@ -679,6 +717,39 @@ def merge_route_rules(self, other_state):
             }
 
 
+def _get_bond_slaves(ifstate):
+    lag = ifstate.get(Bond.CONFIG_SUBTREE, {})
+    return lag.get(Bond.SLAVES, [])
+
+
+def _get_linux_bridge_slaves(ifstate):
```
cathay4t: The `appliers/linux_bridge.py` already has `get_slaves_from_state()`.

I don't like the idea of placing everything related to state to `state.py`. If possible, please
create `iface_state/bridge.py`, `iface_state/ovs.py` and `iface_state/bond.py` files
to hold these interface type specific state related functions.
Not a must, if you don't want it, just remove the duplciation of `_get_linux_bridge_slaves()`
is enough for this PR.

cathay4t: And in `metadata.py` we have:

```
59:        get_slaves_func=_get_bond_slaves_from_state,
60:        set_metadata_func=_set_common_slaves_metadata,
66:        get_slaves_func=_get_ovs_slaves_from_state,
73:        get_slaves_func=linux_bridge.get_slaves_from_state,
```

EdDev: I posted https://github.com/nmstate/nmstate/pull/797 to allow re-using the same getters.

EdDev: Done

PR #799, opened 2020-02-17
nm.team: get the configuration from the device

ffmancera:
The connection profile only contains static information. If an user
modifies the team interface using nmcli or teamd we need to reflect the
changes in nmstate.

### DIFF
--- a/libnmstate/nm/team.py
+++ b/libnmstate/nm/team.py
@@ -20,7 +20,6 @@
 import copy
 import json
 
-from libnmstate.nm import connection as nm_connection
 from libnmstate.nm import nmclient
 from libnmstate.schema import Interface
 from libnmstate.schema import Team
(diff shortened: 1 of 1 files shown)


libnmstate/nm/team.py
```diff
@@ -82,17 +82,8 @@ def get_info(device):
     """
```
tyll: Is the bugzilla still relevant?

cathay4t: You mean https://bugzilla.redhat.com/show_bug.cgi?id=1798947 ?

ffmancera:
> As mentioned in https://bugzilla.redhat.com/show_bug.cgi?id=1792232#c1 the NM provided runtime information is incorrect.
> Before they fix that, we should parse the `teamdctl team0 state dump` json output.

I've discussed this with Beniamino this morning:
* Connection profile contains only static information so we cannot use it.
* We can use `NM.DeviceTeam` because it gets updated when teamd daemon signals its change.

Initially, `NM.DeviceTeam` should work for us. It would be better to have some kind of `nm_device_team_refresh_config()` but until then we should be fine with using `NM.DeviceTeam`

libnmstate/nm/team.py
cathay4t: Never mind. You mean 1792232.

cathay4t:
Approved. Thank you.

cathay4t:
The CI failure is unrelated to this PR and the NM fix is coming soon.

PR #804, opened 2020-02-17
nm, team: Fix reporting of team slaves

EdDev:
When configuring team ports, the team may return a state like this:
`{'device': 'team0', 'ports': {}, 'runner': {'name': 'roundrobin'}}`

such a state includes the `ports` key with an empty dictionary.

The team ports reporting is adjusted to support such a report.

### DIFF
--- a/libnmstate/nm/team.py
+++ b/libnmstate/nm/team.py
@@ -27,6 +27,7 @@ from libnmstate.schema import Team
 
 CAPABILITY = "team"
 TEAMD_JSON_DEVICE = "device"
+TEAMD_JSON_PORTS = "ports"
 
 
 def has_team_capability():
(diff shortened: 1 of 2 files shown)


libnmstate/nm/team.py
```diff
@@ -96,11 +96,9 @@ def get_info(device):
 
 def _convert_teamd_config_to_nmstate_config(team_config):
```
tyll: I do not understand the commit message. What does `the team may return` mean - do you mean teamd?

libnmstate/nm/team.py
```diff
@@ -96,11 +96,9 @@ def get_info(device):
...
     team_config.pop(TEAMD_JSON_DEVICE, None)
-    port_config = team_config.get(Team.PORT_SUBTREE)
-
-    if port_config:
-        team_port = _teamd_port_to_nmstate_port(port_config)
-        team_config[Team.PORT_SUBTREE] = team_port
+    port_config = team_config.get(Team.PORT_SUBTREE, {})
```
tyll: Is `Team.PORT_SUBTREE` correct here? It seems to be the `ports` constant from the teamd config and not the Nmstate port subtree state here. Or is it?

libnmstate/nm/team.py
EdDev: Yes, teamd.

libnmstate/nm/team.py
EdDev: You are right, it is teamd state, not nmstate.

EdDev:
The fix will keep the team tests failing because it takes at least 5 seconds for the teamd to report back an updated state where the slaves are included.

Another PR needs to try and fetch the slaves from the NM device using ‘get_slaves’ kike be do with the other interface types.
This change is still relevant for the other properties of a slave (beyond the name).

libnmstate/nm/team.py
EdDev: I updated the commit message.

libnmstate/nm/team.py
EdDev: I changed it to express this.
We return teamd state adjusted a bit, I hope I expressed it in the code so it will be clear.

PR #807, opened 2020-02-18
nm.team: fetch the slaves using NM.DeviceTeam.get_slaves()

ffmancera:
nm.team: fetch the slaves using NM.DeviceTeam.get_slaves()
    
NM.DeviceTeam configuration might contains outdated information. This
causing a verification error when setting ports in a team interface.
Using the function get_slaves() to fetch the updated information.
    
### DIFF
--- a/libnmstate/nm/team.py
+++ b/libnmstate/nm/team.py
@@ -89,26 +89,30 @@ def get_info(device):
         teamd_json = device.get_config()
         if teamd_json:
             teamd_config = json.loads(teamd_json)
-            team_config = _convert_teamd_config_to_nmstate_config(teamd_config)
+            slave_names = [dev.get_iface() for dev in device.get_slaves()]
+            team_config = _convert_teamd_config_to_nmstate_config(
+                teamd_config, slave_names
+            )
             info[Team.CONFIG_SUBTREE] = team_config
 
     return info
 
 
-def _convert_teamd_config_to_nmstate_config(teamd_config):
+def _convert_teamd_config_to_nmstate_config(teamd_config, slave_names):
     teamd_config.pop(TEAMD_JSON_DEVICE, None)
     port_config = teamd_config.get(TEAMD_JSON_PORTS, {})
-    team_port = _teamd_port_to_nmstate_port(port_config)
+    team_port = _merge_port_config_with_slaves_info(port_config, slave_names)
 
     team_config = teamd_config
     team_config[Team.PORT_SUBTREE] = team_port
     return team_config
 
 
-def _teamd_port_to_nmstate_port(port_config):
+def _merge_port_config_with_slaves_info(port_config, slave_names):
     port_list = []
-    for name, port in port_config.items():
-        port.update({Team.Port.NAME: name})
+    for name in slave_names:
+        port = port_config.get(name, {})
+        port[Team.Port.NAME] = name
         port_list.append(port)
 
     return port_list
--- a/tests/integration/team_test.py
+++ b/tests/integration/team_test.py
@@ -59,7 +59,7 @@ def test_create_team_iface_without_slaves():
 @pytest.mark.xfail(
     raises=NmstateLibnmError, reason="https://bugzilla.redhat.com/1798947"
 )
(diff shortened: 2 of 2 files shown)


libnmstate/nm/team.py
```diff
@@ -88,27 +88,32 @@ def get_info(device):
...
     port_list = []
-    for name, port in port_config.items():
-        port.update({Team.Port.NAME: name})
+    slaves_name = [dev.get_iface() for dev in device.get_slaves()]
+    for name in slaves_name:
+        port = {Team.Port.NAME: name}
+        if port_config:
+            port_options = port_config.get(name)
+            if port_options:
+                port.update(port_options)
```
EdDev: How about this:
```
port = port_config.get(name, {})
port[Team.Port.NAME] = name
```

libnmstate/nm/team.py
```diff
@@ -88,27 +88,32 @@ def get_info(device):
         teamd_json = device.get_config()
         if teamd_json:
             teamd_config = json.loads(teamd_json)
-            team_config = _convert_teamd_config_to_nmstate_config(teamd_config)
+            team_config = _convert_teamd_config_to_nmstate_config(
+                teamd_config, device
```
EdDev: Passing the list of slaves names will be nicer IMO. Letting this functions know about the NM device adds a burden which is not needed.

libnmstate/nm/team.py
ffmancera: Done, thanks!

libnmstate/nm/team.py
ffmancera: I agree, thanks!

ffmancera:
> Commit message:
> 
> > NM.DeviceTeam configuration might contains outdated information. This
> > causing a verification error when setting ports in a team interface.
> > Using the function get_slaves() to fetch the updated information.
> 
> Fetch the list of a team device slaves using the NM DeviceTeam object.
> Attempting to fetch it through the teamd state is not working well because it is not updated quick enough.

Done, thank you!

cathay4t:
@ffmancera Can you provide a example when the `NM.DeviceTeam.get_config()` is incorrect on the slave list?
I need to evaluate the risk when slaves information is not consistent with other option get from `NM.DeviceTeam.get_config()`

From my point of view, provide incorrect data is worse than provide no data.

ffmancera:
> @ffmancera Can you provide a example when the `NM.DeviceTeam.get_config()` is incorrect on the slave list?
> I need to evaluate the risk when slaves information is not consistent with other option get from `NM.DeviceTeam.get_config()`
> 
> From my point of view, provide incorrect data is worse than provide no data.

The examples are the integration tests without this patch they are failing when we are setting a team interface with slaves.

EdDev:
@cathay4t , this change fetches the slaves from the NM device instead of teamd because teamd is not updating its state quick enough. All other data is retrieved from teamd.

A long term solution is to fetch all info from the NM Device. Also, waiting on an updated teamd state could be an option for the future, but for that we will need NM to provide a signal when that happens. Based on local tests, it takes teamd 5sec to be updated with the slaves list.

cathay4t:
> @cathay4t , this change fetches the slaves from the NM device instead of teamd because teamd is not updating its state quick enough. All other data is retrieved from teamd.
> 
> A long term solution is to fetch all info from the NM Device. Also, waiting on an updated teamd state could be an option for the future, but for that we will need NM to provide a signal when that happens. Based on local tests, it takes teamd 5sec to be updated with the slaves list.

Should I remove the test code on checking output of `teamdctl` ?

ffmancera:
> > @cathay4t , this change fetches the slaves from the NM device instead of teamd because teamd is not updating its state quick enough. All other data is retrieved from teamd.
> > A long term solution is to fetch all info from the NM Device. Also, waiting on an updated teamd state could be an option for the future, but for that we will need NM to provide a signal when that happens. Based on local tests, it takes teamd 5sec to be updated with the slaves list.
> 
> Should I remove the test code on checking output of `teamdctl` ?

No, I think it is fine. It is a double check which is always good IMHO. :-)

PR #808, opened 2020-02-18
travis: do not mark integ_slow tests as allowed to fail

ffmancera:


tyll:
Merging without rebase since it only changes the travis file and travis started all jobs as expected.

PR #810, opened 2020-02-18
Docker run fix

EdDev:


automation/run-tests.sh
```diff
@@ -307,7 +307,7 @@ dump_network_info
 
 pyclean
 if [[ "$CI" != "true" ]];then
-    container_exec "cp -rf $CONTAINER_WORKSPACE /root/nmstate-workspace"
+    container_exec "cp -rf $CONTAINER_WORKSPACE /root/nmstate-workspace || true"
```
tyll: This does not seem to be correct. If copying failed, the test should not continue.

EdDev: I explained this in the commit. The copy may fail even if the files copy fully. And if there is a problem, the tests will fail later on.

tyll: can you maybe change this to do this only if `CONTAINER_CMD == docker`?

EdDev: Sure, thanks.

EdDev: Will this suffice?
```bash
    if [ "$CONTAINER_CMD" == "docker" ];then
      container_exec 'rm -rf $CONTAINER_WORKSPACE/*nmstate*.rpm' || true
    else
      container_exec 'rm -rf $CONTAINER_WORKSPACE/*nmstate*.rpm'
    fi
```

tyll: works for me

EdDev: Done

automation/run-tests.sh
```diff
@@ -34,8 +34,12 @@ test -t 1 && USE_TTY="-t"
 function remove_container {
     res=$?
     [ "$res" -ne 0 ] && echo "*** ERROR: $res"
-    container_exec 'rm -rf $CONTAINER_WORKSPACE/*nmstate*.rpm'
-    ${CONTAINER_CMD} rm $CONTAINER_ID -f
+    if [ "$CONTAINER_CMD" == "docker" ];then
```
tyll: `if $CONTAINER_CMD --version | grep -qv podman; then` should run this only if it is not podman. With podman-docker installed, the docker command will actually be podman but show this in the version.

EdDev: Done, thanks.

automation/run-tests.sh
tyll: This conversation was actually for the cp not the rm

EdDev: Ohh, right. Fixed.

EdDev:
CI failures on unrelated issues (OVS related).
The job that tracks NM 1.22-git passes.
Locally it passes.
Force merging this PR.

PR #818, opened 2020-02-19
Travis disable fedora jobs

EdDev:
Depends on #817


tyll:
It launches all the travis jobs as expected, IMHO there is no need to wait for them to complete.

PR #819, opened 2020-02-19
examples, team: add team interface examples

ffmancera:

### DIFF
--- a/examples/team0_with_slaves.yml
+++ b/examples/team0_with_slaves.yml
@@ -1,0 +1,11 @@
+---
+interfaces:
+- name: team0
+  type: team
+  state: up
+  team:
+    ports:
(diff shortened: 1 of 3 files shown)


examples/team0_up.yml
```diff
@@ -0,0 +1,5 @@
+---
```
EdDev: I see no need to show a team without slaves as an example. It is a useless config.

ffmancera: Done, thanks!

examples/team0_with_slaves.yml
```diff
@@ -0,0 +1,11 @@
+---
```
tyll: IMHO you do not need to mention that the interface has slaves in the file name.

PR #820, opened 2020-02-19
CHANGELOG: update change log for 0.2.5 release

ffmancera:

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,32 @@
 # Changelog
+## [0.2.5] - 2020-02-19
+### Breaking changes
+ - Deprecate specifying keyword arguments as positional arguments in the public methods.
+   This will become a hard failure in Nmstate-0.3.0 and later and it affects the following functions:
+   - "libnmstate.apply()"
+   - "libnmstate.commit()"
+   - "libnmstate.rollback()"
+   - "libmstate.show()"
+
+ - NetworkManager 1.22.8 or later is now required.
+
+### New Features
+ - Report missing NetworkManager plugin during validation.
+ - Support OVS Link Aggregation (Bonding).
+
+### Bug Fixes
+ - Fix "AtributeError" when editing an existing OVS interface.
+ - Do not remove slaves implicitly when removing the master interface.
+ - Fix team interface port support.
+ - Prevent virtual interface removal after DHCP or IPV6-RA timeout.
+ - Get Linux Bonding options from kernel instead of reading the NetworkManager configuration.
+
+### Known issues when using NetworkManager 1.22.8
+ - All changes to bridges might disrupt their connectivity.
+ - Cannot set Linux Bond option to "fail_over_mac=1".
+ - Cannot detach a system port from an OVS bridge.
+ - Sometimes, NetworkManager may throw a "Failed to find a compatible device for this connection" error.
+
 ## [0.2.4] - 2020-02-12
 ### Bug Fixes
  - Reduce minimal required MAC address length to 4 octets


CHANGELOG
```diff
@@ -1,4 +1,30 @@
 # Changelog
+## [0.2.5] - 2020-02-19
+### Breaking changes
+ - Deprecate positional arguments in the following methods:
```
tyll: Please reword this like:

Deprecate specifying keyword arguments as positional arguments in the public methods:

This will become a hard failure in Nmstate-0.3.0 and later.

CHANGELOG
```diff
@@ -1,4 +1,30 @@
...
+   - "libnmstate.apply()"
+   - "libnmstate.commit()"
+   - "libnmstate.rollback()"
+   - "libmstate.show()"
+
+### New Features
+ - Report missing NM plugin during validation.
+ - Support OVS Link Aggregation (LAG).
```
tyll: IMHO it should say "Bonding" instead of "LAG".

CHANGELOG
```diff
@@ -1,4 +1,30 @@
 # Changelog
+## [0.2.5] - 2020-02-19
+### Breaking changes
```
tyll: This should also mention that NetworkManager 1.22.8 or later is now required.

CHANGELOG
```diff
@@ -1,4 +1,30 @@
...
+ - Fix "AtributeError" when editing an existing OVS interface.
+ - Do not remove slaves implicitly when removing the master interface.
+ - Fix team interface port support.
+ - Prevent virtual interface removal after DHCP or IPV6-RA timeout.
+
+### Known Issues
+
+In nmstate 0.2.5 there are some known issues:
+ - Cannot activate a connection against unmanaged network interfaces.
+ - Cannot reapply when bridge profile previously does not contain 802-3-ethernet section.
```
tyll: This is too much internal information IMHO. An Nmstate user would not know what this means. I also do not know from this text when exactly this will happen or what this means. I suspect that it means that sometimes the link of interfaces goes down even if it would not be necessary.

CHANGELOG
```diff
@@ -1,4 +1,30 @@
...
+### Bug Fixes
+ - Fix "AtributeError" when editing an existing OVS interface.
+ - Do not remove slaves implicitly when removing the master interface.
+ - Fix team interface port support.
+ - Prevent virtual interface removal after DHCP or IPV6-RA timeout.
+
+### Known Issues
+
+In nmstate 0.2.5 there are some known issues:
+ - Cannot activate a connection against unmanaged network interfaces.
```
tyll: AFAIU, this is worked around somehow. @cathay4t is mentioning a race condition in the workaround but I am not sure how this manifests.

CHANGELOG
```diff
@@ -1,4 +1,30 @@
...
+ - Cannot reapply when bridge profile previously does not contain 802-3-ethernet section.
+ - Cannot set bond option to "arp_validate=active".
```
tyll: This should mention which bond. Is it OVS Bond or Linux Bond?

ffmancera: Is Linux Bond, I am specifying it. Thanks!

CHANGELOG
ffmancera: Hm. In that case I am removing it. Thanks!

CHANGELOG
ffmancera: Should we then remove it or rephrase it? In my opinion, `sometimes the link of interfaces goes down even if it would not be necessary` sounds very generic. Any suggestion? Thanks!

CHANGELOG
ffmancera: Thanks!

CHANGELOG
ffmancera: Done, thank you!

CHANGELOG

CHANGELOG
tyll: @EdDev can you maybe give a high-level description of the bug?

EdDev: How about: `All bridge changes require disruptive actions.` ?

tyll: Reword: All changes to bridges might disrupt their connectivity

CHANGELOG
```diff
@@ -1,4 +1,32 @@
...
+ - Fix "AtributeError" when editing an existing OVS interface.
+ - Do not remove slaves implicitly when removing the master interface.
+ - Fix team interface port support.
+ - Prevent virtual interface removal after DHCP or IPV6-RA timeout.
+
+### Known Issues
+
+In nmstate 0.2.5 there are some known issues:
+ - All changes to bridges might disrupt their connectivity.
+ - Cannot set Linux Bond option to "arp_validate=active".
```
EdDev: This is partially correct, the issue is not really agreed upon.
Nm team argued that this value is irrelevant for some bond modes and that is why it is failing.
I would drop it.

Not sure about the next one with the fail-over.

EdDev: This is partially correct, the issue is not really agreed upon.
Nm team argued that this value is irrelevant for some bond modes and that is why it is failing.
I would drop it.

Not sure about the next one with the fail-over.

CHANGELOG
```diff
@@ -1,4 +1,32 @@
...
+
+### Bug Fixes
+ - Fix "AtributeError" when editing an existing OVS interface.
+ - Do not remove slaves implicitly when removing the master interface.
+ - Fix team interface port support.
+ - Prevent virtual interface removal after DHCP or IPV6-RA timeout.
+
+### Known Issues
+
+In nmstate 0.2.5 there are some known issues:
```
EdDev: What about the issues from NM side?
- Detaching a system port from an OVS bridge.
- Randomly failing with (connection not found or something).

CHANGELOG
ffmancera: I am fine with droping this one if that is the reasoning. I would keep the next one with the fail-over.

ffmancera: I am fine with dropping this one if that is the reasoning. I would keep the next one with the fail-over.

CHANGELOG
EdDev: I think the second point was missed.

It is from:
https://bugzilla.redhat.com/show_bug.cgi?id=1804350

The error we saw is `Failed to find a compatible device for this connection`.

CHANGELOG
```diff
@@ -1,4 +1,32 @@
...
+ - Do not remove slaves implicitly when removing the master interface.
+ - Fix team interface port support.
+ - Prevent virtual interface removal after DHCP or IPV6-RA timeout.
+ - Get Linux Bonding options from kernel instead of reading the NetworkManager configuration.
+
+### Known issues when using NetworkManager 1.22.8
+ - All changes to bridges might disrupt their connectivity.
+ - Cannot set Linux Bond option to "fail_over_mac=1".
+ - Cannot detach a system port from an OVS bridge.
+ - Randomly NM may throw a "Failed to find a compatible device for this connection" error.
```
tyll: Better write NetworkManager here

CHANGELOG
```diff
@@ -1,4 +1,32 @@
...
+   This will become a hard failure in Nmstate-0.3.0 and later and it affects the following functions:
+   - "libnmstate.apply()"
+   - "libnmstate.commit()"
+   - "libnmstate.rollback()"
+   - "libmstate.show()"
+
+ - NetworkManager 1.22.8 or later is now required.
+
+### New Features
+ - Report missing NM plugin during validation.
```
tyll: Please write NetworkManager here

CHANGELOG
tyll: Also "Sometime, ..." would be nicer than "Randomly" because I bet it is not strictly random ;-) - also it needs a comma.

CHANGELOG

CHANGELOG
ffmancera: Done, thank you!

EdDev:
No need to run tests, this is a text file.

PR #823, opened 2020-02-19
python3: Remove Python 2 support leftovers

elvgarrui:


elvgarrui:
> Oops, It seems you forgot to sign the commit. You can do `git commit -s` and in this case you can do `git commit --amend -s`. Thanks!

Fixed, thanks!

PR #828, opened 2020-02-21
validator: allow validation for states without interfaces

ffmancera:
Now it is possible to apply the a DNS configuration without specifying
the interfaces or routes. e.g.

```
---
dns-resolver:
  config:
    search:
    - example.com
    - example.org
    server:
    - 2001:4860:4860::8888
    - 8.8.8.8
```

### DIFF
--- a/libnmstate/validator.py
+++ b/libnmstate/validator.py
@@ -24,7 +24,6 @@ import jsonschema as js
 
 from . import nm
 from . import schema
-from .schema import Constants
 from libnmstate.appliers.bond import is_in_mac_restricted_mode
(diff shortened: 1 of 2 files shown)


cathay4t:
But we have bug which cause you lose the routes of the interface been chose as DNS.

To fix that, the metadata generate stage could take 3 state:
1. Partial merged state. Only interface in desire state merged.
2. Full merged state. All interface in current merged.
3. Original state. Might be used for some checking to make sure certain action is not user oriented.

This requires a lot code changes and better postponed to next major release.

ffmancera:
Rebased!

PR #831, opened 2020-02-21
automation, travis: Specify registry for container images

### LINKED ISSUE #785: Prevent name squatting for container images
Incomplete PR: #673 

To fix name squatting problems, specify the registry in all places (travis config, scripts). This needs to ensure that local rebuilds can still be used and that Travis will still pick up a locally built image when necessary.

elvgarrui:
Container registry is now specified at container images in order to
avoid name squatting attacks or any other related problems.

Ref: https://raesene.github.io/blog/2019/09/25/typosquatting-in-a-multi-registry-world/

This PR fixes #785.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,28 +5,28 @@ env:
     global:
         - use_coveralls=true
     matrix:
-        - CONTAINER_IMAGE=nmstate/centos8-nmstate-dev
+        - CONTAINER_IMAGE=docker.io/nmstate/centos8-nmstate-dev
           testflags="--test-type integ --pytest-args='-x'
               --copr networkmanager/NetworkManager-1.22-git"
           use_codecov=true
           use_coveralls=false
-        - CONTAINER_IMAGE=nmstate/centos8-nmstate-dev
+        - CONTAINER_IMAGE=docker.io/nmstate/centos8-nmstate-dev
           testflags="--test-type integ_slow --pytest-args='-x'
               --copr networkmanager/NetworkManager-1.22-git"
-        - CONTAINER_IMAGE=nmstate/centos8-nmstate-dev
+        - CONTAINER_IMAGE=docker.io/nmstate/centos8-nmstate-dev
           testflags="--test-type integ --pytest-args='-x'
               --copr networkmanager/NetworkManager-master"
-        - CONTAINER_IMAGE=nmstate/centos8-nmstate-dev
+        - CONTAINER_IMAGE=docker.io/nmstate/centos8-nmstate-dev
           testflags="--test-type format"
-        - CONTAINER_IMAGE=nmstate/centos8-nmstate-dev
+        - CONTAINER_IMAGE=docker.io/nmstate/centos8-nmstate-dev
           testflags="--test-type lint"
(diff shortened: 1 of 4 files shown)


tyll:
Hi, thank you very much!

Since this PR fixes #785, it would be great if you would mention this in the comment and add this to the "linked issues" at the right (if you can, not sure if it would be possible for you).

Did you test/know if this change will still allow to run locally built container images as described in [the documentation](https://github.com/nmstate/nmstate/blob/master/automation/README.md#build-a-new-container-image)? We need to ensure that if the CI rebuilds an image because of changes to the container spec, that the updated image will run instead of the old image from the registry.

I noticed that the first instructions seem to be wrong and also sudo should not be necessary anymore. If you can, it would be nice if you could cleanup this in a second commit or pull request.

elvgarrui:
> Since this PR fixes #785, it would be great if you would mention this in the comment and add this to the "linked issues" at the right (if you can, not sure if it would be possible for you).


@tyll Do you mean that I should mention it just in the PR comment, or also in the commit? =)

> Did you test/know if this change will still allow to run locally built container images as described in the documentation? We need to ensure that if the CI rebuilds an image because of changes to the container spec, that the updated image will run instead of the old image from the registry.

Yes, it works correctly! @ffmancera warned me about it when I asked how I should test this.

> I noticed that the first instructions seem to be wrong ...

I'm sorry but I'm not totally sure about what is wrong (apart from the unnecessary sudo) in that first instruction you are referring to. Are you talking about `sudo ../packaging/build-container.sh local/centos8-nmstate-dev`?

EdDev:
> Yes, it works correctly! @ffmancera warned me about it when I asked how I should test this.

In the `run-test` script, `rebuild_container_images` is executing the image build. The resulted image should have the `docker.io` registry included, but I do not see what explicitly sets it.

I suggest to create a draft PR over this PR and check if it indeed runs an image rebuild. You should see that in the CI logs (make sure to update something in the `packaging` folder).

elvgarrui:
> In the `run-test` script, `rebuild_container_images` is executing the image build. The resulted image should have the `docker.io` registry included, but I do not see what explicitly sets it.

That image is created locally from a dockerfile, therefore there is no option of having any kind of name squatting. The 'docker.io' needs to be included when the image is downloaded, not locally created. (Note that when the basic centos-8/fedora image is pulled in the Dockerfile to create that local image, the 'docker.io' prefix is there)

> I suggest to create a draft PR over this PR and check if it indeed runs an image rebuild. You should see that in the CI logs (make sure to update something in the `packaging` folder).
> Maybe make sure to `echo` something in the new image to confirm that it is the one who actually ran.

I created a draft PR with an echo on the Dockerfiles and it seems to work. 
(Line 422 at https://travis-ci.com/nmstate/nmstate/jobs/290835573)

elvgarrui:
@EdDev I'm sorry, you were right! The system was not recognizing the built image because 'docker.io/' was not included. I've fixed it now.
@tyll I've fixed the documentation, thanks for pointing it out!

elvgarrui:
Rebased!

PR #833, opened 2020-02-23
integ: fix missing clean up in vlan integration tests

ffmancera:
The missing clean up in these tests were causing that
tests/integration/timeout_test.py was failing after running
tests/integration/vlan_test.py.

### DIFF
--- a/tests/integration/vlan_test.py
+++ b/tests/integration/vlan_test.py
@@ -74,59 +74,57 @@ def test_add_and_remove_two_vlans_on_same_iface(eth1_up):
 
 
 def test_two_vlans_on_eth1_change_mtu(eth1_up):
-    desired_state = statelib.show_only(("eth1",))
-    eth1_state = desired_state[Interface.KEY][0]
-    vlans_state = create_two_vlans_state()
-    desired_state[Interface.KEY].extend(vlans_state[Interface.KEY])
-    for iface in desired_state[Interface.KEY]:
-        iface[Interface.MTU] = 2000
-    libnmstate.apply(desired_state)
+    with two_vlans_on_eth1() as desired_state:
+        eth1_state = eth1_up[Interface.KEY][0]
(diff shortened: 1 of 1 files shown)


tests/integration/vlan_test.py
```diff
@@ -74,59 +74,54 @@ def test_add_and_remove_two_vlans_on_same_iface(eth1_up):
...
 def test_two_vlans_on_eth1_change_mtu(eth1_up):
-    desired_state = statelib.show_only(("eth1",))
-    eth1_state = desired_state[Interface.KEY][0]
-    vlans_state = create_two_vlans_state()
-    desired_state[Interface.KEY].extend(vlans_state[Interface.KEY])
-    for iface in desired_state[Interface.KEY]:
-        iface[Interface.MTU] = 2000
-    libnmstate.apply(desired_state)
+    with two_vlans_on_eth1() as desired_state:
+        desired_state[Interface.KEY].extend(eth1_up[Interface.KEY])
```
EdDev: You can extract `eth1_up[Interface.KEY][0]` to a variable and reuse it on line 88.
Then, use here `append` instead.

Same for the other tests.

ffmancera: Done, thanks!

ffmancera:
> > integ: fix missing clean up in vlan integration tests
> 
> tests, integ: Fix ...
> 
> > The missing clean up in these tests were causing that
> > tests/integration/timeout_test.py was failing after running
> > tests/integration/vlan_test.py.
> 
> The timeout integration tests fail due to improper vlan tests cleanup.
> To fix the issue, remove explicitly the vlan interfaces in the tests.
> 
> This also shows a real issue that merits a bug. Removing a base interface (of a vlan) does not implicitly remove its dependent vlans. It should, or it should not allow to remove the base (more likely).

Done, thanks! I've created https://github.com/nmstate/nmstate/issues/835 to discuss and track this bug.

PR #834, opened 2020-02-23
nm, bond: Report only non-default settings

EdDev:
Except the `mode` bond option, report only the non-default option
values.

This change is avoiding an undesired merge of default values to the
desired state, which in turn causes defaults to be explicitly set on the
NM profile. Setting some defaults explicitly may cause validation errors
that are introduced by NM.

### DIFF
--- a/libnmstate/nm/bond.py
+++ b/libnmstate/nm/bond.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2018-2019 Red Hat, Inc.
+# Copyright (c) 2018-2020 Red Hat, Inc.
 #
 # This file is part of nmstate
 #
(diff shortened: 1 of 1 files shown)


cathay4t:
I don't like the approach.
Then on-disk profile different the one activated, this show command might showing the
incorrect data.

Regarding enforcing the default explicitly, I think we had this talk a while back: why we
create new profile instead of update changed values: everything should be explicitly
created in case default changes or on-disk profile does not match with run time.

EdDev:
@cathay4t , this change does not avoid setting the defaults explicitly if the user requests to do so. It does not complement defaults on merge nor reports them at all.
If someone change these values on the kernel, they will get reported and therefore explicitly set on-disk.

Per NM behavior, I see no other choice at the moment. NM does not allow setting explicitly all the defaults.

What scenario you see here as problematic?

cathay4t:
Sorry. You are using `NM.SettingBond.get_option_default()` which should have no problems.

The change request would be:

Please don't use `_get_bond_profile_setting(nm_device)` to the `NM.SettingBond` which might provides None. You should use `NM.SettingBond.new()` instead, which is sufficient. For example:
`NM.SettingBond.new().get_option_default('miimon')`

EdDev:
> Please don't use _get_bond_profile_setting(nm_device) to the NM.SettingBond which might provides None. You should use NM.SettingBond.new() instead, which is sufficient. For example:
> NM.SettingBond.new().get_option_default('miimon')

Done, thanks.

PR #836, opened 2020-02-24
verification, bond: prepare "arp_ip_target" option for verification

ffmancera:
The "arp_ip_target" option format for NM is "10.1.1.2,10.1.1.3" but we
are reading the options from sysfs and the format is
"10.1.1.2 10.1.1.3".

### DIFF
--- a/tests/integration/bond_test.py
+++ b/tests/integration/bond_test.py
@@ -501,3 +501,21 @@ def test_create_bond_with_mac(eth1_up, eth2_up):
         extra_iface_state={Interface.MAC: MAC0},
     ) as state:
         assertlib.assert_state_match(state)
+
+
+@pytest.mark.parametrize("ips", ("192.0.2.1,192.0.2.2", "192.0.2.2,192.0.1.1"))
+def test_bond_with_arp_ip_target(eth1_up, eth2_up, ips):
+    with bond_interface(
+        name=BOND99,
+        slaves=[ETH1, ETH2],
+        extra_iface_state={
+            Bond.CONFIG_SUBTREE: {
+                Bond.MODE: BondMode.ACTIVE_BACKUP,
+                Bond.OPTIONS_SUBTREE: {
+                    "arp_interval": 1000,
+                    "arp_ip_target": ips,
+                },
+            },
+        },
+    ) as desired_state:
+        assertlib.assert_state_match(desired_state)
--- a/libnmstate/nm/bond.py
+++ b/libnmstate/nm/bond.py
@@ -77,6 +77,8 @@ def _get_options(nm_device):
             if option == "mode" or value != bond_setting.get_option_default(
                 option
             ):
+                if option == "arp_ip_target":
+                    value = value.replace(" ", ",")
                 options[option] = value
     return options
 


tests/integration/testlib/statelib.py
```diff
@@ -227,6 +227,8 @@ def _convert_lag_numeric_value_options_to_integer(self):
                     for option_name, option_value in options.items():
                         with contextlib.suppress(ValueError):
                             option_value = int(option_value)
+                        if option_name == "arp_ip_target":
```
tyll: Shouldn't this also use the constant?

tests/integration/bond_test.py
```diff
@@ -501,3 +501,20 @@ def test_create_bond_with_mac(eth1_up, eth2_up):
...
+def test_bond_with_arp_ip_target(eth1_up, eth2_up):
+    with bond_interface(
+        name=BOND99,
+        slaves=[ETH1, ETH2],
+        extra_iface_state={
+            Bond.CONFIG_SUBTREE: {
+                Bond.MODE: BondMode.ACTIVE_BACKUP,
+                Bond.OPTIONS_SUBTREE: {
+                    "arp_interval": 1000,
+                    "arp_ip_target": "10.1.1.2,10.1.1.3",
```
tyll: This should probably use the constant

tests/integration/testlib/statelib.py
```diff
@@ -227,6 +227,8 @@ def _convert_lag_numeric_value_options_to_integer(self):
...
+                            option_value = option_value.replace(",", " ")
```
tyll: Why does this need to be normalized in the test specific code? Shouldn't the Nmstate code properly handle the format differences?

ffmancera: As is a difference in the verification phase, we need to also normalize for verification when using assert_state_match(). If not, the desired state won't match against the current state.

tests/integration/bond_test.py
ffmancera: Oops, yes sorry.

tests/integration/testlib/statelib.py
ffmancera: Oops! Thanks!

ffmancera: Done!

tests/integration/bond_test.py
ffmancera: Done!

tests/integration/testlib/statelib.py
tyll: Not sure I follow. Do you want to make Nmstate accept both `"10.1.1.2,10.1.1.3"` and `"10.1.1.2 10.1.1.3"` in the desired state? IMHO it should accept only one of those formats and also show only one of those formats. Then not adjustments for verification should be required.

tyll: @EdDev @cathay4t what do you think?

ffmancera: I agree with you. It does not make sense to accept both formats or accept one and then shows another.

almusil:
Working for vdsm

tests/integration/bond_test.py
tyll: Actually, please do not use these IP addresses but RFC but addresses from https://www.kernel.org/doc/Documentation/networking/bonding.txt

ffmancera: Done, thanks!

PR #842, opened 2020-02-24
README: Remove nmstate-0.1 Copr repos

tyll:
Remove nmstate-0.1 Copr repos because they are going to be removed,
soon.


tyll:
Merging without a rebase since it does not change anything covered by the CI.

PR #843, opened 2020-02-24
VERSION: Bump to 0.2.7

tyll:


EdDev:
If we intend to branch to nmstate-0.2 now, then this should be there and on master we should have 0.3.0.

tyll:
We can also bump master to 0.3.0 once it is branched.

Issue #837, opened 2020-02-24
Need better error message when openvswitch is not started.


cathay4t:
When openvswitch is not started, we got libnm error with:

```
2020-02-24 20:13:04,092 root         ERROR    NM main-loop aborted: 
Connection activation failed on ovs0: reason=<enum 
NM_ACTIVE_CONNECTION_STATE_REASON_DEVICE_DISCONNECTED of type 
NM.ActiveConnectionStateReason>
```

Need better error message.

tyll:
User story:

As an Nmstate API user, when I specify or change an OVS interface but the openvswitch service is unavailable, `libnmstate.apply()` raises an appropriate exception.

Implementation ideas:
During validation, if a ovs interface is specified:
- Check with systemd if openvswitch service is running and raise if not
- or: Check if ovs-vsctl command works and raise if not

ffmancera:
> User story:
> 
> As an Nmstate API user, when I specify or change an OVS interface but the openvswitch service is unavailable, `libnmstate.apply()` raises an appropriate exception.
> 
> Implementation ideas:
> During validation, if a ovs interface is specified:
> 
>     * Check with systemd if openvswitch service is running and raise if not
> 
>     * or: Check if ovs-vsctl command works and raise if not

In my opinion the systemd approach is better. Also we should do this for teamd service. Thanks!

PR #852, opened 2020-02-25
ActiveConnection: Handle uninitialized device

tyll:
Handle ActiveConnection._nmdev being None. This should fix a Traceback
    when the openvswitch service is not running:
    
      File "/usr/lib/python3.6/site-packages/libnmstate/nm/connection.py", line 226, in _active_connection_callback
        ac.devname, ac.reason
      File "/usr/lib/python3.6/site-packages/libnmstate/nm/active_connection.py", line 171, in devname
        return self._nmdev.get_iface()
    AttributeError: 'NoneType' object has no attribute 'get_iface'

This also makes everything use ActiveConnection.devname

### DIFF
--- a/libnmstate/nm/active_connection.py
+++ b/libnmstate/nm/active_connection.py
@@ -86,7 +86,7 @@ class ActiveConnection:
             if self._mainloop.is_action_canceled(e):
                 logging.debug(
                     "Connection deactivation aborted on %s: error=%s",
-                    self._nmdev.get_iface(),
+                    self.devname,
                     e,
                 )
             else:
@@ -105,20 +105,19 @@ class ActiveConnection:
                 else:
                     self._mainloop.quit(
                         "Connection deactivation failed on {}: "
-                        "error={}".format(self._nmdev.get_iface(), e)
+                        "error={}".format(self.devname, e)
                     )
                     return
 
         if success:
             logging.debug(
-                "Connection deactivation succeeded on %s",
-                self._nmdev.get_iface(),
+                "Connection deactivation succeeded on %s", self.devname,
             )
             self._mainloop.execute_next_action()
         else:
             self._mainloop.quit(
                 "Connection deactivation failed on %s: error=unknown"
-                % self._nmdev.get_iface()
+                % self.devname
             )
 
     @property
--- a/libnmstate/nm/active_connection.py
+++ b/libnmstate/nm/active_connection.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2019 Red Hat, Inc.
+# Copyright (c) 2019-2020 Red Hat, Inc.
 #
 # This file is part of nmstate
 #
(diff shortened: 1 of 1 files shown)


libnmstate/nm/active_connection.py
```diff
@@ -168,7 +168,10 @@ def nm_active_connection(self):
 
     @property
     def devname(self):
-        return self._nmdev.get_iface()
+        if self._nmdev:
+            return self._nmdev.get_iface()
+        else:
+            return None
```
ffmancera: Is this fine? If `_nmdev` is None the logs are going to show something like: `Connection deactivation failed on None: error=unknown`. I am not sure if it will help or confuse the user but I don't have any other solution on mind.

EdDev: I agree with @ffmancera on this.

The change itself may be fine, but the reasoning for it (in the commit message) is not IMO.
If `devname` has been queried, it means it was expected to exist, otherwise it would not have been called.
And the reasoning about OVS service is just masking the real problem (or other future problems).

We should probably blow up much earlier or not reach this situation at all.
For the OVS issue itself, I think we should just not proceed if it is not running, like the NM ovs plugin.

tyll: > Is this fine? If `_nmdev` is None the logs are going to show something like: `Connection deactivation failed on None: error=unknown`. I am not sure if it will help or confuse the user but I don't have any other solution on mind.

The problem about the missing openvswitch service is tracked at #837 if it is fixed, this code would not be reached.

The error message could also be improved. I guess it should mention the connection name instead of only the device name to reduce ambiguity.

> The change itself may be fine, but the reasoning for it (in the commit message) is not IMO.
> If `devname` has been queried, it means it was expected to exist, otherwise it would not have been called.
> And the reasoning about OVS service is just masking the real problem (or other future problems).

The ActiveConnection does not guarantee that the device exists. And it also does not seem to make sense for virtual devices that NM tries to create. If the activation fails, it does not make sense that there would be a device IMHO.
 
> We should probably blow up much earlier or not reach this situation at all.

If you still think that ActiveConnection should be changed so that `_nmdev` is always present, please go ahead and close this PR.

> For the OVS issue itself, I think we should just not proceed if it is not running, like the NM ovs plugin.

This is tracked at #837

tyll:
Just verified it myself, with this patch, the hang is not happening anymore.

RamLavi:
tried it on my vm (using `dnf copr enable packit/nmstate-nmstate-852` - big :+1:! )
seems to do the trick. =)

libnmstate/nm/active_connection.py
tyll: > Is this fine? If `_nmdev` is None the logs are going to show something like: `Connection deactivation failed on None: error=unknown`. I am not sure if it will help or confuse the user but I don't have any other solution on mind.

When I wanted to file a bug to track this I noticed there is an easy fix for this: #874

PR #856, opened 2020-02-25
Enable Packit for automatic Copr builds

tyll:
Add configuration for packit-as-a-service to enable automatic Copr builds for pull requests.

We can also use this in the future to simplify updating the Fedora Rawhide package.

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -1,0 +1,15 @@
+---
+specfile_path: nmstate.spec
+upstream_package_name: nmstate
+upstream_project_url: http://nmstate.io
+create_tarball_command: ["python3", "setup.py", "sdist", "--dist-dir", "."]
+current_version_command: ["python3", "setup.py", "--version"]
+actions:
+  post-upstream-clone: "bash -c './packaging/make_spec.sh > nmstate.spec'"
+jobs:
+  - job: copr_build
+    trigger: pull_request
+    metadata:
+      targets:
+        - centos-stream-x86_64
+        - epel-8-x86_64
--- a/packaging/make_srpm.sh
+++ b/packaging/make_srpm.sh
@@ -39,9 +39,10 @@ TAR_FILE="${TMP_DIR}/nmstate-${VERSION}.tar"
 
     ./packaging/make_spec.sh > "${SPEC_FILE}"
     tar --append --file=$TAR_FILE $SPEC_FILE
+    gzip "${TAR_FILE}"
 
     rpmbuild --define "_rpmdir $TMP_DIR/" --define "_srcrpmdir $TMP_DIR/" \
-    -ts $TAR_FILE
+    -ts $TAR_FILE.gz
 ) &> /dev/stderr
 
 SRPM=$(find $TMP_DIR -type f -name \*.src.rpm -exec basename {} \;)
--- a/packaging/nmstate.spec
+++ b/packaging/nmstate.spec
@@ -8,7 +8,7 @@ Release:        @RELEASE@%{?dist}
 Summary:        Declarative network manager API
 License:        LGPLv2+
 URL:            https://github.com/%{srcname}/%{srcname}
-Source0:        https://github.com/%{srcname}/%{srcname}/archive/v%{version}/%{srcname}-%{version}.tar
+Source0:        https://github.com/%{srcname}/%{srcname}/releases/download/v%{version}/%{srcname}-%{version}.tar.gz
 BuildArch:      noarch
 BuildRequires:  python3-devel
 BuildRequires:  python3-setuptools


cathay4t:
Any example how to use this packit? It seems create a copr repo for each PR which I don't see any use of it.

tyll:
> Any example how to use this packit? It seems create a copr repo for each PR which I don't see any use of it.

The Copr repo for each PR makes it easier to test the PR for others to check if their specific bug is fixed. In a next step we can also use it to automate updating the Fedora Rawhide package.

.packit.yml
```diff
@@ -0,0 +1,18 @@
+---
+specfile_path: nmstate.spec
+upstream_package_name: nmstate
+upstream_project_url: http://nmstate.io
+create_tarball_command: ["python3", "setup.py", "sdist", "--dist-dir", "."]
+current_version_command: ["python3", "setup.py", "--version"]
+create_pr: true
```
lachmanfrantisek: This is now the default value and is not needed.

.packit.yml
```diff
@@ -0,0 +1,18 @@
...
+  post-upstream-clone: "bash -c './packaging/make_spec.sh > nmstate.spec'"
+jobs:
+  - job: copr_build
+    trigger: pull_request
+    metadata:
+      targets:
+        - centos-stream-x86_64
+        - epel-8-x86_64
+        - fedora-32-x86_64
+        - fedora-rawhide-x86_64
```
lachmanfrantisek: You can also add the `tests` job (see [the relevant part of the docs](https://packit.dev/testing-farm/) for more info.

- If you only add the job to the config, it will try to install the package from the created copr repo. You can also set up real tests using tmt/fmf.
- This should be enough (the targets will be copied from the build):

```suggestion
        - fedora-rawhide-x86_64
- job: tests
  trigger: pull_request
```

.packit.yml
```diff
@@ -0,0 +1,18 @@
...
+actions:
+  post-upstream-clone: "bash -c './packaging/make_spec.sh > nmstate.spec'"
+jobs:
+  - job: copr_build
+    trigger: pull_request
+    metadata:
+      targets:
+        - centos-stream-x86_64
+        - epel-8-x86_64
+        - fedora-32-x86_64
```
lachmanfrantisek: You can use `fedora-all`, `fedora-devel`, or `fedora-stable` aliases to not need to change anything on each release. ([Updated manually](https://github.com/packit-service/packit/pull/731) on our site for the last time..;)

Aliases will be used with the `x86_64` by default, but you can use something like `fedora-stable-aarch64` as well.

.packit.yml
tyll: To install it on CentOS, it currently needs extra Copr repos for NetworkManager and maybe also openvswitch. Is it possible to specify them?
However, currently the possibility to get Copr repos for pull requests to simplify testing of PRs is the main reason we added Packit, so this is not a a priority.

lachmanfrantisek: > To install it on CentOS, it currently needs extra Copr repos for NetworkManager and maybe also openvswitch. Is it possible to specify them?

- It's currently not possible, but a good suggestion. I'll create an issue for that.
- We can build in the custom copr project, but that's probably not what you want.

lachmanfrantisek: https://github.com/packit-service/packit-service/issues/437

tyll:
@lachmanfrantisek another question:
We have different spec files in Fedora and downstream, because we do GPG verification in the Fedora spec file:
https://src.fedoraproject.org/rpms/nmstate/blob/master/f/nmstate.spec#_45
We cannot do this for downstream CI releases, do you have any recommendations about how to handle this if we also want to enable automatic updates for releases in Fedora Rawhide?

.packit.yml
tyll: We are not targeting Fedora 31 currently, therefore it seems to be the right choice to specify the chroot's explicitly. Actually I guess we do not need Fedora builds currently for the PRs, so I removed them for now.

tyll:
Travis CI succeeded on the optional jobs and RPM building worked on the required job, therefore I will ignore the Travis failure.

lachmanfrantisek:
> We cannot do this for downstream CI releases, do you have any recommendations about how to handle this if we also want to enable automatic updates for releases in Fedora Rawhide?

You can probably download the Fedora version in the `pre-sync` action (https://packit.dev/docs/actions/).

Something like this should work:

```yaml
actions:
  pre-sync:
    - "curl https://src.fedoraproject.org/rpms/nmstate/raw/master/f/nmstate.spec -o nmstate.spec"
```

.packit.yml
thrix: Please see the issue, I believe that could work for this case .... prepare happens before artifact installation ... we now support simple shell steps ....

PR #861, opened 2020-02-26
validator: fix linux bridge validation 

ffmancera:
When creating a linux bridge with an invalid bridge state the cleanup is
not being done.

### DIFF
--- a/libnmstate/validator.py
+++ b/libnmstate/validator.py
@@ -209,9 +209,10 @@ def validate_vxlan(state):
 def validate_bridge(state):
     for iface_state in state.get(schema.Interface.KEY, []):
         if iface_state.get(schema.Interface.TYPE) == LB.TYPE:
-            _assert_vlan_filtering_trunk_tags(
-                iface_state.get(LB.PORT_SUBTREE, [])
+            port_states = iface_state.get(LB.CONFIG_SUBTREE, {}).get(
+                LB.PORT_SUBTREE, []
             )
+            _assert_vlan_filtering_trunk_tags(port_states)
 
 
 def validate_ovs_link_aggregation(state):
--- a/libnmstate/validator.py
+++ b/libnmstate/validator.py
@@ -213,6 +213,15 @@ def validate_bridge(state):
                 LB.PORT_SUBTREE, []
             )
             _assert_vlan_filtering_trunk_tags(port_states)
+            _assert_bridge_port_vlan_not_implemented(port_states)
+
+
+def _assert_bridge_port_vlan_not_implemented(port_states):
+    for port in port_states:
+        if port.get(LB.Port.VLAN_SUBTREE):
+            raise NmstateNotImplementedError(
+                "Linux bridge port vlan filtering is not implemented yet."
+            )
 
 
 def validate_ovs_link_aggregation(state):
--- a/libnmstate/netapplier.py
+++ b/libnmstate/netapplier.py
@@ -66,6 +66,7 @@ def apply(desired_state, verify_change=True, commit=True, rollback_timeout=60):
     validator.validate_dhcp(desired_state)
     validator.validate_dns(desired_state)
     validator.validate_vxlan(desired_state)
+    validator.validate_bridge(desired_state)
 
     checkpoint = _apply_ifaces_state(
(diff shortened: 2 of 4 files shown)


tests/integration/linux_bridge_test.py
```diff
@@ -396,9 +396,9 @@ def test_port_vlan_not_implemented(port0_up):
     }
 
     bridge_state = _create_bridge_subtree_config((port_name,))
-    bridge_state[LinuxBridge.PORT_SUBTREE][0].update(port_vlan_config)
-    with linux_bridge(bridge_name, bridge_state):
```
tyll: It seems to be strange that this is needed. It seems that either `linux_bridge()` should be fixed or there is a bug in Nmstate.

If libnmstate.apply() fails, then Nmstate should revert everything with a rollback. Otherwise `linux_bridge()` should do it.

EdDev:
> When creating a linux bridge with an invalid bridge state the cleanup is
not being done.

What is not cleaning up correctly?
How is the new code fixing it?

tests/integration/linux_bridge_test.py
ffmancera: I am trying to reproduce it out of the test. I don't see anything wrong on linux_bridge or nmstate code..

tyll: What are you seeing inside of tests?

ffmancera:
After some investigation, I think this is a combination of `linux_bridge()` bugged and also an nmstate error. Nmstate should not apply this state. I will investigate it more.

ffmancera:
During the exception handling on tests I have been able to capture the following log:

```
------------------------------------------------------------------------------------------------------------ Captured stderr call ------------------------------------------------------------------------------------------------------------
Exception ignored in: <bound method _NmClient.__del__ of <libnmstate.nm.nmclient._NmClient object at 0x7fad76995b00>>
  File "/usr/lib/python3.6/site-packages/libnmstate/nm/nmclient.py", line 287, in __del__
    "_NmClient.delete_client() should be explicitly called "
libnmstate.error.NmstateInternalError: _NmClient.delete_client() should be explicitly called before python GC clean up the object.
```

ffmancera:
> During the exception handling on tests I have been able to capture the following log:
> 
> ```
> ------------------------------------------------------------------------------------------------------------ Captured stderr call ------------------------------------------------------------------------------------------------------------
> Exception ignored in: <bound method _NmClient.__del__ of <libnmstate.nm.nmclient._NmClient object at 0x7fad76995b00>>
> Traceback (most recent call last):
>   File "/usr/lib/python3.6/site-packages/libnmstate/nm/nmclient.py", line 287, in __del__
>     "_NmClient.delete_client() should be explicitly called "
> libnmstate.error.NmstateInternalError: _NmClient.delete_client() should be explicitly called before python GC clean up the object.
> ```

This is causing the warning because the state is being applied. This is probably happening because we are raising this inside the mainloop which is not good. I will propose a simple patch for this :)

libnmstate/validator.py
```diff
@@ -212,6 +212,16 @@ def validate_bridge(state):
             _assert_vlan_filtering_trunk_tags(
                 iface_state.get(LB.CONFIG_SUBTREE, {}).get(LB.PORT_SUBTREE, [])
             )
+            _assert_bridge_port_vlan_not_implemented(iface_state)
+
+
+def _assert_bridge_port_vlan_not_implemented(ifstate):
+    port_states = ifstate.get(LB.CONFIG_SUBTREE, {}).get(LB.PORT_SUBTREE, [])
```
EdDev: This duplicates line 213.

ffmancera: Yes, you are right. Thanks!

ffmancera:
> There is an empty commit here (`tests, integ: fix clean up on linux bridge test `).
> Only one small comment inline.

Ops, please ignore it. I am going to remove it.

libnmstate/validator.py
ffmancera: Done!

ffmancera:
Please, approve #909. I am going to close this ticket as the CI is bugged.

PR #862, opened 2020-02-26
validator: Drop unnused parameter at validate_interfaces_state()

elvgarrui:
The function validate_interfaces_state does not use the parameter
desired_state, therefore this patch removes it.


cathay4t:
CI failed due to a knonw issue of NetworkManager.

Issue #865, opened 2020-02-27
Retry when verification fails


cathay4t:
The `reapply()` of NetworkManager does not guarantee the state been active.
Currently we wait 0.5 seconds before quitting the mainloop which is not a good practice and 
already approved to problematic at issue #736 .

To fix this problem, need to retry on verification fails with main context iterating.

PR #870, opened 2020-02-27
netapplier: retry verification after failing

ffmancera:
During the verification sometimes ethtool is reporting 'unknown' in
duplex state. This is due a lack of time for some NICs to update the
information. We have noticed it only with Broadcom BCM57840 NetXtreme II
NIC but it could affect to other NICs.

In order to fix this, we are retrying the operation up to 5 times
sleeping 2 seconds for each retry.

Ref: https://bugzilla.redhat.com/1807034

### DIFF
--- a/libnmstate/netapplier.py
+++ b/libnmstate/netapplier.py
@@ -36,9 +36,12 @@ from libnmstate.error import NmstateError
 from libnmstate.error import NmstateLibnmError
 from libnmstate.error import NmstatePermissionError
 from libnmstate.error import NmstateValueError
+from libnmstate.error import NmstateVerificationError
 from libnmstate.nm import nmclient
 
 MAINLOOP_TIMEOUT = 35
+VERIFY_RETRY_INTERNAL = 1
+VERIFY_RETRY_TIMEOUT = 5
 
 
 @_warn_keyword_as_positional
@@ -155,8 +158,18 @@ def _apply_ifaces_state(
                     ifaces2add + ifaces2edit,
                     con_profiles=ifaces_add_configs + ifaces_edit_configs,
                 )
+            verified = False
             if verify_change:
-                _verify_change(desired_state)
+                for _ in range(VERIFY_RETRY_TIMEOUT):
+                    try:
+                        _verify_change(desired_state)
+                        verified = True
+                        break
+                    except NmstateVerificationError:
+                        time.sleep(VERIFY_RETRY_INTERNAL)
+                if not verified:
+                    _verify_change(desired_state)
+
         if not commit:
             return checkpoint
     except nm.checkpoint.NMCheckPointPermissionError:
--- a/libnmstate/netapplier.py
+++ b/libnmstate/netapplier.py
@@ -200,10 +200,11 @@ def _list_new_interfaces(desired_state, current_state):
 
 def _verify_change(desired_state):
     current_state = state.State(netinfo.show())
-    desired_state.verify_interfaces(current_state)
-    desired_state.verify_routes(current_state)
-    desired_state.verify_dns(current_state)
-    desired_state.verify_route_rule(current_state)
+    verifiable_desired_state = copy.deepcopy(desired_state)
+    verifiable_desired_state.verify_interfaces(current_state)
+    verifiable_desired_state.verify_routes(current_state)
+    verifiable_desired_state.verify_dns(current_state)
+    verifiable_desired_state.verify_route_rule(current_state)
 
 
 @contextmanager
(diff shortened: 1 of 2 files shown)


ffmancera:
It seems this should be fixed by #865. @cathay4t is working on that so let's wait until his patch and let's see if it fixes this issue.

libnmstate/netapplier.py
```diff
@@ -156,7 +157,18 @@ def _apply_ifaces_state(
                     con_profiles=ifaces_add_configs + ifaces_edit_configs,
                 )
             if verify_change:
-                _verify_change(desired_state)
+                attempts_left = 5
+                retry = True
+                while(retry):
+                    try:
+                        _verify_change(desired_state)
+                        retry = False
```
tyll: you can use break here and do `while True`.

libnmstate/netapplier.py
```diff
@@ -156,7 +157,18 @@ def _apply_ifaces_state(
...
+                    except NmstateVerificationError:
+                        time.sleep(1)
```
tyll: The sleep should happen after the potential raise or before the verify change. Here it just delays the execution while nothing will change.

libnmstate/netapplier.py
```diff
@@ -156,7 +157,18 @@ def _apply_ifaces_state(
                     con_profiles=ifaces_add_configs + ifaces_edit_configs,
                 )
             if verify_change:
-                _verify_change(desired_state)
+                attempts_left = 5
+                retry = True
+                while(retry):
```
tyll: You might also be able to write it like to avoid a possible endless loop:

```python
for _ in range(5):
    try:
        _verify_change(desired_state)
        break
    except NmstateVerificationError:
        time.sleep(1)
else:
    _verify_change(desired_state)
```

ffmancera: This is better. Thanks!

libnmstate/netapplier.py
```diff
@@ -36,6 +36,7 @@
 from libnmstate.error import NmstateLibnmError
```
tyll: The times in the commit message do not match the time from the code. Maybe the code should be adjusted to sleep only 0.2 seconds.

ffmancera: Oh sorry, I m going to fix that. The time in the code is the needed one. 0.2 seconds is not enough time.

ffmancera:
> Please remove the xfail flag of `test_switch_from_dynamic_ip_without_dhcp_srv_to_static_ipv6` to show that this PR actually fixing the problem.

Done, thank you!

libnmstate/netapplier.py
```diff
@@ -200,10 +209,11 @@ def _list_new_interfaces(desired_state, current_state):
 
 def _verify_change(desired_state):
     current_state = state.State(netinfo.show())
-    desired_state.verify_interfaces(current_state)
-    desired_state.verify_routes(current_state)
-    desired_state.verify_dns(current_state)
-    desired_state.verify_route_rule(current_state)
+    verifiable_desired_state = copy.deepcopy(desired_state)
```
cathay4t: Is there a problem if when not using `copy.deepcopy`?

If no, keep it mutable is OK for me.

libnmstate/netapplier.py
```diff
@@ -156,7 +157,15 @@ def _apply_ifaces_state(
...
                 )
             if verify_change:
-                _verify_change(desired_state)
+                for _ in range(5):
+                    try:
+                        _verify_change(desired_state)
+                        break
+                    except NmstateVerificationError:
+                        time.sleep(1)
+                else:
```
cathay4t: I dislike this complex `for else` syntax, especially when ex-BDFL of python dislike it:

https://mail.python.org/pipermail/python-ideas/2009-October/006157.html

libnmstate/netapplier.py
```diff
@@ -156,7 +157,15 @@ def _apply_ifaces_state(
                     con_profiles=ifaces_add_configs + ifaces_edit_configs,
                 )
             if verify_change:
-                _verify_change(desired_state)
+                for _ in range(5):
+                    try:
+                        _verify_change(desired_state)
+                        break
+                    except NmstateVerificationError:
+                        time.sleep(1)
```
cathay4t: Use constant like `VERIFY_RETRY_INTERNAL` instead of magic number.

libnmstate/netapplier.py
ffmancera: Yes, if we do not use `copy.deepcopy` we could find some issues when retrying the verification.

libnmstate/netapplier.py
ffmancera: Fixed. Thanks!

libnmstate/netapplier.py
ffmancera: Added. Thanks!

cathay4t: `VERIFY_RETRY_INTERNAL` is 1 based on your code. and `VERIFY_RETRY_TIMEOUT` is 5.

ffmancera: Ops. Done, thank you!

PR #878, opened 2020-03-02
validator: Check IP config in bond slaves.

### LINKED ISSUE #774: Need better error message when bond slave has IP settings.
https://bugzilla.redhat.com/show_bug.cgi?id=1782674

elvgarrui:
When a bond interface is created, the slaves cannot have their IP
configuration enabled. In order to fix this problem, this patch
introduces a validation check to see whether the slave interfaces had
their IP enabled and if so, it raises an NmstateValueError.

Ref: https://bugzilla.redhat.com/1782674

This fixes #774 

### DIFF
--- a/libnmstate/validator.py
+++ b/libnmstate/validator.py
@@ -26,6 +26,11 @@ from . import nm
 from . import schema
 from libnmstate.appliers.bond import is_in_mac_restricted_mode
 from libnmstate.appliers.ovs_bridge import is_ovs_running
+from libnmstate.appliers.bond import get_bond_slaves_from_state
+from libnmstate.appliers.team import get_slaves_from_state as get_team_slaves
+from libnmstate.appliers.linux_bridge import get_slaves_from_state
+from libnmstate.appliers.ovs_bridge import get_ovs_slaves_from_state
+from libnmstate.schema import Bond
 from libnmstate.schema import DNS
 from libnmstate.schema import Interface
 from libnmstate.schema import InterfaceIP
@@ -35,11 +40,17 @@ from libnmstate.schema import InterfaceState
 from libnmstate.schema import LinuxBridge as LB
 from libnmstate.schema import OVSBridge
 from libnmstate.schema import VXLAN
-from libnmstate.schema import Bond
 from libnmstate.error import NmstateDependencyError
 from libnmstate.error import NmstateNotImplementedError
 from libnmstate.error import NmstateValueError
 from libnmstate.iplib import is_ipv6_address
+
+IP_DISABLED_SLAVE_GETTERS = {
+    InterfaceType.BOND: get_bond_slaves_from_state,
+    InterfaceType.OVS_BRIDGE: get_ovs_slaves_from_state,
+    InterfaceType.LINUX_BRIDGE: get_slaves_from_state,
+    InterfaceType.TEAM: get_team_slaves,
+}
 
 
 class NmstateRouteWithNoInterfaceError(NmstateValueError):
@@ -103,6 +114,7 @@ def validate_interface_capabilities(ifaces_state, capabilities):
 def validate_interfaces_state(original_desired_state, current_state):
     validate_link_aggregation_state(original_desired_state, current_state)
     _validate_linux_bond(original_desired_state, current_state)
+    _validate_slave_ip_config(original_desired_state)
 
 
 def validate_link_aggregation_state(desired_state, current_state):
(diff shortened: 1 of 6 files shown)


libnmstate/validator.py
```diff
@@ -33,14 +34,34 @@
...
+from libnmstate.schema import Team
 from libnmstate.schema import VXLAN
-from libnmstate.schema import Bond
 from libnmstate.error import NmstateDependencyError
 from libnmstate.error import NmstateNotImplementedError
 from libnmstate.error import NmstateValueError
 from libnmstate.iplib import is_ipv6_address
 
 
+INTERFACE_SLAVE_TREES = {
```
cathay4t: In the `libnmstate/appliers/` foler, there are `get_slaves_from_state()` `get_bond_slaves_from_state()` and etc.

Please use these functions instead.

elvgarrui: I didn't know those existed. Thanks!
Now it is implemented with that. The IPV6 tests were also added.

libnmstate/validator.py
```diff
@@ -24,7 +24,14 @@
 
 from . import nm
 from . import schema
-from libnmstate.appliers.bond import is_in_mac_restricted_mode
+from libnmstate.appliers.bond import (
+    is_in_mac_restricted_mode,
+    get_bond_slaves_from_state,
+)
```
cathay4t: Current code style is not using combined import. Not a must.

libnmstate/validator.py
```diff
@@ -34,12 +41,18 @@
 from libnmstate.schema import LinuxBridge as LB
 from libnmstate.schema import OVSBridge
 from libnmstate.schema import VXLAN
-from libnmstate.schema import Bond
 from libnmstate.error import NmstateDependencyError
 from libnmstate.error import NmstateNotImplementedError
 from libnmstate.error import NmstateValueError
 from libnmstate.iplib import is_ipv6_address
 
+SLAVE_GETTERS = {
```
cathay4t: Maybe `IP_DISABLED_SLAVE_GETTERS`?

For VLAN and OVS Interface, these slaves still can have IP enabled.

libnmstate/validator.py
```diff
@@ -401,3 +415,36 @@ def _assert_bond_config_is_legal(iface_state, original_iface_state):
...
+        if iface_type in SLAVE_GETTERS.keys():
+            slaves = SLAVE_GETTERS[iface_type](iface_state)
+        if slaves:
+            for slave_iface_name in slaves:
+                slave_state = original_desired_state.interfaces.get(
+                    slave_iface_name, {}
+                )
+                if (
+                    slave_state.get(Interface.TYPE)
+                    != InterfaceType.OVS_INTERFACE
```
cathay4t: Actaully, OVS interface has two types: `system` and `internal`. The internal interface can have IP enabled while `system` does not.

dedicate tests for these two OVS interface type are also recommended.

tests/integration/bond_test.py
```diff
@@ -777,3 +778,15 @@ def test_new_bond_uses_mac_of_first_slave_by_name(eth1_eth2_with_no_profile):
         assert get_mac_address(BOND99) == eth1_mac
         _nmcli_simulate_boot(BOND99)
         assert get_mac_address(BOND99) == eth1_mac
+
+
+def test_add_invalid_slave_ip_config(eth1_up):
+    desired_state = eth1_up
+    desired_state[Interface.KEY][0][Interface.IPV4][InterfaceIP.ENABLED] = True
```
cathay4t: Perfer `InterfaceIPv4.ENABLED` as we already know its IPv4. Not a must.

libnmstate/validator.py
```diff
@@ -401,3 +415,36 @@ def _assert_bond_config_is_legal(iface_state, original_iface_state):
...
+                ):
+                    ipv4_state = slave_state.get(Interface.IPV4, {}).get(
+                        InterfaceIP.ENABLED
+                    )
```
cathay4t: Please change it to `InterfaceIP.DISABLED`. We don't want user got `NmstateValueError` when they never defined IP stack on slaves.

libnmstate/validator.py
cathay4t: Also need a test in VLAN showing this is not impacted by new validator.

tests/lib/validator_test.py
```diff
@@ -652,13 +653,100 @@ def test_mac_restriction_in_current_mac_in_desire(self):
...
             )
 
-    def _gen_iface_state(self, bond_name):
-        return {
-            Interface.NAME: bond_name,
-            Interface.TYPE: InterfaceType.BOND,
+
+class TestPortsValidator:
+
+    ipv_up = pytest.mark.parametrize("ipv_up", [(True, False), (False, True)])
```
cathay4t:
```python
parametrize_ip_ver = pytest.mark.parametrize(
    "ip_state",
    [(True, False), (False, True)],
    ids=["ipv4 on, ipv6 off", "ipv4 off, ipv6 on"],
)
```

libnmstate/validator.py
elvgarrui: Fixed!

libnmstate/validator.py
elvgarrui: I've changed the name and made the test

libnmstate/validator.py
elvgarrui: I havent seen any DISABLED property in `libnmstate/schema.py`. InterfaceIP has the ENABLED property, which can be true or false.

tests/lib/validator_test.py
elvgarrui: Fixed this!! Thanks a lot for all your reviews!

libnmstate/validator.py
ffmancera: Maybe I am missing something, but OVS Interface has no type. There is only two kind of OVS Ports "system" and "internal", a port is internal only when it is `ovs-interface` type. So IMO, this check is fine, as it is checking that the port is `OVS_INTERFACE` which means it is "internal" type. Thanks!

tests/integration/bond_test.py
elvgarrui: Fixed this, thanks!

libnmstate/validator.py
ffmancera: Please, see [this commit](https://github.com/nmstate/nmstate/pull/525/commits/dcfb27655871972b93c9559a0a3ace410e44d06b). As you can see, the only one condition to be an internal port is to be `ovs-interface` type. If not, it is a system port. Thanks!

libnmstate/validator.py
cathay4t: You are correct. The code looks good for this part.

libnmstate/validator.py
cathay4t: @elvgarrui @ffmancera The system OVS interface does not listed as `InterfaceType.OVS_INTERFACE`. So this looks good.

cathay4t:
Merged. Thanks.

PR #882, opened 2020-03-03
linux bridge, vlan filtering: add integ tests

maiqueb:
Integration tests for the port vlan functionality.

### DIFF
--- a/tests/integration/linux_bridge_test.py
+++ b/tests/integration/linux_bridge_test.py
@@ -35,6 +35,9 @@ from libnmstate.schema import LinuxBridge
 from .testlib import assertlib
 from .testlib.bondlib import bond_interface
 from .testlib.bridgelib import add_port_to_bridge
+from .testlib.bridgelib import generate_vlan_filtering_config
+from .testlib.bridgelib import generate_vlan_id_config
+from .testlib.bridgelib import generate_vlan_id_range_config
 from .testlib.bridgelib import linux_bridge
 from .testlib.ifacelib import get_mac_address
 from .testlib.iproutelib import ip_monitor_assert_stable_link_up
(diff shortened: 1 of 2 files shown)


tests/integration/linux_bridge_test.py
```diff
@@ -410,6 +425,120 @@ def test_port_vlan_not_implemented(port0_up):
...
+    @staticmethod
+    def _generate_vlan_id_range_config(min_vlan_id, max_vlan_id):
+        return {
+            LinuxBridge.Port.Vlan.TrunkTags.ID_RANGE: {
+                LinuxBridge.Port.Vlan.TrunkTags.MIN_RANGE: min_vlan_id,
+                LinuxBridge.Port.Vlan.TrunkTags.MAX_RANGE: max_vlan_id,
+            }
+        }
+
+
```
ffmancera: IMHO, this should be move to `tests/integration/testlib/bridgelib.py` as you did in unit tests.

tests/integration/linux_bridge_test.py
```diff
@@ -120,6 +120,21 @@ def bond0(port0_up):
...
+        tag=TestVlanFiltering.access_tag,
+        trunk_tags=TestVlanFiltering._generate_vlan_id_config(100, 101, 102),
+    )
+    port_name = port0_up[Interface.KEY][0][Interface.NAME]
+    bridge_state = _create_bridge_subtree_config((port_name,))
+    bridge_state[LinuxBridge.PORT_SUBTREE][0].update(access_port_state)
+    with linux_bridge(TEST_BRIDGE0, bridge_state):
+        yield
+
+
```
ffmancera: I think we are also following "caller before the callee" in integration tests. So please, place this at the buttom. @EdDev What do you think? Does it apply also to fixtures?

maiqueb: AFAIU, fixtures go at top of the tests. Let's way for @EdDev 's say on this.

tests/integration/linux_bridge_test.py
maiqueb: I don't think these functions will be re-used in other integ tests.

If I'm wrong, they can be moved when the need arises. Are you OK with that ?

EdDev: I think it needs to be well integrated into the existing helpers (e.g. `add_port_to_bridge`).
But we can do this later on, it does not worth blocking the PR for it.

tests/integration/linux_bridge_test.py
ffmancera: Looks good, thanks!

tests/integration/linux_bridge_test.py
```diff
@@ -406,6 +409,159 @@ def test_port_vlan_not_implemented(port0_up):
         pass
 
 
+@pytest.mark.xfail(
+    reason="https://nmstate.atlassian.net/browse/NMSTATE-230", strict=True
+)
+class TestVlanFiltering:
+    Access_tag = 300
```
EdDev: nit: CamelCase are reserved for class names.

maiqueb:
Can't this be merged ? It's on the tree of https://github.com/nmstate/nmstate/pull/893 , which removes the XFAILs.

tests/integration/linux_bridge_test.py

PR #888, opened 2020-03-04
Linux bridge vlans merge bridge state

maiqueb:
Enable / disable vlan filtering on linux bridge interfaces via interface state merging.

This allows for a user to partially update the port configuration.

Considering the current state:
```yaml
---
interfaces:
  - name: eth1
    type: ethernet
    state: up
  - name: linux-br0
    type: linux-bridge
    state: up
    bridge:
      port:
        - name: eth1
          stp-hairpin-mode: false
          stp-path-cost: 100
          stp-priority: 32
          vlan:
            mode: trunk
            trunk-tags:
            - id: 101
            - id-range:
                min: 500
                max: 599
            tag: 100
            enable-native: true
```

If the user wants to remove the port vlan configuration on port
eth1 - and as a result disable port vlan filtering on the whose
bridge - the following yaml should be used:
```yaml
---
interfaces:
  - name: linux-br0
    type: linux-bridge
    state: up
    bridge:
      port:
        - name: eth1
          vlan: {}
```

Depends on: #913

### DIFF
--- a/libnmstate/state.py
+++ b/libnmstate/state.py
@@ -379,10 +379,16 @@ class State:
         entries that appear only on one state are ignored.
         This is a reverse recursive update operation.
         """
+        origin_other_state = other_state
         other_state = State(other_state.state)
         for name in self.interfaces.keys() & other_state.interfaces.keys():
             dict_update(other_state.interfaces[name], self.interfaces[name])
             self._ifaces_state[name] = other_state.interfaces[name]
+
+            iface_state = self.interfaces[name]
+            other_iface_state = origin_other_state.interfaces[name]
+            if iface_state.get(Interface.TYPE) == LinuxBridge.TYPE:
+                merge_linux_bridge_ports(iface_state, other_iface_state)
 
     def complement_master_interfaces_removal(self, other_state):
         """
@@ -939,3 +945,39 @@ def _remove_route_rule_default_values(rule):
     if rule.get(RouteRule.ROUTE_TABLE) == RouteRule.USE_DEFAULT_ROUTE_TABLE:
         del rule[RouteRule.ROUTE_TABLE]
     return rule
+
+
+def merge_linux_bridge_ports(desired_iface_state, current_iface_state):
+    """
+    Given a linux bridge desired state, and it's current state, merges
+    those together.
+
+    This extension of the interface merging mechanism simplifies the user's
+    life in scenarios where the user wants to partially update the bridge's
+    configuration - e.g. update only the bridge's port STP configuration -
+    since it enables the user to simply specify the updated values rather
+    than the full current state + the updated value.
+    """
+    desired_bridge_state = desired_iface_state.get(
+        LinuxBridge.CONFIG_SUBTREE, {}
+    )
+    desired_ports = desired_bridge_state.get(LinuxBridge.PORT_SUBTREE, [])
+    current_ports = current_iface_state.get(
+        LinuxBridge.CONFIG_SUBTREE, {}
+    ).get(LinuxBridge.PORT_SUBTREE, [])
+
+    current_indexed_ports = _index_ports(current_ports)
+    desired_indexed_ports = _index_ports(desired_ports)
+
+    merged_ports = desired_indexed_ports
+    for name in desired_indexed_ports.keys() & current_indexed_ports.keys():
+        dict_update(current_indexed_ports[name], desired_indexed_ports[name])
+        merged_ports[name] = current_indexed_ports[name]
+
+    desired_bridge_state[LinuxBridge.PORT_SUBTREE] = list(
+        merged_ports.values()
+    )
+
+
+def _index_ports(ports):
+    return {port[LinuxBridge.Port.NAME]: port for port in ports}
--- a/libnmstate/state.py
+++ b/libnmstate/state.py
@@ -370,6 +370,7 @@ class State:
         self._capitalize_mac()
         self._remove_empty_description()
         self._remove_ip_stack_if_disabled()
+        self._normalize_linux_bridge_port_vlan()
 
     def merge_interfaces(self, other_state):
         """
@@ -770,6 +771,20 @@ class State:
                     other_state._config_route_rules
                 )
             }
+
+    def _normalize_linux_bridge_port_vlan(self):
+        linux_bridges = (
+            iface
+            for iface in self.interfaces.values()
+            if iface[Interface.TYPE] == InterfaceType.LINUX_BRIDGE
+        )
+        for lb in linux_bridges:
+            ports = lb.get(LinuxBridge.CONFIG_SUBTREE, {}).get(
+                LinuxBridge.PORT_SUBTREE, []
+            )
+            for port in ports:
+                if not port.get(LinuxBridge.Port.VLAN_SUBTREE):
+                    port[LinuxBridge.Port.VLAN_SUBTREE] = {}
 
 
 def dict_update(origin_data, to_merge_data):
--- a/tests/integration/testlib/statelib.py
+++ b/tests/integration/testlib/statelib.py
@@ -118,6 +118,7 @@ class State:
         self._sort_ip_addresses()
         self._sort_interfaces_by_name()
         self._canonicalize_iface_ipv6_addresses()
+        self._normalize_linux_bridge_port_vlan()
 
     def match(self, other):
         return _state_match(self.state, other.state)
@@ -228,6 +229,20 @@ class State:
                         with contextlib.suppress(ValueError):
                             option_value = int(option_value)
                         options[option_name] = option_value
+
+    def _normalize_linux_bridge_port_vlan(self):
+        linux_bridges = (
+            iface
+            for iface in self._state.get(Interface.KEY, [])
+            if iface[Interface.TYPE] == LinuxBridge.TYPE
+        )
+        for lb in linux_bridges:
+            ports = lb.get(LinuxBridge.CONFIG_SUBTREE, {}).get(
+                LinuxBridge.PORT_SUBTREE, []
+            )
+            for port in ports:
+                if not port.get(LinuxBridge.Port.VLAN_SUBTREE):
+                    port[LinuxBridge.Port.VLAN_SUBTREE] = {}
 
 
 def _lookup_iface_state_by_name(interfaces_state, ifname):
--- a/tests/lib/state_test.py
+++ b/tests/lib/state_test.py
@@ -36,6 +36,8 @@ from libnmstate.schema import Route
 from libnmstate.schema import RouteRule
 from libnmstate.schema import Team
 
+from .testlib import bridgelib
+
 
 IFACE0 = "foo"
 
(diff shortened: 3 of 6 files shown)


libnmstate/state.py
```diff
@@ -381,9 +381,21 @@ def merge_interfaces(self, other_state):
         This is a reverse recursive update operation.
         """
         other_state = State(other_state.state)
+        origin_other_state = other_state
+        other_state = State(origin_other_state.state)
```
ffmancera: I don't understand why is this needed. You are creating a copy of `other_state` twice.

libnmstate/state.py
```diff
@@ -381,9 +381,21 @@ def merge_interfaces(self, other_state):
...
         for name in self.interfaces.keys() & other_state.interfaces.keys():
             dict_update(other_state.interfaces[name], self.interfaces[name])
             self._ifaces_state[name] = other_state.interfaces[name]
+        # Merge the port list of linux bridge
+        other_state = State(origin_other_state.state)
```
ffmancera: Why is this needed? You can use `origin_other_state` below.

libnmstate/state.py
```diff
@@ -381,9 +381,21 @@ def merge_interfaces(self, other_state):
...
+        for iface_name, iface_state in self.interfaces.items():
+            if (
+                iface_state[Interface.STATE] == InterfaceState.UP
+                and iface_state[Interface.TYPE] == LinuxBridge.TYPE
+            ):
+                other_iface_state = other_state.interfaces.get(iface_name)
+                if other_iface_state:
+                    merge_linux_bridge_ports(iface_state, other_iface_state)
```
ffmancera: As you are not modifying `other_state` here, you can use `origin_other_state` instead of  `other_state`.

cathay4t:
And meanwhile, we need more merge tests:

 * Disable VLAN filtering of certain port.
 * Remove port from bridge.
 * Add port to bridge.

tests/lib/state_test.py
```diff
@@ -1289,3 +1291,235 @@ def _create_desired_state():
...
+            TestPortVlanLinuxBridge.Bridge_name
+        ]
+        desired_bridge_ports = desired_bridge_state.get(
+            LinuxBridge.CONFIG_SUBTREE, {}
+        ).get(LinuxBridge.PORT_SUBTREE, [])
+
+        # this would assure that none of the ports have a port vlan subtree
+        assert not TestPortVlanLinuxBridge._is_vlan_filtering_enabled(
+            desired_bridge_ports
+        )
```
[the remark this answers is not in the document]
maiqueb: I don't understand what you're looking for here.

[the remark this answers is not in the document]
cathay4t: I mean:
```python
        assert desired_bridge_ports[0] == {
            Interface.NAME: TestPortVlanLinuxBridge.Bridge_name,
            Interface.TYPE: InterfaceType.LINUX_BRIDGE,
            LinuxBridge.CONFIG_SUBTREE: {
                LinuxBridge.PORT_SUBTREE: [
                    {LinuxBridge.Port.NAME: TestPortVlanLinuxBridge.Port1},
                    {LinuxBridge.Port.NAME: TestPortVlanLinuxBridge.Port2},
                ]
            },
        }
```

tests/lib/state_test.py
```diff
@@ -1289,3 +1291,235 @@ def _create_desired_state():
...
+            LinuxBridge.Port.NAME: TestPortVlanLinuxBridge.Port1,
+            LinuxBridge.Port.VLAN_SUBTREE: {
+                LinuxBridge.Port.Vlan.MODE: LinuxBridge.Port.Vlan.Mode.TRUNK,
+                LinuxBridge.Port.Vlan.TRUNK_TAGS: [
+                    {LinuxBridge.Port.Vlan.TrunkTags.ID: 100},
+                    {LinuxBridge.Port.Vlan.TrunkTags.ID: 200},
+                    {LinuxBridge.Port.Vlan.TrunkTags.ID: 300},
+                ],
+            },
+        } in desired_bridge_ports
```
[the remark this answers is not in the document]
cathay4t: My bad. I read the code in the wrong way.

tests/lib/testlib/portvlanlib.py
```diff
@@ -0,0 +1,65 @@
...
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+
+
+from libnmstate.schema import LinuxBridge
+
+
+def update_bridge_port_vlan_config(bridge, port_name, vlan_config):
```
EdDev: Currently this is not used, please either use it now or introduce it with the usage.

BTW, I prefer you make it immutable and return a new bridge state (do not bother if this is meant to just extract current code to a function).

libnmstate/state.py
```diff
@@ -935,3 +947,52 @@ def _remove_route_rule_default_values(rule):
...
+    Given a linux bridge desired state, and it's current state, merges
+    those together.
+
+    This extension of the interface merging mechanism simplifies the user's
+    life in scenarios where the user wants to partially update the bridge's
+    configuration - e.g. update only the bridge's port STP configuration -
+    since it enables the user to simply specify the updated values rather
+    than the full current state + the updated value.
+    """
+    ports = iface_state.get(LinuxBridge.CONFIG_SUBTREE, {}).get(
```
EdDev: I'm opening a separated thread to discuss this merge algorithm (instead of continuing this in the  thread on line 998).

I will ask it, because I may be missing here something basic:
Why can't we follow the original interface merging pattern? What is different here?
I mean, the original interface merging has identify the same interfaces based on the names, and for those a recursive merge has been applied. Sets or lists got overwritten but empty dicts have not overwritten.
Seems to me, like the only fix needed is in the `dict_update` implementation, such that a an empty dict (val == {}) will result in overwriting the origin value. Then use it here instead of repeating a similar algorithm.

I have not tried this out yet, will try to do so and update here.

libnmstate/state.py
```diff
@@ -935,3 +947,52 @@ def _remove_route_rule_default_values(rule):
...
+            other_port = other_index_ports.get(port_name)
+
+            if other_port:
+                # when the desired_state features an explicit empty value,
+                # the user is indicating that value is to be deleted. No merge
+                # is to be applied for those attributes.
+                attrs_to_delete = {k for k, v in port.items() if not v}
+                for attr in attrs_to_delete:
+                    other_port.pop(attr)
+                dict_update(port, other_port)
```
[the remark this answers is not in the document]
cathay4t: By using the `dict_update(port, other_port)`, you are using properties
from `other_port`( aka, current state) to override desire state.

debug:
```python
>>> from libnmstate.state import dict_update
>>>
>>> port = {'a': 1}
>>> other_port = {'a': 2, 'b': 3}
>>> dict_update(port, other_port)
{'a': 2, 'b': 3}
>>> port
{'a': 2, 'b': 3}
>>> other_port
{'a': 2, 'b': 3}
```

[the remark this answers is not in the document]
cathay4t: Sorry. I missed the line you do a `dict_update` before this.

libnmstate/state.py
ffmancera: Thank you!

libnmstate/state.py
ffmancera: Thanks!

libnmstate/state.py
ffmancera: Thanks!

libnmstate/state.py
```diff
@@ -935,3 +953,52 @@ def _remove_route_rule_default_values(rule):
...
+    than the full current state + the updated value.
+    """
+    desired_ports = desired_iface_state.get(
+        LinuxBridge.CONFIG_SUBTREE, {}
+    ).get(LinuxBridge.PORT_SUBTREE)
+    current_ports = current_iface_state.get(
+        LinuxBridge.CONFIG_SUBTREE, {}
+    ).get(LinuxBridge.PORT_SUBTREE, [])
+
+    if desired_ports is None and current_ports:
```
EdDev: As I commented before, I do not really like this complexity from this point down.
I want to check if we can simplify it.

- If the port subtree was not mentioned in the desired state, I would expect it to be merged from the current state by the original `merge_interfaces` processing. How come is it not?
- If the port subtree is mentioned in the desired state as an explicit empty list, I would expect it to overwrite the current state as part of the original merge as well.
- We are left with merging individual dicts, for the ones that appear in both the desired and the current. This is exactly what the `merge_interfaces` do at the interface list level, therefore doing the same here on the port list level could work as well. Please help me understand if it cannot.

I have also posted #913 which suggests to treat an explicit empty dict as an explicit request from the user/client to initialize/cleanup a subtree in the state. This fits the idea discussed here to remove the vlan subtree.

libnmstate/state.py
```diff
@@ -381,9 +381,21 @@ def merge_interfaces(self, other_state):
         This is a reverse recursive update operation.
         """
         other_state = State(other_state.state)
+        origin_other_state = State(other_state.state)
```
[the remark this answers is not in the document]
EdDev: I will try to present it fully by code then:
```
def merge_interfaces(self, other):
    origin_other_state = other
    other_state = State(other.state)
    ...
    for ...
        iface_state = self.interfaces[name]
        other_iface_state = origin_other_state.interfaces[name]
        ...
```

EdDev:
> > Review `tests, vlan filtering: add a port vlan manipulation library` commit:
> > > `portvlanlib`
> > 
> > 
> > This does not hint that this is Linux Bridge specific. Port vlan fits OVS ports as well after all.
> > How about `bridgelib`? There you can have a `PortVlan` class as a namespace (with static methods).
> 
> I'm confused. Why wouldn't the ovs bridge use the same subtree ? This library does nothing more than generate the port vlan subtree, that unless I'm missing something can / should be re-used in ovs bridges.

I am good with your intentions, but in practice you used `LinuxBridge` constants in that module.
If the OVS port can share code with the Bridge, it requires proof. At the moment, it does not seem to happen.
I do hope we will manage to be there.

libnmstate/state.py
maiqueb: I've rebased the tree with your PR - removing the interface merging commit - and it does not work.

It fails to update the desired state with the current state on the following tests:
  - https://github.com/nmstate/nmstate/pull/888/files#diff-d74d6ba3d078c4946b2ef63a5e16130fR1409
    - failed to merge vlan configuration from port1 into the desired state - check [0]
  - https://github.com/nmstate/nmstate/pull/888/files#diff-d74d6ba3d078c4946b2ef63a5e16130fR1530
    - failed to merge the vlan configuration from the port - check [1]

Seems the state is not being correctly passed from the current into the desired state.

[0] - https://paste.centos.org/view/3759b702
[1] - https://paste.centos.org/view/38ff380b

These pastebins will expire after 1 day.

libnmstate/state.py
```diff
@@ -939,3 +945,39 @@ def _remove_route_rule_default_values(rule):
...
+    This extension of the interface merging mechanism simplifies the user's
+    life in scenarios where the user wants to partially update the bridge's
+    configuration - e.g. update only the bridge's port STP configuration -
+    since it enables the user to simply specify the updated values rather
+    than the full current state + the updated value.
+    """
+    desired_bridge_state = desired_iface_state.get(
+        LinuxBridge.CONFIG_SUBTREE, {}
+    )
+    desired_ports = desired_bridge_state.get(LinuxBridge.PORT_SUBTREE)
```
EdDev: Isn't it suppose to default to and empty tuple/list?

libnmstate/state.py
```diff
@@ -771,6 +772,20 @@ def merge_route_rules(self, other_state):
                 )
             }
 
+    def _normalize_linux_bridge_port_vlan(self):
+        linux_bridges = [
+            iface
+            for iface in self.interfaces.values()
+            if iface[Interface.TYPE] == InterfaceType.LINUX_BRIDGE
+        ]
```
EdDev: This fits a generator. It can be converted to one by just replacing the square brackets with regular ones (`()`).

tests/integration/testlib/statelib.py
```diff
@@ -229,6 +230,20 @@ def _convert_lag_numeric_value_options_to_integer(self):
                             option_value = int(option_value)
                         options[option_name] = option_value
 
+    def _normalize_linux_bridge_port_vlan(self):
+        linux_bridges = [
```
EdDev: Same here, you can use a generator.

tests/lib/state_test.py
```diff
@@ -1293,3 +1295,299 @@ def _create_desired_state():
                 ]
             }
         )
+
+
+class TestLinuxBridgeState:
+    Bridge_name = "bridge0"
+    Port1 = "eth1"
+    Port2 = "eth2"
```
EdDev: CamelCase is reserved for classes, please use upper or lower cases.

tests/lib/state_test.py
```diff
@@ -1293,3 +1295,299 @@ def _create_desired_state():
...
+            LinuxBridge.CONFIG_SUBTREE: {
+                LinuxBridge.PORT_SUBTREE: [
+                    {LinuxBridge.Port.NAME: TestLinuxBridgeState.Port1},
+                ]
+            },
+        }
+        desired_state = state.State({Interface.KEY: [desired_bridge_state]})
+        desired_state.merge_interfaces(current_state)
+
+        desired_bridge_ports = desired_bridge_state.get(
```
EdDev: I think you need to query the `desired_state` object and not on the original row data/dict. They are not the same after the merge (the State constructor copies it).

tests/lib/state_test.py
```diff
@@ -1293,3 +1295,299 @@ def _create_desired_state():
...
+    Port1 = "eth1"
+    Port2 = "eth2"
+
+    def test_add_port_to_bridge(self):
+        current_bridge_state = {
+            Interface.NAME: TestLinuxBridgeState.Bridge_name,
+            Interface.TYPE: InterfaceType.LINUX_BRIDGE,
+            Interface.STATE: InterfaceState.UP,
+            LinuxBridge.CONFIG_SUBTREE: {LinuxBridge.PORT_SUBTREE: []},
+        }
```
EdDev: Consider using a helper function to create a basic bridge interface state.

tests/lib/state_test.py
maiqueb: Done.

This exposed a bug on the port merge algorithm - I had to initialize `merged_ports` to the ports coming from the desired state in https://github.com/nmstate/nmstate/pull/888/files#diff-09946053413e41502a142f4bfe77875bR987 

Good catch :)

PR #892, opened 2020-03-05
nm, linux bridge, vlan filtering: implement port vlan for linux bridges

maiqueb:
This patch implements trunk ports and access ports as per the
port vlan schema, introduced in [0].

[0] - #440

This is part of the split from PR #576 .

PR #893, opened 2020-03-05
Activate port vlan functionality

maiqueb:
Activate port vlan functionality in nmstate.

Depends on: #882 
Depends on: #888 
Depends on: #892 

### DIFF
--- a/libnmstate/nm/applier.py
+++ b/libnmstate/nm/applier.py
@@ -398,10 +398,11 @@ def _build_connection_profile(
     if bond_opts:
         settings.append(bond.create_setting(bond_opts, wired_setting))
     elif iface_type == bridge.BRIDGE_TYPE:
-        bridge_options = iface_desired_state.get(bridge.BRIDGE_TYPE, {}).get(
-            LB.OPTIONS_SUBTREE
-        )
-        if bridge_options:
+        bridge_config = iface_desired_state.get(bridge.BRIDGE_TYPE, {})
+        bridge_options = bridge_config.get(LB.OPTIONS_SUBTREE)
+        bridge_ports = bridge_config.get(LB.PORT_SUBTREE)
+
+        if bridge_options or bridge_ports:
             linux_bridge_setting = bridge.create_setting(
                 bridge_options, base_profile
             )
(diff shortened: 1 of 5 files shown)


cathay4t:
@maiqueb Don't worry about the `backport candidate` flag. We(me or @ffmancera ) will do the backport.

maiqueb:
> Previously we created the settings only if something changed on the bridge interface. Most-likely this new code is correct, but at the same time it changes logic that is not a must.
IMO, it will be safer to check if the options or the port subtree changed and only then create the settings.

> An example where this changed the current case: Previously, if one defined a bridge without any special configuration (only the name, type and state), we would not have created this nm setting at all and some default would have been used probably. Now, we do create it and therefore this has the potential to destabilize the other scenarios.

> Lets do this thing as a separate commit and for now just check also the port subtree explicitly. We should not create the setting if no options or port subtree was specified.
My main concern here is that we may need to revert this change and I will prefer to do it with a simple git commit revert. So it may as well be placed in this PR, just before this commit.

I've noticed that we could introduce a less traumatic change; take a short look @ https://github.com/nmstate/nmstate/pull/893/commits/d282cf2d116ec61904c66393f31981e286b82494 

Would that option be OK for you ? If so, I'll just squash that commit into the others.

EdDev:
> I've noticed that we could introduce a less traumatic change; take a short look @ d282cf2
> Would that option be OK for you ? If so, I'll just squash that commit into the others.

Could you please explain me why this complexity is needed? I lost you there.
I just asked to check in addition to the options the port subtree, if one of them exists, call the setting. Isn't that simpler?
Alternatively, you could just  do what you have now, but as a separate commit and then add the commit with the vlan filtering.

I surely do not think it is a good idea for the caller to check if vlan filtering is enabled or not, that was an older version we all recommended not to go with.

libnmstate/nm/applier.py
```diff
@@ -401,7 +401,10 @@ def _build_connection_profile(
         bridge_options = iface_desired_state.get(bridge.BRIDGE_TYPE, {}).get(
             LB.OPTIONS_SUBTREE
         )
-        if bridge_options:
+        bridge_ports = iface_desired_state.get(bridge.BRIDGE_TYPE, {}).get(
```
EdDev: nit: It should be enough to do the `iface_desired_state.get(bridge.BRIDGE_TYPE, {})` once.

PR #894, opened 2020-03-05
automation: compare packaging againts upstream branch

ffmancera:
Travis is only taking the contributor branch, in order to do the diff
over the packaging folder we must compare againts the upstream repository.

Signed-off-by: Fernando Fernandez Mancera <ffmancera@riseup.net>

PR #887, opened 2020-03-04
containers: replace docker.io hub with quay.io hub

ffmancera:
Signed-off-by: Fernando Fernandez Mancera <ffmancera@riseup.net>

PR #897, opened 2020-03-05
Backport bulk for nmstate-0.2.8

ffmancera:
This is backporting the following PRs: #894, #887, #886, #870, #863, #860


ffmancera:
> Back-porting to a stable branch should be carefully reviewed and decided upon.
> Please do not do this in a bulk, please propose each separately in order to avoid mistakes and allow progress even if some backports are not agreed upon.
> 
> In general, I would suggest to follow some of the downstream rules, even if it is an overhead. Specifically, having an issue for each backport that explains why it merits it.
> (A BZ, if it exists, is enough for me)

I agree with that. But I think this is going to overload us with a lot of PRs to review. In addition, @cathay4t requested a Backport bulk instead of a PR for each patch backport. What do you think? @cathay4t

ffmancera:
Some patches have been splitted in #927

ffmancera:
I am closing this because I am going to split the commits in several PRs.

PR #913, opened 2020-03-09
state: An empty dict value is treated as an overwrite

EdDev:
The recursive dict update is used to merge two dicts that represent
states.
When the data to-be-merge includes an empty dict as the value, it will
overwrite any existing data.

This change allows a client to overwrite a whole subtree from the existing
state.

### DIFF
--- a/libnmstate/state.py
+++ b/libnmstate/state.py
@@ -200,7 +200,6 @@ class State:
         self._state = copy.deepcopy(state)
 
         self._ifaces_state = self._index_interfaces_state_by_name()
-        self._complement_interface_empty_ip_subtrees()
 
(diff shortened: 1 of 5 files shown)


libnmstate/state.py
```diff
@@ -768,6 +769,9 @@ def merge_route_rules(self, other_state):
 def dict_update(origin_data, to_merge_data):
     """Recursevely performes a dict update (merge)"""
 
+    if not to_merge_data:
```
maiqueb: shouldn't this be `if to_merge_data == {} ` instead ?

EdDev: `not to_merge_data` includes this case and more.
As it can only be an instance of `Mapping`, we should be fine with the generic version.

PR #915, opened 2020-03-10
automation: rebuild rpmdb if copr repo is being used

ffmancera:
Travis might fail with

```
error: cannot open Packages database in /var/lib/rpm
Error: Error: rpmdb open failed
Error: non zero exit code: 1: OCI runtime error
```

Rebuilding the rpmdb when copr repositories are being used should fix
this issue.


ffmancera:
I've been investigating a lot here. I have not been able to find a different solution rather than what @cathay4t did in https://github.com/nmstate/nmstate/commit/818149c63f5377c6a39c0b9794b2528fcf22f4da. As it seems the rpmdb error is solved, I am going to close this until a new error rpmdb error.

PR #918, opened 2020-03-10
Merge linux bridge ports via port_update

maiqueb:
code share . WHATEVER YOU DO, DON'T MERGE:

Depends on: #913 
Alternative implementation to: #888

### DIFF
--- a/libnmstate/state.py
+++ b/libnmstate/state.py
@@ -380,9 +380,15 @@ class State:
         This is a reverse recursive update operation.
         """
         other_state = State(other_state.state)
+        origin_other_state = State(other_state.state)
         for name in self.interfaces.keys() & other_state.interfaces.keys():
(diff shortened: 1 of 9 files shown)


libnmstate/state.py
```diff
@@ -939,3 +945,53 @@ def _remove_route_rule_default_values(rule):
...
+
+
+def _index_ports(ports):
+    indexed_ports = {}
+    for port in ports:
+        port_name = port.get(LinuxBridge.Port.NAME)
+
+        if port_name:
+            indexed_ports[port_name] = port
+    return indexed_ports
```
EdDev: it must have a port name, so just following the interfaces indexing, this should work:
```
        return {
            port[LinuxBridge.Port.NAME]: port for port in ports
        }
```

maiqueb:
Abandoned in favor of https://github.com/nmstate/nmstate/pull/888

Issue #903, opened 2020-03-06
Checking libnm/NM version mismatch


ffmancera:
```
def nm_decode_version(v = None):
    if v is None:
        v = NM.utils_version()
    return ((v >> 16) & 0xFF,
            (v >>  8) & 0xFF,
            (v) & 0xFF)
```

You can also print it in in hex, which gives 0x11503, meaning
0x1.0x15.0x3, which is 1.21.3.

The major benefit of this encoding is that you can compare versions by
number.

```
def nm_encode_version(major, minor, micro):
    return (major << 16) | (minor << 8) | micro
...
if NM.utils_version() < nm_encode_version(1, 20, 0):
    warn("version missmatch")
```
or just
```
if NM.utils_version() < 0x11400: # 1.20.0
    warn("version missmatch")
```

tanuja3:
I'm learning about how the code base works for GSoC project. Can I work on this issue?

ffmancera:
Yes please, feel free to take it. I suggest you to read the CONTRIBUTING.md documentation before creating a pull request. Thanks! Let me know if you have any question.

PR #921, opened 2020-03-11
Checking the libnm and nm version mismatch

tanuja3:
Adding nm_encode_version() and libnm_nm_version_mismatch_check() functions to give warning in case of mismatch between libnm and nm versions.

### DIFF
--- a/libnmstate/__init__.py
+++ b/libnmstate/__init__.py
@@ -32,6 +32,8 @@ from .prettystate import PrettyState
 
 ROOT_DIR = os.path.dirname(os.path.abspath(__file__))
 
+_MINIMUM_NM_VERSION = "1.22.8"
+
 __all__ = [
     "show",
     "apply",
--- a/libnmstate/nm/nmclient.py
+++ b/libnmstate/nm/nmclient.py
@@ -37,6 +37,7 @@ from gi.repository import GLib
 from gi.repository import GObject
 
 from libnmstate import error
+from libnmstate import _MINIMUM_NM_VERSION
 
 GObject
 
@@ -52,10 +53,34 @@ def nm_version():
 
 
 def libnm_nm_version_mismatch_check():
-    if StrictVersion(NM.utils_version()) < StrictVersion(
+    if StrictVersion(NM.Client.get_version()) < StrictVersion(
+        _MINIMUM_NM_VERSION
+    ):
+        warnings.warn(
+            "NetworkManager version ("
+            + NM.Client.get_version()
+            + ") is too old, atleast version ("
+            + _MINIMUM_NM_VERSION
+            + ") is required."
+        )
+    if StrictVersion(NM.utils_version()) < StrictVersion(_MINIMUM_NM_VERSION):
+        warnings.warn(
+            "Libnm version ("
+            + NM.utils_version()
+            + ") is too old, atleast version ("
+            + _MINIMUM_NM_VERSION
+            + ") is required."
+        )
+    if StrictVersion(NM.utils_version()) != StrictVersion(
         NM.Client.get_version()
     ):
-        warnings.warn("Version mismatch")
+        warnings.warn(
+            "NetworkManager version ("
+            + NM.Client.get_version()
+            + ") does not match libnm version ("
+            + NM.utils_version()
+            + ")."
+        )
 
 
 def _delete_client():
@@ -257,6 +282,7 @@ class _MainLoop:
 class _NmClient:
     def __init__(self):
         self._client = NM.Client.new(None)
+        libnm_nm_version_mismatch_check()
 
     @property
     def client(self):
--- a/libnmstate/nm/nmclient.py
+++ b/libnmstate/nm/nmclient.py
@@ -23,6 +23,7 @@ import logging
 import sys
 
 import gi
+import warnings
 
 try:
     gi.require_version("NM", "1.0")  # NOQA: F402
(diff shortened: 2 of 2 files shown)


libnmstate/nm/nmclient.py
```diff
@@ -49,6 +51,13 @@ def nm_version():
     return NM.Client.get_version(client())
 
 
+def libnm_nm_version_mismatch_check():
+    if StrictVersion(NM.utils_version()) < StrictVersion(
```
tyll: I think this should check for inequality, we want to warn if they are not equal in this case.

libnmstate/nm/nmclient.py
```diff
@@ -49,6 +51,13 @@ def nm_version():
     return NM.Client.get_version(client())
 
 
+def libnm_nm_version_mismatch_check():
```
tyll: something needs to call this function to ensure that the check will actually happen, maybe when the client is created, @ffmancera what do you think?

libnmstate/nm/nmclient.py
```diff
@@ -49,6 +51,13 @@ def nm_version():
...
+    if StrictVersion(NM.utils_version()) < StrictVersion(
+        NM.Client.get_version()
+    ):
+        warnings.warn("Version mismatch")
```
tyll: The error message should be more detailed, something like `NetworkManager version (1.20.2) does not match libnm version (1.23.3)`.

libnmstate/nm/nmclient.py
tyll: I re-read the issue. Unfortunately it was not very clear, I am sorry for the confusion. There are actually three warnings

1) Warn if the versions do not match
2) Warn if the NM version is too old, currently we require at least 1.22.8 (the version should be a constant, probably in `libnmstate/__init__.py` with a name like `_MINIMUM_NM_VERSION`. @ffmancera what do you think?
3) (optional, not sure if it also makes sense) Warn if the utils version is too old

I hope things are more clear now. What do you think?

tanuja3: Yes, things are clearer now. I'll implement the first two warnings. For the 3rd warning what should the lower limit of the utils_version be?

tanuja3: I have another query about the warning with comparison between the versions. Are the version numbers of libnm and nm supposed to be equal for sure for them to be compatible? Is there something like the versions are compatible even if libnm version is greater to nm version? Should the warning be given even for the compatible versions of libnm and nm?

libnmstate/nm/nmclient.py
tanuja3: Okay I'll do this.

libnmstate/nm/nmclient.py
tyll: > Yes, things are clearer now. I'll implement the first two warnings. For the 3rd warning what should the lower limit of the utils_version be?

the same version

> I have another query about the warning with comparison between the versions. Are the version numbers of libnm and nm supposed to be equal for sure for them to be compatible? Is there something like the versions are compatible even if libnm version is greater to nm version? Should the warning be given even for the compatible versions of libnm and nm?

Good question. For now, please warn if it is not an exact match. 1.22.x is probably compatible with 1.22.y but we do not expect anyone to do want this intentionally, therefore I favor the exact match

libnmstate/nm/nmclient.py
ffmancera: I think it would be fine to check this when the client is created. Before doing anything else to avoid unexpected errors.

tanuja3:
I don't understand why the build is failing. Could you please guide me on how to resolve this?

libnmstate/nm/nmclient.py
```diff
@@ -49,6 +52,37 @@ def nm_version():
     return NM.Client.get_version(client())
 
 
+def libnm_nm_version_mismatch_check():
+    if StrictVersion(NM.Client.get_version()) < StrictVersion(
+        _MINIMUM_NM_VERSION
+    ):
+        warnings.warn(
+            "NetworkManager version ("
```
tyll: Please use str.format() to construct the messages.

tanuja3:
Did I handle `_MINIMUM_NM_VERSION` constant - which I used to store the minimum required nm version in `libnmstate/__init__.py` incorrectly?

libnmstate/__init__.py
```diff
@@ -32,6 +32,8 @@
 
 ROOT_DIR = os.path.dirname(os.path.abspath(__file__))
 
+_MINIMUM_NM_VERSION = "1.22.8"
```
tyll: Sorry for the confusion, this does not work because this causes a cyclic dependency. Please move this constant into the nmclient.py file.

libnmstate/nm/nmclient.py
```diff
@@ -49,6 +52,31 @@ def nm_version():
     return NM.Client.get_version(client())
 
 
+def libnm_nm_version_mismatch_check():
+    if StrictVersion(NM.Client.get_version()) < StrictVersion(
+        _MINIMUM_NM_VERSION
+    ):
+        warnings.warn(
+            "NetworkManager version ({}) is too old, atleast version ({}) is required.".format(
```
tyll:
```suggestion
            "NetworkManager version ({}) is too old, at least version ({}) is required.".format(
```

libnmstate/nm/nmclient.py
```diff
@@ -49,6 +52,31 @@ def nm_version():
...
+                NM.Client.get_version(), _MINIMUM_NM_VERSION
+            )
+        )
+    if StrictVersion(NM.utils_version()) < StrictVersion(_MINIMUM_NM_VERSION):
+        warnings.warn(
+            "Libnm version ({}) is too old, atleast version ({}) is required.".format(
```
tyll:
```suggestion
            "Libnm version ({}) is too old, at least version ({}) is required.".format(
```

libnmstate/__init__.py
tanuja3: Thank you, I'll change that. But also getting another error because `NM.utils_version()` function will return the version ID of libnm but not the version. Can you please suggest what function I should use to get the libnm version?

tyll: #903 contains example code to decode/encode the version, please take a look

tanuja3:
I'm not being able to make progress on this issue. So I'm closing this PR, thank you.

Issue #902, opened 2020-03-06
travis: create optional job running VDSM tests


ffmancera:
It would be nice to add a script to Nmstate that runs the VDSM script with the current Nmstate GIT repo and also adding it as an optional job to Travis.

How to run VDSM tests:

```
git clone https://gerrit.ovirt.org/vdsm
cd vdsm
sudo bash -c "TEST_NMSTATE=1 ./tests/network/functional/run-tests.sh"

To run only certain tests:
bash -c "TEST_NMSTATE=1 ./tests/network/functional/run-tests.sh --shell"

Once you are in the container:
cd /vdsm-tmp/vdsm/
pytest -vv --log-level=DEBUG --target-lib --skip-stable-link-monitor -m "legacy_switch and nmstate" tests/network/functional/(test_file)
```

PR #923, opened 2020-03-11
automation: Add travis job to run vdsm tests

almusil:


almusil:
Once https://gerrit.ovirt.org/c/107553/ is merged I can proceed

automation/run-tests.sh
```diff
@@ -259,11 +259,16 @@ while true; do
...
+        shift
+        echo "Testing vdsm"
+        exit
+        ;;
     --help)
         set +x
         echo -n "$0 [--copr=...] [--customize=...] [--debug-shell] [--el8] "
         echo -n "[--help] [--pytest-args=...] "
-        echo "[--test-type=<TEST_TYPE>]"
+        echo "[--test-type=<TEST_TYPE>] [--test-vdsm=<PR_NUMBER>]"
```
tyll: It would be great if this could use the local Nmstate git checkout, maybe by adding a volume to the VDSM container. Then it would also work on locally instead of only for published PRs.

almusil: I was thinking about that, let's see what I will manage to do with vdsm side

automation/run-tests.sh
```diff
@@ -227,8 +227,16 @@ function modprobe_ovs {
     lsmod | grep -q ^openvswitch || modprobe openvswitch || { echo 1>&2 "Please run 'modprobe openvswitch' as root"; exit 1; }
 }
 
+function vdsm_tests {
+  git clone https://gerrit.ovirt.org/vdsm
+  cd vdsm
+  ./tests/network/functional/run-tests.sh --nmstate-source="$PROJECT_PATH"
+  cd -
+  rm -rf vdsm
```
almusil: I am not sure what we can do about the vdsm clone. Any idea?

.travis.yml
```diff
@@ -22,13 +22,16 @@ env:
           testflags="--test-type lint"
         - CONTAINER_IMAGE=quay.io/nmstate/centos8-nmstate-dev
           testflags="--test-type unit_py36"
+        - testflags="--test-vdsm=$TRAVIS_PULL_REQUEST"
```
almusil: And I forgot to remove the PR number

automation/run-tests.sh
tyll: What exactly is the problem that you see here? Cloning with git in general or the possible problems about how it is implemented here?

automation/run-tests.sh
```diff
@@ -227,8 +227,16 @@ function modprobe_ovs {
     lsmod | grep -q ^openvswitch || modprobe openvswitch || { echo 1>&2 "Please run 'modprobe openvswitch' as root"; exit 1; }
 }
 
+function vdsm_tests {
+  git clone https://gerrit.ovirt.org/vdsm
```
tyll: This might make it faster:

```suggestion
  git clone --depth 1 https://gerrit.ovirt.org/vdsm
```

automation/run-tests.sh
```diff
@@ -227,8 +227,16 @@ function modprobe_ovs {
...
+  cd vdsm
+  ./tests/network/functional/run-tests.sh --nmstate-source="$PROJECT_PATH"
```
tyll: If you only need the shell script but nothing else from the repo, something like this is possible:

`curl https://raw.githubusercontent.com/oVirt/vdsm/master/tests/network/functional/run-tests.sh | bash -s - --nmstate-source="$PROJECT_PATH"`

 but since `run-test.sh` from vdsm does not use a main function that is called at the end, this execute only part of the script if there is a transmission error.

automation/run-tests.sh
almusil: The possible solution how to remove any leftover. There should be logic that will remove it no matter if the test fail or not.

automation/run-tests.sh
almusil: Unfortunately we need whole repo because it is mounted into the container.

automation/run-tests.sh
tyll: Here is my suggestion:

1) Always create a temp directory for the tests:
```bash
NMSTATE_TEMPDIR=$(mktemp -d /tmp/nmstate-test-XXXX)
```

2) Update run_exit to always remove it

3) Clone vdsm into the temp dir and run the test from there if requested
```bash
git clone -C "${NMSTATE_TEMPDIR}" --depth 1 https://gerrit.ovirt.org/vdsm
"${NMSTATE_TEMPDIR}/vdsm/tests/network/functional/run-tests.sh" --nmstate-source="$PROJECT_PATH"
```

What do you think?

almusil: Sounds great, thanks

automation/run-tests.sh
almusil: Done, thanks

almusil:
Closes #902

automation/run-tests.sh
```diff
@@ -227,8 +230,20 @@ function modprobe_ovs {
...
 }
 
+function remove_tempdir {
+   rm -rf "$NMSTATE_TEMPDIR"
+}
+
+function vdsm_tests {
+  trap remove_tempdir EXIT
+  git -C "$NMSTATE_TEMPDIR" clone --depth 1 https://gerrit.ovirt.org/vdsm
+  cd "$NMSTATE_TEMPDIR/vdsm"
```
tyll: Is the `cd` necessary?

tyll:
The new job fails in the CI, is this expected? https://travis-ci.com/github/nmstate/nmstate/jobs/297323042

automation/run-tests.sh
almusil: Done

almusil:
> The new job fails in the CI, is this expected? https://travis-ci.com/github/nmstate/nmstate/jobs/297323042

It might be actually regression on master. I haven't tested older versions of nmstate master but this was not failing before on 0.2 branch for sure.

automation/run-tests.sh
almusil: It actually is necessary. Otherwise we would end up with wrong configuration for the vdsm func tests.

cathay4t:
> > The new job fails in the CI, is this expected? https://travis-ci.com/github/nmstate/nmstate/jobs/297323042
> 
> It might be actually regression on master. I haven't tested older versions of nmstate master but this was not failing before on 0.2 branch for sure.

Interesting failure. I will investigate on it.

cathay4t:
Created github issue to tracking the issue of:

https://github.com/nmstate/nmstate/issues/959

automation/run-tests.sh
```diff
@@ -234,8 +234,16 @@ function remove_tempdir {
    rm -rf "$NMSTATE_TEMPDIR"
 }
 
+function vdsm_tests {
+  trap remove_tempdir EXIT
```
ffmancera: Please, use 4 spaces for each indentation. Thanks!

almusil: Done, thanks

ffmancera:
> Created github issue to tracking the issue of:
> 
> #959

In my opinion, we should merge the tests and then continue investigating the failure.

tyll:
Will merge without rebase since it does not touch actual code.

PR #927, opened 2020-03-11
Update Travis CI on nmstate-0.2

ffmancera:
This is needed in order to avoid CI blocking when backporting patches to nmstate-0.2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,32 +5,32 @@ env:
     global:
         - use_coveralls=true
     matrix:
-        - CONTAINER_IMAGE=nmstate/centos8-nmstate-dev
+        - CONTAINER_IMAGE=quay.io/nmstate/centos8-nmstate-dev:nmstate-0.2
           testflags="--test-type integ --pytest-args='-x'"
           use_codecov=true
           use_coveralls=false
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -6,13 +6,12 @@ env:
         - use_coveralls=true
     matrix:
         - CONTAINER_IMAGE=quay.io/nmstate/centos8-nmstate-dev:nmstate-0.2
-          testflags="--test-type integ --pytest-args='-x'"
+          testflags="--test-type integ --pytest-args='-x'
```
tyll: @EdDev are you ok with this change?

EdDev: No, I am not. We need an NM release otherwise no one can consume our releases.

ffmancera:
> travis: use NM-1.22-git Copr repository for integration tests: this change has no reasoning and I do not think we can release nmstate when no underly nm version exists to back it up.

Currently, NM-1.22-git contains internal fixes like the NM crash when activating OVS interfaces, this is blocking us on CI. Upgrading the CI to 1.22-git will solve this.

ffmancera:
This is not needed anymore as NM has released 1.22.10. I am going to work on the backport.

PR #930, opened 2020-03-12
integ: add MAC behaviour tests for linux bonds

ffmancera:
The linux bond should keep the MAC address even if the slaves are being
reordered, removed, added or replaced when fail_over_mac is 0.

### DIFF
--- a/tests/integration/bond_test.py
+++ b/tests/integration/bond_test.py
@@ -101,7 +101,7 @@ def bond88_with_slave(eth1_up):
 
 
 @pytest.fixture
-def bond99_with_slave(eth2_up):
+def bond99_with_eth2(eth2_up):
     slaves = [eth2_up[Interface.KEY][0][Interface.NAME]]
     with bond_interface(BOND99, slaves) as state:
         yield state
@@ -302,9 +302,9 @@ def test_remove_one_of_the_bond_slaves(eth1_up, eth2_up):
 
 
 @pytest.mark.tier1
-def test_swap_slaves_between_bonds(bond88_with_slave, bond99_with_slave):
+def test_swap_slaves_between_bonds(bond88_with_slave, bond99_with_eth2):
     bonding88 = bond88_with_slave[Interface.KEY][0][Bond.CONFIG_SUBTREE]
-    bonding99 = bond99_with_slave[Interface.KEY][0][Bond.CONFIG_SUBTREE]
+    bonding99 = bond99_with_eth2[Interface.KEY][0][Bond.CONFIG_SUBTREE]
 
     bonding88[Bond.SLAVES], bonding99[Bond.SLAVES] = (
         bonding99[Bond.SLAVES],
@@ -312,7 +312,7 @@ def test_swap_slaves_between_bonds(bond88_with_slave, bond99_with_slave):
     )
 
     state = bond88_with_slave
-    state.update(bond99_with_slave)
+    state.update(bond99_with_eth2)
     libnmstate.apply(state)
 
     assertlib.assert_state(state)
@@ -371,8 +371,8 @@ def test_bond_with_empty_ipv6_static_address(eth1_up):
 
 
 @pytest.mark.tier1
-def test_create_vlan_over_a_bond_slave(bond99_with_slave):
-    bond_ifstate = bond99_with_slave[Interface.KEY][0]
+def test_create_vlan_over_a_bond_slave(bond99_with_eth2):
+    bond_ifstate = bond99_with_eth2[Interface.KEY][0]
     bond_slave_ifname = bond_ifstate[Bond.CONFIG_SUBTREE][Bond.SLAVES][0]
     vlan_id = 102
     vlan_iface_name = "{}.{}".format(bond_slave_ifname, vlan_id)
@@ -380,11 +380,11 @@ def test_create_vlan_over_a_bond_slave(bond99_with_slave):
         vlan_iface_name, vlan_id, bond_slave_ifname
     ) as desired_state:
         assertlib.assert_state(desired_state)
-    assertlib.assert_state(bond99_with_slave)
+    assertlib.assert_state(bond99_with_eth2)
 
 
 @pytest.mark.tier1
-def test_create_linux_bridge_over_bond(bond99_with_slave):
+def test_create_linux_bridge_over_bond(bond99_with_eth2):
     port_state = {
         "stp-hairpin-mode": False,
         "stp-path-cost": 100,
@@ -399,24 +399,24 @@ def test_create_linux_bridge_over_bond(bond99_with_slave):
 
 
 @pytest.mark.tier1
-def test_preserve_bond_after_bridge_removal(bond99_with_slave):
+def test_preserve_bond_after_bridge_removal(bond99_with_eth2):
     bridge_name = "linux-br0"
     bridge_state = add_port_to_bridge(create_bridge_subtree_state(), BOND99)
     with linux_bridge(bridge_name, bridge_state) as desired_state:
         assertlib.assert_state_match(desired_state)
-    assertlib.assert_state(bond99_with_slave)
+    assertlib.assert_state(bond99_with_eth2)
 
 
 @pytest.mark.tier1
-def test_create_vlan_over_a_bond(bond99_with_slave):
-    vlan_base_iface = bond99_with_slave[Interface.KEY][0][Interface.NAME]
+def test_create_vlan_over_a_bond(bond99_with_eth2):
+    vlan_base_iface = bond99_with_eth2[Interface.KEY][0][Interface.NAME]
     vlan_id = 102
     vlan_iface_name = "{}.{}".format(vlan_base_iface, vlan_id)
     with vlan_interface(
         vlan_iface_name, vlan_id, vlan_base_iface
     ) as desired_state:
         assertlib.assert_state(desired_state)
-    assertlib.assert_state(bond99_with_slave)
+    assertlib.assert_state(bond99_with_eth2)
 
 
 @pytest.mark.tier1
@@ -430,7 +430,7 @@ def test_change_bond_option_miimon(bond99_with_2_slaves):
 
 
 @pytest.mark.tier1
-def test_change_bond_option_with_an_id_value(bond99_with_slave):
+def test_change_bond_option_with_an_id_value(bond99_with_eth2):
     option_name = "xmit_hash_policy"
     desired_state = statelib.show_only((BOND99,))
     iface_state = desired_state[Interface.KEY][0]
--- a/tests/integration/bond_test.py
+++ b/tests/integration/bond_test.py
@@ -251,7 +251,9 @@ def test_add_slave_to_bond_without_slaves(eth1_up):
         assert bond_cur_state[Bond.CONFIG_SUBTREE][Bond.SLAVES] == [slave_name]
 
 
-@pytest.mark.xfail(strict=True, reason="Jira issue # NMSTATE-143")
+@pytest.mark.xfail(
+    strict=True, reason="https://github.com/nmstate/nmstate/issues/932"
+)
 def test_remove_all_slaves_from_bond(eth1_up):
     slave_name = (eth1_up[Interface.KEY][0][Interface.NAME],)
     with bond_interface(name=BOND99, slaves=[slave_name]) as state:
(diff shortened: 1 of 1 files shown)


tests/integration/bond_test.py
```diff
@@ -353,6 +353,71 @@ def test_changing_slave_order_keeps_mac_of_existing_bond(bond99_with_2_slaves):
     )
 
 
+def test_adding_slaves_keeps_mac_of_existing_bond(bond99_with_slave, eth1_up):
```
tyll:
```suggestion
def test_adding_a_slave_keeps_mac_of_existing_bond(bond99_with_slave, eth1_up):
```

tests/integration/bond_test.py
```diff
@@ -353,6 +353,71 @@ def test_changing_slave_order_keeps_mac_of_existing_bond(bond99_with_2_slaves):
...
+    assert_mac_address(current_state)
+
+    libnmstate.apply(bond99_with_slave)
+    modified_state = statelib.show_only((BOND99,))
+    assert_mac_address(
+        modified_state, current_state[Interface.KEY][0][Interface.MAC]
+    )
+
+
+def test_adding_to_empty_slaves_keeps_mac_of_existing_bond():
```
tyll:
```suggestion
def test_adding_slaves_to_empty_bond_keeps_mac():
```

tests/integration/bond_test.py
```diff
@@ -353,6 +353,71 @@ def test_changing_slave_order_keeps_mac_of_existing_bond(bond99_with_2_slaves):
...
+        assert_mac_address(current_state)
+
+        libnmstate.apply(state)
+        modified_state = statelib.show_only((BOND99,))
+        assert_mac_address(
+            modified_state, current_state[Interface.KEY][0][Interface.MAC]
+        )
+
+
+def test_replacing_slaves_keeps_mac_of_existing_bond(
```
tyll: what is replaced to what here?

tests/integration/bond_test.py
```diff
@@ -353,6 +353,71 @@ def test_changing_slave_order_keeps_mac_of_existing_bond(bond99_with_2_slaves):
...
+    bond99_with_slave, eth1_up
```
tyll: IMHO it is not very reader-friendly that using `bond99_with_slave` means that `eth2` is the slave. Also it is very dangerous because it is not obvious that the test relies on this.

ffmancera: I could do it using `with bond_interface():` but in my opinion we should use existing features. Maybe, we can rename bond99.. but I don't think we should do this in this PR. Would you like another PR?

tyll: IMHO it is fine to rename it in this PR but probably in a different commit. It could be something like `bond99_with_eth2`.
@EdDev what do you think?

tyll: If you prefer to do the renaming in a different PR, it is also fine with me.

EdDev: Sounds fine, but you could solve it by depending on both slaves and swapping them without knowing that already exists.
In many tests you do not really care what slave is used, here you do.. this one is an exception.
But I am fine with either ways.

tests/integration/bond_test.py
ffmancera: Done, thanks!

tests/integration/bond_test.py
ffmancera: Done, thank you!

tests/integration/bond_test.py
ffmancera: Done, thank you!

tests/integration/bond_test.py
```diff
@@ -353,6 +353,71 @@ def test_changing_slave_order_keeps_mac_of_existing_bond(bond99_with_2_slaves):
...
+    assert_mac_address(current_state)
+
+    libnmstate.apply(bond99_with_eth2)
+    modified_state = statelib.show_only((BOND99,))
+    assert_mac_address(
+        modified_state, current_state[Interface.KEY][0][Interface.MAC]
+    )
+
+
+@pytest.mark.xfail(reason="Jira issue # NMSTATE-143")
```
cathay4t: Please limit the xfail with `raises=`.

ffmancera: Done, thanks!

tests/integration/bond_test.py
```diff
@@ -353,6 +353,73 @@ def test_changing_slave_order_keeps_mac_of_existing_bond(bond99_with_2_slaves):
...
+
+    libnmstate.apply(bond99_with_eth2)
+    modified_state = statelib.show_only((BOND99,))
+    assert_mac_address(
+        modified_state, current_state[Interface.KEY][0][Interface.MAC]
+    )
+
+
+@pytest.mark.xfail(
+    raises=NmstateVerificationError, reason="Jira issue # NMSTATE-143"
```
tyll: Please use full URLs to issues and ideally make it a GitHub issue

tests/integration/bond_test.py
```diff
@@ -353,6 +353,73 @@ def test_changing_slave_order_keeps_mac_of_existing_bond(bond99_with_2_slaves):
     )
 
 
+def test_adding_a_slave_keeps_mac_of_existing_bond(bond99_with_eth2, eth1_up):
```
tyll: please mark this and the other working tests as tier1

tests/integration/bond_test.py
ffmancera: Done, thanks!

tests/integration/bond_test.py
ffmancera: Done, thanks!

tests/integration/bond_test.py
```diff
@@ -353,6 +355,77 @@ def test_changing_slave_order_keeps_mac_of_existing_bond(bond99_with_2_slaves):
...
+)
+@pytest.mark.parametrize("slaves", ([ETH2], []))
+def test_removing_slaves_keeps_mac_of_existing_bond(
+    bond99_with_2_slaves, slaves
+):
+    bond_state = bond99_with_2_slaves[Interface.KEY][0]
+    bond_state[Bond.CONFIG_SUBTREE][Bond.SLAVES] = slaves
+
+    current_state = statelib.show_only((BOND99,))
+    assert_mac_address(current_state)
```
EdDev: I do not understand this assert. You check that the current mac equals the current mac?

ffmancera: If I am not wrong. This is checking that the current_state contains a mac address.

EdDev: `assert_mac_address` checks that all mac addresses in the provide state are the same.
As you provide only one, it has no value.
If you provide it with an expected mac as input, it will have value (like if you want to check it did not changed after an action).

ffmancera: Done, thanks!

tests/integration/bond_test.py
```diff
@@ -353,6 +355,73 @@ def test_changing_slave_order_keeps_mac_of_existing_bond(bond99_with_2_slaves):
...
 
+@pytest.mark.tier1
+def test_adding_a_slave_keeps_mac_of_existing_bond(bond99_with_eth2, eth1_up):
+    bond_state = bond99_with_eth2[Interface.KEY][0]
+    bond_slaves = bond_state[Bond.CONFIG_SUBTREE][Bond.SLAVES]
+    bond_slaves.insert(0, ETH1)
+
+    current_state = statelib.show_only((BOND99,))
+
+    libnmstate.apply(bond99_with_eth2)
```
EdDev: This is very confusing when reading. It is no longer a bond with one slave. Maybe do an assignment at the top: `desired_state = bond99_with_eth2`.

tests/integration/bond_test.py
```diff
@@ -353,6 +355,73 @@ def test_changing_slave_order_keeps_mac_of_existing_bond(bond99_with_2_slaves):
     )
 
 
+@pytest.mark.tier1
+def test_adding_a_slave_keeps_mac_of_existing_bond(bond99_with_eth2, eth1_up):
+    bond_state = bond99_with_eth2[Interface.KEY][0]
+    bond_slaves = bond_state[Bond.CONFIG_SUBTREE][Bond.SLAVES]
+    bond_slaves.insert(0, ETH1)
```
EdDev: nit: It will be nicer to extract the slave name from `eth1_up` directly.

tests/integration/bond_test.py
```diff
@@ -353,6 +355,73 @@ def test_changing_slave_order_keeps_mac_of_existing_bond(bond99_with_2_slaves):
...
+
+    current_state = statelib.show_only((BOND99,))
```
EdDev: nit: The bond name can also be extracted from the original state.

tests/integration/bond_test.py
```diff
@@ -353,6 +355,73 @@ def test_changing_slave_order_keeps_mac_of_existing_bond(bond99_with_2_slaves):
...
+
+    libnmstate.apply(bond99_with_eth2)
+    modified_state = statelib.show_only((BOND99,))
+    assert_mac_address(
+        modified_state, current_state[Interface.KEY][0][Interface.MAC]
+    )
+
+
+@pytest.mark.tier1
+def test_adding_slaves_to_empty_bond_keeps_mac(eth1_up):
```
EdDev: It is not expected to pass, why you name it as if it should?

tests/integration/bond_test.py
```diff
@@ -353,6 +355,73 @@ def test_changing_slave_order_keeps_mac_of_existing_bond(bond99_with_2_slaves):
...
+        bond_state[Bond.CONFIG_SUBTREE][Bond.SLAVES] = [ETH1]
+
+        current_state = statelib.show_only((BOND99,))
+
+        libnmstate.apply(state)
+        modified_state = statelib.show_only((BOND99,))
+        with pytest.raises(AssertionError):
+            assert_mac_address(
+                modified_state, current_state[Interface.KEY][0][Interface.MAC]
+            )
```
EdDev: `current_state[Interface.KEY][0][Interface.MAC] ! = modified_state[Interface.KEY][0][Interface.MAC]`

Is more appropriate and direct IMO.

tests/integration/bond_test.py
```diff
@@ -353,6 +355,73 @@ def test_changing_slave_order_keeps_mac_of_existing_bond(bond99_with_2_slaves):
...
+    bond_slaves = bond_state[Bond.CONFIG_SUBTREE][Bond.SLAVES]
+    bond_slaves.insert(0, ETH1)
+
+    current_state = statelib.show_only((BOND99,))
+
+    libnmstate.apply(bond99_with_eth2)
+    modified_state = statelib.show_only((BOND99,))
+    assert_mac_address(
+        modified_state, current_state[Interface.KEY][0][Interface.MAC]
+    )
```
EdDev: It is questionable why there is a need to use this helper.
A direct approach could have worked as well.
`current_state[Interface.KEY][0][Interface.MAC] ! = modified_state[Interface.KEY][0][Interface.MAC]`

tests/integration/bond_test.py
```diff
@@ -353,6 +355,73 @@ def test_changing_slave_order_keeps_mac_of_existing_bond(bond99_with_2_slaves):
...
+    bond99_with_eth2, eth1_up
+):
+    bond_state = bond99_with_eth2[Interface.KEY][0]
+    bond_state[Bond.CONFIG_SUBTREE][Bond.SLAVES] = [ETH1]
+
+    current_state = statelib.show_only((BOND99,))
+
+    libnmstate.apply(bond99_with_eth2)
+    modified_state = statelib.show_only((BOND99,))
+    assert_mac_address(
```
EdDev: Interesting, I did not expect this to pass. I expected the bond to get back its previous mac when all nics got disconnected from it.

tests/integration/bond_test.py
```diff
@@ -353,6 +355,73 @@ def test_changing_slave_order_keeps_mac_of_existing_bond(bond99_with_2_slaves):
...
+        modified_state, current_state[Interface.KEY][0][Interface.MAC]
+    )
+
+
+@pytest.mark.xfail(
+    raises=NmstateVerificationError,
+    reason="https://github.com/nmstate/nmstate/issues/932",
+)
+@pytest.mark.parametrize("slaves", ([ETH2], []))
+def test_removing_slaves_keeps_mac_of_existing_bond(
```
EdDev: Does this work with plain `ip link` or sysfs commands?

tests/integration/bond_test.py
ffmancera: You are right, `_doesn't keep_mac` fits better.

tests/integration/bond_test.py
ffmancera: The problem is , right now we are not able to remove all the slaves from a linux bond. It is a nmstate issue. I've not tried it with sysfs or `ip link`

tests/integration/bond_test.py
ffmancera: Done!

tests/integration/bond_test.py
ffmancera: Thanks, done!

tests/integration/bond_test.py
ffmancera: Done, thank you!

tests/integration/bond_test.py
ffmancera: Done!

tests/integration/bond_test.py
ffmancera: Done!

tests/integration/bond_test.py
ffmancera: Done!

tests/integration/bond_test.py
ffmancera: Done!

ffmancera: tbh, me neither. :-)

tests/integration/bond_test.py
ffmancera: Tested with `ip link`. It is not keeping the mac address.

PR #934, opened 2020-03-12
nmstate 0.3: Enforce keyword args for apply(), commit(), rollback() and show()

### LINKED ISSUE #786: nmstate 0.3: Enforce keyword args for apply(), commit(), rollback() and show()
This is a follow-up, to #635 once nmstate 0.3 is branched. This needs to revert the changes for warnings from #635 and add `*` to the function definitions instead.

adwait-thattey:
resolves #786


tests/lib/netinfo_test.py
```diff
@@ -135,41 +135,3 @@ def test_netinfo_show_bond_iface(nm_mock, nm_dns_mock):
...
-    nm_mock.ipv4.get_info.return_value = current_iface0["ipv4"]
-    nm_mock.ipv6.get_info.return_value = current_iface0["ipv6"]
-    nm_mock.ipv4.get_route_running.return_value = []
-    nm_mock.ipv4.get_route_config.return_value = []
-    nm_mock.ipv6.get_route_running.return_value = []
-    nm_mock.ipv6.get_route_config.return_value = []
-    nm_dns_mock.get_running.return_value = current_config[DNS.KEY][DNS.RUNNING]
-    nm_dns_mock.get_config.return_value = current_config[DNS.KEY][DNS.CONFIG]
-
-    with pytest.warns(FutureWarning) as record:
```
tyll: Please change this to expect an exception instead of removing the test.

tests/lib/netapplier_test.py
```diff
@@ -178,38 +178,3 @@ def test_edit_existing_bond(netinfo_nm_mock, netapplier_nm_mock):
     m_apply_changes.assert_called_once_with(
         desired_config[INTERFACES], State(desired_config)
     )
-
-
-@mock.patch.object(netapplier, "_apply_ifaces_state", lambda *_: None)
-def test_warning_apply():
-    with pytest.warns(FutureWarning) as record:
```
tyll: Please change this to expect an exception instead of removing the test, also in the other instances.

libnmstate/netapplier.py
```diff
@@ -28,7 +28,6 @@
 from libnmstate.nm.nmclient import nmclient_context
 from libnmstate import state
 from libnmstate import validator
-from libnmstate.deprecation import _warn_keyword_as_positional
```
tyll: Please also remove the code for `_warn_keyword_as_positional`, it seems to be kept.

tests/integration/timeout_test.py
```diff
@@ -55,5 +55,5 @@ def test_lot_of_vlans_with_bridges(eth1_up):
             ]
         )
     checkpoint = libnmstate.apply({Interface.KEY: interfaces}, commit=False)
-    libnmstate.rollback(checkpoint)
+    libnmstate.rollback(checkpoint=checkpoint)
```
tyll: Good, please keep this as its own commit but reorder it to be the first commit. Also follow the commit template from the repo.

libnmstate/netapplier.py
tyll: Please use the commit message template, squash this commit and the one for updating the integration tests for this change and consider using "Adwait Thattey <coderdude1999@gmail.com>" for the signed-of-by tag.

tyll:
I do not see the changes, did you maybe forget to push?

tests/lib/netinfo_test.py
adwait-thattey: Should I just check if `TypeError` exception is raised or should I also check if the error string contains things like "takes .. positional arguments but ... was given"

adwait-thattey:
The PR test is failing because of pylint. 
For example: the test has `netinfo.show(None)` to ensure that error is thrown. 
But pylint is flagging it as incorrect as the API has changed. 
The actual test is running correctly. 

How to handle this?

tests/lib/netinfo_test.py
tyll: checking the exception class is enough, thank you!

tyll:
Regarding pylint, please see: https://stackoverflow.com/questions/28829236/is-it-possible-to-ignore-one-single-specific-line-with-pylint
Thank you!

adwait-thattey:
All the tests seem to be passing now

adwait-thattey:
Change the messages. The 2nd one seemed long to me too. But couldn't figure what else to write

ffmancera:
> Changed the messages. The 2nd one seemed long to me too. But couldn't figure what else to write

If I am not wrong, CI is checking that too. If it doesn't complain we are good. :-)

adwait-thattey:
> > Changed the messages. The 2nd one seemed long to me too. But couldn't figure what else to write
> 
> If I am not wrong, CI is checking that too. If it doesn't complain we are good. :-)

The CI seems to have passed

tyll:
@adwait-thattey when you rebase, please reword the commit message to wrap the long lines at 72 characters, Thank you.

adwait-thattey:
rebased

adwait-thattey:
Rebased

adwait-thattey:
Can this be merged now?

tests/lib/netinfo_test.py
```diff
@@ -168,8 +168,6 @@ def test_warning_show(nm_mock, nm_dns_mock):
     nm_dns_mock.get_running.return_value = current_config[DNS.KEY][DNS.RUNNING]
     nm_dns_mock.get_config.return_value = current_config[DNS.KEY][DNS.CONFIG]
 
-    with pytest.warns(FutureWarning) as record:
+    with pytest.raises(TypeError):
+        # pylint: disable=E1121
```
cathay4t: Use proper name instead of E1121.

Please re-enable the pylint after the line you want  to disable.

libnmstate/netapplier.py
```diff
@@ -42,9 +41,10 @@
 VERIFY_RETRY_TIMEOUT = 5
 
 
-@_warn_keyword_as_positional
```
cathay4t: I prefer raise an exception instead of silently ignore arguments.

adwait-thattey: So, instead of *, should I create a decorator that checks for positional args and raises an exception?

tyll: I think there is a misunderstanding. @cathay4t this will not ignore any arguments. If extra positional arguments are provided, Python will raise a TypeError. The tests check this.

adwait-thattey: what should I do?

tyll: For this we wait for @cathay4t feedback, but you also need to address the comment about pylint and rebase to resolve a conflict.

adwait-thattey: @tyll I already addressed the pylint issue. I was waiting to push both together.

adwait-thattey:
wrt pylint issue, I changed it to use names instead of codes. But putting the disable command on independent lines is creating issues.
For example:
`netapplier.apply(         #pylint: disable=too-many-function-args
    {"interfaces": []}, True, True, 0
)`
causes black to reformat it to put the comment at the end of third line. But pylint requires it in the first line to disable the error. For this I will need to add another exception for black, which feels like a hack. 

Instead AFAIU, the disable command only runs in the scope. So adding it like,
`
with pytest.raises(TypeError):
      # pylint: disable=too-many-function-args
      ...
`
should fix the issue without causing issues with black. 
This is what I have done as of now. 

Is there any other way to resolve this issue?

ffmancera:
> wrt pylint issue, I changed it to use names instead of codes. But putting the disable command on independent lines is creating issues.
> For example:
> 
> ```python
> netapplier.apply(   #pylint: disable=too-many-function-args   
>     {"interfaces": []}, True, True, 0
> )
> ```
> 
> causes black to reformat it to put the comment at the end of third line. But pylint requires it in the first line to disable the error. For this I will need to add another exception for black, which feels like a hack.
> 
> Instead AFAIU, the disable command only runs in the scope. So adding it like,
> 
> ```python
> with pytest.raises(TypeError):  
>       # pylint: disable=too-many-function-args  
>       ...
> ```
> 
> should fix the issue without causing issues with black.
> This is what I have done as of now.
> 
> Is there any other way to resolve this issue?

Is there another way to fix this @tyll? If we need to disable pylint each time we use netapplier.commit/apply/rollback it doesn't seem good to me.

In addition, @adwait-thattey you need to re-enable it after the line you want to disable. Thanks!

tyll:
> Is there another way to fix this @tyll? If we need to disable pylint each time we use netapplier.commit/apply/rollback it doesn't seem good to me.

It is only needed here because the test deliberately uses the methods in a wrong way and the pylint warning is actually expected here.

> In addition, @adwait-thattey you need to re-enable it after the line you want to disable. Thanks!

+1

libnmstate/netapplier.py
cathay4t: My fault. The `TypeError` is enough for me.

adwait-thattey:
Enabled pylint again after each func call.

Is this okay?

cathay4t:
Merged. Thanks for the help.

PR #956, opened 2020-03-16
Backport: nm sriov: Properly return when device not exists

ffmancera:
When interface does not exists, the `nm.sriov._has_sriov_capability()`
will raise:

```python
    def _has_sriov_capability(ifname):
        dev = device.get_device_by_name(ifname)
>       if nmclient.NM.DeviceCapabilities.SRIOV & dev.props.capabilities:
E       AttributeError: 'NoneType' object has no attribute 'props'
```

Fixed by add additional check on whether device is None.

Also fixed the integration test which incorrectly using eth0 which is
not test NIC, changed to eth1.

Signed-off-by: Gris Ge <fge@redhat.com>

### DIFF
--- a/examples/eth1_with_sriov.yml
+++ b/examples/eth1_with_sriov.yml
@@ -1,0 +1,13 @@
+---
+interfaces:
+- name: eth1
+  type: ethernet
+  state: up
+  ethernet:
+    sr-iov:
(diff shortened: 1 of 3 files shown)


ffmancera:
Rebased, thanks!

PR #965, opened 2020-03-17
validator: raise if the necessary services are not running

ffmancera:
OVS interfaces require openvswitch service started and running. If it is
not running and the desired state contains OVS interfaces then nmstate
is going to raise an error in the validation.

### DIFF
--- a/libnmstate/netinfo.py
+++ b/libnmstate/netinfo.py
@@ -20,6 +20,7 @@ from operator import itemgetter
 
 from libnmstate import nm
 from libnmstate import validator
+from libnmstate.appliers.ovs_bridge import is_ovs_running
 from libnmstate.deprecation import _warn_keyword_as_positional
 from libnmstate.nm import dns as nm_dns
 from libnmstate.nm.nmclient import nmclient_context
@@ -78,7 +79,7 @@ def show(include_status_data=False):
 def capabilities():
     caps = set()
 
-    if nm.ovs.has_ovs_capability():
+    if nm.ovs.has_ovs_capability() and is_ovs_running():
         caps.add(nm.ovs.CAPABILITY)
 
     if nm.team.has_team_capability():
--- a/libnmstate/validator.py
+++ b/libnmstate/validator.py
@@ -25,6 +25,7 @@ import jsonschema as js
 from . import nm
 from . import schema
 from libnmstate.appliers.bond import is_in_mac_restricted_mode
+from libnmstate.appliers.ovs_bridge import is_ovs_running
 from libnmstate.schema import DNS
 from libnmstate.schema import Interface
 from libnmstate.schema import InterfaceIP
@@ -84,11 +85,16 @@ def validate_interface_capabilities(ifaces_state, capabilities):
             nm.ovs.INTERNAL_INTERFACE_TYPE,
         )
         if is_ovs_type and not has_ovs_capability:
-            raise NmstateDependencyError(
-                "Open vSwitch NetworkManager support not installed "
-                "and started"
-            )
-        if iface_type == InterfaceType.TEAM and not has_team_capability:
+            if not is_ovs_running():
+                raise NmstateDependencyError(
+                    "openvswitch service is not started."
+                )
+            else:
+                raise NmstateDependencyError(
+                    "Open vSwitch NetworkManager support not installed "
+                    "and started"
+                )
+        elif iface_type == InterfaceType.TEAM and not has_team_capability:
             raise NmstateDependencyError(
                 "NetworkManager-team plugin not installed and started"
             )
--- a/tests/integration/ovs_test.py
+++ b/tests/integration/ovs_test.py
@@ -34,6 +34,7 @@ from .testlib import statelib
 from .testlib.nmplugin import disable_nm_plugin
 from .testlib.ovslib import Bridge
 from .testlib.ovslib import get_proxy_port_name_of_ovs_interface
+from .testlib.servicelib import disable_service
 from .testlib.vlan import vlan_interface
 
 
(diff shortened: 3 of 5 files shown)


libnmstate/netapplier.py
```diff
@@ -62,6 +62,7 @@ def apply(desired_state, verify_change=True, commit=True, rollback_timeout=60):
     desired_state = copy.deepcopy(desired_state)
     validator.validate(desired_state)
     validator.validate_capabilities(desired_state, netinfo.capabilities())
+    validator.validate_services(desired_state)
```
cathay4t: The code should merge into `validator.validate_capabilities()` as that function is checking the OVS capability and this PR is expand that capability check.

libnmstate/nm/ovs.py
```diff
@@ -50,6 +51,14 @@ def has_ovs_capability():
     return nmclient.NM.Capability.OVS in nm_client.get_capabilities()
 
 
+def is_ovs_running():
+    proc = subprocess.run(
+        ("systemctl", "status", "openvswitch"), stdout=subprocess.PIPE,
+    )
```
cathay4t: The `subprocess.PIPE` is incorrect. Should use [`DEVNULL`][1] instead.

Meanwhile, as a library, we should not hang on this command line call, please use `timeout=SYSTEMCTL_TIMEOUT_SECONDS`.

with `check=True`, you can simply put it like:

```python
try:
    subprocess.run(
        cmds,
        stdout=subprocess.DEVNULL,
        stderr=subprocess.DEVNULL,
        check=True,
        timeout=SYSTEMCTL_TIMEOUT_SECONDS,
    )
    return True
except Exception:
    return False
```

[1]: https://docs.python.org/3.4/library/subprocess.html#subprocess.DEVNULL

libnmstate/validator.py
```diff
@@ -94,6 +94,23 @@ def validate_interface_capabilities(ifaces_state, capabilities):
             )
 
 
+def validate_services(state):
+    validate_interface_services(state.get(Interface.KEY, []))
```
cathay4t: Why create alias of a function?

tests/integration/ovs_test.py
```diff
@@ -193,6 +194,22 @@ def test_nm_ovs_plugin_missing():
...
+                {
+                    Interface.KEY: [
+                        {
+                            Interface.NAME: BRIDGE1,
+                            Interface.TYPE: InterfaceType.OVS_BRIDGE,
+                            Interface.STATE: InterfaceState.UP,
+                        }
+                    ]
+                }
+            )
```
cathay4t: If some how `libnmstate.apply()` succeeded(for example the `disable_service()` failed to disable the service), you have a leftover OVS bridge. Please remove it also.

tests/integration/testlib/servicelib.py
```diff
@@ -0,0 +1,33 @@
...
+
+from contextlib import contextmanager
+import time
+
+from . import cmdlib
+
+
+@contextmanager
+def disable_service(service):
+    cmdlib.exec_cmd(("systemctl", "stop", service))
```
cathay4t: Both start and stop command should be protected by `check=True`.

tests/integration/testlib/servicelib.py
```diff
@@ -0,0 +1,33 @@
...
+    try:
+        yield
+    finally:
+        cmdlib.exec_cmd(("systemctl", "start", service))
+        time.sleep(1)
```
cathay4t: CI system might be slow. Please use a retry on `systemctl status openvswitch` return 0 with a timeout protected.

cathay4t: I have some code at PR #967 :

```python

def retry_till_true_or_timeout(timeout, func, *args, **kwargs):
    ret = func(*args, **kwargs)
    while timeout > 0:
        if ret:
            break
        time.sleep(1)
        timeout -= 1
        ret = func(*args, **kwargs)
    return ret
```

libnmstate/netapplier.py

libnmstate/nm/ovs.py
ffmancera: Done, thanks!

libnmstate/validator.py
ffmancera: Removed.

tests/integration/ovs_test.py
ffmancera: I am removing it after at the end. Thanks!

tests/integration/testlib/servicelib.py
ffmancera: Done!

tests/integration/testlib/servicelib.py
ffmancera: You are right. I fixed it. Thanks for the code example.

libnmstate/nm/ovs.py
```diff
@@ -50,6 +53,20 @@ def has_ovs_capability():
     return nmclient.NM.Capability.OVS in nm_client.get_capabilities()
 
 
+def is_ovs_running():
```
cathay4t: This function does not related to NetworkManager, please place it in `appliers/ovs.py` initially(I will find better place for it later).

libnmstate/nm/ovs.py
```diff
@@ -50,6 +53,20 @@ def has_ovs_capability():
...
+    try:
+        subprocess.run(
+            ("systemctl", "status", "openvswitch"),
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
+            check=True,
+            timeout=SYSTEMCTL_TIMEOUT_SECONDS,
+        )
+        return True
+    except subprocess.CalledProcessError:
```
cathay4t: You have to handle the timeout exception also. I would suggest simple `try Exception`.

libnmstate/netinfo.py
```diff
@@ -81,6 +81,9 @@ def capabilities():
     if nm.ovs.has_ovs_capability():
         caps.add(nm.ovs.CAPABILITY)
 
+    if nm.ovs.is_ovs_running():
+        caps.add(nm.ovs.SERVICE)
```
cathay4t: I don't think is required. The `nm.ovs.CAPABILITY` is enough, it's user's job to make sure both daemon and plugin installed. Our job is raise proper error guiding them to do so.

ffmancera: I am doing this in order to raise different error messages. IMO, we should tell the user what is missing. What do you think? @tyll Thanks!

libnmstate/nm/ovs.py
ffmancera: Done, thanks!

libnmstate/nm/ovs.py
ffmancera: Done, thanks! Maybe we could create something like `libnmstate/utils/ovs.py` and do so for other interface types.  What do you think? I move it to `appliers/ovs_bridge.py`. Thanks!

libnmstate/netinfo.py
cathay4t: I think `nm.ovs.CAPABILITY` indicate OVS is functional, or you need to rename it to `nm.ovs.NM_PLUGIN_READY`. For proper error message, you can check `is_ovs_running()` before raise exception.

ffmancera: I see. Will do, thanks.

ffmancera: Changes done, thanks!

PR #966, opened 2020-03-17
tests, integ: split tier1 integration tests

ffmancera:
This patch is introducing integ_tier1 tests and using it in travis CI.
This will allow us to run integration test quickly as both jobs, integ
and integ_tier1  are going to run at the same time.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -44,13 +44,23 @@ def pytest_addoption(parser):
 
 
 def pytest_collection_modifyitems(config, items):
-    if config.getoption("--runslow"):
-        # --runslow is given in cli: do not skip slow tests
-        return
+    if not config.getoption("--runslow"):
+        # --runslow is not in cli: skip slow tests
+        _mark_skip_slow_tests(items)
+    _mark_tier2_tests(items)
+
+
+def _mark_skip_slow_tests(items):
     skip_slow = pytest.mark.skip(reason="need --runslow option to run")
     for item in items:
         if "slow" in item.keywords:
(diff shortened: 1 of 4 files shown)


ffmancera:
> You are removing tier1 test cases from `runtest.sh --test-type integ` which is incorrect.
> 
> The `runtest.sh --test-type integ` should run all(except slow) integration test cases.
> 
> I was expecting `runtest.sh --test-type integ` == tier1 + tier2.
> 
> The travis just use `integ_tier1` and `integ_tier2`.

Done. I have introduced `integ_tier1` and `integ_tier2`. Travis CI is using integ_tier1 and integ_tier2. You can run tier1+tier2 tests using `integ`. Thanks!

tests/integration/conftest.py
```diff
@@ -38,18 +38,53 @@ def pytest_addoption(parser):
     parser.addoption(
         "--runslow", action="store_true", default=False, help="run slow tests"
     )
+    parser.addoption(
+        "--runtier1",
```
cathay4t: This is not required.

`pytest -m tier1` could help.

You just add tier2 marker to test case which does not tier1 marker.

cathay4t: Something like:

```python
def pytest_collection_modifyitems((config, items):
    for item in items:
        # Set test as tier 2 if not tier assigned.
        if (
            "tier1" not in item.keywords
        ):
            item.add_marker(pytest.mark.tier2)

```

ffmancera: Done, thanks!

ffmancera:
Done, thanks!

PR #977, opened 2020-03-20
nm.team: Fixed 'hwaddr' should be not included in team option

UnversedBlood:
Modified team.py to pop the hardware address from the Json.
Also made an integration test to test the changes in team_test.py.
Ref: https://bugzilla.redhat.com/1808990

Signed-off-by: Víctor Javier Díaz Garrido <vicdigar@hotmail.com>

### DIFF
--- a/libnmstate/nm/team.py
+++ b/libnmstate/nm/team.py
@@ -28,6 +28,7 @@ from libnmstate.schema import Team
 CAPABILITY = "team"
 TEAMD_JSON_DEVICE = "device"
 TEAMD_JSON_PORTS = "ports"
+TEAMD_JSON_HWADDR = "hwaddr"
 
 
 def has_team_capability():
@@ -100,6 +101,7 @@ def get_info(device):
 
 def _convert_teamd_config_to_nmstate_config(teamd_config, slave_names):
     teamd_config.pop(TEAMD_JSON_DEVICE, None)
+    teamd_config.pop(TEAMD_JSON_HWADDR, None)
     port_config = teamd_config.get(TEAMD_JSON_PORTS, {})
     team_port = _merge_port_config_with_slaves_info(port_config, slave_names)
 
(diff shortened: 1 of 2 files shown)


UnversedBlood:
> 
> 
> Please, could you add `Ref: https://bugzilla.redhat.com/1808990` in the commit message? Thanks!

Sure, thanks!

libnmstate/nm/team.py
```diff
@@ -100,6 +101,7 @@ def get_info(device):
 
 def _convert_teamd_config_to_nmstate_config(teamd_config, slave_names):
     teamd_config.pop(TEAMD_JSON_DEVICE, None)
+    teamd_config.pop(TEAMD_JSON_HWADDR, None)
```
smyle10: Will this remove 'hwaddr' in any cases? If so, I think now we want to remove 'hwaddr' from the output only when it is not explicitly set by user.  If people set it manually, it should be printed.

ffmancera: This will remove the `hwaddr` always. This value is from teamd and is set automatically. In our case, it is useless because we are managing the MAC address in the interface state and not in team configuration. Please, note that we are doing the same with `device` as we do not want to show it because we are already managing the interface name without teamd.

IMHO, an user should not use those parameters and they should use the parameters provided by our API. Thanks.

UnversedBlood:
> I am expecting a mechanism not a case by case fix.
> 
> With curent fix, when `teamd` add new extra properties in the future, we have to deal with them again.
> 
> So, please find a way to solve this problem once for all. Investigating why this properties got included in `libnmstate.show()` could be a good start.

Ok i will try to solve the problem as you want, thanks for the tip!

ffmancera:
@cathay4t is going to work on that as it could be too complex for a good first issue. I will try to find you another task. Thanks for your work.

PR #1011, opened 2020-04-01
[nmstate-0.2] Backport vlan filtering

ffmancera:
This must be approved by @maiqueb

### DIFF
--- a/tests/lib/state_test.py
+++ b/tests/lib/state_test.py
@@ -306,6 +306,21 @@ class TestAssertIfaceState:
 
         desired_state.verify_interfaces(current_state)
 
+    def test_initialize_subtree(self):
+        current_state = self._base_state
+        desired_state = state.State(
+            {
+                Interface.KEY: [
+                    {Interface.NAME: "foo-name", OVSBridge.CONFIG_SUBTREE: {}}
+                ]
+            }
+        )
+
+        desired_state.merge_interfaces(current_state)
+
+        desired_iface_state = desired_state.interfaces["foo-name"]
+        assert desired_iface_state[OVSBridge.CONFIG_SUBTREE] == {}
+
     @property
     def _base_state(self):
         return state.State(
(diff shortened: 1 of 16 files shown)


tests/lib/validator_test.py
```diff
@@ -640,6 +640,7 @@ def test_mac_restriction_in_current_mac_in_desire(self):
         iface_state[Interface.MAC] = MAC0
         bond_config = iface_state[Bond.CONFIG_SUBTREE]
         bond_config.pop(Bond.MODE)
+        bond_config.pop(Bond.OPTIONS_SUBTREE)
```
cathay4t, line 643: Why this is included?

tests/lib/state_test.py
```diff
@@ -1293,3 +1295,265 @@ def _create_desired_state():
                 ]
             }
         )
+
+
+class TestLinuxBridgeState:
```
maiqueb, line 1300: shouldn't this test be skipped / xfailed until the implementation is added ?

ffmancera: I did `cherry-pick` on all the patches except for `linux bridge, vlan filtering: activate port vlan feature`. AFAICS, these tests are not testing any kind of NM implementation so they are not failing without the implementation.

Indeed, they were not marked as skipped/xfailed originally so.. I don't think so.

tests/lib/validator_test.py
ffmancera: This is part of `state: An empty dict value is treated as an overwrite`. The patch is needed because the unit tests are using the empty dict as an overwrite. I could remove the change from this patch but I would prefer a complete backport rather than a modified backport.

Thanks!

tests/lib/state_test.py
maiqueb: touchè :)

ffmancera:
Rebased, thanks!

PR #1014, opened 2020-04-03
nm.active_connection: wait until NM.ActiveConnection is deleted

ffmancera:
If any interface is marked as absent, then nmstate should wait until the
NM.ActiveConnection is deleted. This allow nmstate to do not check if
the profile exists when getting the current state eliminating all the
race conditions at that point.

### DIFF
--- a/libnmstate/nm/active_connection.py
+++ b/libnmstate/nm/active_connection.py
@@ -26,6 +26,9 @@ from .nmclient import GLib
 from .nmclient import NM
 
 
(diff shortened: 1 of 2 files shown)


ffmancera:
I've tested this manually and it is working :-)

ffmancera:
There is a better way to do that using the signals. Will update the PR soon. Thanks!

ffmancera:
This is working well.

PR #1029, opened 2020-04-17
sriov: normalize vf mac address before verification

ffmancera:
The VF mac address is being read from the sysfs, it is always in
lowercase. In order to avoid verification failures, nmstate is setting
the VF mac to lowercase in the desired state before verifying.

Ref: https://bugzilla.redhat.com/1818750

### DIFF
--- a/libnmstate/state.py
+++ b/libnmstate/state.py
@@ -32,6 +32,7 @@ from libnmstate import metadata
 from libnmstate.appliers import bond
 from libnmstate.appliers import linux_bridge
 from libnmstate.appliers import ovs_bridge
+from libnmstate.appliers import sriov
 from libnmstate.appliers import team
 from libnmstate.error import NmstateNotImplementedError
 from libnmstate.error import NmstateValueError
@@ -746,10 +747,17 @@ class State:
                 )
 
     def _capitalize_mac(self):
+        """
+        The MAC address is upper case but case insensitive. This is
+        capitalizing the MAC address. In addition, this is unifying the output
+        format MAC for SR-IOV VFs.
+        """
         for ifstate in self.interfaces.values():
             mac = ifstate.get(Interface.MAC)
             if mac:
                 ifstate[Interface.MAC] = mac.upper()
+            if ifstate[Interface.TYPE] == InterfaceType.ETHERNET:
+                sriov.capitalize_vf_mac(ifstate)
 
     def _remove_empty_description(self):
         for ifstate in self.interfaces.values():
--- a/tests/integration/sriov_test.py
+++ b/tests/integration/sriov_test.py
@@ -109,6 +109,17 @@ def test_sriov_remove_vf_config(sriov_iface_vf):
     assertlib.assert_state(sriov_interface)
 
 
+@pytest.mark.xfail(
+    raises=NmstateNotSupportedError,
+    reason="The device does not support SR-IOV.",
+)
+def test_sriov_vf_mac_mixed_case(sriov_iface_vf):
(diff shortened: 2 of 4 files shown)


libnmstate/state.py
```diff
@@ -812,6 +813,28 @@ def _normalize_linux_bridge_port_vlan(self):
                 if not port.get(LinuxBridge.Port.VLAN_SUBTREE):
                     port[LinuxBridge.Port.VLAN_SUBTREE] = {}
 
+    def _normalize_sriov_vf_mac(self):
+        """
+        SR-IOV properties are being read from the sysfs. The MAC address is
+        always being stored in lowercase, in order to avoid verification errors
+        the MAC address should be always in lowercase.
```
cathay4t, line 820: In document we said MAC address is upper case but case insensitive.

I think to unify the output format of MAC address. Please make all MAC address upper case.

ffmancera:
This has been tested with SR-IOV HW.

libnmstate/state.py
ffmancera: Done, thanks!

libnmstate/state.py
```diff
@@ -745,10 +745,25 @@ def _sort_ip_addresses(self):
...
+            if ifstate[Interface.TYPE] == InterfaceType.ETHERNET:
+                vfs = (
+                    ifstate.get(Ethernet.CONFIG_SUBTREE, {})
+                    .get(Ethernet.SRIOV_SUBTREE, {})
+                    .get(Ethernet.SRIOV.VFS_SUBTREE, [])
+                )
+                for vf in vfs:
+                    vf_mac = vf.get(Ethernet.SRIOV.VFS.MAC_ADDRESS)
+                    if mac:
+                        vf[Ethernet.SRIOV.VFS.MAC_ADDRESS] = vf_mac.upper()
```
cathay4t, line 766: Please move these lines to a function in `appliers/sriov.py`, no need to make the bloated `state.py` bigger.

tests/integration/sriov_test.py
```diff
@@ -109,6 +109,17 @@ def test_sriov_remove_vf_config(sriov_iface_vf):
     assertlib.assert_state(sriov_interface)
 
 
+@pytest.mark.xfail(
```
cathay4t, line 112: The xfail has a `condition`, I am wondering whether that could be used to check whether `eth1` support SRIOV or not via sysfs. Not related to this PR.

tests/integration/sriov_test.py
```diff
@@ -109,6 +109,17 @@ def test_sriov_remove_vf_config(sriov_iface_vf):
...
+    raises=NmstateNotSupportedError,
+    reason="The device does not support SR-IOV.",
+)
+def test_sriov_vf_mac_uppercase(sriov_iface_vf):
```
cathay4t, line 116: `test_sriov_vf_mac_mixed_case`

tests/integration/sriov_test.py
ffmancera: If I am not wrong, there was a command to check if an interface supports SR-IOV. Maybe we could use it and check the result. Usinf skipif instead of xfail will improve the CI time a bit when SR-IOV is not supported.

libnmstate/state.py
ffmancera: Done, thanks!

tests/integration/sriov_test.py
ffmancera: Done, thanks!

ffmancera:
I've tested this with SR-IOV hardware.

PR #1034, opened 2020-04-20
travis: switch to docker as podman 1.9.0 is broken in Ubuntu 

ffmancera:
We are installing podman from the opensuse repository [1], 16th April
they have updated the podman package from podman 1.8.2 to podman 1.9.0
which is broken. As the repository does not contain podman
1.8.2 anymore, we are switching to docker again temporarily.

[1] http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_18.04/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ language: generic
 env:
     global:
         - use_coveralls=true
+        - CONTAINER_CMD=docker
     matrix:
         - CONTAINER_IMAGE=docker.io/nmstate/centos8-nmstate-dev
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -6,53 +6,56 @@ env:
         - use_coveralls=true
     matrix:
         - CONTAINER_IMAGE=docker.io/nmstate/centos8-nmstate-dev
+          CONTAINER_CMD=docker
```
cathay4t, line 9: Please place this variable in `global` section.

ffmancera: Done, thanks!

Issue #904, opened 2020-03-06
Support run integration on real hardwares.


cathay4t:
With commands:

```
NIC1=enp8s0
NIC2=enp9s0

sudo ip link set $NIC1 down
sudo ip link set $NIC1 name eth1
sudo ip link set $NIC1 up

sudo ip link set $NIC2 down
sudo ip link set $NIC2 name eth2
sudo ip link set $NIC2 up
```

We can run integration test cases on real hardware.

I intend to use this issue to track the support of 
`TEST_NIC1=enp8s0 TEST_NIC2=enp9s0 ./runtest.sh`

cathay4t:
I tried on e1000e(qemu).

Got some failure regarding the link status change for reapply which is caused by e1000e reset link when MTU changes.

So I should also update CI test cases which might impacted by NIC driver by using veth created at the test time.

PR #1053, opened 2020-04-28
automation: allow to run the tests out of the container

ffmancera:
This PR needs some more work but I would like to get some feedback. It is working on my Fedora 32 VM but some tests are failing. I need to investigate if it is related to something missing in the script or anything else.

In addition, I need to add documentation for the run-test-system.sh script.

Thanks!

### DIFF
--- a/automation/run-tests.sh
+++ b/automation/run-tests.sh
@@ -20,6 +20,9 @@ TEST_TYPE_INTEG_SLOW="integ_slow"
 FEDORA_IMAGE_DEV="docker.io/nmstate/fedora-nmstate-dev"
 CENTOS_IMAGE_DEV="docker.io/nmstate/centos8-nmstate-dev"
 
+CREATED_INTERFACES=""
+INTERFACES="eth1 eth2"
+
 PYTEST_OPTIONS="--verbose --verbose \
         --log-level=DEBUG \
         --log-date-format='%Y-%m-%d %H:%M:%S' \
@@ -36,39 +39,26 @@ NMSTATE_TEMPDIR=$(mktemp -d /tmp/nmstate-test-XXXX)
 : ${CONTAINER_CMD:=podman}
 
 test -t 1 && USE_TTY="-t"
-
-function remove_container {
-    res=$?
-    [ "$res" -ne 0 ] && echo "*** ERROR: $res"
-    container_exec 'rm -rf $CONTAINER_WORKSPACE/*nmstate*.rpm' || true
-    ${CONTAINER_CMD} rm -f $CONTAINER_ID
-}
+source automation/tests-container-utils.sh
+source automation/tests-machine-utils.sh
 
 function pyclean {
+    exec_cmd '
         find . -type f -name "*.py[co]" -delete
         find . -type d -name "__pycache__" -delete
-}
-
-function container_exec {
-    ${CONTAINER_CMD} exec $USE_TTY -i $CONTAINER_ID \
-        /bin/bash -c "cd $CONTAINER_WORKSPACE && $1"
-}
-
-function add_extra_networks {
-    container_exec '
-      ip link add eth1 type veth peer eth1peer && \
-      ip link add eth2 type veth peer eth2peer && \
-      ip link set eth1peer up && \
-      ip link set eth2peer up
-      # Due to https://nmstate.atlassian.net/browse/NMSTATE-279
-      # Mandually set test NICs as managed by NetworkManager.
-      nmcli device set eth1 managed yes
-      nmcli device set eth2 managed yes
     '
 }
 
+function exec_cmd {
+    if [ -z ${RUN_BAREMETAL} ];then
+        container_exec "$1"
+    else
+        bash -c "$1"
+    fi
+}
+
 function dump_network_info {
-    container_exec '
+    exec_cmd '
       nmcli dev; \
       # Use empty PAGER variable to stop nmcli send output to less which hang \
       # the CI. \
@@ -81,20 +71,22 @@ function dump_network_info {
 }
 
 function install_nmstate {
-    container_exec '
-      rpm -ivh `./packaging/make_rpm.sh|tail -1 || exit 1`
-    '
+    if [ $INSTALL_NMSTATE == "true" ];then
+        exec_cmd '
+          rpm -ivh `./packaging/make_rpm.sh|tail -1 || exit 1`
+        '
+    fi
 }
 
 function run_tests {
     if [ $TEST_TYPE == $TEST_TYPE_ALL ] || \
        [ $TEST_TYPE == $TEST_TYPE_FORMAT ];then
-        container_exec "tox -e black"
+        exec_cmd "tox -e black"
     fi
 
     if [ $TEST_TYPE == $TEST_TYPE_ALL ] || \
        [ $TEST_TYPE == $TEST_TYPE_LINT ];then
-        container_exec "tox -e flake8,pylint,yamllint"
+        exec_cmd "tox -e flake8,pylint,yamllint"
     fi
 
     if [ $TEST_TYPE == $TEST_TYPE_ALL ] || \
@@ -103,10 +95,10 @@ function run_tests {
             # Due to https://github.com/pypa/virtualenv/issues/1009
             # Instruct virtualenv not to upgrade to the latest versions of pip,
             # setuptools, wheel and etc
-            container_exec 'env VIRTUALENV_NO_DOWNLOAD=1 \
+            exec_cmd 'env VIRTUALENV_NO_DOWNLOAD=1 \
                             tox --sitepackages -e py36'
         else
-            container_exec 'tox -e py36'
+            exec_cmd "tox -e py36"
         fi
     fi
 
@@ -116,15 +108,15 @@ function run_tests {
             echo "Running unit test in $CONTAINER_IMAGE container is not " \
                  "support yet"
         else
-            container_exec 'tox -e py38'
+            exec_cmd "tox -e py38"
         fi
     fi
 
     if [ $TEST_TYPE == $TEST_TYPE_ALL ] || \
        [ $TEST_TYPE == $TEST_TYPE_INTEG ];then
-        container_exec "
-          cd $CONTAINER_WORKSPACE &&
-          pytest \
+        exec_cmd "cd $CONTAINER_WORKSPACE"
+        exec_cmd "
+            pytest \
             $PYTEST_OPTIONS \
             --cov-report=html:htmlcov-integ \
             tests/integration \
@@ -132,8 +124,8 @@ function run_tests {
     fi
 
     if  [ $TEST_TYPE == $TEST_TYPE_INTEG_TIER1 ];then
-        container_exec "
-          cd $CONTAINER_WORKSPACE &&
+        exec_cmd "cd $CONTAINER_WORKSPACE"
+        exec_cmd "
           pytest \
             $PYTEST_OPTIONS \
             --cov-report=html:htmlcov-integ_tier1 \
@@ -143,8 +135,8 @@ function run_tests {
     fi
 
     if  [ $TEST_TYPE == $TEST_TYPE_INTEG_TIER2 ];then
-        container_exec "
-          cd $CONTAINER_WORKSPACE &&
+        exec_cmd "cd $CONTAINER_WORKSPACE"
+        exec_cmd "
           pytest \
             $PYTEST_OPTIONS \
             --cov-report=html:htmlcov-integ_tier2 \
@@ -155,8 +147,8 @@ function run_tests {
 
     if [ $TEST_TYPE == $TEST_TYPE_ALL ] || \
        [ $TEST_TYPE == $TEST_TYPE_INTEG_SLOW ];then
-        container_exec "
-          cd $CONTAINER_WORKSPACE &&
+        exec_cmd "cd $CONTAINER_WORKSPACE"
+        exec_cmd "
           pytest \
             $PYTEST_OPTIONS \
             --cov-report=html:htmlcov-integ_slow \
@@ -164,15 +156,6 @@ function run_tests {
             tests/integration \
             ${nmstate_pytest_extra_args}"
     fi
-}
-
-function collect_artifacts {
-    container_exec "
-      journalctl > "$CONT_EXPORT_DIR/journal.log" && \
-      dmesg > "$CONT_EXPORT_DIR/dmesg.log" && \
-      cat /var/log/openvswitch/ovsdb-server.log > "$CONT_EXPORT_DIR/ovsdb-server.log" && \
-      mv "$CONTAINER_WORKSPACE/pytest-run.log" "$CONT_EXPORT_DIR/pytest-run.log" || true
-    "
 }
 
 function write_separator {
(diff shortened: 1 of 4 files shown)


automation/run-tests-system.sh
```diff
@@ -0,0 +1,168 @@
+#!/bin/bash -ex
+
+EXEC_PATH=$(dirname "$(realpath "$0")")
+PROJECT_PATH="$(dirname $EXEC_PATH)"
+
+TEST_TYPE_ALL="all"
```
tyll, line 6: In my opinion, there is too much duplication between this script and the container script. Maybe consider to make it one script that accepts a flag like `--test-environment bare-metal` which makes it do the preparation as done in this script. An alternative is to move the common code into a library.

automation/run-tests-system.sh
```diff
@@ -0,0 +1,168 @@
...
+    local wraplength=$(((cols - textlength) / 2))
+
+    eval printf %.1s "${char}"'{1..'"${wraplength}"\}
+    echo -n "${text}"
+    wraplength=$((wraplength + ((cols - textlength) % 2)))
+    eval printf %.1s "${char}"'{1..'"${wraplength}"\}
+    echo
+}
+
+function run_exit {
```
tyll, line 74: This does not seem to be called

automation/run-tests-system.sh
```diff
@@ -0,0 +1,168 @@
...
+modprobe_ovs
+
+prepare_network_environment
+
+(source /etc/os-release; echo $PRETTY_NAME); rpm -q NetworkManager
+
+pyclean
+run_tests
+
+teardown_network_environment
```
tyll, line 168: This should be called via trap to ensure it is also called when there is a failure.

automation/run-tests-system.sh
```diff
@@ -0,0 +1,168 @@
...
+    dump_network_info
+}
+
+function modprobe_ovs {
+    lsmod | grep -q ^openvswith || modprobe openvswitch || { echo 1>&2 "Please run 'modprobe openvswitch' as root" exit 1; }
+}
+
+function check_iface_exist {
+    set +e
+    ip a s | grep -q $1
```
tyll, line 85: The set setting should probably be restored after the call or not be touched.

automation/run-tests-system.sh
ffmancera: There is not that much code duplicated. They have a common structure but the container script is handling a lot of stuff related with the container (all the commands are being done with `container_exec`) and it would be quite complex to unify them.

In my opinion, they should be in different scripts to facilitate the code readability and the maintenance in the future. This script will allow more options so it will be even bigger.

What do you think? Thanks!

automation/run-tests-system.sh
ffmancera: Ooops!

automation/run-tests-system.sh
ffmancera: Yes, you are right. Thanks!

automation/run-tests-system.sh
ffmancera: I will move that out. The `set +e` and `set -e` is needed because if grep does not find anything it will stop the whole run and we do not want that.

automation/run-tests-system.sh
tyll: More than half of the lines from `run-tests-system.sh` are copied from `run-tests.sh`, so there is quite a lot duplication in that way. You are correct that most code from `run-tests.sh` is not duplicated, though.
The container_exec function actually makes it easier to unify them because you could rename them to something generic and depending on whether it is for bare-metal or container, it would wrap the command for the container command or just execute it directly, in my opinion.

automation/run-tests-system.sh
ffmancera: Fixed.

automation/run-tests-system.sh
ffmancera: Done, thanks!

automation/run-tests-system.sh
ffmancera: Done, thanks!

automation/tests-baremetal-utils.sh
```diff
@@ -0,0 +1,51 @@
+VETH_ETH1_CREATED="false"
+VETH_ETH2_CREATED="false"
+
+function check_iface_exist {
+    ip a s | grep -q $1
+}
+
+function prepare_baremetal_network {
```
cathay4t, line 8: `prepare_baremetal_network` should be use the same command with container did. Some thing like

```
fun setup_test_nics {
 if [ $NMSTATE_TEST_NIC1 ] and [$NMSTATE_TEST_NIC2]; then
   ip link rename $NMSTATE_TEST_NIC1 to eth1
   ip link rename $NMSTATE_TEST_NIC2 to eth2
else:
  create veth eth1 and eth2 and set them as managed.
 fi
}

exec_cmd setup_test_nics

File: automation/tests-baremetal-utils.sh [hunk 6]
```diff
@@ -0,0 +1,51 @@
...
+    if [ $VETH_ETH1_CREATED == "yes" ]; then
+        ip link del eth1
+    fi
+
+    if [ $VETH_ETH2_CREATED == "yes" ]; then
+        ip link del eth2
+    fi
+}
+
+function install_baremetal_dependencies {
```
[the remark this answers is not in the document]
cathay4t: And this stop us supporting no-dnf based linux system.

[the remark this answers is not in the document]
ffmancera: I agree, removed. Thanks!

automation/tests-baremetal-utils.sh
ffmancera: Done, they are using the same function now.

automation/tests-baremetal-utils.sh
```diff
@@ -0,0 +1,51 @@
...
+    dnf install -y NetworkManager-ovs NetworkManager-team
+    pip3 install pytest pytest-cov
+}
+
+function start_baremetal_services {
+    systemctl start openvswitch
+    systemctl restart NetworkManager
```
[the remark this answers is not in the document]
ffmancera: Done, thanks!

automation/tests-baremetal-utils.sh
```diff
@@ -0,0 +1,51 @@
+VETH_ETH1_CREATED="false"
+VETH_ETH2_CREATED="false"
+
+function check_iface_exist {
+    ip a s | grep -q $1
+}
+
+function prepare_baremetal_network {
+    echo 'ENV{ID_NET_DRIVER}=="veth", ENV{INTERFACE}=="eth[0-9]|eth[0-9]*[0-9]", ENV{NM_UNMANAGED}="0"' >/etc/udev/rules.d/88-veths.rules
```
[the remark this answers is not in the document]
ffmancera: Done, thanks!

automation/run-tests.sh
```diff
@@ -283,8 +215,40 @@ function upload_coverage {
     fi
 }
 
+function check_iface_exist {
+    exec_cmd "ip a s | grep -q $1"
+}
+
+function prepare_network_environment {
+    set +e
+    check_iface_exist 'eth1'
```
tyll, line 224: It would be nicer to use a loop here:

```
for device in eth1 eth2;
do
    if !check_iface_exist $device
         CREATED_DEVICES = "${CREATED_DEVICES} $device
...
```

This also allows to move "eth1 eth2" into a variable and simplifies changing them in the future.

automation/run-tests.sh
```diff
@@ -283,8 +215,40 @@ function upload_coverage {
...
+    if [ $? -eq 1 ]; then
+        VETH_ETH2_CREATED="yes"
+        exec_cmd 'ip link add eth2 type veth peer name eth2peer && ip link set eth2peer up'
+	exec_cmd 'nmcli device set eth2 managed yes'
+    fi
+    set -e
+}
+
+function teardown_network_environment {
+    if [ $VETH_ETH1_CREATED == "yes" ]; then
```
tyll, line 241: If you add the devices to a list (see above), then you can also use a loop here.

automation/tests-container-utils.sh
```diff
@@ -0,0 +1,147 @@
...
+    if is_file_changed "$PROJECT_PATH/packaging"; then
+        if [ $CONTAINER_IMAGE == $CENTOS_IMAGE_DEV ];then
+            ${PROJECT_PATH}/packaging/build-container.sh centos8-nmstate-dev
+        elif [ $CONTAINER_IMAGE == $FEDORA_IMAGE_DEV ];then
+            ${PROJECT_PATH}/packaging/build-container.sh fedora-nmstate-dev
+        fi
+    fi
+}
+
+function upgrade_nm_from_copr {
```
tyll, line 51: updating copr might also be interesting for bare-metal or virtual machines.

automation/run-tests.sh
ffmancera: I agree. Will do. Thanks!

automation/tests-container-utils.sh
ffmancera: Yes but then this will break compatibility with other systems than Fedora/Centos/RHEL. Could we split this for a future task?

automation/run-tests.sh
ffmancera: Done, thanks!

automation/run-tests.sh
ffmancera: Done, thanks!

automation/tests-container-utils.sh
tyll: Not sure what is breaking here because it requires a command-line switch to runs this code. If you think it is too complicated and you don't need to be able to update VMs, I am fine with delaying this.

ffmancera: Copr is not available in other distributions different than Fedora/CentOS/RHEL. @cathay4t wants the baremetal script to be compatible with all the distributions, AFAIK.

tyll: Since Copr is optional, the script is still compatible as long as the flag is not used.

cathay4t:
The CI failed with `automation/tests-machine-utils.sh: No such file or directory`

ffmancera:
> The CI failed with `automation/tests-machine-utils.sh: No such file or directory`

Oops. I didn't commit the file. Fixed. Thanks!

automation/run-tests.sh
```diff
@@ -36,39 +39,26 @@ NMSTATE_TEMPDIR=$(mktemp -d /tmp/nmstate-test-XXXX)
...
-      ip link add eth2 type veth peer eth2peer && \
-      ip link set eth1peer up && \
-      ip link set eth2peer up
-      # Due to https://nmstate.atlassian.net/browse/NMSTATE-279
-      # Mandually set test NICs as managed by NetworkManager.
-      nmcli device set eth1 managed yes
-      nmcli device set eth2 managed yes
-    '
+function exec_cmd {
+    if [ -z ${RUN_BAREMETAL+x} ];then
```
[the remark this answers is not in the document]
ffmancera: Fixed, thanks!

automation/run-tests.sh
```diff
@@ -283,8 +215,34 @@ function upload_coverage {
...
+function check_iface_exist {
+    exec_cmd "ip a s | grep -q $1"
+}
+
+function prepare_network_environment {
+    set +e
+    for device in $INTERFACES;
+    do
+        peer="${device}peer"
+        check_iface_exist $device
```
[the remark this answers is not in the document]
ffmancera: Yes, LLDP tests is going to use a dedicated veth.

automation/tests-container-utils.sh
ffmancera: Done, thanks!

ffmancera:
Rebased and ready for review. Thanks!

tyll:
Does this fix #904, too?

ffmancera:
> Does this fix #904, too?

Not sure. If the interfaces are named as "eth1" and "eth2" before running the tests, yes. If not, this is not going to work because the script is creating eth1 and eth2 veths.

tests/integration/conftest.py
```diff
@@ -116,6 +116,12 @@ def eth2_up():
         yield ifstate
 
 
+@pytest.fixture(scope="function")
+def lldptest_up():
```
cathay4t, line 120: Can you use fixture to crate veth pair for LLDP test in the run time like DHCP test did?

ffmancera: Done, LLDP tests is creating a veth pair in the runtime. Thanks!

automation/run-tests.sh
```diff
@@ -283,8 +215,52 @@ function upload_coverage {
     fi
 }
 
+function check_iface_exist {
+    exec_cmd "ip a s | grep -q $1"
```
cathay4t, line 219: `ip address` is not a good command for checking existence of NIC. Please use `ip link` instead.

automation/run-tests.sh
```diff
@@ -319,11 +295,17 @@ while true; do
         vdsm_tests
         exit
         ;;
+    --machine)
```
cathay4t, line 298: how about `--baremetal`?

automation/tests-container-utils.sh
```diff
@@ -0,0 +1,126 @@
...
+          container_exec "cp -rf $CONTAINER_WORKSPACE /root/nmstate-workspace || true"
+        else
+          container_exec "cp -rf $CONTAINER_WORKSPACE /root/nmstate-workspace"
+        fi
+        # Change workspace to keep the original one clean
+        CONTAINER_WORKSPACE="/root/nmstate-workspace"
+    fi
+}
+
+function pass_env_var {
```
cathay4t, line 105: This is not passing environment but actually start the container.

And you have two `docker run` in this script which is incorrect.

automation/tests-container-utils.sh
```diff
@@ -0,0 +1,126 @@
...
+function collect_artifacts {
+    container_exec "
+      journalctl > "$CONT_EXPORT_DIR/journal.log" && \
+      dmesg > "$CONT_EXPORT_DIR/dmesg.log" && \
+      cat /var/log/openvswitch/ovsdb-server.log > "$CONT_EXPORT_DIR/ovsdb-server.log" && \
+      mv "$CONTAINER_WORKSPACE/pytest-run.log" "$CONT_EXPORT_DIR/pytest-run.log" || true
+    "
+}
+
+function prepare_container_environment {
```
cathay4t, line 73: This is not prepare the container as the container is running now.

You may name them as:
 * `create_container`
 * `container_pre_test_setup`
 * `container_post_test_cleanup`

automation/run-tests.sh
ffmancera: According to @tyll, "baremetal" means a real machine and we are supporting running on a VM too. IMO, both are valid so I think you should discuss this.

automation/run-tests.sh
ffmancera: Done, thanks!

automation/tests-container-utils.sh
ffmancera: I agree about it is starting the container. But sorry, I am not able to find the `docker run` that you mentioned.

automation/tests-container-utils.sh
ffmancera: I've renamed them to:
* `create_container`
* `container_pre_test_setup`
But IMO, `container_post_test_cleanup` does not fit as we are removing the container after the tests. Please see `remove_container` function.

Thanks!

automation/tests-container-utils.sh
ffmancera: I've renamed the function to `create_container`. Thanks!

automation/run-tests.sh
cathay4t: Then machine is ok for me.

automation/tests-container-utils.sh
```diff
@@ -0,0 +1,126 @@
...
+
+function container_pre_test_setup {
+    ${CONTAINER_CMD} --version && cat /etc/resolv.conf
+
+    if [[ "$CI" == "true" ]];then
+        rebuild_container_images
+    fi
+
+    mkdir -p $EXPORT_DIR
+    CONTAINER_ID="$(${CONTAINER_CMD} run --privileged -d -e CI=$CI -v /sys/fs/cgroup:/sys/fs/cgroup:ro -v $PROJECT_PATH:$CONTAINER_WORKSPACE -v $EXPORT_DIR:$CONT_EXPORT_DIR $CONTAINER_IMAGE)"
```
cathay4t, line 81: Please break the long line.

ffmancera: I've removed that line as it is not needed. As you said, there were two `run` lines, this one is not needed. Thanks!

ffmancera:
Changes done, thanks! @cathay4t

PR #1055, opened 2020-04-30
Validate vlan mtu not greater than base iface

### LINKED ISSUE #1047: Need more clear error message on failure of changing vlan's mtu greater than the base interface
Downstream bug: https://bugzilla.redhat.com/show_bug.cgi?id=1788763

As this decorative, IMO, is low priority.

adwait-thattey:
fixes #1047 
A new validator is added that verifies that the
MTU of VLAN is less than or equal to MTU of base iface

Signed-off-by: Adwait Thattey <coderdude1999@gmail.com>

### DIFF
--- a/libnmstate/ifaces/ifaces.py
+++ b/libnmstate/ifaces/ifaces.py
@@ -123,6 +123,7 @@ class Ifaces:
 
     def _pre_edit_validation_and_cleanup(self):
         self._validate_over_booked_slaves()
+        self._validate_vlan_mtu()
         self._handle_master_slave_list_change()
         self._match_child_iface_state_with_parent()
         self._mark_orphen_as_absent()
@@ -163,6 +164,31 @@ class Ifaces:
                             f"OVS patch port peer {iface.peer} must be an OVS"
                             " patch port"
                         )
+
+    def _validate_vlan_mtu(self):
+        """
+        Validate that mtu of vlan or vxlan is less than
+        or equal to it's base interface's MTU
+
+        If base MTU is not present, set same as vlan MTU
+        """
+        for iface in self._ifaces.values():
+
+            if (
+                iface.type in [InterfaceType.VLAN, InterfaceType.VXLAN]
+                and iface.is_up
+                and iface.mtu
+            ):
+                base_iface = self._ifaces.get(iface.parent)
+                if not base_iface.mtu:
+                    base_iface.mtu = iface.mtu
+                if iface.mtu > base_iface.mtu:
+                    raise NmstateValueError(
+                        f"Interface {iface.name} has bigger "
+                        f"MTU({iface.mtu}) "
+                        f"than its base interface: {iface.parent} "
+                        f"MTU({base_iface.mtu})"
+                    )
 
     def _handle_master_slave_list_change(self):
         """
--- a/tests/lib/ifaces/vlan_iface_test.py
+++ b/tests/lib/ifaces/vlan_iface_test.py
@@ -20,7 +20,9 @@
 import pytest
 
 from libnmstate.error import NmstateValueError
+from libnmstate.ifaces import Ifaces
 from libnmstate.schema import VLAN
+from libnmstate.schema import Interface
 from libnmstate.schema import InterfaceType
 
 from libnmstate.ifaces.vlan import VlanIface
@@ -31,6 +33,13 @@ BASE_IFACE_NAME = "base1"
 
 
 class TestVlanIface:
+    def _gen_base_iface_info(self):
+        iface_info = gen_foo_iface_info(iface_type=InterfaceType.ETHERNET)
+        iface_info[Interface.NAME] = BASE_IFACE_NAME
+        iface_info[Interface.MTU] = 1500
+
+        return iface_info
+
     def _gen_iface_info(self):
         iface_info = gen_foo_iface_info(iface_type=InterfaceType.VLAN)
         iface_info[VLAN.CONFIG_SUBTREE] = {
@@ -58,3 +67,29 @@ class TestVlanIface:
         iface = VlanIface(iface_info)
         with pytest.raises(NmstateValueError):
             iface.pre_edit_validation_and_cleanup()
+
+    def test_add_vlan_with_mtu_greater_than_base(self):
+        base_iface_info = self._gen_base_iface_info()
+        vlan_iface_info = self._gen_iface_info()
+        vlan_iface_info[Interface.MTU] = base_iface_info[Interface.MTU] + 1
+
+        with pytest.raises(NmstateValueError):
+            Ifaces(
+                des_iface_infos=[base_iface_info, vlan_iface_info],
+                cur_iface_infos=[],
+            )
+
+    def test_add_vlan_with_base_mtu_undefined(self):
+        base_iface_info = self._gen_base_iface_info()
+        base_iface_info[Interface.MTU] = None
+        vlan_iface_info = self._gen_iface_info()
+        vlan_iface_info[Interface.MTU] = 1501
+
+        ifaces = Ifaces(
+            des_iface_infos=[base_iface_info, vlan_iface_info],
+            cur_iface_infos=[],
+        )
+
+        base_iface = ifaces.get(base_iface_info.get(Interface.NAME))
+
+        assert base_iface.mtu == vlan_iface_info[Interface.MTU]
--- a/tests/lib/ifaces/vxlan_iface_test.py
+++ b/tests/lib/ifaces/vxlan_iface_test.py
@@ -20,7 +20,9 @@
 import pytest
 
 from libnmstate.error import NmstateValueError
+from libnmstate.ifaces import Ifaces
 from libnmstate.schema import VXLAN
+from libnmstate.schema import Interface
 from libnmstate.schema import InterfaceType
 
 from libnmstate.ifaces.vxlan import VxlanIface
@@ -32,6 +34,13 @@ BASE_IFACE_NAME = "base1"
 
 
 class TestVxlanIface:
+    def _gen_base_iface_info(self):
+        iface_info = gen_foo_iface_info(iface_type=InterfaceType.ETHERNET)
+        iface_info[Interface.NAME] = BASE_IFACE_NAME
+        iface_info[Interface.MTU] = 1500
+
+        return iface_info
+
     def _gen_iface_info(self):
         iface_info = gen_foo_iface_info(iface_type=InterfaceType.VXLAN)
         iface_info[VXLAN.CONFIG_SUBTREE] = {
@@ -63,3 +72,29 @@ class TestVxlanIface:
         iface = VxlanIface(iface_info)
         with pytest.raises(NmstateValueError):
             iface.pre_edit_validation_and_cleanup()
+
+    def test_add_vxlan_with_mtu_greater_than_base(self):
+        base_iface_info = self._gen_base_iface_info()
+        vxlan_iface_info = self._gen_iface_info()
+        vxlan_iface_info[Interface.MTU] = base_iface_info[Interface.MTU] + 1
+
+        with pytest.raises(NmstateValueError):
+            Ifaces(
+                des_iface_infos=[base_iface_info, vxlan_iface_info],
+                cur_iface_infos=[],
+            )
+
+    def test_add_vxlan_with_base_mtu_undefined(self):
+        base_iface_info = self._gen_base_iface_info()
+        base_iface_info[Interface.MTU] = None
+        vxlan_iface_info = self._gen_iface_info()
+        vxlan_iface_info[Interface.MTU] = 1501
+
+        ifaces = Ifaces(
+            des_iface_infos=[base_iface_info, vxlan_iface_info],
+            cur_iface_infos=[],
+        )
+
+        base_iface = ifaces.get(base_iface_info.get(Interface.NAME))
+
+        assert base_iface.mtu == vxlan_iface_info[Interface.MTU]
--- a/libnmstate/ifaces/base_iface.py
+++ b/libnmstate/ifaces/base_iface.py
@@ -298,6 +298,10 @@ class BaseIface:
     def mtu(self):
         return self._info.get(Interface.MTU)
 
+    @mtu.setter
+    def mtu(self, value):
+        self._info[Interface.MTU] = value
+
     def _capitalize_mac(self):
         if self.mac:
             self._info[Interface.MAC] = self.mac.upper()


adwait-thattey:
@ffmancera Can u please check if this is what is required.
If yes, I will make changes to the tests (if required)

libnmstate/validator.py
```diff
@@ -400,6 +402,18 @@ def _validate_linux_bond(original_desired_state, current_state):
             _assert_bond_config_is_legal(iface_state, original_iface_state)
 
 
+def _validate_vlan(desired_state, current_state):
+    for iface_state in desired_state.interfaces.values():
+        if iface_state[Interface.TYPE] == InterfaceType.VLAN:
+            base_iface_name = iface_state.get(VLAN.CONFIG_SUBTREE).get(VLAN.BASE_IFACE)
+            base_iface_cur_state = current_state.interfaces.get(base_iface_name)
+            if base_iface_cur_state[Interface.TYPE] == InterfaceType.ETHERNET:
```
ffmancera, line 410: IMO, this check is not required. You can remove it, I think.

libnmstate/validator.py
```diff
@@ -400,6 +402,18 @@ def _validate_linux_bond(original_desired_state, current_state):
...
+                base_iface_mtu = base_iface_cur_state.get(Interface.MTU)
+                vlan_mtu = iface_state.get(Interface.MTU)
+                if vlan_mtu > base_iface_mtu:
+                    raise ValueError("MTU of VLAN can not be more than MTU of base Interface")
```
ffmancera, line 414: It would be nice if the error shows up the information, something like: "vlan1 mtu (number) cannot be greater than its base interface eth1 mtu (number)". It should be easy because you already have all the information stored in variables.

adwait-thattey:
Done. I am running tests locally. Will push after that

One doubt:
Is it okay to directly put the validation in `_validate_vlan` or should I create another method
`_assert_vlan_mtu_not_greater_than_base_iface` and call that from the `_validate_vlan`

because I find the latter approach used at multiple places in `validator.py`

adwait-thattey:
I had to make a few changes considering a few existing tests

1. Added condition to check only if VLAN is up. 
2. Changed an existing test that was setting MTU greater than the default value of base eth

After this, some tests are still failing. In these tests, the base and the VLAN are being created at the same time. So the base exist in the desired_state and not the current_state and the error is thrown.
How do you propose I handle this? 

Should I check for the base first in the desired state and then in the current state?
In such a situation, if the value of base MTU is not provided in desired_state, should I default it to 1500?

adwait-thattey:
squashed the commits

Also, I am thinking of adding a test that deliberately adds a VLAN with MTU greater than the base to check if an error is thrown

ffmancera:
> squashed the commits
> 
> Also, should I add a test that deliberately adds a VLAN with MTU greater than the base to check if an error is thrown?

Yes, you should do it.

adwait-thattey:
Updated with validator
I tried writing the test for this like this in [vlan_test.py](https://github.com/nmstate/nmstate/blob/master/tests/integration/vlan_test.py)

```python
def test_add_vlan_with_mtu_greater_than_base(eth1_up):
    desired_state = {
        Interface.KEY: [
            {
                Interface.NAME: VLAN_IFNAME,
                Interface.TYPE: InterfaceType.VLAN,
                VLAN.TYPE: {VLAN.ID: 101, VLAN.BASE_IFACE: eth1_up[Interface.KEY][0][Interface.NAME]},
                Interface.MTU: eth1_up[Interface.MTU] + 1
            }
        ]
    }

    with pytest.raises(ValueError):
```
The problem I am facing is that `eth1_up` does not seem to have a parameter "mtu". Thus it gives a KeyError. How do I get the mtu of eth1?

adwait-thattey:
Added test and Rebased

ffmancera:
Looking into this tomorrow. Thanks for your contribution! :-)

tests/integration/ethernet_mtu_test.py
```diff
@@ -158,7 +158,7 @@ def test_set_mtu_on_two_vlans_with_a_shared_base(eth1_up):
             ]
         }
         for iface_state in desired_state[Interface.KEY]:
-            iface_state[Interface.MTU] = 2000
+            iface_state[Interface.MTU] = 1500
```
ffmancera, line 161: Why are you modifying this test?

libnmstate/validator.py.bak
```diff
@@ -0,0 +1,498 @@
...
+    exceeds the limit specified in MAX_SUPPORTED_INTERFACES
+    """
+    num_of_interfaces = len(
+        [intface for intface in data.get(schema.Interface.KEY, ())]
+    )
+    if num_of_interfaces > MAX_SUPPORTED_INTERFACES:
+        logging.warning(
+            "Interfaces count exceeds the limit %s in desired state",
+            MAX_SUPPORTED_INTERFACES,
+        )
```
ffmancera, line 498: I think this file is wrong.

libnmstate/validator.py
```diff
@@ -357,6 +359,40 @@ def _assert_vlan_filtering_trunk_tag(trunk_tag_state):
...
+                )
+                if not base_iface_state:
+                    base_iface_state = current_state.interfaces.get(
+                        base_iface_name
+                    )
+                if not base_iface_state:
+                    raise ValueError(
+                        f"base interface {base_iface_name} for vlan \
+                        {iface_name} does not exist"
+                    )
```
ffmancera, line 385: This if statement could be inside of the one above, this way it improves readability.

libnmstate/validator.py
```diff
@@ -400,6 +436,12 @@ def _validate_linux_bond(original_desired_state, current_state):
             _assert_bond_config_is_legal(iface_state, original_iface_state)
 
 
+def _validate_vlan(original_desired_state, current_state):
+    _assert_vlan_mtu_not_greater_than_base_iface(
+        original_desired_state, current_state
+    )
+
+
```
ffmancera, line 444: In this case the function only contains another function. There is no need for this alias, IMO.

libnmstate/validator.py
```diff
@@ -357,6 +359,40 @@ def _assert_vlan_filtering_trunk_tag(trunk_tag_state):
...
+                    )
+                if not base_iface_state:
+                    raise ValueError(
+                        f"base interface {base_iface_name} for vlan \
+                        {iface_name} does not exist"
+                    )
+
+                base_iface_mtu = base_iface_state.get(Interface.MTU)
+                vlan_mtu = iface_state.get(Interface.MTU)
+                if vlan_mtu and base_iface_mtu and (vlan_mtu > base_iface_mtu):
```
ffmancera, line 389: I think `vlan_mtu` and `base_iface_mtu` cannot be `None` at this point. I suggest you to get them directly with `[]` and there is no need to check them here. Thanks!

libnmstate/validator.py
```diff
@@ -357,6 +359,40 @@ def _assert_vlan_filtering_trunk_tag(trunk_tag_state):
...
+):
+    # validate if vlan mtu is greater than base interface mtu
+    for iface_name, iface_state in original_desired_state.interfaces.items():
+        if iface_state.get(Interface.TYPE) == InterfaceType.VLAN:
+            if iface_state.get(Interface.STATE) == InterfaceState.UP:
+                base_iface_name = iface_state.get(VLAN.CONFIG_SUBTREE).get(
+                    VLAN.BASE_IFACE
+                )
+                # check if base_iface is in desired state.
+                # if not then get from current state
```
ffmancera, line 373: Please, remove the comments.

libnmstate/validator.py
adwait-thattey: In one (or maybe more) test cases they are coming as None
`test_create_vlan_over_a_bond_slave` is one of the tests failing if I don't check that.

Full past failure log: https://pastebin.com/Jf4nVPNB 
from https://travis-ci.com/github/nmstate/nmstate/jobs/336141093

adwait-thattey:
Made the changes

libnmstate/validator.py
ffmancera: Fair enough, thanks!

ffmancera:
Travis CI seems blocked, I am restarting it..

ffmancera:
For some reason, CI is not working on this PR. All the jobs are passing but it is not reporting back. Do you why? @tyll @cathay4t?

tyll:
> For some reason, CI is not working on this PR. All the jobs are passing but it is not reporting back. Do you know why? @tyll @cathay4t?

I guess it is a Travis problem, saw this with the network role, too. Best would be to rebase which needs to be done anyhow IMHO.

adwait-thattey:
@ffmancera  Rebased
But one of the builds is failing with 
```docker exec -t -i f4165a6c906e2d2bb56edca298ca15b689d44c45a762bd685d1517be538f9897 /bin/bash -c 'cd /workspace/nmstate && dnf clean all'
error: db5 error(17) from dbenv->open: File exists
error: cannot open Packages index using db5 - File exists (17)
error: cannot open Packages database in /var/lib/rpm
Error: Error: rpmdb open failed
```
Not sure what this is due to...

ffmancera:
We have made a lot of changes in the validator, could you rebase this an adapt it to the new code? The code looks good but it needs to be adapted to the new design. Thanks!

adwait-thattey:
@ffmancera rebased and adapted the code for the new design
Please review

libnmstate/ifaces/ifaces.py
```diff
@@ -165,6 +166,35 @@ def _validate_ovs_patch_peers(self):
                             " patch port"
                         )
 
+    def _validate_vlan_mtu_lt_base_mtu(self):
+        for iface in self._ifaces.values():
+            if iface.iface_type == InterfaceType.VLAN:
```
cathay4t, line 171: Should also handle `Vxlan`.

libnmstate/ifaces/ifaces.py
```diff
@@ -165,6 +166,35 @@ def _validate_ovs_patch_peers(self):
                             " patch port"
                         )
 
+    def _validate_vlan_mtu_lt_base_mtu(self):
+        for iface in self._ifaces.values():
```
cathay4t, line 170: Should check whether `iface.mtu` is valid first.

libnmstate/ifaces/ifaces.py
```diff
@@ -165,6 +166,35 @@ def _validate_ovs_patch_peers(self):
...
+                            raise ValueError(
+                                f"base interface {base_iface_name} for vlan \
+                                                {iface.name} does not exist"
+                            )
+
+                    base_iface_mtu = base_iface.mtu
+                    vlan_mtu = iface.mtu
+                    if (
+                        vlan_mtu
+                        and base_iface_mtu
```
cathay4t, line 189: When base_iface/parent interface does not have MTU defined while VLAN/VxLAN does, we should set parent interface MTU as vlan's.
But it could be done in another PR if you don't want to do it.

tests/integration/vlan_test.py
```diff
@@ -192,6 +192,26 @@ def test_add_new_base_iface_with_vlan():
         libnmstate.apply(desired_state, verify_change=False)
 
 
+def test_add_vlan_with_mtu_greater_than_base(eth1_up):
```
cathay4t, line 195: Also need unit test case to cover all the cases.

libnmstate/ifaces/ifaces.py
adwait-thattey: Should I raise NmStateValueError if mtu of vlan is not present? Or just skip the check then

libnmstate/ifaces/ifaces.py
adwait-thattey: I have added this in the validator itself. 
is that fine or should it be added somewhere else?

adwait-thattey:
@cathay4t 
If mtu for vlan doesn't exist, I am raising an exception. [Ref. ifaces/ifaces.py, Line 184](https://github.com/adwait-thattey/nmstate/blob/fix-issue-1047/libnmstate/ifaces/ifaces.py#L184)

This is causing some [other tests](https://travis-ci.com/github/nmstate/nmstate/jobs/351085970#L1246) to fail 

Instead of raising an exception should I just skip the validator if mtu for vlan doesn't exist?

libnmstate/ifaces/ifaces.py
```diff
@@ -165,6 +166,43 @@ def _validate_ovs_patch_peers(self):
...
+                            f"paramater mtu missing for vlan {iface.name}"
+                        )
+
+                    base_iface = self._ifaces.get(iface.parent)
+                    if not base_iface:
+                        raise NmstateValueError(
+                            f"base interface {iface.parent} for vlan \
+                                            {iface.name} does not exist"
+                        )
+
```
cathay4t, line 195: This is my second time to say it:


This is not required.
The `Ifaces._validate_unknown_parent()` already checked it. No need to do it again.

libnmstate/ifaces/ifaces.py
```diff
@@ -165,6 +166,43 @@ def _validate_ovs_patch_peers(self):
...
+                    if not base_iface.mtu:
+                        base_iface.mtu = iface.mtu
+
+                    if iface.mtu > base_iface.mtu:
+                        raise NmstateValueError(
+                            f"{iface.name} mtu {iface.mtu} cannot be \
```
cathay4t, line 201: The use of `\` is incorrect: by using it, you add a lot spaces.

You should use:

```python
raise NmstateValueError(
    f"Interface {iface.name} has bigger "
    f"MTU({iface.mtu}) "
    f"than its base interface: {iface.parent} "
    f"MTU({parent_iface.mtu})"
)
```

libnmstate/ifaces/ifaces.py
```diff
@@ -165,6 +166,43 @@ def _validate_ovs_patch_peers(self):
...
+                iface.iface_type == InterfaceType.VLAN
+                or iface.iface_type == InterfaceType.VXLAN
+            ):
+
+                if iface.state == InterfaceState.UP:
+
+                    if not iface.mtu:
+                        raise NmstateValueError(
+                            f"paramater mtu missing for vlan {iface.name}"
+                        )
```
cathay4t, line 187: Should not raise exception as MTU is OK to not defined.

libnmstate/ifaces/ifaces.py
```diff
@@ -165,6 +166,43 @@ def _validate_ovs_patch_peers(self):
...
+
+            If base MTU is not present, set same as vlan MTU
+        """
+        for iface in self._ifaces.values():
+            if (
+                iface.iface_type == InterfaceType.VLAN
+                or iface.iface_type == InterfaceType.VXLAN
+            ):
+
+                if iface.state == InterfaceState.UP:
```
cathay4t, line 182: Use `iface.is_up`.

cathay4t:
> @cathay4t
> If mtu for vlan doesn't exist, I am raising an exception. [Ref. ifaces/ifaces.py, Line 184](https://github.com/adwait-thattey/nmstate/blob/fix-issue-1047/libnmstate/ifaces/ifaces.py#L184)
> 
> This is causing some [other tests](https://travis-ci.com/github/nmstate/nmstate/jobs/351085970#L1246) to fail
> 
> Instead of raising an exception should I just skip the validator if mtu for vlan doesn't exist?

I will look into the failures once you fixed all my changes requests.

adwait-thattey:
@cathay4t made the changes. Tests are also passing

libnmstate/ifaces/ifaces.py
```diff
@@ -165,6 +166,32 @@ def _validate_ovs_patch_peers(self):
                             " patch port"
                         )
 
+    def _validate_vlan_mtu_lte_base_mtu(self):
+        """
+            Validate that mtu of vlan or vxlan is less than
+            or equal to it's base interface's MTU
+
+            If base MTU is not present, set same as vlan MTU
```
cathay4t, line 174: Not indent correctly.

libnmstate/ifaces/ifaces.py
```diff
@@ -165,6 +166,32 @@ def _validate_ovs_patch_peers(self):
                             " patch port"
                         )
 
+    def _validate_vlan_mtu_lte_base_mtu(self):
```
cathay4t, line 169: lte is not a common abbreviation. Use `_validate_vlan_mtu` is enough.

libnmstate/ifaces/ifaces.py
```diff
@@ -165,6 +166,32 @@ def _validate_ovs_patch_peers(self):
...
+            If base MTU is not present, set same as vlan MTU
+        """
+        for iface in self._ifaces.values():
+            if (
+                iface.iface_type == InterfaceType.VLAN
+                or iface.iface_type == InterfaceType.VXLAN
+            ):
+
+                if iface.is_up:
+                    if iface.mtu:
```
cathay4t, line 183: Please merge all these conditions into single `if`

tests/lib/ifaces/vlan_iface_test.py
```diff
@@ -20,7 +20,8 @@
 import pytest
 
 from libnmstate.error import NmstateValueError
-from libnmstate.schema import VLAN
+from libnmstate.ifaces import Ifaces
+from libnmstate.schema import VLAN, Interface
```
cathay4t, line 24: Current code style is import one thing per line. Please follow that.

tests/lib/ifaces/vxlan_iface_test.py
```diff
@@ -20,7 +20,8 @@
 import pytest
 
 from libnmstate.error import NmstateValueError
-from libnmstate.schema import VXLAN
+from libnmstate.ifaces import Ifaces
+from libnmstate.schema import VXLAN, Interface
```
cathay4t, line 24: Current code style is import one thing per line. Please follow that.

adwait-thattey:
@cathay4t Done

adwait-thattey:
also rebased

PR #1059, opened 2020-05-03
sriov: raise proper exception when SR-IOV parameter is not supported

ffmancera:
Some SR-IOV parameters could be not supported by the device driver.
Nmstate must return a proper error message instead of a generic one.

```
2020-05-03 12:35:19,728 root         ERROR    NM main-loop aborted:
The device=eno1 does not support one or more of the SR-IOV parameters
set.
```

Ref: https://bugzilla.redhat.com/1819588


ffmancera:
Rebased, thanks!

PR #64, opened 2018-08-28
schema: Add LLDP information

EdDev:

PR #1063, opened 2020-05-07
Add LLDP support

ffmancera:
This patch introduces LLDP support. Currently, integration tests are missing.


ffmancera:
I am trying to figure out how to add integration tests for this. I cannot even add unit tests using MagicMock because NM.LldpNeighbor object store the properties values in attributes but currently the object does not support setting attributes through libnm.

Any suggestion on this? Thanks!

ffmancera:
Yay! Tests are passing. This patch is ready for reviewing. :-)

libnmstate/nm/lldp.py
```diff
@@ -0,0 +1,334 @@
...
+    if lldp_status == -1 or not lldp_status:
+        info[LLDP.ENABLED] = False
+    else:
+        info[LLDP.ENABLED] = True
+        _get_neighbors_info(info, nmdev)
+
+    return {LLDP.CONFIG_SUBTREE: info}
+
+
+def _get_lldp_status(nm_client, nmdev):
```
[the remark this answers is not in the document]
ffmancera: This function is being used in `get_info()`.

libnmstate/nm/connection.py
```diff
@@ -437,10 +437,11 @@ def create(self, con_name, iface_name, iface_type):
         con_setting.props.autoconnect_slaves = (
             NM.SettingConnectionAutoconnectSlaves.YES
         )
+        con_setting.props.lldp = lldp
 
         self._setting = con_setting
 
-    def import_by_profile(self, con_profile):
+    def import_by_profile(self, con_profile, lldp=None):
```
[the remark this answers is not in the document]
ffmancera: Done. Thanks!

libnmstate/nm/lldp.py
```diff
@@ -0,0 +1,334 @@
...
+}
+
+
+def get_info(nm_client, nmdev):
+    """
+    Provides the current LLDP neighbors information
+    """
+    lldp_status = _get_lldp_status(nm_client, nmdev)
+    info = {}
+    if lldp_status == -1 or not lldp_status:
```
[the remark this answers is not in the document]
ffmancera: Default means NM global config file value which is by default disabled. According to Beniamino, there is no way from libnm to know if lldp is enable or not with libnm if the value in the profile is `default`. Therefore, the best option is to force the users to enable it explicitly. This is going to be solved by a property in the NM.Device object to know if the device is listening on LLDP. Please see: https://bugzilla.redhat.com/show_bug.cgi?id=1832273

libnmstate/nm/lldp.py
```diff
@@ -0,0 +1,334 @@
...
+    7       DOCSIS cable device
+    8       Station Only
+    9       C-VLAN
+    10      S-VLAN
+    11      TPMR
+    =============================
+    """
+    capabilities = []
+    for bit, capability in LLDP_BIT_TO_CAPABILITY.items():
+        bit = bin(code)[-bit]
```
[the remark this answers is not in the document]
ffmancera: This is not working. We need to read the bits in the reverse order. I will propose a better way this night. Thanks

libnmstate/nm/lldp.py
```diff
@@ -0,0 +1,334 @@
...
+    3       MAC bridge
+    4       WLAN AC
+    5       Router
+    6       Telephone
+    7       DOCSIS cable device
+    8       Station Only
+    9       C-VLAN
+    10      S-VLAN
+    11      TPMR
+    =============================
```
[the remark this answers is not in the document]
ffmancera: Removed, thanks.

libnmstate/nm/lldp.py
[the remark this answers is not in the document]
ffmancera:
```
def _decode_sys_caps(code):
    capabilities = []
    bits = bin(code)[len("0b"):]
    for i in range(1, len(bits)+1):
        try:
            if bits[-i] == '1':
                capabilities.append(LLDP_CAP_NAMES[i-1])
        except IndexError:
            pass

    return capabilities
```

libnmstate/schema.py
```diff
@@ -299,3 +299,86 @@ class Runner:
...
+
+        VLAN_SUBTREE = "vlan"
+
+        class VLAN:
+            ID = "id"
+            NAME = "name"
+
+        MAC_PHY_SUBTREE = "mac-phy"
+
+        class MACPHY:
```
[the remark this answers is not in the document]
ffmancera: Sure, done!

libnmstate/nm/lldp.py
```diff
@@ -0,0 +1,334 @@
...
+    3: LLDP.Neighbors.SYSTEM_CAPABILITY_MAC_BRIDGE,
+    4: LLDP.Neighbors.SYSTEM_CAPABILITY_WLAN_AC,
+    5: LLDP.Neighbors.SYSTEM_CAPABILITY_ROUTER,
+    6: LLDP.Neighbors.SYSTEM_CAPABILITY_TELEPHONE,
+    7: LLDP.Neighbors.SYSTEM_CAPABILITY_DOCSIS,
+    8: LLDP.Neighbors.SYSTEM_CAPABILITY_STATION,
+    9: LLDP.Neighbors.SYSTEM_CAPABILITY_CVLAN,
+    10: LLDP.Neighbors.SYSTEM_CAPABILITY_SVLAN,
+    11: LLDP.Neighbors.SYSTEM_CAPABILITY_TPMR,
+}
```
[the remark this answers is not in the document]
ffmancera: Done, thanks!

libnmstate/nm/lldp.py
```diff
@@ -0,0 +1,334 @@
...
+    if code == 2:
+        return "ifindex"
+    elif code == 3:
+        return "system-port"
+    else:
+        return "unknown"
+
+
+def _decode_management_address_type(code, address):
+    if code == 1:
```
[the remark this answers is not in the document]
ffmancera: Done, thank you!

libnmstate/nm/lldp.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+def _decode_sys_caps(code):
+    capabilities = []
+    prefix_length = len("0b")
+    bits = bin(code)[prefix_length:]
+    for i in range(1, len(bits) + 1):
+        try:
+            if bits[-i] == "1":
+                capabilities.append(LLDP_CAP_NAMES[i - 1])
```
tyll, line 161: In my opinion, it would be more robust and easier to understand if `LLDP_CAP_NAMES` contains the actual bitmasks like

```
LDDP_CAPABILITY_NAMES = {
0b1:     LLDP.Neighbors.SYSTEM_CAPABILITY_OTHER,
0b10:￼    LLDP.Neighbors.SYSTEM_CAPABILITY_REPEATER,
0b100:￼    LLDP.Neighbors.SYSTEM_CAPABILITY_MAC_BRIDGE,
0b1000:￼    LLDP.Neighbors.SYSTEM_CAPABILITY_WLAN_AC,
0b1_0000:￼    LLDP.Neighbors.SYSTEM_CAPABILITY_ROUTER,
...
}
```

and then use something like

```
capabilities = []
for mask, capability in LLDP_CAPABILITY_NAMES:
    if code & mask:
        capabilities.append(capability)
return capabilities
```

What do you think?

libnmstate/nm/lldp.py
```diff
@@ -0,0 +1,330 @@
...
+        port_info[LLDP.Neighbors.Port.TYPE] = _decode_port_type(
+            port_type.get_uint32()
+        )
+
+    if port_info:
+        info[LLDP.Neighbors.PORT_SUBTREE] = port_info
+
+
+def _decode_port_type(code):
+    return LLDP_PORT_TYPE_TO_NMSTATE.get(code, "unknown")
```
tyll, line 205: `"unknown"` should be a constant.

libnmstate/nm/lldp.py
```diff
@@ -0,0 +1,330 @@
...
+        chassis_info[LLDP.Neighbors.Chassis.TYPE] = _decode_chassis_type(
+            chassis_id_type.get_uint32()
+        )
+
+    if chassis_info:
+        info[LLDP.Neighbors.CHASSIS_SUBTREE] = chassis_info
+
+
+def _decode_chassis_type(code):
+    return LLDP_CHASSIS_TYPE_TO_NMSTATE.get(code, "unknown")
```
tyll, line 185: `"unknown"` should be a constant.

libnmstate/nm/lldp.py
ffmancera: Hm. That is a nice option. I agree with this. Thanks!

libnmstate/nm/lldp.py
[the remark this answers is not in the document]
cathay4t: @ffmancera You still don't use constant to store the magic number -1. And also please put above comments in code.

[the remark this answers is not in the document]
ffmancera: Opps, missed the -1. I am sorry.

libnmstate/nm/lldp.py
```diff
@@ -0,0 +1,330 @@
...
+    LLDP.Neighbors.SYSTEM_CAPABILITY_TELEPHONE,
+    LLDP.Neighbors.SYSTEM_CAPABILITY_DOCSIS,
+    LLDP.Neighbors.SYSTEM_CAPABILITY_STATION,
+    LLDP.Neighbors.SYSTEM_CAPABILITY_CVLAN,
+    LLDP.Neighbors.SYSTEM_CAPABILITY_SVLAN,
+    LLDP.Neighbors.SYSTEM_CAPABILITY_TPMR,
+]
+
+
+LLDP_CHASSIS_TYPE_TO_NMSTATE = {
```
cathay4t, line 57: You may put 0 as `reserved` as spec file said and use list as above.

libnmstate/nm/lldp.py
```diff
@@ -0,0 +1,330 @@
...
+    3: LLDP.Neighbors.Chassis.TYPE_PORT,
+    4: LLDP.Neighbors.Chassis.TYPE_MAC,
+    5: LLDP.Neighbors.Chassis.TYPE_NETWORK_ADDRESS,
+    6: LLDP.Neighbors.Chassis.TYPE_INTERFACE_NAME,
+    7: LLDP.Neighbors.Chassis.TYPE_LOCAL,
+}
+
+
+LLDP_PORT_TYPE_TO_NMSTATE = {
+    1: LLDP.Neighbors.Port.TYPE_INTERFACE_ALIAS,
```
cathay4t, line 69: You may put 0 as reserved as spec file said and use list as above.

libnmstate/nm/lldp.py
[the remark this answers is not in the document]
ffmancera: Done, thanks!

libnmstate/nm/lldp.py
ffmancera: Done! Thanks :-)

libnmstate/nm/lldp.py
ffmancera: Thanks!

libnmstate/nm/lldp.py
ffmancera: Done, thanks!

libnmstate/nm/lldp.py
ffmancera: Done, thanks!

libnmstate/nm/lldp.py
ffmancera: Done, thanks!

packaging/Dockerfile.centos8-nmstate-dev
```diff
@@ -28,11 +28,14 @@ RUN dnf -y install dnf-plugins-core epel-release && \
...
                    python3-requests \
                    python3-docopt \
+                   lldpd \
+                   tcpreplay \
                    && \
     alternatives --set python /usr/bin/python3 && \
     ln -s /usr/bin/pytest-3 /usr/bin/pytest && \
     dnf clean all && \
     systemctl enable openvswitch && \
+    systemctl enable lldpd && \
```
tyll, line 38: Why is this service needed?

libnmstate/nm/lldp.py
```diff
@@ -0,0 +1,339 @@
...
+NM_INTERFACE_TYPE_SYSTEM_PORT = 3
+NM_LLDP_STATUS_DEFAULT = -1
+
+
+LLDP_CAP_NAMES = {
+    0b1: LLDP.Neighbors.SYSTEM_CAPABILITY_OTHER,
+    0b10: LLDP.Neighbors.SYSTEM_CAPABILITY_REPEATER,
+    0b100: LLDP.Neighbors.SYSTEM_CAPABILITY_MAC_BRIDGE,
+    0b1000: LLDP.Neighbors.SYSTEM_CAPABILITY_WLAN_AC,
+    0b10000: LLDP.Neighbors.SYSTEM_CAPABILITY_ROUTER,
```
tyll, line 48:
```suggestion
    0b1_0000: LLDP.Neighbors.SYSTEM_CAPABILITY_ROUTER,
```

By separating the nibbles, it is easier to read the value because counting the zeroes is not so nice.

packaging/Dockerfile.centos8-nmstate-dev
ffmancera: That service is needed in order to listen on lldp packets

libnmstate/nm/lldp.py
ffmancera: Sorry, I am confused by that. How does this work? Thanks!

ffmancera: Sorry I am confused by that, so do you propose to split the last bit from the other zeroes? Sorry if I am missing something, thanks!

tyll: Just add an underscore after 4 binary digits like `0b1_0000` or `0b10_0000` `0b1_0000_0000` it is easier to read than `0b100000000`. It is like adding commas in English numbers 1,000,000.

packaging/Dockerfile.centos8-nmstate-dev
tyll: It is not needed by NetworkManager, so something else seems to be the problem here.

libnmstate/nm/lldp.py
ffmancera: Thanks for explaining :-)

ffmancera: Done, thanks!

packaging/Dockerfile.centos8-nmstate-dev
ffmancera: I am investigating this. I have a system where this is working without lldpd, now I am trying to know why is not working on the container or my VM.

ffmancera: Indeed, it is working in VM and centos container without lldpd. I might get confused or something because I thought I tried this without lldpd and it was not working. Never mind, thanks for pointing this out :-)

packaging/Dockerfile.centos8-nmstate-dev
```diff
@@ -28,6 +28,7 @@ RUN dnf -y install dnf-plugins-core epel-release && \
                    python3-coveralls \
                    python3-requests \
                    python3-docopt \
+                   tcpreplay \
```
tyll, line 31: I guess the Fedora container spec needs to be updated, too. It is not in the CI anymore but still in the packaging directory.

ffmancera: Oops. You are right.

ffmancera: Done, thanks!

ffmancera:
> There is no changed on the verification stage.
> 
> When user applying the same yml from `show()` with LLDP status changed, there will be a verification failure.

Done, it should not fail on verification now.

ffmancera:
Rebased with the new metadata and merge state flow. Thanks!

libnmstate/ifaces/base_iface.py
```diff
@@ -376,6 +378,11 @@ def _remove_empty_description(state):
         del state[Interface.DESCRIPTION]
 
 
+def _remove_lldp_neighbors(state):
+    if state.get(LLDP.CONFIG_SUBTREE):
+        state[LLDP.CONFIG_SUBTREE][LLDP.NEIGHBORS_SUBTREE] = []
```
cathay4t, line 383: Should be `state[LLDP.CONFIG_SUBTREE].pop(LLDP.NEIGHBORS_SUBTREE, None)`

dominikholler:
Looks good and works for me, great work!
For me it even worked in combination with lldpad.
I did not check if it would work with OVS and FCoE, but I ensure that this is checked this in the long run.
For oVirt it would be helpful if the following TLVs would be added:

- TTL (tlv type 3), to show document how old the lldp information might be,
- Port Description (tlv type 4)
- VLAN Name (tlv tpe 0x7f, oui 0x0080C2  (similar to Port port-vlan), subtype 3)
- Link Aggregation ( oui 0x0080C2, subtype 7 )
- Link Aggregation ( oui 0x00120F, subtype 3 )
- MTU ( oui 0x00120F, subtype 4 )

Would be nice, if there would be a way to distinguish between the two Link Aggregation TLVs.

EdDev:
The suggested API mimics NM and lldpad outputs, giving informal naming for the data, assuming the end user is the one consuming it.
But even products like oVirt do not assume this, exposing the data in a formal unambiguous manner with the addition of human readable namings.

I think this is the main difference from the original spec https://github.com/nmstate/nmstate/pull/64 .

I would prefer to see a more formal output that applications may consume easily without the need to interpret nmstate naming.
But if you still prefer the current approach, please document the exact details of the data source (numeric type, subtypes and oid/s).

ffmancera:
> The suggested API mimics NM and lldpad outputs, giving informal naming for the data, assuming the end user is the one consuming it.
> But even products like oVirt do not assume this, exposing the data in a formal unambiguous manner with the addition of human readable namings.
> 
> I think this is the main difference from the original spec #64 .
> 
> I would prefer to see a more formal output that applications may consume easily without the need to interpret nmstate naming.
> But if you still prefer the current approach, please document the exact details of the data source (numeric type, subtypes and oid/s).

As we are not using adding everything on the schema I propose: 
1. Merge this so RHV and other users can test it and use it.
2. Implement missing features as @dominikholler requested, when possible from NM.
3. We can create a field like "formal" in the lldp subtree, if true then we can use formal names if False then we use the human readable.

What do you think?

ffmancera:
@cathay4t I've done all your requested changes. Thanks!

ffmancera:
According to the feedback received, I've implemented a new schema approach. Now lldp has a `neighbors` subtree that contains a list of neighbors. Each neighbor is represented by a list of TLVs. 

Each TLV is a dict containing the tlv-type, the tlv-subtype and the oid if needed. In addition, it contains the informal name and the value. The value could be a single value(string or integer) or a dict because some TLV contains different fields.

```
    interfaces:
    - name: eth1
      type: ethernet
      state: up
      ipv4:
        dhcp: false
        enabled: false
      ipv6:
        autoconf: false
        dhcp: false
        enabled: false
      lldp:
        enabled: true
        neighbors:
        - - name: system-name
            tlv-type: 5
            value: Summit300-48
          - name: system-description
            tlv-type: 6
            value: Summit300-48 - Version 7.4e.1 (Build 5) by Release_Master 05/27/05
              04:53:11
          - name: system-capabilities
            tlv-type: 7
            value:
            - mac-bridge
            - router
          - name: chassis-id
            tlv-type: 1
            value:
              chassis-id: 00:01:30:F9:AD:A0
              chassis-id-type: mac
          - name: port-id
            tlv-type: 2
            value:
              port-id: 1/1
              port-id-type: interface-name
          - name: ieee-802-1-vlans
            oid: 00:80:c2
            tlv-subtype: 1
            tlv-type: 127
            value:
            - name: v2-0488-03-0505
              vid: 488
          - name: ieee-802-3-mac-phy-conf
            oid: 00:12:0f
            tlv-subtype: 1
            tlv-type: 127
            value:
              autoneg: true
              operational-mau-type: 16
              pmd-autoneg-cap: 27648
          - name: ieee-802-1-ppvids
            oid: 00:80:c2
            tlv-subtype: 2
            tlv-type: 127
            value:
            - ppvid: 0
          - name: management-addresses
            tlv-type: 8
            value:
            - address: 00::01::30::F9::AD::A0
              address-subtype: ipv6
              interface-number: 1001
              interface-number-subtype: ifindex
          - name: ieee-802-3-max-frame-size
            oid: 00:12:0f
            tlv-subtype: 4
            tlv-type: 127
            value: 1522
      mac-address: 36:E6:3A:97:AF:C9
      mtu: 1500
```

I hope you like more this approach. Thanks for the feedback and reviews! @cathay4t @EdDev @tyll

tyll:
> ```
>     interfaces:
>     - name: eth1
>       type: ethernet
>       state: up
>       ipv4:
>         dhcp: false
>         enabled: false
>       ipv6:
>         autoconf: false
>         dhcp: false
>         enabled: false
>       lldp:
>         enabled: true
>         neighbors:
>         - - name: system-name
>             tlv-type: 5

The `tlv` prefix seems to be redundant, since this would mean: type-length-value-type, just `type` seems to be sufficient for me.


>             value: Summit300-48
>           - name: system-description
>             tlv-type: 6
>             value: Summit300-48 - Version 7.4e.1 (Build 5) by Release_Master 05/27/05 04:53:11
>           - name: system-capabilities
>             tlv-type: 7
>             value:
>             - mac-bridge
>             - router

Are these values (mac-bridge, router) official values or should they use IDs as well?

>           - name: chassis-id
>             tlv-type: 1
>             value:
>               chassis-id: 00:01:30:F9:AD:A0
>               chassis-id-type: mac
>           - name: port-id
>             tlv-type: 2
>             value:
>               port-id: 1/1
>               port-id-type: interface-name

Are port-id(-type) official names and interface-name?

>           - name: ieee-802-1-vlans
>             oid: 00:80:c2
>             tlv-subtype: 1
>             tlv-type: 127
>             value:
>             - name: v2-0488-03-0505
>               vid: 488
>           - name: ieee-802-3-mac-phy-conf
>             oid: 00:12:0f
>             tlv-subtype: 1
>             tlv-type: 127
>             value:
>               autoneg: true
>               operational-mau-type: 16
>               pmd-autoneg-cap: 27648

Are autoneg etc. offician names?

>           - name: ieee-802-1-ppvids
>             oid: 00:80:c2
>             tlv-subtype: 2
>             tlv-type: 127
>             value:
>             - ppvid: 0

Is this an official name?

>           - name: management-addresses
>             tlv-type: 8
>             value:
>             - address: 00::01::30::F9::AD::A0

Are the double colons correct? Are the keys here official names?
>               address-subtype: ipv6
>               interface-number: 1001
>               interface-number-subtype: ifindex

Are the names here official names?

>           - name: ieee-802-3-max-frame-size
>             oid: 00:12:0f
>             tlv-subtype: 4
>             tlv-type: 127
>             value: 1522
>       mac-address: 36:E6:3A:97:AF:C9
>       mtu: 1500
> ```
> 
> I hope you like more this approach. Thanks for the feedback and reviews! @cathay4t @EdDev @tyll

Thank you. The `tlv-` prefix seems to be redundant and I wonder if the values should use IDs, too.

tests/integration/lldp_test.py
```diff
@@ -0,0 +1,188 @@
...
+
+def test_lldp_system(eth1_up):
+    with lldp_enabled(eth1_up):
+        _send_lldp_packet()
+        dstate = statelib.show_only((ETH1,))
+        lldp_config = dstate[Interface.KEY][0][LLDP.CONFIG_SUBTREE]
+        test_neighbor = _get_lldp_test(lldp_config[LLDP.NEIGHBORS_SUBTREE])
+
+        for tlv in test_neighbor:
+            if tlv["name"] == "system-name":
```
tyll, line 54: Since the important part is the type/subtype/oid and not the human-readable name, the tests should check that those are correct.

tests/integration/lldp_test.py
```diff
@@ -0,0 +1,188 @@
...
+                assert tlv["value"] == "Summit300-48"
```
tyll, line 55: `"value"` should probably be a constant, since it will be part of the schema/official API.

tests/integration/lldp_test.py
```diff
@@ -0,0 +1,188 @@
...
+def test_lldp_chassis(eth1_up):
+    with lldp_enabled(eth1_up):
+        _send_lldp_packet()
+        dstate = statelib.show_only((ETH1,))
+        lldp_config = dstate[Interface.KEY][0][LLDP.CONFIG_SUBTREE]
+        test_neighbor = _get_lldp_test(lldp_config[LLDP.NEIGHBORS_SUBTREE])
+
+        for tlv in test_neighbor:
+            if tlv["name"] == "chassis-id":
+                assert tlv["value"]["chassis-id"] == "00:01:30:F9:AD:A0"
```
tyll, line 71: `"chassis-i"` should probably be a constant if it is an official name or it needs to be changed to be accessed via an official ID.

ffmancera:
About official name questions, I have used the official names from the specifications all the time.

tests/integration/lldp_test.py
ffmancera: There is no official ID for that kind of values. I took the names from the specification so I assume this is the official name.

tyll, line 71:
```suggestion
            if tlv["type"] == 1:
                assert tlv["value"] == "00:01:30:F9:AD:A0"
```

Does this make sense?

tests/integration/lldp_test.py
```diff
@@ -0,0 +1,319 @@
...
+    with lldp_enabled(eth1_up) as dstate:
+        assertlib.assert_state_match(dstate)
+
+
+def test_lldp_yaml(eth1_up):
+    with lldp_enabled(eth1_up):
+        _send_lldp_packet()
+        dstate = statelib.show_only((ETH1,))
+        lldp_config = dstate[Interface.KEY][0][LLDP.CONFIG_SUBTREE]
+        test_neighbor = _get_lldp_test(lldp_config[LLDP.NEIGHBORS_SUBTREE])
```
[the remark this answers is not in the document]
ffmancera: We cannot assume that there is no more neighbors that the one we are injecting. When testing LLDP in different platforms I have found that we got more that one if there a switch or something providing LLDP information.

This is only true when using our container.

tests/integration/lldp_test.py
```diff
@@ -0,0 +1,319 @@
...
+            if (
+                tlv[LLDP.Neighbors.TLV_TYPE] == 127
+                and tlv[LLDP.Neighbors.ORGANIZATION_CODE] == "00:12:0f"
+                and tlv[LLDP.Neighbors.TLV_SUBTYPE] == 4
+            ):
+                seen.add(tlv[LLDP.Neighbors.TLV_TYPE])
+                seen.add(tlv[LLDP.Neighbors.ORGANIZATION_CODE])
+                seen.add(tlv[LLDP.Neighbors.TLV_SUBTYPE])
+                assert tlv[MFS_KEY] == 1522
+        assert set([127, 4, "00:12:0f"]).issubset(seen)
```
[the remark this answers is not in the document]
ffmancera: Done, thanks!

tests/integration/lldp_test.py
```diff
@@ -0,0 +1,319 @@
...
+
+
+def _get_lldp_test(neighbors):
+    for neighbor in neighbors:
+        for tlv in neighbor:
+            if (
+                tlv[LLDP.Neighbors.TLV_TYPE] == 5
+                and tlv[SYSTEM_NAME] == "Summit300-48"
+            ):
+                return neighbor
```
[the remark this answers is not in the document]
ffmancera: Done, thanks!

tests/integration/lldp_test.py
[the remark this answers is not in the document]
tyll: tcpreplay is injecting on eth1peer, which implies that it is a veth interface. Maybe the test should then setup its custom veth interfaces that it would also use on other platforms to avoid conflicts when eth1 is actuall not a veth.

libnmstate/ifaces/base_iface.py
```diff
@@ -376,6 +378,11 @@ def _remove_empty_description(state):
         del state[Interface.DESCRIPTION]
 
 
+def _remove_lldp_neighbors(state):
+    if state.get(LLDP.CONFIG_SUBTREE):
+        state[LLDP.CONFIG_SUBTREE].pop(LLDP.NEIGHBORS_SUBTREE, None)
```
cathay4t, line 383: `state.get(LLDP.CONFIG_SUBTREE, {}).pop(LLDP.NEIGHBORS_SUBTREE, None)`

libnmstate/nm/applier.py
```diff
@@ -458,10 +459,12 @@ def _build_connection_profile(
     if base_profile:
         con_setting.import_by_profile(base_con_profile)
         con_setting.set_profile_name(iface_name)
+        lldp.apply_lldp_setting(con_setting, iface_desired_state)
     else:
         con_setting.create(
             con_name=iface_name, iface_name=iface_name, iface_type=iface_type,
         )
+        lldp.apply_lldp_setting(con_setting, iface_desired_state)
```
cathay4t, line 467: Two duplicated lines is not needed, single one with upper indent would works.

tests/integration/lldp_test.py
```diff
@@ -0,0 +1,318 @@
...
+def _get_lldp_test(neighbors):
+    for neighbor in neighbors:
+        tlvs = list(
+            filter(
+                lambda tlv: tlv[LLDP.Neighbors.TLV_TYPE] == 5
+                and tlv[SYSTEM_NAME] == LLDP_TEST_SYSTEM_NAME,
+                neighbor,
+            )
+        )
+        if len(tlvs) == 1
```
cathay4t, line 315: Missing a `:`.

libnmstate/nm/applier.py
ffmancera: You are right, thanks! Done!

tests/integration/lldp_test.py
ffmancera: oops

ffmancera: Fixed!

ffmancera:
The CI error is caused by a bug on NM 1.22-git and master. NM folks are investigating it.

tests/integration/lldp_test.py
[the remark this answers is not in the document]
ffmancera: All of the platforms of NMCI is using veth interfaces, I like this idea but could we do it in another PR? That would require specific testing in platforms with not veths. Thanks!

libnmstate/ifaces/base_iface.py
```diff
@@ -376,6 +378,11 @@ def _remove_empty_description(state):
         del state[Interface.DESCRIPTION]
 
 
+def _remove_lldp_neighbors(state):
+    if state.get(LLDP.CONFIG_SUBTREE):
```
cathay4t, line 382: This is not required any more.

tests/integration/lldp_test.py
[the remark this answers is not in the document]
tyll: You wrote that a switch might inject LLDP information on different platforms but you also state that veths are used on every platform but the switches should not be able to inject LLDP information in the veth. Therefore, it seems to me that you can just add the assertion for one neighbor. If I am missing something and it is too complicated, I am fine with revising the test. Please track this in an issue, then.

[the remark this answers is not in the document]
ffmancera: I will add the test checking that there is only one neighbor. You are right in your statement, thanks!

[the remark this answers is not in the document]
ffmancera: Done, I've added a test. Thanks!

libnmstate/ifaces/base_iface.py
ffmancera: Done, thanks!

tests/integration/lldp_test.py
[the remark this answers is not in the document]
tyll: All the `len()` asserts are happening after a filter AFAICS, so there is no assertion to check for only one neighbor.

PR #1089, opened 2020-05-25
nm.active_connection: fix incorrect use of GLib.Error

ffmancera:
The exception raise by libnm on activation is GLib.Error and
its code is NM.ManagerError as its error domain `nm-manager-error`.

Use `GLib.Error.matches()` to check on error types.

### DIFF
--- a/libnmstate/nm/active_connection.py
+++ b/libnmstate/nm/active_connection.py
@@ -26,9 +26,6 @@ from .common import GObject
 from .common import NM
 
 
-NM_MANAGER_ERROR_DOMAIN = "nm-manager-error-quark"
-
-
 NM_AC_STATE_CHANGED_SIGNAL = "state-changed"
 
 
@@ -108,13 +105,9 @@ class ActiveConnection:
 
         try:
             success = src_object.deactivate_connection_finish(result)
-        except Exception as e:
-            if (
-                isinstance(e, GLib.GError)
-                # pylint: disable=no-member
-                and e.domain == NM_MANAGER_ERROR_DOMAIN
-                and e.code == NM.ManagerError.CONNECTIONNOTACTIVE
-                # pylint: enable=no-member
+        except GLib.Error as e:
+            if e.matches(
+                NM.ManagerError.quark(), NM.ManagerError.CONNECTIONNOTACTIVE
             ):
                 success = True
                 logging.debug(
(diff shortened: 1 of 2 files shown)


libnmstate/nm/active_connection.py
```diff
@@ -108,13 +105,10 @@ def _deactivate_connection_callback(self, src_object, result, user_data):
...
         try:
             success = src_object.deactivate_connection_finish(result)
-        except Exception as e:
-            if (
-                isinstance(e, GLib.GError)
-                # pylint: disable=no-member
-                and e.domain == NM_MANAGER_ERROR_DOMAIN
-                and e.code == NM.ManagerError.CONNECTIONNOTACTIVE
-                # pylint: enable=no-member
+        except common.GLib.Error as e:
```
tyll, line 108: What should happen if a different exception than `common.GLib.Error` is raised? Does this never happen?

ffmancera: According to libnm documentation it always raises GLib.Error exception. https://lazka.github.io/pgi-docs/#NM-1.0/classes/Client.html#NM.Client.deactivate_connection_finish

libnmstate/nm/active_connection.py
```diff
@@ -108,13 +105,9 @@ def _deactivate_connection_callback(self, src_object, result, user_data):
...
         try:
             success = src_object.deactivate_connection_finish(result)
-        except Exception as e:
-            if (
-                isinstance(e, GLib.GError)
-                # pylint: disable=no-member
-                and e.domain == NM_MANAGER_ERROR_DOMAIN
-                and e.code == NM.ManagerError.CONNECTIONNOTACTIVE
-                # pylint: enable=no-member
+        except GLib.Error as e:
```
EdDev, line 108: This now changes the `else` to be included under this type as well. Is that indeed the case?

ffmancera: Yes, the exception must be always a `GLib.Error` one, the else is only handling other `GLib.Error exceptions`. If we get any other kind of exception here then it is a bug from libnm.

EdDev: That is fine, but you do not cleanup on such a failure. It will just raise an unknown exception, the handler will not get disconnected, etc...

ffmancera: Done, thanks!

PR #1091, opened 2020-05-26
nm.checkpoint: check if the connection is activated

ffmancera:
Before initiate the interface rollback wait action, nmstate is checking
if the connection is already activated.


ffmancera:
I will run the CI several times in order to see if there are random errors left. Thanks.

PR #1092, opened 2020-05-27
test, integ: add test for renaming existing profiles

ffmancera:
If an interface has an existing profile active with a different name
than the interface name, then nmstate should rename the profile when
editting the interface.


tests/integration/profile_test.py
```diff
@@ -61,6 +61,22 @@
     "eth1",
 ]
 
+NMCLI_CLONE_CON_CMD = [
```
EdDev, line 64: I think these are too specific to be placed here.
It will fit well under the relevant place it is used.

(unless you think you will need this soon)

tests/integration/profile_test.py
```diff
@@ -119,6 +145,21 @@ def dummy_inactive_profile():
         cmdlib.exec_cmd(_nmcli_delete_connection("testProfile"))
 
 
+@contextmanager
+def create_active_profile(con_name):
```
EdDev, line 149: It does much more than the name implies and it is hard to understand what is the test doing unless diving in this implementation.
How about this as a fixture:
```
cloned_profile_name = "test-profile
with cloned_active_profile(name=cloned_profile_name, source=con_name):
    assert _profile_exists(ETH_PROFILE_DIRECTORY + "ifcfg-" + cloned_profile_name)
    cmdlib.exec_cmd(_nmcli_delete_connection(con_name))
    yield cloned_profile_name
```

tests/integration/profile_test.py
```diff
@@ -86,6 +102,16 @@ def test_delete_existing_interface_inactive_profiles(eth1_up):
...
 
 
+def test_rename_existing_interface_active_profile(eth1_up):
+    with create_active_profile(eth1_up[Interface.KEY][0][Interface.NAME]):
+        eth1_up[Interface.KEY][0][Interface.MTU] = 2000
+        libnmstate.apply(eth1_up)
+        profile_exists = _profile_exists(
+            ETH_PROFILE_DIRECTORY + "ifcfg-testProfile"
+        )
+        assert profile_exists
```
EdDev, line 112: nit: I think you can assert directly the `_profile_exists` result.

tests/integration/profile_test.py
ffmancera: I changed the name to "cloned_active_profile". I found hard to understand using it as a fixture. Please, look at the new approach I think it is clearer now :-)

tests/integration/profile_test.py
ffmancera: You are right, I moved it to the function that is using it. Thanks!

tests/integration/profile_test.py
ffmancera: Done, thanks!

PR #1093, opened 2020-05-28
test, integ: clean auto_routes and auto_dns only in the needed tests

ffmancera:
In some platforms it is not possible to lose the connectivity during a
long time. In order to fix this, we are only cleaning auto_dns and
auto_routes in the tests when necessary.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -25,8 +25,6 @@ import pytest
 
 import libnmstate
 from libnmstate.schema import DNS
-from libnmstate.schema import Interface
-from libnmstate.schema import InterfaceIP
 from libnmstate.schema import Route
 from libnmstate.schema import RouteRule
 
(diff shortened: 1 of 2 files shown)


EdDev:
BTW, you could also add this as a pytest argument, so if mentioned, it will cleanup and if not it will avoid it.
Adding this inside the tests themselfes can be left as is, maybe just extracting it out to a fixture and the top level (with function scope).

ffmancera:
> The change itself looks fine as long as we indeed only need it for a single test.
> We have been using this "clean" just because of one single test? It sounds a bit weird to me.
> Have you tracked down the reason for adding it in the first place?
> 
> It is worth also mentioning this in the commit message, as anyone seeing this change in the future may ask the same question.

We were doing it because a test were failing. Initially it was not an issue to lose the connectivity but openstack is having some issues. IMO, a fixture is not needed because it is only one test.

EdDev:
> IMO, a fixture is not needed because it is only one test.

Fixtures separate setup/teardown of tests from the actual test itself. Failing in a fixture expresses differently as well. It does not really matter how many users it has.
If you want to limit it to only this module, you can define it there and not globally.

ffmancera:
Done, thanks!

ffmancera:
> Thank you!
> 
> Please note that the other cleanup (DNS, routes and rules) are still in place at the session level.

Yes, that is intended :-)

PR #1120, opened 2020-06-17
test, integ: do not restore static routes for dynamic ifaces

ffmancera:
Nmstate does not support static routes for interfaces configured with
dynamic IP. In some environments like CI gating and NMCI there are
interfaces configured like that.

In order to fix that, nmstate is not restoring the static routes of the
interfaces configured with dynamic ip.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -25,6 +25,9 @@ import pytest
 
 import libnmstate
 from libnmstate.schema import DNS
+from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceIP
+from libnmstate.schema import InterfaceIPv6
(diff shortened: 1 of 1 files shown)


ffmancera:
> I dislike this approach.
> We are not restoring thing up, but give user a modified original state.
> 
> If we cannot restore to certain state, neither we support it or ask the test starter script to do change before and after our test.

I agree with this but that is not possible in some limited scenarios like CI gating machine. I will try to investigate why are we reporting an static IP address because according to the system the IP address is dynamic and not static but nmstate is not reporting that.

ffmancera:
I am closing this PR because the solution proposed is not good. I am working in a better one.

PR #1142, opened 2020-06-21
nm.ipv4, nm.ipv6: add auto-route-table-id support for IPv4 and IPv6

ffmancera:
This patch is adding support for `table_id` on IPv4 and IPv6. `table_id`
is setting `NM.SettingIPConfig.route-table` which decides in which table
the route will be configured.

### DIFF
--- a/libnmstate/nm/ipv4.py
+++ b/libnmstate/nm/ipv4.py
@@ -66,6 +66,10 @@ def create_setting(config, base_con_profile):
             setting_ipv4.props.ignore_auto_dns = not config.get(
                 InterfaceIPv4.AUTO_DNS, True
             )
+            setting_ipv4.props.route_table = config.get(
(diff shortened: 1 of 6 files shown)


cathay4t:
I prefer `auto-route-table-id` over `table-id` as this is designed for dynamic routes only.

And we need to cover:
 * Currently gateway use `table_id`, so we should hide the `table_id` when interface been configured as static.

ffmancera:
> I prefer `auto-route-table-id` over `table-id` as this is designed for dynamic routes only.
> 

Sure, will do.

> And we need to cover:
> 
>     * Currently gateway use `table_id`, so we should hide the `table_id` when interface been configured as static.

Maybe I am missing something but this what we are doing now. When the interface is configured as static we are hiding `table_id`.

ffmancera:
Changes done, thanks! @cathay4t

ffmancera:
Rebased, thanks!

PR #1148, opened 2020-06-23
base_iface: rename iface_type property to type

ffmancera:
In order to follow the property naming, it makes sense to rename
"iface_type" property to "type".


ffmancera:
Rebased, thanks!

PR #1174, opened 2020-07-02
nm.profile: rewrite NM profile activation

ffmancera:
This patch is rewriting the NM profile activation in order to use UUID
to identify which profile should be activated instead of using connection
id or device.

In addition, the ConnectionProfile class is being renamed to Profile and moved
to profile.py. A following PR will rewrite nm.active_connection.py into
profile.py in order to unify the NM profile management.


ffmancera:
@cathay4t you can do a pre-review. The patch needs some adjustments for tests and for the activation but the structure is quite stable now. Thanks!

libnmstate/nm/device.py
```diff
@@ -22,15 +22,12 @@
...
 
-def activate(context, dev=None, connection_id=None):
-    """Activate the given device or remote connection profile."""
-    conn = connection.ConnectionProfile(context)
-    conn.nmdevice = dev
-    conn.con_id = connection_id
-    conn.activate()
+def activate(context, dev=None, profile=None):
+    """Activate the given remote connection profile."""
+    profile.activate()
```
cathay4t, line 30: If this function is used by test only, please remove it and find a better way for test usage.

The same for other `nm.device` functions.

libnmstate/nm/profile.py
```diff
@@ -21,14 +21,948 @@
...
+        self._net_state = net_state
+        self._preapply_dns_fix()
+        self._profiles = [
+            Profile(context, iface)
+            for iface in net_state.ifaces.values()
+            if iface.is_changed or iface.is_desired
+        ]
+        self._save_to_disk = save_to_disk
+
+    def apply_changes(self):
```
cathay4t, line 81: There are two `apply_changes()`.

libnmstate/nm/profile.py
```diff
@@ -21,14 +21,948 @@
...
+        if preserve_old_dns_config:
+            for iface in self._net_state.ifaces.values():
+                if iface.is_changed or iface.is_desired:
+                    iface.remove_dns_metadata()
+
+
+class Profile:
+    def __init__(self, context, iface=None, profile=None):
+        self._ctx = context
+        self.rofile = profile
```
cathay4t, line 342: typo here

libnmstate/nm/profile.py
```diff
@@ -21,14 +21,948 @@
...
+            self._ctx.register_async(action, fast=True)
+            self.profile.delete_async(
+                self._ctx.cancellable,
+                self._delete_connection_callback,
+                user_data,
+            )
+
+    def activate(self):
+        if self.uuid:
+            self.import_by_uuid()
```
cathay4t, line 650: You should differetiate `NM.RemoteConnection` and `NM.SimpleConnection` like `self._remote_conn`  and `self._simple_conn`.

We only do import when `self._remote_conn` is undefined, not just import when uuid is defined.

You have to made it clear when we import:
 * Newly created profile.
 * Activation against `NM.Device` only like (checkpoint rollback).

libnmstate/nm/profile.py
```diff
@@ -21,14 +21,948 @@
...
+                iface.mark_as_changed()
+
+        if preserve_old_dns_config:
+            for iface in self._net_state.ifaces.values():
+                if iface.is_changed or iface.is_desired:
+                    iface.remove_dns_metadata()
+
+
+class Profile:
+    def __init__(self, context, iface=None, profile=None):
```
cathay4t, line 340: The name profile is blur when the Class is named as `Profile`. Try `nm_remote_conn` or `remote_conn`.

I use `iface_info` for dict iface state, `iface_state` for `BaseIface` object. Maybe `iface_state` here and correct the
`iface_state` used in this file.

libnmstate/nm/profile.py
```diff
@@ -21,14 +21,948 @@
...
+            self._active_connection_callback,
+            user_data,
+        )
+
+    @property
+    def profile(self):
+        return self.profile
+
+    @profile.setter
+    def profile(self, con_profile):
```
cathay4t, line 683: In stead exposing old interfaces, you should think about expose minimum interfaces required, so that you could know which is old unused function.

libnmstate/nm/profile.py
```diff
@@ -21,14 +21,948 @@
...
+        assert self._con_id is None
+        self._con_id = connection_id
+
+    @property
+    def uuid(self):
+        uuid = self.profile.get_uuid() if self._con_profile else None
+        return self._uuid or uuid
+
+    @uuid.setter
+    def uuid(self, connection_uuid):
```
cathay4t, line 718: there is code out side of `nm/profile.py` use/set this property.

ffmancera:
Found a better way for doing this. Will update soon. Thanks for all the comments.

PR #1175, opened 2020-07-02
tests: add assertion that only one neighbor is reported

UnversedBlood:
Delete `_get_lldp_test` function, instead use an assertion that checks the number of neighbors.

Signed-off-by: Víctor Javier Díaz Garrido <vicdigar@hotmail.com>

### DIFF
--- a/tests/integration/lldp_test.py
+++ b/tests/integration/lldp_test.py
@@ -122,8 +122,8 @@ def test_lldp_yaml(eth1_up):
         _send_lldp_packet()
         dstate = statelib.show_only((ETH1,))
         lldp_config = dstate[Interface.KEY][0][LLDP.CONFIG_SUBTREE]
-        test_neighbor = _get_lldp_test(lldp_config[LLDP.NEIGHBORS_SUBTREE])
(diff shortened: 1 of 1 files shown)


tests/integration/lldp_test.py
```diff
@@ -122,7 +122,10 @@ def test_lldp_yaml(eth1_up):
         _send_lldp_packet()
         dstate = statelib.show_only((ETH1,))
         lldp_config = dstate[Interface.KEY][0][LLDP.CONFIG_SUBTREE]
-        test_neighbor = _get_lldp_test(lldp_config[LLDP.NEIGHBORS_SUBTREE])
+
+        assert len(lldp_config[LLDP.NEIGHBORS_SUBTREE]) == 1
+
+        test_neighbor = lldp_config[LLDP.NEIGHBORS_SUBTREE][0]
```
ffmancera, line 128: Please, remove the introduced blank lines. Thanks!

UnversedBlood: Done, thanks!

PR #1186, opened 2020-07-06
tests: fix error on test_ovs_remove_port

ffmancera:
In the ovs_remove_port test we are using nmcli directly using cmdlib.
nmcli need a second to refresh the information if not, it will be
outdated and then the test will not raise the exception.

This was not happening before because Nmstate was waiting 5 seconds
after applying the state.

### DIFF
--- a/tests/integration/ovs_test.py
+++ b/tests/integration/ovs_test.py
@@ -18,6 +18,7 @@
 #
 
 from subprocess import CalledProcessError
+import time
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


cathay4t:
> In the ovs_remove_port test we are using nmcli directly using cmdlib.
> nmcli need a second to refresh the information if not, it will be
> outdated and then the test will not raise the exception.
> 
> This was not happening before because Nmstate was waiting 5 seconds
> after applying the state.

Can you elaborate this? I don't recall we wait 5 seconds after apply.


Please check with NM team first in stead of simple `sleep`, sounds like we are hiding some issue
by this.

cathay4t:
@ffmancera Any update on this issue?

cathay4t:
Never mind. I am debugging it as it impact my nispor pull requests.

PR #1189, opened 2020-07-06
[nmstate-0.3] dns: fix remove dns config

ffmancera:
This patch is fixing the support of removing the dns config. In
addition, is fixing the `test_remove_dns_config` test by using the
parametrize variable.

Ref: https://bugzilla.redhat.com/1850698


cathay4t:
Thanks.

PR #1198, opened 2020-07-10
automation: extend --customize argument for no container support

ffmancera:
Some CI gating platforms need to install dependencies using pip3.

### DIFF
--- a/automation/run-tests.sh
+++ b/automation/run-tests.sh
@@ -263,6 +263,13 @@ function upgrade_nm_from_copr {
     exec_cmd "systemctl restart NetworkManager"
 }
 
+function run_customize_command {
+    if [[ -n "$customize_cmd" ]];then
+        clean_dnf_cache
+        exec_cmd "${customize_cmd}"
+    fi
+}
+
 options=$(getopt --options "" \
     --long customize:,pytest-args:,help,debug-shell,test-type:,el8,copr:,artifacts-dir:,test-vdsm,machine,use-installed-nmstate\
     -- "${@}")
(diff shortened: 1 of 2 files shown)


automation/run-tests.sh
```diff
@@ -343,6 +350,7 @@ if [ -z ${RUN_BAREMETAL} ];then
     container_pre_test_setup
 else
     CONTAINER_WORKSPACE="."
+    run_customize_command
```
tyll, line 353: This seems to change running the customize command only on non-container machines instead of only in containers.

Btw. IMHO the if statement could be more clean if you avoid negations:

```
if [ -n "${USE_CONTAINER}" ]
then

...

else
...
fi
```

would make it more clear that the first branch is for containers and the "else" branch for non-containers. Another alternative would be

```
if [ -n "${USE_MACHINE}" ]
    # do non container stuff
else
   # do container stuff
fi
```

But from our meeting discussion I remember that the distinction was "container" vs. "non container", therefore using "container" in the variable name makes more sense to me.

automation/run-tests.sh
```diff
@@ -263,6 +263,13 @@ function upgrade_nm_from_copr {
     exec_cmd "systemctl restart NetworkManager"
 }
 
+function run_customize_command {
```
tyll, line 266: Mentioning "pip" in the commit message seems to be unmotivated since the code change does not involve pip at all. Also CI gating is not directly related to this change. I would motivate it as:

Using a customize command is useful when running tests outside a container, too. Therefore always run it when specified. This is for example useful when installing unpackaged dependencies in CI based gating.

Btw. I assumed the gating platforms provide their own way of preparing the test machine for running the test and installing dependencies, so it is probably not hard requirement to have this in there. Nevertheless it seems to be useful to allow this both in containers and non-containers (but this is not yet, what the code seems to do).

ffmancera: Modified, thanks!

automation/run-tests.sh
ffmancera: I removed the negation and it is more clean now. Thanks!

Issue #1203, opened 2020-07-15
nmstate-plugin-ovsdb should require python3dist(ovs)


ffmancera:
## Describe the bug
In `nmstate.spec`, nmstate-plugin-ovsdb subpackage is recommending `python3dist(ovs)` instead it should require it. In addition, python3-nmstate is recommending `NetworkManager` as we are not using dbus directly anymore, we should require it too.

Ref: http://ftp.rpm.org/max-rpm/s1-rpm-build-creating-spec-file.html

Razaloc:
Can I work on this?

ffmancera:
> Can I work on this?

Sure!

PR #1221, opened 2020-07-21
Packaging: adjust dependecy requirements #1203

Razaloc:
-Modify nmstate.spec

Signed-off-by: Rafael G. Ruiz <llerrak@hotmail.com>

### DIFF
--- a/packaging/nmstate.spec
+++ b/packaging/nmstate.spec
@@ -39,7 +39,7 @@ Summary:        nmstate plugin for OVS database manipulation
 Requires:       python3-%{libname} = %{?epoch:%{epoch}:}%{version}-%{release}
 # The python-openvswitch rpm pacakge is not in the same repo with nmstate,
 # hence state it as Recommends, no requires.
-Recommends:     python3dist(ovs)
+Requires:     python3dist(ovs)
 
 %description -n python3-%{libname}
 This package contains the Python 3 library for Nmstate.


packaging/nmstate.spec
```diff
@@ -26,7 +26,7 @@ Summary:        nmstate Python 3 API library
 Requires:       NetworkManager-libnm >= 1:1.22.10
 # Use Recommends for NetworkManager because only access to NM DBus is required,
 # but NM could be running on a different host
-Recommends:     NetworkManager
+Requires:     NetworkManager
```
cathay4t, line 29: As the above comments states, please still use `Recommends` for NM.

ffmancera: The above statement is now false. We now need full access to NM AFAIK, because we are now not using dbus directly for checkpoint support. Could we use nmstate without NM installed? Am I missing something? Thanks!

EdDev: nmstate just needs the dbus socket to work with NM, not NM itself.
kubernetes-nmstate once worked in an isolates manner, where they deployed k-nmstate and nmstate on a container and used the dbus socket to communicate with NM. As this introduces several limitations (for the parts that NM is not involved) the approach was dropped.

Although I think no one else uses it like this anymore, this may still be a valid deployment. You will need to explicitly drop support for such a deployment and with it this specific dependency type on NM.

ffmancera: Okay, I agree with this. Thanks for explaining :-)

Razaloc: Thanks for the comments. I've changed that part back.

PR #1223, opened 2020-07-22
ifaces: validate SR-IOV VF MTU

ffmancera:
Nmstate will raise a NmstateValueError if a VF has a greater MTU than
its PF. In order to do this, we are using sysfs and the PCI address to
identify the PF of each VF.

Ref: https://github.com/intel/SDN-NFV-Hands-on-Samples/blob/master/SR-IOV_Network_Virtual_Functions_in_KVM/listvfs_by_pf.sh
Ref: https://bugzilla.redhat.com/1856256

### DIFF
--- a/libnmstate/ifaces/ifaces.py
+++ b/libnmstate/ifaces/ifaces.py
@@ -18,6 +18,8 @@
 #
 
 import logging
+import os
+import subprocess
 
 from libnmstate.error import NmstateKernelIntegerRoundedError
 from libnmstate.error import NmstateValueError
@@ -127,6 +129,7 @@ class Ifaces:
     def _pre_edit_validation_and_cleanup(self):
         self._validate_over_booked_slaves()
         self._validate_vlan_mtu()
+        self._validate_sriov_mtu()
         self._handle_master_slave_list_change()
         self._match_child_iface_state_with_parent()
         self._mark_orphen_as_absent()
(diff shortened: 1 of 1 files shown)


ffmancera:
This patch has been tested using SR-IOV hardware.

libnmstate/ifaces/ifaces.py
```diff
@@ -194,6 +197,53 @@ def _validate_vlan_mtu(self):
                         f"MTU({base_iface.mtu})"
                     )
 
+    def _validate_sriov_mtu(self):
```
EdDev, line 200: Please split this method so it will be clear what is the high level logic without entering all the details.
E.g. building the pf-vf tree is one thing, calculating the max mtu is another and setting it on the PF a 3rd.

cathay4t:
> Is there any reference on this mtu dependency between the PF and VF? It sounds strange to me.
> If this is HW dependent, then it surely cannot be under nmstate responsibility.
> 
> Please note that the PF is another interface that can be used in addition to the VF/s, having this dependency is strange.

According to Intel employee at https://community.intel.com/t5/Ethernet-Products/L2-Unicast-not-sent-on-ixgbevf/m-p/383563/highlight/true#M6138 PF should hold bigger or equal MTU than VFs.

The kernel code seem only says "The 82599 cannot support a mix of jumbo and non-jumbo PF/VFs".

So to simplify it, asking VF MTU smaller or equal to PF is reasonable in nmstate and also fit to real use cases.

libnmstate/ifaces/ifaces.py
cathay4t: For detail,

You may change `EthernetIface.__init__()` to detect whether certain NIC is VF and set the `EthernetIface.parent` property to its PF.

Then rename `Ifaces._validate_vlan_mtu()` to `Ifaces._validate_child_mtu` with small modification, go get consistant
behavior like we did for VLAN/VxLAN.

EdDev:
> > Is there any reference on this mtu dependency between the PF and VF? It sounds strange to me.
> > If this is HW dependent, then it surely cannot be under nmstate responsibility.
> > Please note that the PF is another interface that can be used in addition to the VF/s, having this dependency is strange.
> 
> According to Intel employee at https://community.intel.com/t5/Ethernet-Products/L2-Unicast-not-sent-on-ixgbevf/m-p/383563/highlight/true#M6138 PF should hold bigger or equal MTU than VFs.
> 
> The kernel code seem only says "The 82599 cannot support a mix of jumbo and non-jumbo PF/VFs".
> 
> So to simplify it, asking VF MTU smaller or equal to PF is reasonable in nmstate and also fit to real use cases.

This is one specific vendor with one specific model, if there is no overall rule or if the others do not limit this, it do not think it is nmstate role to enforce this.
Can we confirm this for other vendors (e.g. Melanox) and models?

ffmancera:
> > > Is there any reference on this mtu dependency between the PF and VF? It sounds strange to me.
> > > If this is HW dependent, then it surely cannot be under nmstate responsibility.
> > > Please note that the PF is another interface that can be used in addition to the VF/s, having this dependency is strange.
> > 
> > 
> > According to Intel employee at https://community.intel.com/t5/Ethernet-Products/L2-Unicast-not-sent-on-ixgbevf/m-p/383563/highlight/true#M6138 PF should hold bigger or equal MTU than VFs.
> > The kernel code seem only says "The 82599 cannot support a mix of jumbo and non-jumbo PF/VFs".
> 
> Thanks for this info!
> 
> > So to simplify it, asking VF MTU smaller or equal to PF is reasonable in nmstate and also fit to real use cases.
> 
> This is one specific vendor with one specific model, if there is no overall rule or if the others do not limit this, it do not think it is nmstate role to enforce this.
> Can we confirm this for other vendors (e.g. Melanox) and models?

I have tried this with a Broadcomm NIC and others Mellanox drivers and it is working well. I think we should not raise this error because it would limit all the users.

ffmancera:
As this is a specific NIC/driver issue, I am closing this PR because Nmstate should not handle driver/NIC limitations. As I said above, this is working well with other NICs.

PR #1242, opened 2020-07-25
nm.wired: do not report MTU if it is 0

ffmancera:
If an interface contains an MTU with value 0, Nmstate should not report
it because it is an special interface like OVS patch port interfaces.

Added a test case for this.


cathay4t:
@ffmancera Should we backport this patch?

ffmancera:
> @ffmancera Should we backport this patch?

Yes, I will do it. Sorry I forgot to add the the label. Thanks!

PR #1254, opened 2020-07-30
nm.profile: rewrite NM profile activation

ffmancera:
Created new classes:

* NmProfiles handle the NmProfile creation and keeps the list of
  profiles. In addition, it is used to apply the config of each
  NmProfile in the correct order.

* NmProfile handle the activation, deactivation and deletion of the
  profile. In addition it keeps the
  NM.RemoteConnection/NM.SimpleConnection and handles the profile creation
  of proxy interfaces.

* The logic of profile activation from libnmstate/nm/applier.py and
  libnmstate/nm/connection.py has been unified in
  libnmstate/nm/profile.py.

Moved functions:

* Move _preapply_dns_fix() from nm/applier.py to libnmstate/dns.py and
  rename it to preapply_dns_fix_for_profiles().

* Move create_proxy_profile to NmProfile class. For each profile,
  Nmstate will check if it is needed to create a proxy profile.

* Move _is_ovs_lag_port() from nm/applier.py to libnmstate/ifaces/ovs.py


libnmstate/nm/profile.py
```diff
@@ -21,9 +21,828 @@
...
+            profiles_to_delete,
+            devs_to_delete,
+        ]
+
+    def _prepare_state_for_profiles(self, net_state):
+        nm_dns.preapply_dns_fix_for_profiles(self._ctx, net_state)
+
+
+class NmProfile:
+    def __init__(self, context, save_to_disk, iface=None, iface_state=None):
```
cathay4t, line 275: The `save_to_disk` should be argument of `create_proxy_profile()` or `apply_config()`.

libnmstate/nm/profile.py
```diff
@@ -21,9 +21,828 @@
...
+        self._ac_handlers = set()
+        self._dev_handlers = set()
+        self._remote_conn = None
+        self._simple_conn = None
+        self.to_activate = False
+        self.to_deactivate = False
+        self.to_delete = False
+        self.to_deactivate_beforehand = False
+        self.to_modify = False
+        self.delete_dev = False
```
cathay4t, line 291: Change to `self.pending_actions = ['activate', 'delete_profile', etc]`

ffmancera:
> You don't have clear line between `NmProfiles` and `NmProfile`.
> 
> For example. the `profile.to_activate = True` line show the `NmProfiles` is determine the state of
> `NmProfile` while `NmProfile` can know it via `iface_state`.

I wanted to unify this. In that specific case this is true but for other cases it must be done in NmProfiles because the others NmProfile objects affect the action that must be taken. I could split this up but in my opinion it will make it less clear because then NmProfiles will determine some of the actions that must be done on the NmProfile objects and at the sime time NmProfile object will determine others too.

> 
> I think:
> 
>     * `NmProfiles` is for handling actions require knowledge **between** profiles.
> 
>     * `NmProfile` is taking actions based on given iface_state.
> 
> 
> Meanwhile, I would like to see the object of `BaseIface` been passing to `NmProfile` instead of dict,
> so that you can use `BaseIface.master` etc properties.

NmProfiles are being created using the BaseIface object. But from `create_proxy_profile()` I am creating it using the dictionary but I will change that. Thanks!

> 
> There is also unclear line between `nm/profile.py` and `nm/connection.py`.
> I would like to suggest:
> 
>     * `nm/profile.py` handle profile activation/deactivation/deletion.
> 
>     * `nm/connection.py` handle converting iface_state dict to `NM.SimpleConnection` or updating `NM.RemoteConnection`.

I am removing a lot of code from nm/connection.py. nm/connection.py will handle only NM.SettingConnection.

PR #1316, opened 2020-09-04
packaging: make varlink as hard requirement

ffmancera:
Varlink is only used for varlink support so it should not be imported
when doing other operations using nmstatectl.

```
File "/usr/lib/python3.6/site-packages/nmstatectl/nmstatectl.py", line 40, in <module>
  from nmstatectl.nmstate_varlink import start_varlink_server
File "/usr/lib/python3.6/site-packages/nmstatectl/nmstate_varlink.py", line 31, in <module>
  raise libnmError.NmstateDependencyError("python3 varlink module not found")
libnmstate.error.NmstateDependencyError: python3 varlink module not found
```

### DIFF
--- a/packaging/nmstate.spec
+++ b/packaging/nmstate.spec
@@ -43,19 +43,11 @@ Requires:       python3-%{libname} = %{?epoch:%{epoch}:}%{version}-%{release}
 # hence state it as Recommends, no requires.
 Requires:     python3dist(ovs)
 
-%package -n nmstate-varlink
-Summary:        Varlink support for libnmstate
-Requires:       python3dist(varlink)
-
-
 %description -n python3-%{libname}
 This package contains the Python 3 library for Nmstate.
 
 %description -n nmstate-plugin-ovsdb
 This package contains the nmstate plugin for OVS database manipulation.
-
-%description -n nmstate-varlink
-This package provides varlink support for libnmstate.
 
 %prep
 %setup -q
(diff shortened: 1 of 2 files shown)


nmstatectl/nmstatectl.py
```diff
@@ -31,13 +31,18 @@
...
 from libnmstate.error import NmstatePermissionError
 from libnmstate.error import NmstateValueError
 from libnmstate.schema import Interface
 from libnmstate.schema import InterfaceIP
 from libnmstate.schema import Route
 from libnmstate.schema import RouteRule
-from nmstatectl.nmstate_varlink import start_varlink_server
+
+try:
+    from nmstatectl.nmstate_varlink import start_varlink_server
```
tyll: if varlink is not available, the help should also not show the varlink support

packaging/nmstate.spec
```diff
@@ -35,6 +35,7 @@ Recommends:     NetworkManager-config-server
 Suggests:       NetworkManager-ovs
 Suggests:       NetworkManager-team
 Requires:       python3-nispor
+Requires:       python3dist(varlink)
```
tyll: This should not be necessary since requirements.txt are considered by rpmbuild.

ffmancera: Removed, thanks.

tyll:
It would be nice to also have CI jobs that check whether the package is installable on all our target distributions. Since the container images include all the anticipated dependencies, it is not tested, there.

cathay4t:
> It would be nice to also have CI jobs that check whether the package is installable on all our target distributions. Since the container images include all the anticipated dependencies, it is not tested, there.

That sounds like a job of downstream packager's work.

cathay4t (approved):

PR #1324, opened 2020-09-07
vlan, vxlan: iface cannot have ip when subordinated

ffmancera:


EdDev (approved):
I can't thik of any type that can have a stack when it is subordinated.
I guess the only exception is the OVS internal interface which is connected to the switch.

cathay4t (approved):

PR #1326, opened 2020-09-09
nm.translator: use constants from the schema

ffmancera:
Use the constants from the schema instead of hardcoded strings.


cathay4t (approved):
LGTM. Thanks.

PR #1330, opened 2020-09-10
inclusive language: deprecate "slaves" from the API

### LINKED ISSUE #1200: Replace master/slave term with more inclusive terminology in the API
Proposed alternatives for master/slave include:

    primary/secondary
    main/replica or subordinate
    initiator/target
    requester/responder
    controller/device
    host/worker or proxy
    leader/follower
    director/performer

ffmancera:
* Bond.SLAVES is now deprecated by Bond.PORT

* OVSBridge.Port.LinkAggregation.SLAVES_SUBTREE is now deprecated by
  OVSBridge.Port.LinkAggregation.PORT_SUBTREE

* OVSBridge.Port.LinkAggregation.Slave.NAME is now deprecated by
  OVSBridge.Port.LinkAggregation.Port.NAME

* Team.PORT_SUBTREE is now "port" instead of "ports"

### DIFF
--- a/libnmstate/ifaces/bond.py
+++ b/libnmstate/ifaces/bond.py
@@ -19,6 +19,7 @@
 
 import contextlib
 import logging
+import warnings
 
 from libnmstate.error import NmstateValueError
 from libnmstate.schema import Bond
@@ -28,21 +29,25 @@ from libnmstate.schema import Interface
 from .base_iface import BaseIface
 
 
+DEPRECATED_SLAVES = "slaves"
+
+
 class BondIface(BaseIface):
     _MODE_CHANGE_METADATA = "_bond_mode_changed"
 
     def sort_slaves(self):
         if self.slaves:
-            self.raw[Bond.CONFIG_SUBTREE][Bond.SLAVES].sort()
+            self.raw[Bond.CONFIG_SUBTREE][Bond.PORT].sort()
 
     def __init__(self, info, save_to_disk=True):
         super().__init__(info, save_to_disk)
         self._normalize_options_values()
         self._fix_bond_option_arp_monitor()
+        self._replace_deprecated_terms()
 
     @property
     def slaves(self):
-        return self.raw.get(Bond.CONFIG_SUBTREE, {}).get(Bond.SLAVES, [])
+        return self.raw.get(Bond.CONFIG_SUBTREE, {}).get(Bond.PORT, [])
 
     @property
     def is_master(self):
@@ -125,6 +130,12 @@ class BondIface(BaseIface):
                 "please disable one of them by setting to 0"
             )
 
+    def _replace_deprecated_terms(self):
+        bond_cfg = self.raw.get(Bond.CONFIG_SUBTREE)
+        if bond_cfg and bond_cfg.get(DEPRECATED_SLAVES):
+            bond_cfg[Bond.PORT] = bond_cfg.pop(DEPRECATED_SLAVES)
+            warnings.warn("Using 'slaves' is deprecated, use 'port' instead.")
+
     @staticmethod
     def is_mac_restricted_mode(mode, bond_options):
         return (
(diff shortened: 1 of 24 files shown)


libnmstate/schema.py
```diff
@@ -325,7 +368,7 @@ class Team:
     TYPE = InterfaceType.TEAM
     CONFIG_SUBTREE = InterfaceType.TEAM
 
-    PORT_SUBTREE = "ports"
+    PORT_SUBTREE = "port"
```
cathay4t: This is breaking the API. Please do in another PR.

Possible approach for not breaking API as old working yaml will not pass the schema check.

Include both `ports` and `port` in schema.

In `libnmstate/ifaces/team.py`, change the `ports` in desire state to `port` and raise a warnning.

Both `ports` and `port` in desire state should works.

cathay4t (changes requested):

cathay4t:
Only did partial reviews. Have to manually test whether old API still works.

libnmstate/schema.py
EdDev: Deprecation warnings are supposed to be handled by ‘ DEPRECATED_CONSTANTS’, no? Is there another part that breaks?

ffmancera: The problem here is that DEPRECATED_CONSTANTS mechanism works for removing/replacing constants. In this case, we are NOT modifying the constant but its value so it's different. IMO, @cathay4t proposed mechanism should work. I will try it out in another PR.

ffmancera:
> Only did partial reviews. Have to manually test whether old API still works.

Will test it manually too. I think it should work as the deprecation test is passing.

libnmstate/schema.py
ffmancera: I've removed it from this PR.

ffmancera:
I've tested this manually, here is the output:

```
Python 3.6.8 (default, Apr 16 2020, 01:36:27) 
[GCC 8.3.1 20191121 (Red Hat 8.3.1-5)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from libnmstate.schema import Bond
>>> from libnmstate.schema import OVSBridge
>>> Bond.SLAVES
sys:1: FutureWarning: Using 'Bond.SLAVES' is deprecated, use 'Bond.PORT' instead.
'port'
>>> OVSBridge.Port.LinkAggregation.SLAVES_SUBTREE
sys:1: FutureWarning: Using 'LinkAggregation.SLAVES_SUBTREE' is deprecated, use 'OVSBridge.Port.LinkAggregation.PORT_SUBTREE' instead.
'port'
>>> OVSBridge.Port.LinkAggregation.Slave.NAME
sys:1: FutureWarning: Using 'LinkAggregation.Slave' is deprecated, use 'OVSBridge.Port.LinkAggregation.Port' instead.
'name'
```

cathay4t (changes requested):
When yml file still using old `slave` term, the apply will fail, this is not acceptable.

We should still support old term of `slave` regardless user use it as constant or hard-coded string.

ffmancera:
> When yml file still using old `slave` term, the apply will fail, this is not acceptable.
> 
> We should still support old term of `slave` regardless user use it as constant or hard-coded string.

I've fixed this. Thanks!

libnmstate/ifaces/bond.py
```diff
@@ -125,6 +130,12 @@ def _validate_miimon_conflict_with_arp_interval(self):
                 "please disable one of them by setting to 0"
             )
 
+    def _replace_deprecated_terms(self):
+        lag_info = self.raw.get(Bond.CONFIG_SUBTREE)
```
cathay4t: Use `bond_cfg` or `bond_info` please, as there is no other code in this file using term of `lag`.

cathay4t (changes requested):
The `schemas/operational-state.yaml` is not updated.
And it should allows both old and new terms.

libnmstate/ifaces/bond.py
ffmancera: Done, thanks!

ffmancera:
> The `schemas/operational-state.yaml` is not updated.
> And it should allows both old and new terms.

Done, thank you!

cathay4t (approved):
LGTM. Thanks.

PR #1270, opened 2020-08-07
libnmstate-varlink: Implementing varlink support for libnmstate

SudhanAruna:
Implemented varlink support for libnmstate functions (show, apply, commit and rollback)
Added test cases for varlink service
Integrated with setup.py
Added README file

Signed-off-by: Arunagirinadan Sudharshan <sudhan.aruna.97@gmail.com>

PR #1342, opened 2020-09-22
libnmstate-varlink: Implementing varlink support for libnmstate

qinqon:
Backport from https://github.com/nmstate/nmstate/pull/1270

Implemented varlink support for libnmstate functions (show, apply, commit and rollback)
Added test cases for varlink service
Integrated with setup.py
Added README file

Signed-off-by: Arunagirinadan Sudharshan <sudhan.aruna.97@gmail.com>

### DIFF
--- a/nmstatectl/README.varlink.md
+++ b/nmstatectl/README.varlink.md
@@ -1,0 +1,154 @@
+# Varlink support for libnmstate
+Varlink protocol ([Varlink.org](https://varlink.org/)) can used to communicate with libnmstate via stdin/out and varlink client implementation ([Existing Language binding](https://varlink.org/Language-Bindings#how-to-test-new-language-bindings)). Varlink protocol encodes all messages as JSON object and communicates over unix and tcp socket connections. Nmstate service is defined in io.nmstate interface address and libnmstate functions should called in as specified in varlink interface defination file `io.nmstate.varlink`. This implementation supports python3-varlink version 29.0.0 and unix socket connection only.
+
+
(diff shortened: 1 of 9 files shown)


ffmancera:
Thanks for the contribution, why is this needed? Backporting varlink support will require a great effort. Thanks!

qinqon:
We (kubernetes-nmstate) were investigating ways to keep nmstate alive while running commands since it affect CPU load starting it up everytime, so one option is to use varlink but we are at nmstate 0.2 and we cannot upgrade since nm at nodes is old, so I did try a backport to check how difficult would be.

ffmancera:
> We (kubernetes-nmstate) were investigating ways to keep nmstate alive while running commands since it affect CPU load starting it up everytime, so one option is to use varlink but we are at nmstate 0.2 and we cannot upgrade since nm at nodes is old, so I did try a backport to check how difficult would be.

I think I could help you a little bit with this. Let me ask to the team about what do they think on backporting this. The ideal solution would be to upgrade to 0.4 or 0.3 at least. Thanks!

qinqon:
@ffmancera I don't see varlink here https://github.com/nmstate/nmstate/tree/nmstate-0.3/nmstatectl, maybe is easier to backport to 0.3 ? also I doubt we are going to be able to have new nmstate 0.2 on time for us, so maybe a backport to 0.3 is more feasible so for us.

qinqon:
Going to try to backport it to nmstate 0.3

PR #1361, opened 2020-10-06
packaging: use a 'Recommends' dep for OVS if building for RHEL

acardace:
Keep using a Requires for fedora.

This fixes an issue when building the package for ELN.
https://docs.fedoraproject.org/en-US/eln/

Signed-off-by: Antonio Cardace <acardace@redhat.com>


cathay4t (approved):
Thanks. LGTM.

cathay4t:
Closed as dup of #1374

PR #1363, opened 2020-10-07
[nmstate-0.2] nm, ovs: Fix report crash when OVS has dup iface names

EdDev:
In case of an existing OVS deployment which uses an identical name for
the bridge, port and interface, libnmstate.show() exploded.

It is now possible to report such deployments.

### DIFF
--- a/libnmstate/nm/ovs.py
+++ b/libnmstate/nm/ovs.py
@@ -21,7 +21,6 @@ import logging
 from libnmstate.schema import OVSBridge as OB
 
 from . import connection
-from . import device
 from . import nmclient
 
 
@@ -137,7 +136,12 @@ def get_bridge_info(bridge_device, devices_info):
 
 
 def get_ovs_info(bridge_device, devices_info):
-    port_profiles = _get_slave_profiles(bridge_device, devices_info)
+    ovs_ports_info = (
+        info
+        for info in devices_info
+        if is_ovs_port_type_id(info[1]["type_id"])
+    )
+    port_profiles = _get_slave_profiles(bridge_device, ovs_ports_info)
     ports = _get_bridge_ports_info(port_profiles, devices_info)
     options = _get_bridge_options(bridge_device)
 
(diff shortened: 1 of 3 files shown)


tyll:
Thank you. What about the base branch? Is this fixed there? Would be great to add a test there that shows this and backport it here to ensure there is no regression.
Also, the CI fails, please take a look.

libnmstate/nm/ovs.py
```diff
@@ -137,7 +136,12 @@ def get_bridge_info(bridge_device, devices_info):
 
 
 def get_ovs_info(bridge_device, devices_info):
-    port_profiles = _get_slave_profiles(bridge_device, devices_info)
+    ovs_ports_info = (
+        info
+        for info in devices_info
+        if is_ovs_port_type_id(info[1]["type_id"])
```
tyll, line 142: Seems like `type_id` should be a constant.

libnmstate/nm/ovs.py
```diff
@@ -172,8 +176,21 @@ def _get_bridge_port_info(port_profile, devices_info):
     vlan_mode = port_setting.props.vlan_mode
 
     port_name = port_profile.get_interface_name()
-    port_device = device.get_device_by_name(port_name)
-    port_slave_profiles = _get_slave_profiles(port_device, devices_info)
+    port_device = next(
+        dev
+        for dev, devinfo in devices_info
+        if devinfo["name"] == port_name
+        and is_ovs_port_type_id(devinfo["type_id"])
```
tyll, line 183: name/type_id should probably be constants.

libnmstate/nm/ovs.py
```diff
@@ -172,8 +176,21 @@ def _get_bridge_port_info(port_profile, devices_info):
...
+    )
+    devices_info_excluding_bridges_and_ports = (
+        info
+        for info in devices_info
+        if not is_ovs_bridge_type_id(info[1]["type_id"])
+        and not is_ovs_port_type_id(info[1]["type_id"])
```
tyll, line 189: 1 and type_id should be constants.

EdDev:
> Thank you. What about the base branch? Is this fixed there?

No. On base it requires more work because the report itself is no longer supporting interfaces with the same name.

> Would be great to add a test there that shows this and backport it here to ensure there is no regression.

I am missing an integration test on this change.
The same could be taken to base when we will fix it there.

> Also, the CI fails, please take a look.

I shared the idea with @ffmancera , it was not ready yet.
I fixed the annoying unit tests (I should be blamed for adding those, so this is a self inflicting pain), I will add now the integration one.

libnmstate/nm/ovs.py
EdDev: Yest it should, but as it was not done so far I do not think we should do this now.

libnmstate/nm/ovs.py
EdDev: They should, but these internal keys are not well managed at the moment and adding this under this PR will unfocus it. As we are now in maintenance  mode on this branch, I suggest targeting such change on base only.

libnmstate/nm/ovs.py
EdDev: See previous responses.

libnmstate/nm/ovs.py
ffmancera: I agree with @EdDev here. I would do this on base branch but not on 0.2 or 0.3.

ffmancera (approved):
This looks good to me, thanks for this.

cathay4t (approved):

PR #1366, opened 2020-10-07
[nmstate-0.3] nm, bridge, ovs: Collect only existing profiles

EdDev:
During the reporting flow, connections that are in teardown process
no longer point to a valid profile. Avoid collecting such profiles (in
practice, these are actually `None` objects).


cathay4t (approved):
Approved. Thanks.

PR #1367, opened 2020-10-07
nm, ovs: Fix report crash when OVS has dup iface names

EdDev:
In case of an existing OVS deployment which uses an identical name for
the bridge, port and interface, libnmstate.show() exploded.

It is now possible to report such deployments.

### DIFF
--- a/libnmstate/nm/ovs.py
+++ b/libnmstate/nm/ovs.py
@@ -147,7 +147,12 @@ def get_port_by_port(nmdev):
 
 
 def get_ovs_info(context, bridge_device, devices_info):
-    port_profiles = _get_port_profiles(bridge_device, devices_info)
+    ovs_ports_info = (
+        info
+        for info in devices_info
+        if is_ovs_port_type_id(info[1]["type_id"])
+    )
+    port_profiles = _get_port_profiles(bridge_device, ovs_ports_info)
     ports = _get_bridge_ports_info(context, port_profiles, devices_info)
     options = _get_bridge_options(context, bridge_device)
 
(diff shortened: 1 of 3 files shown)


ffmancera:
Please, let me prepare a full solution for this. Thanks!

tyll:
> Please, let me prepare a full solution for this. Thanks!

A more comprehensive solution can still be added as a follow-up. IMHO it is important to ensure this is fixed on base before merging the PRs on the other branches.

EdDev:
> > Please, let me prepare a full solution for this. Thanks!
> 
> A more comprehensive solution can still be added as a follow-up. IMHO it is important to ensure this is fixed on base before merging the PRs on the other branches.

I am not sure this is possible, the changes needed here are much different and the fixes I used on previous version are not enough, on base another layer fails. So there are two points that need caring here beyond what I just posted:
- Handle the new error seen here (coming from `register_async`).
- Support reporting duplicate names.

If we manage to fix the 1st now, great. If we cannot do it in a few days, I prefer the fixes on older branched to get in and work on this in parallel.

cathay4t:
This PR does not works when ovs internal interface holding the same name with ovs bridge.

The upper level fix has been working at https://github.com/nmstate/nmstate/pull/1370

cathay4t (changes requested):
In stead of filtering `ovs_ports_info`, we should not use interface name as search rule in `_get_port_profiles`.

I will try to work out a better approach.

cathay4t:
Closed as dup of #1370

PR #1369, opened 2020-10-08
Add support to MacVtap interfaces

### LINKED ISSUE #1332: Add MacvTap interface support

ffmancera:
This patch needs a release from Nispor.

### DIFF
--- a/libnmstate/ifaces/macvtap.py
+++ b/libnmstate/ifaces/macvtap.py
@@ -1,0 +1,32 @@
+#
+# Copyright (c) 2020 Red Hat, Inc.
+#
+# This file is part of nmstate
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+
+from libnmstate.schema import MacVtap
+
+from .macvlan import MacVlanIface
+
+
+class MacVtapIface(MacVlanIface):
+    @property
+    def parent(self):
+        return self.config_subtree.get(MacVtap.BASE_IFACE)
+
+    @property
+    def config_subtree(self):
+        return self.raw.get(MacVtap.CONFIG_SUBTREE, {})
--- a/libnmstate/nm/macvlan.py
+++ b/libnmstate/nm/macvlan.py
@@ -18,7 +18,10 @@
 #
 
 from libnmstate.error import NmstateValueError
+from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceType
 from libnmstate.schema import MacVlan
+from libnmstate.schema import MacVtap
 from .common import NM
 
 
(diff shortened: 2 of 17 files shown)


ffmancera:
@cathay4t Ths is now ready for review. Thanks.

libnmstate/nm/plugin.py
```diff
@@ -139,6 +140,9 @@ def get_interfaces(self):
             iface_info.update(nm_lldp.get_info(self.client, dev))
             iface_info.update(nm_team.get_info(dev))
             iface_info.update(get_infiniband_info(applied_config))
+            # This is a workaround needed until all the interface querying
+            # is done by Nispor.
```
cathay4t: Nispor python binding already supported `NisporMacVtap`, while not use it?

cathay4t (changes requested):
The `MacVtapIface` is 99% identical to `MacVlanIface`. How about use `class MacVtapIface(MacVlanIface)` and only redefined the difference?

The same duplication also exists for nispor and test case code.

libnmstate/nm/plugin.py
ffmancera: This is a workaround needed due to Nmstate gathering the interface type from NetworkManager, as we are deciding the interface type using the setting name. If the interface type is not adjusted, Nmstate will fail during verification as NM and Nispor interfaces will not be merged correctly.

I've added this explanation to the function.

ffmancera:
> The `MacVtapIface` is 99% identical to `MacVlanIface`. How about use `class MacVtapIface(MacVlanIface)` and only redefined the difference?
> 
> The same duplication also exists for nispor and test case code.

Done, thank you!

libnmstate/ifaces/macvtap.py
```diff
@@ -0,0 +1,32 @@
...
+from .macvlan import MacVlanIface
+
+
+class MacVtapIface(MacVlanIface):
+    @property
+    def parent(self):
+        return self._config.get(MacVtap.BASE_IFACE)
+
+    @property
+    def _config(self):
```
cathay4t: Redefine a internal property is not good.  Just change that to public interface as `config_subtree`.

libnmstate/nm/macvlan.py
```diff
@@ -31,11 +34,10 @@
...
 
-def create_setting(iface_state, base_con_profile):
-    macvlan = iface_state.get(MacVlan.CONFIG_SUBTREE)
-    if not macvlan:
-        return None
-
+def create_setting(iface_state, base_con_profile, tap=False):
+    macvlan = iface_state.get(MacVlan.CONFIG_SUBTREE) or iface_state.get(
+        MacVtap.CONFIG_SUBTREE
+    )
```
cathay4t, line 42: `macvlan = iface_state.get(MacVtap.CONFIG_SUBTREE) if tap else iface_state.get(MacVlan.CONFIG_SUBTREE)`

libnmstate/nm/plugin.py
cathay4t: Then you need to reword the comment to say NM is require extra attention to get interface type right for mac_vtap and mac_vlan, it has nothing to do with nispor.

libnmstate/nm/profile.py
```diff
@@ -579,7 +587,10 @@ def classify_profile_for_actions(self, groups):
                             ACTION_DEACTIVATE_BEFOREHAND
                         )
                         groups["profiles_to_deactivate_beforehand"].add(self)
-                elif self.iface.type == InterfaceType.MAC_VLAN:
+                elif (
+                    self.iface.type == InterfaceType.MAC_VLAN
+                    or self.iface.type == InterfaceType.MAC_VTAP
+                ):
```
cathay4t: A comment is required why and when it should be deactivated first.

And this seems not a good way to deactivate interface every time it has been included in desire state.

cathay4t (changes requested):

libnmstate/nm/plugin.py
ffmancera: I've moved the comment to the function and fixed the comment. Thanks!

libnmstate/ifaces/macvtap.py
ffmancera: Done. Thank you!

libnmstate/nm/macvlan.py
ffmancera: Done, thanks!

libnmstate/nm/profile.py
ffmancera: I've added the comment and in addition I've added another condition to the statement. It is only being deactivated if the macvlan/macvtap interface is modified not only included in desired state.

cathay4t (approved):

PR #1384, opened 2020-10-18
route_rules: add "state: absent" support

### LINKED ISSUE #1139: Allow partial editing for route rules
Introduce support of `state: absent` to `RouteRule`.

ffmancera:

### DIFF
--- a/libnmstate/route_rule.py
+++ b/libnmstate/route_rule.py
@@ -1,7 +1,25 @@
+#
+# Copyright (c) 2020 Red Hat, Inc.
+#
+# This file is part of nmstate
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+
 from collections import defaultdict
 import logging
 
-from libnmstate.error import NmstateNotImplementedError
 from libnmstate.error import NmstateVerificationError
 from libnmstate.error import NmstateValueError
 from libnmstate.iplib import KERNEL_MAIN_ROUTE_TABLE_ID
@@ -13,6 +31,7 @@ from libnmstate.schema import InterfaceIP
 from libnmstate.schema import RouteRule
 from libnmstate.schema import Route
 
+from .ifaces.base_iface import BaseIface
 from .state import StateEntry
 from .state import state_match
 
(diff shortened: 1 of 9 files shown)


cathay4t (changes requested):
You patch only allows removing specific route rule without support of wildcards.

Please refer to route entry for example implementation.

ffmancera:
@cathay4t This is ready for review now. I've added more integration tests

cathay4t (changes requested):
```yml
---
routes:
  config:
  - destination: 192.168.2.0/24
    metric: 108
    next-hop-address: 192.168.1.3
    next-hop-interface: eth1
    table-id: 200
  - destination: 2001:db8:a::/64
    metric: 108
    next-hop-interface: eth1
    table-id: 200
route-rules:
  config:
    - ip-from: 192.168.3.2/32
      route-table: 200
    - ip-from: 2001:db8:b::/64
      route-table: 200
interfaces:
  - name: eth1
    type: ethernet
    state: up
    mtu: 1500
    ipv4:
      enabled: true
      dhcp: false
      address:
      - ip: 192.168.1.1
        prefix-length: 24
    ipv6:
      enabled: true
      dhcp: false
      autoconf: false
      address:
      - ip: 2001:db8:1::1
        prefix-length: 64
```

```yml
---
route-rules:
  config:
    - state: absent
      route-table: 200
interfaces:
  - name: eth1
```
Does not works.

ffmancera:
> ```yaml
> ---
> routes:
>   config:
>   - destination: 192.168.2.0/24
>     metric: 108
>     next-hop-address: 192.168.1.3
>     next-hop-interface: eth1
>     table-id: 200
>   - destination: 2001:db8:a::/64
>     metric: 108
>     next-hop-address: 2001:db8:1::2
>     next-hop-interface: eth1
>     table-id: 200
> route-rules:
>   config:
>     - ip-from: 192.168.3.2/32
>       route-table: 200
>     - ip-from: 2001:db8:b::/64
>       route-table: 200
> interfaces:
>   - name: eth1
>     type: ethernet
>     state: up
>     mtu: 1500
>     ipv4:
>       enabled: true
>       dhcp: false
>       address:
>       - ip: 192.168.1.1
>         prefix-length: 24
>     ipv6:
>       enabled: true
>       dhcp: false
>       autoconf: false
>       address:
>       - ip: 2001:db8:1::1
>         prefix-length: 64
> ```
> 
> ```yaml
> ---
> route-rules:
>   config:
>     - state: absent
>       route-table: 200
> interfaces:
>   - name: eth1
> ```
> 
> Does not works.

This was happening because the interface were not being marked as changed therefore, if the user set the minimum required state it was not working. Now the interface is being marked as changed. Test case added.

In addition, I've found a bug on IPv6 route rules. I've created a fix and an integration test in a separated commit because it needs to be backported.

cathay4t (changes requested):
All good.

Please add unit test cases testing the wildcard match on all supporting propertied not just
route table id.

ffmancera:
> All good.
> 
> Please add unit test cases testing the wildcard match on all supporting propertied not just
> route table id.

Done. I've added 3 new cases testing all the features with wildcard. I hope that is good :-)

libnmstate/route_rule.py
```diff
@@ -84,38 +102,64 @@ def is_valid(self, config_iface_routes):
...
+                continue
+            new_rules = set()
+            for rule in rule_set:
+                if not rl.match(rule):
+                    new_rules.add(rule)
+            if new_rules != rule_set:
+                self._rules[route_table] = new_rules
+                for iface in ifaces.values():
+                    if iface.is_ipv4_enabled or iface.is_ipv6_enabled:
+                        iface.mark_as_changed()
```
cathay4t: You are marking all interfaces as changed regardless they are holding route rule or not.

Since at this point, we don't know which interface will be used for saving this route rule, I think you can do
it in the `nm` level where you know current route rules and desire route rules or do it in `RouteRuleState.gen_metadata()`

cathay4t (changes requested):

libnmstate/route_rule.py
ffmancera: Done, it is being marked in `RouteRuleState.gen_metadata()`. Thanks and nice catch!

libnmstate/route_rule.py
```diff
@@ -162,6 +202,7 @@ def gen_metadata(self, route_state, ifaces):
             for rule in rules:
                 family = Interface.IPV6 if rule.is_ipv6 else Interface.IPV4
                 route_rule_metadata[iface_name][family].append(rule.to_dict())
+                ifaces[iface_name].mark_as_changed()
```
cathay4t: With `self._cur_rules`, you can check whether desired rule changed from current rule for this table id, and only mark as changed if so.

Meanwhile, if someone removed the all rule from certain table ID, that interface will not be marked as changed. But this is little bit complex and can wait in another PR.

cathay4t (changes requested):

libnmstate/route_rule.py
```diff
@@ -159,6 +199,8 @@ def gen_metadata(self, route_state, ifaces):
                 Interface.IPV4: [],
                 Interface.IPV6: [],
             }
+            if rules != self._cur_rules[route_table]:
+                ifaces[iface_name].mark_as_changed()
```
ffmancera: That handles the case if all the rules has been removed from a table. The structure `self._rules` is {table: [rules]}, therefore when all the rules has been removed it will be `{table: []}` and the generated metadata will be {iface_name: {ipv4: [], ipv6: []}} which means that all the rules has been removed.

In addition, the code is comparing self._rules with self._cur_rules for that table, so if it changed the interface will be marked as changed.

cathay4t (changes requested):
Please use `RouteRuleState.RULES_CHANGED = True` instead of two constants.

ffmancera:
> Please use `RouteRuleState.RULES_CHANGED = True` instead of two constants.

That is not possible. We need to identify it per interface not for the whole RouteRuleState. I've replaced it for a "_changed" = True.

cathay4t (approved):

PR #1385, opened 2020-10-19
Adjusted the comment in nmstate.spec file

### LINKED ISSUE #1378: Outdated comment in specfile or missing Recommends
https://github.com/nmstate/nmstate/blob/2a1632f3f8e8ad4a04721254ebbe9751596c5d16/packaging/nmstate.spec#L44
The specfile mentions that recommends is used but it is not. So the comment needs to be adjusted or the Requires added.

96RadhikaJadhav:
In the fix changed the comment in nmstate.spec file to remove the unused recommends.

### DIFF
--- a/packaging/nmstate.spec
+++ b/packaging/nmstate.spec
@@ -41,8 +41,8 @@ Requires:       python3-nispor
 Summary:        nmstate plugin for OVS database manipulation
 Requires:       python3-%{libname} = %{?epoch:%{epoch}:}%{version}-%{release}
 %if 0%{?rhel}
-# The python-openvswitch rpm package is not in the same repo with nmstate,
-# require only if openvswitch is installed.
+# The python-openvswitch rpm package is not in the same repo with nmstate, require only
+# if openvswitch is installed. If not installed, then it is recommended.
 Requires:       (python3dist(ovs) if openvswitch)
 Recommends:     python3dist(ovs)
 %else


packaging/nmstate.spec
```diff
@@ -41,10 +41,8 @@ Requires:       python3-nispor
 Summary:        nmstate plugin for OVS database manipulation
 Requires:       python3-%{libname} = %{?epoch:%{epoch}:}%{version}-%{release}
 %if 0%{?rhel}
-# The python-openvswitch rpm package is not in the same repo with nmstate,
-# require only if openvswitch is installed.
+# The python-openvswitch rpm package required only if openvswitch is installed.
```
ffmancera: The comment is wrong it should be something like: `The python-openvswitch rpm package is not in the same repo with nmstate, require only if openvswitch is installed. If not installed, then it is recommended`

packaging/nmstate.spec
```diff
@@ -41,10 +41,8 @@ Requires:       python3-nispor
...
 Requires:       (python3dist(ovs) if openvswitch)
-Recommends:     python3dist(ovs)
```
ffmancera, line 47: Please do not remove this. If openvswitch is not present we need to recommend the installation.

ffmancera (changes requested):
Hello, thank for your contribution! I made some comments below, please check them. In addition I would suggest to rewrite the commit message to `packaging: adjust openvswitch comment`. 

Thanks!

packaging/nmstate.spec
```diff
@@ -41,8 +41,9 @@ Requires:       python3-nispor
 Summary:        nmstate plugin for OVS database manipulation
 Requires:       python3-%{libname} = %{?epoch:%{epoch}:}%{version}-%{release}
 %if 0%{?rhel}
-# The python-openvswitch rpm package required only if openvswitch is installed.
+#The python-openvswitch rpm package is not in the same repo with nmstate, require only #if openvswitch is installed. If not installed, then it is recommended.
```
ffmancera: Please, break the line as it was done before and keep the blank space between the comment and the `#`.

ffmancera (changes requested):
It looks good, just a nit. Please, squash the commits in order to have only one with the title `packaging: adjust openvswitch comment`. If you need any help let me know, thanks! :-)

ffmancera (approved):
LGTM, thanks!

PR #1393, opened 2020-10-22
team: deprecate 'ports' in favor of 'port'

### LINKED ISSUE #1328: Please use `port` for `Team.PORT_SUBTREE`
Please deprecate the use of `ports` and align with `LinuxBridge` and `OvsBridge`.

ffmancera:
In order to unify the API keys and make it more consistent, this patch
is deprecating 'ports' in favor of 'port' which is already used by
other interfaces like bond, ovs..

### DIFF
--- a/libnmstate/nm/team.py
+++ b/libnmstate/nm/team.py
@@ -63,12 +63,12 @@ def _convert_team_config_to_teamd_format(team_config, ifname):
     team_config = copy.deepcopy(team_config)
     team_config[TEAMD_JSON_DEVICE] = ifname
 
-    team_ports = team_config.get(Team.PORT_SUBTREE, ())
+    team_ports = team_config.pop(Team.PORT_SUBTREE, ())
     team_ports_formatted = {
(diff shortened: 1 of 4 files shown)


cathay4t (approved):

libnmstate/nm/team.py
```diff
@@ -63,12 +63,12 @@ def _convert_team_config_to_teamd_format(team_config, ifname):
     team_config = copy.deepcopy(team_config)
     team_config[TEAMD_JSON_DEVICE] = ifname
 
-    team_ports = team_config.get(Team.PORT_SUBTREE, ())
+    team_ports = team_config.pop(Team.PORT_SUBTREE, ())
```
EdDev, line 66: Isn’t this part of a different change? Looks like a fix, but I am unsure.

ffmancera: No this is not part of a different change. We need to provide NM the teamd json configuration, on teamd the keyword is  `ports` so before we just needed to collect it and pass it. Now, we need to pop it out as using `port` is not correct on teamd json formatting.

I hope this answers your question, thanks!

PR #1409, opened 2020-11-03
route: mark interface as changed on gen_route_state()

ffmancera:
As discussed for RouteRule, interfaces object should not be modified in
other modules.


cathay4t (approved):
LGTM. Thanks.

PR #1425, opened 2020-11-20
nmstatectl: TOML support for nmstatectl show, edit and set.

### LINKED ISSUE #1362: RFE: Add toml input/output support to nmstatectl
@Conan-Kudo proposed to add toml support for nmstatectl besides yaml and json. It would be nice to check if this is possible and to implement this as an optional format (maybe as an optional dependency).

SudhanAruna:
Added toml format support for nmstatectl functions(show, edit and set).

Signed-off-by: Arunagirinadan Sudharshan <sudhan.aruna.97@gmail.com>
fix #1362

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,9 +96,10 @@ libnmstate.apply(state)
 ```
 
 ```shell
-# use yaml or json formats
+# use yaml, toml or json formats
 nmstatectl set desired-state.yml
 nmstatectl set desired-state.json
+nmstatectl set desired-state.toml
 ```
 
 Edit the current state(python/shell):
--- a/libnmstate/prettystate.py
+++ b/libnmstate/prettystate.py
@@ -22,6 +22,7 @@ from collections.abc import Sequence
 from copy import deepcopy
 import difflib
 import json
+import toml
 
 import yaml
 
@@ -83,6 +84,10 @@ class PrettyState:
     @property
     def json(self):
         return json.dumps(self.state, indent=4, separators=(",", ": "))
+
+    @property
+    def toml(self):
+        return toml.dumps(self.state)
 
 
 def represent_dict(dumper, data):
--- a/nmstatectl/nmstatectl.py
+++ b/nmstatectl/nmstatectl.py
@@ -27,6 +27,7 @@ import sys
 import tempfile
 
 import yaml
+import toml
 
 import libnmstate
 from libnmstate import PrettyState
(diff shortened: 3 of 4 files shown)


SudhanAruna:
Sorry for this late PR. Please let me know if any changes required.

cathay4t (changes requested):
Sorry, I have to be the bad guy to reject this, because:

* TOML does not use indent to identify the ownership of section which does not fit showing network state. For example, toml says:
```toml
[interfaces.ipv4]
enabled = false
address = []
dhcp = false
```
from above I have to scroll back to see which interface this config belong to.

 * There is no real user request toml format.
 * The dependency `python3-toml` does not exists in RHEL 8 official repo. I don't want to be maintainer of `python3-toml` in RHEL 8 just for a CLI output format.
 * `nmstatectl` users can simple use tools to convert JSON/YAML to toml by themselves.

cathay4t:
Sorry, I should state my comments before the work been started.

Conan-Kudo:
> TOML does not use indent to identify the ownership of section which does not fit showing network state.

That's correct. You use namespacing in the section header to do it. This more cleanly maps to how NetworkManager itself represents the configuration internally and via tools like `nmcli`.

> The dependency `python3-toml` does not exists in RHEL 8 official repo. I don't want to be maintainer of `python3-toml` in RHEL 8 just for a CLI output format.

Good news, I suppose: Most of the Python ecosystem increasingly depends on that module anyway, and the Python team already maintains it in Fedora for supporting things like [PEP 518 `pyproject.toml`](https://www.python.org/dev/peps/pep-0518/). So I don't think you'd be alone in maintaining it for RHEL anyway.

EdDev:
Unfortunately, I concur with @cathay4t here.
I have seen no good reason to add this support to the `nmstatectl` when we already have JSON and YAML support.

I guess this can be provided as an external tool for converting TOML to one of the existing formats if one needs such a feature.

Conan-Kudo:
@EdDev There are three reasons, as I brought up with the team in DevConf.US:

1. The general attitude around YAML based configuration that most systems administrators that I've talked to is unblemished hatred. YAML makes it difficult for people to be sure they have gotten the configuration file correct. It's not just a matter of whitespace indentation, but all the other weird properties with YAML. At least at work, that experience has been formed by the problems experienced with Netplan configuration, which *is* YAML.

2. JSON does not support comments. This makes leaving notes and conveying meaning essentially impossible.

3. TOML is very similar to NetworkManager's own native configuration format, while still internally representing as a dictionary structure. That makes it very easy to plug into `nmstate` and make it useful for people to take their configurations and translate it a level up if they're using `nmstate` to configure NetworkManager.

EdDev:
Thank you for the clarification and use case.
I continued the discussion on the issue itself.

ovirt-infra:

PR #1426, opened 2020-11-24
 ovs: Ignore OVS capabilities at get interfaces

qinqon:
included in the schema. This change remove the OVS check so ovs-port
is always excluded.

In case we want to have OVS working at containerize nmstate then we have to find an alternative to `systemctl status openvswitch`

### DIFF
--- a/libnmstate/nm/plugin.py
+++ b/libnmstate/nm/plugin.py
@@ -114,7 +114,6 @@ class NetworkManagerPlugin(NmstatePlugin):
 
     def get_interfaces(self):
         info = []
-        capabilities = self.capabilities
 
         applied_configs = self._applied_configs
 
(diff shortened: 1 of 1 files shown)


libnmstate/nm/plugin.py
```diff
@@ -141,6 +141,9 @@ def get_interfaces(self):
             iface_info.update(get_infiniband_info(applied_config))
             iface_info.update(get_current_macvlan_type(applied_config))
 
+            if iface_info[Interface.TYPE] == InterfaceType.OVS_PORT:
```
EdDev: I think this will explode if the NM ovs plugin is not installed.

libnmstate/nm/plugin.py
```diff
@@ -141,6 +141,9 @@ def get_interfaces(self):
...
+                continue
+
             if NmstatePlugin.OVS_CAPABILITY in capabilities:
```
EdDev: The fix is to replace this condition with `nm_ovs.has_ovs_capability(self.client)`.
We should not process the body only if the NM plugin is not installed, if the OVS is running or not is irrelevant.

EdDev (changes requested):

libnmstate/nm/plugin.py
qinqon: If OVS runnig is irrelevant why was it added ?

EdDev: OVS running check was added to the capabilities report and the condition was left as it.
Based on your findings and the input from the NM team, it may not be needed at all.

cathay4t (changes requested):
The code looks good to me. But the CI failed.

EdDev (approved):
Thank you!

cathay4t (approved):

cathay4t:
@qinqon Could you backport this patch to 0.3 branch?

EdDev:
> @qinqon Could you backport this patch to 0.3 branch?

He did: https://github.com/nmstate/nmstate/pull/1429

tyll:
Since this change is considered important enough to be backported it would be great to also have a test for this scenario to ensure it will continue to work in the future.

PR #1427, opened 2020-11-24
ovs: Ignore ovs-port always

qinqon:
included in the schema. This change just ignore ovs-port even if
openvswitch appear as not running.


cathay4t (approved):
LGTM. Thanks.

EdDev:
@cathay4t this was a backport, #1426 was not merged yet as the solution is probably to just drop the whole ovs capability check here.

PR #1429, opened 2020-11-25
ovs, 0.3: Ignore OVS capabilities at get interfaces

qinqon:
included in the schema. This change remove the OVS check so ovs-port
is always excluded.


EdDev (commented):
nit: The commit message now is not exact, as this has been already done. This change removed the whole OVS compatibility check when processing the interfaces reporting. It is not longer needed.

EdDev (approved):
Thank you!

cathay4t (approved):
LGTM. Thanks.

PR #1434, opened 2020-11-30
veth: add support to veth interfaces

ffmancera:


libnmstate/ifaces/veth.py
```diff
@@ -0,0 +1,32 @@
...
+#
+
+from libnmstate.schema import Veth
+
+from .base_iface import BaseIface
+
+
+class VethIface(BaseIface):
+    @property
+    def config_subtree(self):
```
cathay4t: this is not used.

libnmstate/nm/connection.py
```diff
@@ -206,6 +208,11 @@ def create_new_nm_simple_conn(iface, nm_profile):
             create_macvlan_setting(iface_info, nm_profile, tap=True)
         )
 
+    if Veth.CONFIG_SUBTREE in iface_info:
```
cathay4t: how about `if iface.type == InterfaceType.VETH`.

libnmstate/nm/connection.py
```diff
@@ -206,6 +208,11 @@ def create_new_nm_simple_conn(iface, nm_profile):
...
+        veth_setting = create_veth_setting(iface_info, nm_profile)
```
cathay4t: By passing `VethIface`, you can save `nm/veth` from parsing dictionary. Let's say the `create_veth_setting()` can just use `veth_setting.props.peer = iface.peer`

libnmstate/nmstate.py
```diff
@@ -166,6 +166,10 @@ def _get_interface_info_from_plugins(plugins):
             iface[IFACE_PLUGIN_SRC_METADATA] = [plugin.name]
             iface_name = iface[Interface.NAME]
             iface_type = iface.get(Interface.TYPE, InterfaceType.UNKNOWN)
+            # Veth interface is not allowed to have duplicate names with
+            # ethernet interfaces.
```
cathay4t: This comment does not explain why you are changing the type from veth to ethernet.

cathay4t (changes requested):
Still missing integration test case for add/remove/change

ffmancera:
Not stale :(

libnmstate/ifaces/veth.py
ffmancera: Fixed!

libnmstate/nm/connection.py
ffmancera: Done, I will open a PR modifying this for the other interfaces. I think it is important to keep consistency, thanks!

libnmstate/nm/connection.py
ffmancera: Done! Thanks!

libnmstate/nmstate.py
ffmancera: Modified, the new comment is: 

```
# When NetworkManager does not support Veth interfaces will report  
# them as Ethernet. Therefore, the information will not be merged   
# correctly from Nispor and NetworkManager, in order to solve that, 
# we are considering Veth as Ethernet when merging.  
```

tests/integration/veth_test.py
```diff
@@ -0,0 +1,138 @@
...
+                    Veth.PEER: peer,
+                },
+            }
+        ]
+    }
+    try:
+        libnmstate.apply(d_state)
+        yield d_state
+    finally:
+        d_state[Interface.KEY][0][Interface.STATE] = InterfaceState.ABSENT
```
cathay4t, line 204: You should also remove the peer interface.

cathay4t (changes requested):
Still have a problem: 
You cannot actually bring a veth up(in ip link) through a single yaml file.

tests/integration/veth_test.py
ffmancera: When working with veth interfaces removing a veth interface it should remove the peer interface too. IMO, this is checking there is not a big bug here, if the tests fail because the peer still present means something very bad is happening.

ffmancera:
> Still have a problem:
> You cannot actually bring a veth up(in ip link) through a single yaml file.

I do not think this is going to be possible. NetworkManager is creating the veth interface and the peer during the activation. At that point we are not able to fetch the existing profile of the peer and bring it up.

I am investigating if that is something possible but right now I think Nmstate is not able to do it. Thanks!

tests/integration/veth_test.py
ffmancera: Done, thanks!

ffmancera:
@cathay4t I've added a new test bringing up both interfaces. Basically, we need to bring up one of them and then the other. I have been looking to our work flow and I don't know how it could be possible to do this in one state..

Indeed, I don't think this is possible using NetworkManager or ip link. Because the creation of one interface implies the creation of the peer. So we need to wait until the peer is created to bring it up.

cathay4t (changes requested):
```yml
---
interfaces:
- name: veth1
  type: veth
  state: up
  veth:
    peer: veth2
```

Still has `veth1` as `LOWERLAYERDOWN`.

ffmancera:
> ```yaml
> ---
> interfaces:
> - name: veth1
>   type: veth
>   state: up
>   veth:
>     peer: veth2
> ```
> 
> Still has `veth1` as `LOWERLAYERDOWN`.

That is correct. IMO, nmstate should not bring the peer up if not specified. If an user want to have the veth interface and the peer up it should use the following state:

```yaml
    ---
    interfaces:
    - name: veth1
      type: veth
      state: up
      veth:
        peer: veth1peer
    - name: veth1peer
      type: veth
      state: up
      veth:
        peer: veth1
```

What do you think? I am preparing a patch for this anyway, so if you do not agree I will apply it. Thank you!

ffmancera:
> > ```yaml
> > ---
> > interfaces:
> > - name: veth1
> >   type: veth
> >   state: up
> >   veth:
> >     peer: veth2
> > ```
> > 
> > 
> > Still has `veth1` as `LOWERLAYERDOWN`.
> 
> That is correct. IMO, nmstate should not bring the peer up if not specified. If an user want to have the veth interface and the peer up it should use the following state:
> 
> ```yaml
>     ---
>     interfaces:
>     - name: veth1
>       type: veth
>       state: up
>       veth:
>         peer: veth1peer
>     - name: veth1peer
>       type: veth
>       state: up
>       veth:
>         peer: veth1
> ```
> 
> What do you think? I am preparing a patch for this anyway, so if you do not agree I will apply it. Thank you!

> > ```yaml
> > ---
> > interfaces:
> > - name: veth1
> >   type: veth
> >   state: up
> >   veth:
> >     peer: veth2
> > ```
> > 
> > 
> > Still has `veth1` as `LOWERLAYERDOWN`.
> 
> That is correct. IMO, nmstate should not bring the peer up if not specified. If an user want to have the veth interface and the peer up it should use the following state:
> 
> ```yaml
>     ---
>     interfaces:
>     - name: veth1
>       type: veth
>       state: up
>       veth:
>         peer: veth1peer
>     - name: veth1peer
>       type: veth
>       state: up
>       veth:
>         peer: veth1
> ```
> 
> What do you think? I am preparing a patch for this anyway, so if you do not agree I will apply it. Thank you!

Never mind. I just did the fix. I did not find any use case when this is explicitly needed and if the user want to set one of the interfaces as down can do it later.

ffmancera:
@cathay4t ready for review! Thanks!

ffmancera:
I've added a new test case where the following state is setting the interface link as up.

```yml
---
interfaces:
- name: veth1
  type: veth
  state: up
  veth:
    peer: veth2
```

cathay4t:
> > ```yaml
> > ---
> > interfaces:
> > - name: veth1
> >   type: veth
> >   state: up
> >   veth:
> >     peer: veth2
> > ```
> > 
> > 
> > Still has `veth1` as `LOWERLAYERDOWN`.
> 
> That is correct. IMO, nmstate should not bring the peer up if not specified. If an user want to have the veth interface and the peer up it should use the following state:
> 
> ```yaml
>     ---
>     interfaces:
>     - name: veth1
>       type: veth
>       state: up
>       veth:
>         peer: veth1peer
>     - name: veth1peer
>       type: veth
>       state: up
>       veth:
>         peer: veth1
> ```
> 
> What do you think? I am preparing a patch for this anyway, so if you do not agree I will apply it. Thank you!

Yes. I agree. User need to include both end in desire state.

cathay4t (approved):
LGTM. Thanks.

cathay4t:
@ffmancera Please also fix issue https://github.com/nmstate/nmstate/issues/1461 in another PR.

PR #1437, opened 2020-12-07
ifaces: do not remove unmanaged orphan interfaces

ffmancera:
If there are unmanaged OVS interface present in the network state, NM
may report uncomplete information. Therefore, nmstate could consider
them as orphan and remove it when modifying the network state.

In order to fix this, nmstate should not consider it orphan and remove
it when it is not on desired state or marked as changed.


cathay4t (approved):
LGTM. Thanks.

PR #1450, opened 2020-12-19
Add wireguard support

### LINKED ISSUE #1075: Does it support wireguard interfaces?
Hi,

I was just looking for a project like this :). 
Does it support wireguard interfaces? 
How can they be configured?

shreyagupta30:
This adds wireguard schema and it's tests. 

Signed-off-by : Shreya Gupta <shreyagupta3011@gmail.com>

### DIFF
--- a/libnmstate/ifaces/ifaces.py
+++ b/libnmstate/ifaces/ifaces.py
@@ -43,6 +43,7 @@ from .veth import VethIface
 from .vlan import VlanIface
 from .vxlan import VxlanIface
 from .vrf import VrfIface
+from .wireguard import WireguardIface
 
 
 class _UserSpaceIfaces:
@@ -865,5 +866,7 @@ def _to_specific_iface_obj(info, save_to_disk):
         return MacVtapIface(info, save_to_disk)
     elif iface_type == InterfaceType.VETH:
         return VethIface(info, save_to_disk)
+    elif iface_type == InterfaceType.WIREGUARD:
+        return WireguardIface(info, save_to_disk)
     else:
         return BaseIface(info, save_to_disk)
--- a/libnmstate/ifaces/wireguard.py
+++ b/libnmstate/ifaces/wireguard.py
@@ -1,0 +1,48 @@
+#
+# Copyright (c) 2021 Red Hat, Inc.
+#
+# This file is part of nmstate
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+
+from libnmstate.schema import Wireguard
+
+from .base_iface import BaseIface
+
+
+class WireguardIface(BaseIface):
+    @property
+    def is_virtual(self):
+        return True
+
+    @property
+    def config_subtree(self):
+        return self.raw.get(Wireguard.CONFIG_SUBTREE, {})
+
+    @property
+    def fwmark(self):
+        return self.config_subtree.get(Wireguard.FWMARK)
+
+    @property
+    def listen_port(self):
+        return self.config_subtree.get(Wireguard.LISTEN_PORT)
+
+    @property
+    def private_key(self):
+        return self.config_subtree.get(Wireguard.PRIVATE_KEY)
+
+    @property
+    def peers(self):
+        return self.config_subtree.get(Wireguard.Peer.CONFIG_SUBTREE, [])
--- a/libnmstate/nm/connection.py
+++ b/libnmstate/nm/connection.py
@@ -55,6 +55,7 @@ from .vlan import create_setting as create_vlan_setting
 from .vrf import create_vrf_setting
 from .vxlan import create_setting as create_vxlan_setting
 from .wired import create_setting as create_wired_setting
+from .wireguard import create_wireguard_setting
 
 
 class _ConnectionSetting:
@@ -232,6 +233,8 @@ def create_new_nm_simple_conn(iface, nm_profile):
 
     if iface.ieee_802_1x_conf:
         settings.append(create_802_1x_setting(iface.ieee_802_1x_conf))
+    if iface.type == InterfaceType.WIREGUARD:
+        settings.append(create_wireguard_setting(iface, nm_profile))
 
     ethtool_setting = create_ethtool_setting(iface, nm_profile)
     if ethtool_setting:
--- a/libnmstate/nm/ieee_802_1x.py
+++ b/libnmstate/nm/ieee_802_1x.py
@@ -22,6 +22,7 @@ from libnmstate.schema import Ieee8021X
 
 from .common import NM
 from .common import GLib
+from .secrets import get_secrets
 
 _NM_PROP_MAP = {
     NM.SETTING_802_1X_IDENTITY: Ieee8021X.IDENTITY,
@@ -42,7 +43,12 @@ def get_802_1x_info(context, nm_ac):
     if not nm_setting:
         return {}
 
-    secrets = _get_secrets(context, nm_profile)
+    secrets = get_secrets(
+        context,
+        nm_profile,
+        f"Retrieve 802.1x secrets of profile {nm_profile.get_uuid()}",
+        NM.SETTING_802_1X_SETTING_NAME,
+    )
 
     info = {}
     for nm_prop, key_name in _NM_PROP_MAP.items():
@@ -102,30 +108,3 @@ def _file_path_to_glib_bytes(file_path):
     file_path_bytes = bytearray(f"file://{file_path}".encode("utf8"))
     file_path_bytes.append(0)
     return GLib.Bytes.new(file_path_bytes)
-
-
-def _get_secrets(context, nm_profile):
-    secrets = {}
-    action = f"Retrieve 802.1x secrets of profile {nm_profile.get_uuid()}"
-    context.register_async(action, fast=True)
-    user_data = (context, secrets, action)
-    nm_profile.get_secrets_async(
-        NM.SETTING_802_1X_SETTING_NAME,
-        context.cancellable,
-        _get_secrets_callback,
-        user_data,
-    )
-    context.wait_all_finish()
-    return secrets
-
-
-def _get_secrets_callback(nm_profile, result, user_data):
-    context, secrets, action = user_data
-
-    try:
-        nm_secrets = nm_profile.get_secrets_finish(result)
-    except GLib.Error as e:
-        context.fail(NmstateLibnmError(f"{action} failed: error={e}"))
-
-    context.finish_async(action)
-    secrets.update(nm_secrets[NM.SETTING_802_1X_SETTING_NAME])
--- a/libnmstate/nm/ieee_802_1x.py
+++ b/libnmstate/nm/ieee_802_1x.py
@@ -17,7 +17,6 @@
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 #
 
-from libnmstate.error import NmstateLibnmError
 from libnmstate.schema import Ieee8021X
 
 from .common import NM
--- a/libnmstate/nm/wireguard.py
+++ b/libnmstate/nm/wireguard.py
@@ -1,0 +1,148 @@
+#
+# Copyright (c) 2021 Red Hat, Inc.
+#
+# This file is part of nmstate
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 2.1 of the License, or
+# (at your option) any later version.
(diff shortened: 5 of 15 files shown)


ovirt-infra:

libnmstate/schema.py
```diff
@@ -437,3 +439,14 @@ class Mode:
...
     CONFIG_SUBTREE = "mac-vtap"
+
+class Wireguard:
+    TYPE = InterfaceType.WIREGUARD
+    CONFIG_SUBTREE = "wireguard"
+    FWMARK = "fwmark" 
+    IPV4_AUTO_DEFAULT_ROUTE = "ipv4-auto-default-route"
+    IPV6_AUTO_DEFAULT_ROUTE = "ipv6-auto-default-route"	 
+    LISTEN_PORT = "listen-port" 	 	 
+    MTU = "mtu"
```
cathay4t: The MTU is already included as `Interface.MTU`. There is no need to mention it here.

libnmstate/schema.py
```diff
@@ -437,3 +439,14 @@ class Mode:
 class MacVtap(MacVlan):
     TYPE = InterfaceType.MAC_VTAP
     CONFIG_SUBTREE = "mac-vtap"
+
+class Wireguard:
+    TYPE = InterfaceType.WIREGUARD
+    CONFIG_SUBTREE = "wireguard"
+    FWMARK = "fwmark" 
+    IPV4_AUTO_DEFAULT_ROUTE = "ipv4-auto-default-route"
+    IPV6_AUTO_DEFAULT_ROUTE = "ipv6-auto-default-route"
```
cathay4t: These two are just a wrapper done by NetworkManager.
The wireguard natively does not have them.

IMHO, user can manually get the same result using `fwmark` and `route rule` as mentioned in https://www.wireguard.com/netns/#routing-all-your-traffic

Please list the use cases which these options are required mandatory.

libnmstate/schema.py
```diff
@@ -437,3 +439,14 @@ class Mode:
...
+
+class Wireguard:
+    TYPE = InterfaceType.WIREGUARD
+    CONFIG_SUBTREE = "wireguard"
+    FWMARK = "fwmark" 
+    IPV4_AUTO_DEFAULT_ROUTE = "ipv4-auto-default-route"
+    IPV6_AUTO_DEFAULT_ROUTE = "ipv6-auto-default-route"	 
+    LISTEN_PORT = "listen-port" 	 	 
+    MTU = "mtu" 	 	 
+    PEER_ROUTES = "peer-routes"
```
cathay4t: The peer-routes can be done by existing route support of nmstate. No need to depend NM design for this.

cathay4t (changes requested):
We are not child project of NetworkManager, please don't copy whatever NetworkManager provides to our schema. Nmstate has plugin design which allows other backends.

And nmstate schema should not depend on implementation of certain backend, so you should look on wireguard options itself instead of what NM provides.

The `ipv4-auto-default-route`, `peer-routes` and etc are just convenient stuffs created by NM which can be done via existing nmstate route and route rule schema.

You also missed a lot options:
 * Peer
 * KeepAlive
 * Allowed IP

We will not merge schema-only PR without implementation/test code showing it actual works.

We seems don't have intention to support wireguard server, please also mention that in commit message indication this is only for wireguard client.

ffmancera:
> We are not child project of NetworkManager, please don't copy whatever NetworkManager provides to our schema. Nmstate has plugin design which allows other backends.
> 
> And nmstate schema should not depend on implementation of certain backend, so you should look on wireguard options itself instead of what NM provides.
> 

While I do agree on this, I see an issue here. NetworkManager provides certain options and some of them are wrappers that does not exist on wireguard. I will look at them and help @shreyagupta30 on this :-)

> The `ipv4-auto-default-route`, `peer-routes` and etc are just convenient stuffs created by NM which can be done via existing nmstate route and route rule schema.
> 
> You also missed a lot options:
> 
>     * Peer
> 
>     * KeepAlive
> 
>     * Allowed IP
> 

We are going to collect them and add these.

> 
> We will not merge schema-only PR without implementation/test code showing it actual works.

We are not planning on merging them now. This a PR for initial feedback which is nice.

> 
> We seems don't have intention to support wireguard server, please also mention that in commit message indication this is only for wireguard client.

Yep. That is missing.

By the way, if we want to support other options than the supported from NetworkManager we are going to need WIreguard support on Nispor and therefore, generic netlink support on rtnetlink. I will investigate this.

Thank you! @cathay4t @shreyagupta30

ffmancera:
Not stale :-)

libnmstate/ifaces/ifaces.py
```diff
@@ -44,6 +44,7 @@
 from .vlan import VlanIface
 from .vxlan import VxlanIface
 from .vrf import VrfIface
+from .wireguard import WireguardIface
```
thom311, line 46: the WireGuard maintainers have a strong opinion about the spelling and styling of WireGuard and the API.

In particular, they wish that `WireGuard` is spelled with upper `G`. NetworkManager's API does that too, at places where there is a CamelCase identifier (but it violates that are places like `NM_SETTING_WIREGUARD_SETTING_NAME` or `nm_wireguard_peer_new()`. By doing that, NetworkManager already violates it's own style, because from `NMSettingWireGuard` one would normally expect `NM_SETTING_WIRE_GUARD_SETTING_NAME`.

They probably would also like to see that the configuration options are spelled like why are in `man wg`. That is, `AllowedIPs` and `PresharedKey` instead of `allowed-ips` and `preshared-key`. However, that style already conflicted with NetworkManager's API, so NetworkManager does not do that either. So never mind that.

It's only about `WireGuard` instead of `Wireguard`.

shreyagupta30: Hi @thom311, I will make sure to correct it in my next commit. Thanks a lot for the review. :)

ffmancera: @cathay4t What do you think? Should we break our existing style? I do not like the idea.. but it's true that they have strong opinions on that.

shreyagupta30:
Not stale :)

ffmancera:
Not stale :.-)

ffmancera:
Not stale again :(

shreyagupta30:
Not stale

libnmstate/ifaces/ifaces.py
cathay4t: For internal stuff like class name, you can follow the guidance of WireGuard. But for public property, please use lower case with dash for consistency

libnmstate/nm/ieee_802_1x.py
```diff
@@ -17,11 +17,11 @@
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 #
 
-from libnmstate.error import NmstateLibnmError
 from libnmstate.schema import Ieee8021X
 
 from .common import NM
 from .common import GLib
+from .utils import get_secrets
```
cathay4t: If we have file called `utils.py`, it will eventually grow bloated. Please use `secret.py` instead.

libnmstate/nm/wireguard.py
```diff
@@ -0,0 +1,145 @@
...
+        for allowed_ip in peer.get(Wireguard.Peer.ALLOWED_IPS, []):
+            wireguard_peer.append_allowed_ip(allowed_ip, True)
+
+        wireguard_setting.append_peer(wireguard_peer)
+
+
+def get_info(context, nm_ac):
+    if not nm_ac:
+        return {}
+    nm_profile = nm_ac.get_connection()
```
cathay4t, line 78: Even this is for temporary use(we will move to Nispor for querying the kernel interface), but please use `applied_config` if possible. Please check `nm/ipv4.py` for example.

libnmstate/nm/connection.py
```diff
@@ -142,7 +143,6 @@ def create_new_nm_simple_conn(iface, nm_profile):
 
         if wired_setting:
             settings.append(wired_setting)
-
```
cathay4t: This is unrelated change.

libnmstate/nm/utils.py
```diff
@@ -0,0 +1,48 @@
...
+    return secrets
+
+
+def _get_secrets_callback(nm_profile, result, user_data):
+    context, secrets, action, setting_name = user_data
+
+    try:
+        nm_secrets = nm_profile.get_secrets_finish(result)
+    except GLib.Error as e:
+        context.fail(NmstateLibnmError(f"{action} failed: error={e}"))
```
cathay4t: A `return` is required here.

libnmstate/schema.py
```diff
@@ -459,3 +461,19 @@ class Ieee8021X:
...
     CA_CERT = "ca-cert"
+
+
+class Wireguard:
+    TYPE = InterfaceType.WIREGUARD
+    CONFIG_SUBTREE = "wireguard"
+    FWMARK = "fwmark"
+    LISTEN_PORT = "listen-port"
+    PRIVATE_KEY = "private-key"
+    PEERS = "peers"
```
cathay4t: The `Wireguard.PEERS` should changed to `Wireguard.Peer.CONFIG_SUBTREE` for consistency.

libnmstate/schemas/operational-state.yaml
```diff
@@ -716,6 +717,45 @@ definitions:
...
+                  endpoint:
+                    type: string
+                  persistent-keepalive:
+                    type: integer
+                  preshared-key:
+                    type: string
+                  public-key:
+                    type: string
+                  flags:
+                    type: integer
```
cathay4t: Both `flags` and `peer.flags` are not used. Please remove them.

libnmstate/schemas/operational-state.yaml
```diff
@@ -716,6 +717,45 @@ definitions:
...
+      properties:
+        type:
+          type: string
+          enum:
+            - wireguard
+        wireguard:
+          type: object
+          properties:
+            fwmark:
+              type: integer
```
cathay4t: The wireguard wg CLI tool supports string like `0xf1`, please support it also.

tests/integration/wireguard_test.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def test_modify_wireguard():
+    with wireguard_interface(WG0) as desired_state:
+        wg_state = desired_state[Interface.KEY][0][Wireguard.CONFIG_SUBTREE]
+        wg_state[Wireguard.PRIVATE_KEY] = WG1_PRIVATE_KEY
+        wg_state[Wireguard.LISTEN_PORT] = 9500
+
+        libnmstate.apply(desired_state)
+        assertlib.assert_state(desired_state)
```
cathay4t, line 57: Please use `assert_state_match`, in the future there might be more properties added to wireguard, the `assert_state` will fail then.

tests/integration/wireguard_test.py
```diff
@@ -0,0 +1,139 @@
...
+        wg_state[Wireguard.PEERS] = [
+            {
+                Wireguard.Peer.ALLOWED_IPS: ["192.0.2.1"],
+                Wireguard.Peer.ENDPOINT: "my-wg.example.com:4001",
+                Wireguard.Peer.PRESHARED_KEY: PEER0_PRESHARED_KEY,
+                Wireguard.Peer.PUBLIC_KEY: PEER0_PUBLIC_KEY,
+            },
+        ]
+        libnmstate.apply(desired_state)
+        assertlib.assert_state_match(desired_state)
```
cathay4t, line 87: Please use `assert_state_match` instead.

tests/integration/wireguard_test.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def test_remove_wireguard_peer():
```
cathay4t, line 94: Currently, you are trusting `libnmstate.apply()` to do the proper verification. But your patch is ask NM to write a profile, then we verify it by reading this NM profile in verification stage. That does not sounds like trustable.

Could you consider to use `wg showconf` to actual check whether we are doing correct?

libnmstate/ifaces/wireguard.py
```diff
@@ -0,0 +1,48 @@
...
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+
+from libnmstate.schema import Wireguard
+
+from .base_iface import BaseIface
+
+
+class WireguardIface(BaseIface):
```
cathay4t, line 25: Need unit test case for this class.

cathay4t (changes requested):
In summery, these changes required:
 * Rename `utils.py` to `secret.py` to be more specific.
 * Support hex string for `fwmark`.
 * Improve the test to use `wg showconf` for validation.
 * Need unit test case for `WireguardIface`.

shreyagupta30:
not stale

libnmstate/nm/wireguard.py
ffmancera: Hi Gris, this won't work. applied_config is a `NM.SimpleConnection` and that class does not contain `get_secrets_async()` that is needed to get the secrets of Wireguard. The class required is `NM.RemoteConnection` which is the one that `nm_ac.get_connection()` is provided. I've been working with Shreya on this but I think unfortunately we cannot use `applied_config` here.

shreyagupta30:
Not stale

ffmancera:
This was not stale, we are looking at it. Sorry for the inconvenience.

libnmstate/nm/wireguard.py
ffmancera: What do you think about my comment @cathay4t?

cathay4t: If applied connection does not provide such info, I am OK with saved configure.

PR #1462, opened 2021-01-07
nm.connection: use iface type and object on setting creation

ffmancera:
When creating the setting Nmstate should use the interface type to
decide which setting should create. In addition, the create setting
functions should use `BaseIface` object instead of the dict in order to
simplify the information parsing when creating the setting.

### DIFF
--- a/libnmstate/nm/connection.py
+++ b/libnmstate/nm/connection.py
@@ -24,11 +24,8 @@ import uuid
 from libnmstate.schema import Interface
 from libnmstate.schema import InterfaceType
 from libnmstate.schema import LinuxBridge as LB
-from libnmstate.schema import MacVlan
-from libnmstate.schema import MacVtap
 from libnmstate.schema import OVSBridge as OvsB
 from libnmstate.schema import OVSInterface
-from libnmstate.schema import VRF
 
 from libnmstate.ifaces.bridge import BridgeIface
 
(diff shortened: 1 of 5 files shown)


libnmstate/nm/connection.py
```diff
@@ -196,16 +193,14 @@ def create_new_nm_simple_conn(iface, nm_profile):
...
-        settings.append(create_macvlan_setting(iface_info, nm_profile))
+    if iface.type == InterfaceType.MAC_VLAN:
+        settings.append(create_macvlan_setting(iface, nm_profile))
 
-    if MacVtap.CONFIG_SUBTREE in iface_info:
-        settings.append(
-            create_macvlan_setting(iface_info, nm_profile, tap=True)
-        )
+    if iface.type == InterfaceType.MAC_VTAP:
+        settings.append(create_macvlan_setting(iface, nm_profile, tap=True))
```
cathay4t, line 203: This is incorrect. The `create_mactap_setting()` is taking dict instead of `BaseIface` object.

And it should be __vtap__ not __vlan__

cathay4t (changes requested):

libnmstate/nm/connection.py
ffmancera: In order to re-use code we are using the same function for macvlan and macvtap. I did not modify the if statement logic. About taking dict instead of `BaseIface` object, this patch is modifying that function.

ffmancera:
I've modified the commit message to describe better what has been done in this commit.

libnmstate/nm/connection.py
cathay4t: Oh I forgot that. Then we are good.

cathay4t (approved):
LGTM.

PR #1464, opened 2021-01-07
veth: deactivate first only if peer is changed

ffmancera:
Veth interfaces should deactivate first only if the veth peer is
modified.

### DIFF
--- a/libnmstate/ifaces/veth.py
+++ b/libnmstate/ifaces/veth.py
@@ -22,10 +22,12 @@ from libnmstate.schema import Veth
 from .ethernet import EthernetIface
 
 
-_IS_PEER_METADATA = "_is_peer"
-
-
 class VethIface(EthernetIface):
+    def __init__(self, info, save_to_disk=True):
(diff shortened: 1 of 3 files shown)


libnmstate/ifaces/veth.py
```diff
@@ -38,10 +39,14 @@ def peer(self):
     def is_peer(self):
         return self.raw.get(_IS_PEER_METADATA) is True
 
+    @property
+    def is_peer_changed(self):
+        return self.raw.get(_PEER_CHANGED_METADATA) is True
```
cathay4t: In stead of using `_PEER_CHANGED_METADATA`, you may just set `self._peer_changed = False` in `__init__()` which save a constant.

cathay4t (changes requested):

libnmstate/ifaces/veth.py
ffmancera: Done, thanks!

cathay4t (approved):

PR #1465, opened 2021-01-09
veth: retry import device when activating new veth peer

ffmancera:
NetworkManager takes some extra time on creating the device object for
the new veth peer. Trying to activate an interface without importing the
device will end up on NetworkManager trying to create a new device for
the interface and that will end up on an error.

In order to fix this, Nmstate will retry to import the device up to 5
times with an interval of 0.5 seconds. If the device is not imported
during this time, then nmstate will try to activate it anyway.

### DIFF
--- a/libnmstate/nm/profile.py
+++ b/libnmstate/nm/profile.py
@@ -22,6 +22,8 @@
 
 from distutils.version import StrictVersion
 
+import time
+
 from libnmstate.error import NmstateLibnmError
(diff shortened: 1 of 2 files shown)


ffmancera:
@cathay4t This is removing the random error. Veth support is fully working now! Yay! :-)

cathay4t (approved):
Tested 30 times of the `tests/integration/veth_test.py` with LIBNM 1.30. No failure.

LGTM.

PR #1476, opened 2021-01-14
nm.context: remove refresh_content() calls

### LINKED ISSUE #1398: Maybe we don't need `NetworkManagerPlugin.refresh_content()` any more
We are using nispor for 99% query data, please check whether we still need to refresh content of `NM.Client`.

ffmancera:
NetworkManager plugin refresh_content() is not needed anymore because
most of the information is being retrieved from Nispor.

The function has been moved to `tests/integration/nm/checkpoint.py`
because there is one integration test that need it.

### DIFF
--- a/tests/integration/nm/checkpoint.py
+++ b/tests/integration/nm/checkpoint.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2019 Red Hat, Inc.
+# Copyright (c) 2019-2021 Red Hat, Inc.
 #
 # This file is part of nmstate
 #
@@ -24,6 +24,7 @@ import pytest
 from libnmstate.nm.checkpoint import CheckPoint
 from libnmstate.nm.checkpoint import get_checkpoints
 from libnmstate.error import NmstateConflictError
+from libnmstate.error import NmstateLibnmError
 from libnmstate.error import NmstateValueError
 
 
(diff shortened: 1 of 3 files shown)


tests/integration/nm/checkpoint.py
```diff
@@ -127,11 +126,18 @@ def test_plugin_create_and_rollback_checkpoint(nm_plugin):
...
+            pass
+
+
 def test_plugin_create_checkpoint_and_timeout(nm_plugin):
     plugin = nm_plugin
     plugin.create_checkpoint(timeout=1)
     time.sleep(2)
 
-    nm_plugin.refresh_content()
+    _refresh_content(plugin)
```
cathay4t: Could you explain why this is need in test while the production code does not?

cathay4t (changes requested):

tests/integration/nm/checkpoint.py
ffmancera: The test code does not represent a real behaviour. On Nmstate if the user try to do rollback() and the checkpoint has expired, Nmstate will fail with the following error:

```
E           libnmstate.error.NmstateLibnmError: Checkpoint rollback failed: error=nm-manager-error-quark: checkpoint /org/freedesktop/NetworkManager/Checkpoint/1015 does not exist (12)
```

I've modified the integration test so `_refresh_content()` is not needed anymore.

cathay4t (approved):

PR #1481, opened 2021-01-15
[nmstate-0.3] nm: Use fallback checker on profile deactivation and delete

ffmancera:
When NM is under heave loads, NM might raise timeout error when
try to deactivate or delete a profile, to solve that this patch
introduce the same method in 2407f98
to have a fallback checker on whether profile is deactivated/deleted
every 15 seconds.

No test case required as current `test_lot_of_vlans_with_bridges` test
case has `state: absent` which is good enough for testing this patch.

### DIFF
--- a/libnmstate/nm/active_connection.py
+++ b/libnmstate/nm/active_connection.py
@@ -21,12 +21,14 @@ import logging
 
 from libnmstate.error import NmstateLibnmError
 
+from .common import Gio
 from .common import GLib
 from .common import GObject
 from .common import NM
 
 
 NM_AC_STATE_CHANGED_SIGNAL = "state-changed"
+FALLBACK_CHECKER_INTERNAL = 15
 
 
 class ActivationError(Exception):
@@ -37,6 +39,8 @@ class ActiveConnection:
     def __init__(self, context=None, nm_ac_con=None):
         self._ctx = context
         self._act_con = nm_ac_con
+        self._signal_handler = None
+        self._fallback_checker = None
 
         nmdevs = None
         if nm_ac_con:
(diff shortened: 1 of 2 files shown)


cathay4t (approved):

cathay4t:
I have ammend your patch with:

```patch
diff --git a/libnmstate/nm/connection.py b/libnmstate/nm/connection.py
index 3b8e3dd..af7296f 100644
--- a/libnmstate/nm/connection.py
+++ b/libnmstate/nm/connection.py
@@ -541,7 +541,7 @@ class ConnectionProfile:
     def _delete_fallback_checker_callback(self, action):
         if self.profile:
             for nm_profile in self._ctx.client.get_connections():
-                if nm_profile.get_id() == self.profile.get_id():
+                if nm_profile.get_uuid() == self.profile.get_uuid():
                     return GLib.SOURCE_CONTINUE

         self._fallback_checker_cleanup()
```

ffmancera:
> I have ammend your patch with:
> 
> ```diff
> diff --git a/libnmstate/nm/connection.py b/libnmstate/nm/connection.py
> index 3b8e3dd..af7296f 100644
> --- a/libnmstate/nm/connection.py
> +++ b/libnmstate/nm/connection.py
> @@ -541,7 +541,7 @@ class ConnectionProfile:
>      def _delete_fallback_checker_callback(self, action):
>          if self.profile:
>              for nm_profile in self._ctx.client.get_connections():
> -                if nm_profile.get_id() == self.profile.get_id():
> +                if nm_profile.get_uuid() == self.profile.get_uuid():
>                      return GLib.SOURCE_CONTINUE
> 
>          self._fallback_checker_cleanup()
> ```

As nmstate-0.3 was using the id to identify the profile I thought it could be good to continue using it. Anyway, uuid is much better. Thank you!

PR #1482, opened 2021-01-17
ifaces: mark iface as changed when only when modified

### LINKED ISSUE #1475: Need review all `mark_as_desired()` and `mark_as_changed()`.
1. Try your best not to include interfaces to desire state.
 2. Don't be too smart on DNS interface for NM.

ffmancera:
When matching the child interface state with parent, the child must been
marked as changed only if the state has been modified.


ffmancera:
@cathay4t I did not find any other instance of `mark_as_changed` and `mark_as_desired` that could be removed/modified.

cathay4t (approved):

PR #1492, opened 2021-01-19
nm.active_connetion: retry on profile activation if libnm error happened

### LINKED ISSUE #1477: Retry profile activation if failed once
According to NetworkManager, when the profile activation fails Nmstate should retry the activation in order to avoid certain race condition problems.

ffmancera:
When activating a profile if NetworkManager fails during the activation,
Nmstate should retry it once.

### DIFF
--- a/libnmstate/nm/active_connection.py
+++ b/libnmstate/nm/active_connection.py
@@ -27,7 +27,6 @@ from .common import Gio
 from .common import NM
 from .device import get_nm_dev
 from .device import get_iface_type
-from .device import mark_device_as_managed
 from .ipv4 import is_dynamic as is_ipv4_dynamic
 from .ipv6 import is_dynamic as is_ipv6_dynamic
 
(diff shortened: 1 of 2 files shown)


libnmstate/nm/active_connection.py
```diff
@@ -148,13 +144,27 @@ def wait(ctx, nm_ac, nm_dev):
 
     def _activate_profile_callback(self, nm_client, result, _user_data):
         nm_ac = None
+        retry = _user_data
```
cathay4t: You may just rename the `_user_data` to `retry` in the function argument.

cathay4t (changes requested):

libnmstate/nm/active_connection.py
ffmancera: Done, thank you!

cathay4t (approved):

PR #1502, opened 2021-02-03
[nmstate-0.3] connection: retry on profile activation if libnm error happened

ffmancera:
When activating a profile if NetworkManager fails during the activation,
Nmstate should retry it once.


cathay4t:
@ffmancera To minimize the changes, could you only include the retry code and still keep the `set_managed` workaround?

cathay4t (approved):

PR #1509, opened 2021-02-09
route: workaround for removing IPv6 default gateway

ffmancera:
Due to NM bug [1], Nmstate is not removing IPv6 default gateway. In
order to workaround this, Nmstate is deactivating first the interface if
an IPv6 route has been removed.

Integration test added.

[1] https://bugzilla.redhat.com/1837254


cathay4t (changes requested):
The test case still passes without the fix in libnmstate.

ffmancera:
#1510

ffmancera:
> The test case still passes without the fix in libnmstate.

Done, fixed!

cathay4t (approved):
LGTM. Thanks.

PR #1526, opened 2021-02-22
SR-IOV: increase the verification timeout if SR-IOV is present

ffmancera:
Certain drivers like i40e take a long time to modify the VFs in the
kernel. Nmstate is timing out on verification because of that. In order
to fix this, nmstate is incresing the verification time if SR-IOV is
present on the desired state.

### DIFF
--- a/libnmstate/netapplier.py
+++ b/libnmstate/netapplier.py
@@ -24,6 +24,7 @@ import time
 
 from libnmstate import validator
 from libnmstate.error import NmstateVerificationError
+from libnmstate.schema import InterfaceType
 
 from .nmstate import create_checkpoints
 from .nmstate import destroy_checkpoints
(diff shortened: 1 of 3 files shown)


cathay4t (approved):

libnmstate/netapplier.py
```diff
@@ -109,7 +110,13 @@ def _apply_ifaces_state(plugins, net_state, verify_change, save_to_disk):
         plugin.apply_changes(net_state, save_to_disk)
     verified = False
     if verify_change:
-        for _ in range(VERIFY_RETRY_TIMEOUT):
+        if _net_state_contains_sriov_interface(net_state):
+            # If SR-IOV is present, the verification timeout is being increased
+            # to avoid timeouts due to slow drivers like i40e.
+            verify_retry = VERIFY_RETRY_TIMEOUT * 4
```
tyll: `4` should be a constant. And why 4?

ffmancera: We are retrying up to 4 times more. Yes, I will set it as a constant. 20 seconds should be enough time.

ffmancera: Fixed!

cathay4t:
oVirt has confirmed the fix. Merging.

cathay4t:
The CI is failed on `AttributeError: 'NoneType' object has no attribute 'props'`

PR #1524, opened 2021-02-21
[nmstate-0.3] SR-IOV: Do not create VF profiles automatically

ffmancera:
As VFs are Ethernet type and Nmstate is automatically adding them to the
desired state and marking them as desired, the profile is being created.
This is wrong as Nmstate should only wait for verification of the VFs.

In order to fix this, the new VF interfaces are being marked as so. This
way Nmstate is not creating the profiles anymore.

Integration test case added.

Signed-off-by: Fernando Fernandez Mancera <ffmancera@riseup.net>

PR #1528, opened 2021-02-23
[nmstate-0.3] SR-IOV: fail on verification if `total_vfs` does not match vfs len

ffmancera:


ffmancera:
This PR depends on #1524

cathay4t (approved):

PR #1529, opened 2021-02-24
[nmstate-0.3]  bridge: do not bring up existing ports 

ffmancera:
Nmstate should not manage linux bridge unmanaged ports if they are not
being removed from the linux bridge port list.

Nmstate will set the interface as managed in the following situations:

  * The unmanaged port is being removed from the Linux bridge.
  * The Linux bridge is being marked as absent, therefore Nmstate needs
    to manage the unmanaged port in order to deattach it.

Ref: https://bugzilla.redhat.com/1932247

### DIFF
--- a/libnmstate/ifaces/ifaces.py
+++ b/libnmstate/ifaces/ifaces.py
@@ -194,7 +194,12 @@ class Ifaces:
         """
         for iface in self._ifaces.values():
             if iface.is_up and iface.is_master:
+                cur_iface = self.current_ifaces.get(iface.name)
                 for slave_name in iface.slaves:
+                    if cur_iface and slave_name in cur_iface.slaves:
+                        # Nmstate should not touch the port interface which has
+                        # already been included in current interface.
+                        continue
                     slave_iface = self._ifaces[slave_name]
                     if not slave_iface.is_desired and not slave_iface.is_up:
                         slave_iface.mark_as_up()
--- a/tests/integration/nm/linux_bridge_test.py
+++ b/tests/integration/nm/linux_bridge_test.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2018-2020 Red Hat, Inc.
+# Copyright (c) 2018-2021 Red Hat, Inc.
 #
 # This file is part of nmstate
 #
@@ -26,6 +26,7 @@ import pytest
 import libnmstate
 from libnmstate import nm
 from libnmstate import schema
+from libnmstate.schema import Bridge
 from libnmstate.schema import Interface
 from libnmstate.schema import InterfaceIPv4
 from libnmstate.schema import InterfaceIPv6
@@ -41,6 +42,7 @@ from .testlib import main_context
 
 BRIDGE0 = "brtest0"
 DUMMY1 = "dummy1"
+ETH1 = "eth1"
 
 
 @pytest.fixture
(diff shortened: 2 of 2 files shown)


ffmancera:
Integration tests missing, working on them.

tyll:
It seems that integration tests are missing.

tyll:
> It seems that integration tests are missing.

Also, the integration tests should be added to the newer branches first and then be backported to show that the issues are actually fixed in the newer versions and to avoid regressions.

ffmancera:
Integration tests has been added.

libnmstate/ifaces/ifaces.py
```diff
@@ -177,6 +178,25 @@ def _mark_vf_interface_as_absent_when_sriov_vf_decrease(self):
                     if vf_iface:
                         vf_iface.mark_as_absent_by_desire()
 
+    def _mark_bridge_ports_as_removed(self):
+        """
+        When a Linux bridge port is unmanaged and is removed Nmstate should
```
cathay4t: The unmanaged is dedicated to NetworkManager. The `BaseIface` does not need to aware of that.
Please place the workaround code in `libnmstate/nm`.

Current code in `Ifaces` already mark removed port `BaseIface` as `is_changed` in function `Ifaces._handle_master_slave_list_change()`.

cathay4t (changes requested):

libnmstate/ifaces/ifaces.py
ffmancera: Different solution.

libnmstate/ifaces/ifaces.py
```diff
@@ -194,7 +194,12 @@ def _bring_slave_up_if_not_in_desire(self):
         """
         for iface in self._ifaces.values():
             if iface.is_up and iface.is_master:
+                cur_iface = self.current_ifaces.get(iface.name)
                 for slave_name in iface.slaves:
+                    if cur_iface and slave_name in cur_iface.slaves:
+                        # Nmstate should bring up the port interface if it has
+                        # been added to the state not in all transactions
```
cathay4t: Nmstate should not touch the port interface which has already been included in current interface.

cathay4t (approved):

tests/integration/nm/linux_bridge_test.py
```diff
@@ -313,3 +313,85 @@ def test_bridge_enslave_unmanaged_interface(nm_unmanaged_dummy1):
...
+@pytest.fixture
+def nm_down_unmanaged_dummy1():
+    cmdlib.exec_cmd(
+        f"ip link add name {DUMMY1} type dummy".split(" "), check=True
+    )
+    yield
+    cmdlib.exec_cmd(
+        f"nmcli dev set {DUMMY1} managed yes".split(" "), check=True
+    )
+    time.sleep(1)  # Wait device became managed
```
dankenigsberg, line 329: in a follow up PR: cannot we wait for an event? sleep() is inherently raceful.

tests/integration/nm/linux_bridge_test.py
```diff
@@ -313,3 +313,85 @@ def test_bridge_enslave_unmanaged_interface(nm_unmanaged_dummy1):
...
+    with linux_bridge(
+        BRIDGE0,
+        bridge_subtree_state={
+            LB.OPTIONS_SUBTREE: {LB.STP_SUBTREE: {LB.STP.ENABLED: False}}
+        },
+    ) as desired_state:
+        cmdlib.exec_cmd(
+            f"ip link set {DUMMY1} master {BRIDGE0}".split(), check=True
+        )
+        libnmstate.apply(desired_state)
```
dankenigsberg, line 354: based on the usage of `sleep()` above, don't we need to have one here, too, so that NM has a chance to manage the interface?

tests/integration/nm/linux_bridge_test.py
ffmancera: This is just for testing. On Nmstate implementation it is working with events/signals and async. calls. But on the integration test we are just invoking a shell command.

tests/integration/nm/linux_bridge_test.py
ffmancera: Here we are using Nmstate implementation which uses async. calls and event/signals. So Nmstate is waiting until all the operations finished.

tests/integration/nm/linux_bridge_test.py
dankenigsberg: thanks for the clarification. In a follow-up PR, please consider adding a similar logic to the tests. You don't want the test to be needlessly slow (why wait for 1 second, if NM did its thing in 100mS?) and you don't want the test to succeed just because NM was too slow to set the device to managed. The fact that you need a comment to explain the `sleep()` hints that something like `wait_for_device_is_managed(dummy1, timeout=1)` is in place.

ffmancera: Sure, that is a nice improvement. I will add this to the idea lists we do have.

PR #1538, opened 2021-03-12
test, vlan: add test to change vlan id

smyle10:
Support to change vlan id, after changing, all the old configures and
subordination of the vlan iface should be preserved.

Signed-off-by: Mingyu Shi <mshi@redhat.com>

### DIFF
--- a/tests/integration/vlan_test.py
+++ b/tests/integration/vlan_test.py
@@ -24,10 +24,13 @@ import pytest
 
 import libnmstate
 from libnmstate.error import NmstateVerificationError
+from libnmstate.error import NmstateLibnmError
 from libnmstate.schema import VLAN
 from libnmstate.schema import Interface
 from libnmstate.schema import InterfaceState
 from libnmstate.schema import InterfaceType
+from libnmstate.schema import InterfaceIPv4
+from libnmstate.schema import InterfaceIPv6
 
 from .testlib import assertlib
 from .testlib import statelib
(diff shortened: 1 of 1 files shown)


tests/integration/vlan_test.py
```diff
@@ -234,3 +236,87 @@ def create_two_vlans_state():
             },
         ]
     }
+
+
+@pytest.mark.xfail(reason="https://bugzilla.redhat.com/1722352")
```
ffmancera: This should be marked as strict. It should be something like:

```
@pytest.mark.xfail(reason="https://bugzilla.redhat.com/1722352", strict=True)
```

tests/integration/vlan_test.py
```diff
@@ -234,3 +236,87 @@ def create_two_vlans_state():
...
+                    InterfaceIPv6.DHCP: True,
+                    InterfaceIPv6.AUTOCONF: True,
+                    InterfaceIPv6.AUTO_DNS: True,
+                    InterfaceIPv6.AUTO_GATEWAY: True,
+                    InterfaceIPv6.AUTO_ROUTES: True,
+                    InterfaceIPv6.AUTO_ROUTE_TABLE_ID: 100,
+                },
+            },
+        ]
+    }
```
ffmancera: Instead of using this, I recommend you to do:
```
def test_change_vlan_id(vlan_on_eth1):
  vlan_state = vlan_on_eth1[Interface.KEY][1]
  # Here you can modify the state (change vlan id)
```

tests/integration/vlan_test.py
```diff
@@ -234,3 +236,87 @@ def create_two_vlans_state():
...
+            },
+            {
+                Interface.NAME: vlanb_name,
+                VLAN.CONFIG_SUBTREE: {VLAN.ID: 1002},
+            },
+        ]
+    }
+    libnmstate.apply(desired_state)
+    new_state = libnmstate.show()
+    assert base_state == new_state
```
ffmancera: Instead of this, you can do:
```
assertlib.assert_state(vlan_on_eth1)
```

ffmancera (changes requested):
The test looks good but you should use the existing fixtures and helper functions. Thanks!

tests/integration/vlan_test.py
smyle10: > Instead of using this, I recommend you to do:
> 
> ```
> def test_change_vlan_id(vlan_on_eth1):
>   vlan_state = vlan_on_eth1[Interface.KEY][1]
>   # Here you can modify the state (change vlan id)
> ```

IIUC, vlan_on_eth1 is hard coded using VLAN_IFNAME = "eth1.101", right? But in my case, a name like "eth1.$vlan_id" cannot work: bz1902976. So... days ago I asked you about using global variable to change VLAN_IFNAME :)
If vlan_on_eth1 is really restricted, I think I can use vlan_interface("vlan.a", 101, eth1_up[Interface.KEY][0][Interface.NAME]) ?

tests/integration/vlan_test.py
smyle10: Instead of verify the vlan interface only, I meant to verify the whole state, in case it impacts the others. As this ever happened in certain old versions.

smyle10: But sometimes routes, DNS from DHCP do have a chance to vary automatically...
At least I can use assertlib.assert_state(base_state) to verify all `interfaces`. What do you think?

tests/integration/vlan_test.py
```diff
@@ -234,3 +236,87 @@ def create_two_vlans_state():
             },
         ]
     }
+
+
+@pytest.mark.xfail(reason="https://bugzilla.redhat.com/1722352", strict=True)
```
cathay4t: Please use `raises=AssertError` also

cathay4t (changes requested):

tests/integration/vlan_test.py
```diff
@@ -234,3 +236,87 @@ def create_two_vlans_state():
...
+                    VLAN.CONFIG_SUBTREE: {VLAN.ID: 1001},
+                },
+                {
+                    Interface.NAME: vlanb_name,
+                    VLAN.CONFIG_SUBTREE: {VLAN.ID: 1002},
+                },
+            ]
+        }
+        libnmstate.apply(desired_state)
+        assertlib.assert_state(base_state)
```
cathay4t: Please use `assertlib.assert_state_match` instead.

cathay4t (dismissed):
Not ideal but good enough for first contributor.

smyle10:
> Not ideal but good enough for first contributor.

Thanks, can you elaborate the better solution? I'm glad to learn from it.

cathay4t:
> > Not ideal but good enough for first contributor.
> 
> Thanks, can you elaborate the better solution? I'm glad to learn from it.

 * the test is about changing VLAN ID, setting up two vlans should be considered as fixture(environment setup).
 * No need to manually specify all the states, you can just save the state using `show_only()` before changing VLAN ID. And them verify it after changed the ID.
 * This test is focusing on checking whether changing VLAN ID works or not. Checking whether change vlan ID impact exists settings should be in another test case.

cathay4t:
@smyle10 Fernando has finished both the code and test work in #1587 . Can we close this PR and move on to another task?
Thank you!

ffmancera:
> @smyle10 Fernando has finished both the code and test work in #1587 . Can we close this PR and move on to another task?
> Thank you!

Hello, I am sorry for the overlapping on this. I will make sure it won't happen again. Thank you very much for your contribution and efforts.

cathay4t:
Closing as the test is already merged.

PR #1540, opened 2021-03-19
[nmstate-0.3] ovs: do not set ovs-bridge as master if duplicate names are present

ffmancera:
When OVS interfaces are present with duplicate names i.e ovs-bridge,
ovs-interface and ovs-port with the same name, Nmstate is merging the
information to show it.

As Nmstate needs to store the DNS configuration as part of the
configuration of an interface, When the ovs-bridge is being reported
first than the ovs-interface, the IP configuration is being removed
internally. If the DNS configuration was stored on the ovs-bridge, it
will fail because the IP configuration is required.

This solution only solves the following error and does not provide
duplicate names support.

```
  File "/usr/bin/nmstatectl", line 11, in <module>
    load_entry_point('nmstate==0.3.4', 'console_scripts', 'nmstatectl')()
  File "/usr/lib/python3.6/site-packages/nmstatectl/nmstatectl.py", line 67, in main
    return args.func(args)
  File "/usr/lib/python3.6/site-packages/nmstatectl/nmstatectl.py", line 267, in apply
    args.save_to_disk,
  File "/usr/lib/python3.6/site-packages/nmstatectl/nmstatectl.py", line 289, in apply_state
    save_to_disk=save_to_disk,
  File "/usr/lib/python3.6/site-packages/libnmstate/netapplier.py", line 71, in apply
    net_state = NetState(desired_state, current_state, save_to_disk)
  File "/usr/lib/python3.6/site-packages/libnmstate/net_state.py", line 58, in __init__
    self._ifaces.gen_dns_metadata(self._dns, self._route)
  File "/usr/lib/python3.6/site-packages/libnmstate/ifaces/ifaces.py", line 408, in gen_dns_metadata
    iface_metadata = dns_state.gen_metadata(self, route_state)
  File "/usr/lib/python3.6/site-packages/libnmstate/dns.py", line 116, in gen_metadata
    "name servers: %s" % server
libnmstate.error.NmstateValueError: Failed to find suitable interface for saving DNS name servers: 8.8.8.8
```

Ref: https://bugzilla.redhat.com/1939557

### DIFF
--- a/libnmstate/ifaces/base_iface.py
+++ b/libnmstate/ifaces/base_iface.py
@@ -295,8 +295,15 @@ class BaseIface:
     def gen_metadata(self, ifaces):
         if self.is_master and not self.is_absent:
             for slave_name in self.slaves:
-                slave_iface = ifaces[slave_name]
-                slave_iface.set_master(self.name, self.type)
+                if slave_name != self.name:
+                    slave_iface = ifaces[slave_name]
+                    slave_iface.set_master(self.name, self.type)
+                else:
+                    logging.warning(
+                        f"The interface {self.name} is setting {slave_name} as"
+                        " port. Multiple interfaces with names are not"
+                        " supported and unexpected errors may occur."
+                    )
 
     def update(self, info):
         self._info.update(info)
--- a/tests/integration/nm/ovs_test.py
+++ b/tests/integration/nm/ovs_test.py
@@ -34,12 +34,16 @@ from libnmstate.schema import VLAN
 from .testlib import main_context
 from ..testlib import cmdlib
 from ..testlib import assertlib
+from ..testlib.retry import retry_till_true_or_timeout
 
 
 BRIDGE0 = "brtest0"
 ETH1 = "eth1"
 ETH2 = "eth2"
 BOND0 = "bond0"
+OVS_DUP_NAME = "br-ex"
+
+VERIFY_RETRY_TMO = 5
 
 
 @pytest.fixture
@@ -378,6 +382,105 @@ def ovs_bridge_over_bond_system_iface_with_same_name(eth1_up, eth2_up):
     time.sleep(1)
 
 
+def _nmcli_ovs_bridge_with_ipv4_dns():
+    nmcli_ovs_interface = (
+        "nmcli",
+        "connection",
+        "add",
+        "type",
+        "ovs-interface",
+        "slave-type",
+        "ovs-port",
+        "conn.interface",
+        "br-ex",
+        "master",
+        "ovs-port-br-ex",
+        "con-name",
+        "ovs-if-br-ex",
+        "ipv4.method",
+        "manual",
+        "ipv4.addr",
+        "192.0.2.2/24",
+        "ipv4.dns",
+        "192.0.2.1",
+        "ipv4.routes",
+        "0.0.0.0/0 192.0.2.1",
+    )
+    cmdlib.exec_cmd(nmcli_ovs_interface, check=True)
+
+
+def _verify_ovs_activated(ovs_name):
+    ret, out, err = cmdlib.exec_cmd(
+        f"nmcli --field GENERAL.STATE device show {ovs_name}".split(),
+        check=True,
+    )
+    connected = "connected" in out
+    ret, out, err = cmdlib.exec_cmd(
+        f"nmcli --field IP4.ADDRESS device show {ovs_name}".split(),
+        check=True,
+    )
+    ipv4_configured = "192.0.2.2/24" in out
+    ret, out, err = cmdlib.exec_cmd(
+        f"nmcli --field IP4.ROUTE device show {ovs_name}".split(), check=True,
+    )
+    route_configured = "0.0.0.0/0" in out
+    return connected and ipv4_configured and route_configured
+
+
+@pytest.fixture
+def ovs_bridge_first_and_ovs_interface_with_same_name_ipv4():
+    # The order on this function is important. The OVS bridge must be defined
+    # before the OVS interface.
+    cmdlib.exec_cmd(
+        "nmcli connection add type ovs-port conn.interface br-ex master br-ex "
+        "con-name ovs-port-br-ex".split(),
+        check=True,
+    )
+    cmdlib.exec_cmd(
+        "nmcli connection add type ovs-bridge con-name br-ex conn.interface "
+        "br-ex".split(),
+        check=True,
+    )
+    _nmcli_ovs_bridge_with_ipv4_dns()
+    # Wait a little bit for NM to activate above interfaces to do not hit race
+    # problems.
+    assert retry_till_true_or_timeout(
+        VERIFY_RETRY_TMO, _verify_ovs_activated, OVS_DUP_NAME
+    )
+    yield
+    cmdlib.exec_cmd(
+        "nmcli connection del ovs-port-br-ex br-ex ovs-if-br-ex".split(),
+        check=True,
+    )
+
+
+@pytest.fixture
+def ovs_interface_first_and_ovs_bridge_with_same_name_ipv4():
+    # The order on this function is important. The OVS interface must be
+    # defined before the OVS bridge.
+    cmdlib.exec_cmd(
+        "nmcli connection add type ovs-port conn.interface br-ex master br-ex "
+        "con-name ovs-port-br-ex".split(),
+        check=True,
+    )
+    _nmcli_ovs_bridge_with_ipv4_dns()
+    cmdlib.exec_cmd(
+        "nmcli connection add type ovs-bridge con-name br-ex conn.interface "
+        "br-ex".split(),
+        check=True,
+    )
+    # Wait a little bit for NM to activate above interfaces to do not hit race
+    # problems.
+    assert retry_till_true_or_timeout(
+        VERIFY_RETRY_TMO, _verify_ovs_activated, OVS_DUP_NAME
+    )
+    yield
+    cmdlib.exec_cmd(
+        "nmcli connection del ovs-port-br-ex br-ex ovs-if-br-ex".split(),
+        check=True,
+    )
+
+
 @pytest.mark.tier1
 def test_create_vlan_over_ovs_system_interface_bond_with_same_name(
     ovs_bridge_over_bond_system_iface_with_same_name,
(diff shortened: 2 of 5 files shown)


tyll:
It seems to fix the problem with this reproducer:

```bash
nmcli c add type ovs-port conn.interface br-ex master br-ex con-name ovs-port-br-ex
nmcli c add type ovs-bridge \
            con-name br-ex \
            conn.interface br-ex
nmcli c add type ovs-interface slave-type ovs-port conn.interface br-ex master ovs-port-br-ex con-name \
            ovs-if-br-ex ipv4.method manual ipv4.addr 192.0.2.2/24 ipv4.dns 192.0.2.1 ipv4.routes "0.0.0.0/0 192.0.2.1"

nmstatectl set <<EOF
interfaces:
  - name: test
    type: ethernet
    state: absent
EOF

```

tyll:
@ffmancera packit is not working

ffmancera:
> @ffmancera packit is not working

I think this is a packit issue. I will ask them.

cathay4t (changes requested):
The patch looks dangerous to me as it impact on more use cases than we are working around.

Could you investigate how hard to enable real duplicate name support in the upper layer?

ffmancera:
> The patch looks dangerous to me as it impact on more use cases than we are working around.
> 
> Could you investigate how hard to enable real duplicate name support in the upper layer?

Supporting duplicate name in the upper layer will be even more dangerous because it requires the following changes:

* Address the interfaces by ifname+type.
* Modify set_controller and other gen_metadata functions in order to generate the controller metadata correctly.

I think these changes affects even more use cases and it is dangerous. IMO, it is an unnecessary risk, what do you think?

tests/integration/nm/ovs_test.py
```diff
@@ -378,6 +378,32 @@ def ovs_bridge_over_bond_system_iface_with_same_name(eth1_up, eth2_up):
...
+        "br-ex".split(),
+        check=True,
+    )
+    nmcli_ovs_interface = (
+        "nmcli c add type ovs-interface slave-type ovs-port "
+        "conn.interface br-ex master ovs-port-br-ex con-name ovs-if-br-ex "
+        "ipv4.method manual ipv4.addr 192.0.2.2/24 ipv4.dns 192.0.2.1 "
+        "ipv4.routes".split()
+    )
+    nmcli_ovs_interface.append("0.0.0.0/0 192.0.2.1")
```
tyll: using a tuple instead of a string with split() would make this simpler, then no append is needed. Why do you prefer the string+split?

tests/integration/nm/ovs_test.py
```diff
@@ -378,6 +378,32 @@ def ovs_bridge_over_bond_system_iface_with_same_name(eth1_up, eth2_up):
     time.sleep(1)
 
 
+@pytest.fixture
+def ovs_bridge_and_ovs_interface_with_same_name_ipv4():
+    cmdlib.exec_cmd(
+        "nmcli c add type ovs-port conn.interface br-ex master br-ex con-name "
```
tyll: IMHO it is better style to use non-abbreviated names in scripts/documentation, such as "nmcli connection add ...."

tests/integration/nm/ovs_test.py
```diff
@@ -406,3 +432,19 @@ def test_create_vlan_over_ovs_system_interface_bond_with_same_name(
                 ]
             }
         )
+
+
+@pytest.mark.tier1
+def test_modify_state_with_ovs_bridge_interface_with_ipv4_dns(
```
tyll: The test does not make it clear that the duplicate interfaces need to be reported by the NM plugin in a certain order to trigger the error. Ideally, this should be enforced to ensure that the actual error condition is tested.

libnmstate/ifaces/base_iface.py
```diff
@@ -295,8 +295,9 @@ def master_type(self):
     def gen_metadata(self, ifaces):
         if self.is_master and not self.is_absent:
             for slave_name in self.slaves:
-                slave_iface = ifaces[slave_name]
-                slave_iface.set_master(self.name, self.type)
+                if slave_name != self.name:
```
tyll, line 298: The else case should probable log a warning to make the problematic state more visible.

tyll:
This issue was hard to reproduce because the NM plugin does not report interfaces in a stable order, what do you think about sorting the interfaces based on their name, too, to make nmstate more deterministic?

tests/integration/nm/ovs_test.py
ffmancera: Well, in normal situations is easier to read IMO. But yes, here I will change it for a tuple.

tests/integration/nm/ovs_test.py
ffmancera: +1

tests/integration/nm/ovs_test.py
ffmancera: Done.

tests/integration/nm/ovs_test.py
ffmancera: I've changed the test name, please could you re-check it? Thanks!

libnmstate/ifaces/base_iface.py
ffmancera: Done!

ffmancera:
> This issue was hard to reproduce because the NM plugin does not report interfaces in a stable order, what do you think about sorting the interfaces based on their name, too, to make nmstate more deterministic?

Well, sorting the interfaces is hard. We should do it about name, maybe type? IMO, nmstate should not care about the order here..

cathay4t:
> > The patch looks dangerous to me as it impact on more use cases than we are working around.
> > Could you investigate how hard to enable real duplicate name support in the upper layer?
> 
> Supporting duplicate name in the upper layer will be even more dangerous because it requires the following changes:
> 
>     * Address the interfaces by ifname+type.
> 
>     * Modify set_controller and other gen_metadata functions in order to generate the controller metadata correctly.
> 
> 
> I think these changes affects even more use cases and it is dangerous. IMO, it is an unnecessary risk, what do you think?

Then I will support keep doing workaround for duplicate name in nmstate-0.3.

tests/integration/nm/ovs_test.py
```diff
@@ -378,6 +378,102 @@ def ovs_bridge_over_bond_system_iface_with_same_name(eth1_up, eth2_up):
...
+        "con-name ovs-port-br-ex".split(),
+        check=True,
+    )
+    cmdlib.exec_cmd(
+        "nmcli connection add type ovs-bridge con-name br-ex conn.interface "
+        "br-ex".split(),
+        check=True,
+    )
+    nmcli_ovs_interface = (
+        "nmcli",
```
tyll, line 387: Does black require all items individually? If not, you can make this more readable as

```
nmcli connection add
type ovs-interface
slave-type ovs-port
conn.interface br-ex
...
```

tests/integration/nm/ovs_test.py
```diff
@@ -378,6 +378,102 @@ def ovs_bridge_over_bond_system_iface_with_same_name(eth1_up, eth2_up):
...
+        "192.0.2.2/24",
+        "ipv4.dns",
+        "192.0.2.1",
+        "ipv4.routes",
+        "0.0.0.0/0 192.0.2.1",
+    )
+    cmdlib.exec_cmd(nmcli_ovs_interface, check=True)
+    # Wait a little bit for NM to activate above interfaces to do not hit race
+    # problems.
+    time.sleep(1)
```
tyll: It would be nicer not to sleep but check the NM state regularly with a timeout.

tests/integration/nm/ovs_test.py
```diff
@@ -378,6 +378,102 @@ def ovs_bridge_over_bond_system_iface_with_same_name(eth1_up, eth2_up):
...
+@pytest.fixture
+def ovs_interface_first_and_ovs_bridge_with_same_name_ipv4():
+    # The order on this function is important. The OVS interface must be
+    # defined before the OVS bridge.
+    cmdlib.exec_cmd(
+        "nmcli connection add type ovs-port conn.interface br-ex master br-ex "
+        "con-name ovs-port-br-ex".split(),
+        check=True,
+    )
+    nmcli_ovs_interface = (
```
tyll: seems like these should be functions now

tests/integration/nm/ovs_test.py
```diff
@@ -378,6 +378,102 @@ def ovs_bridge_over_bond_system_iface_with_same_name(eth1_up, eth2_up):
...
+    time.sleep(1)
+    yield
+    cmdlib.exec_cmd(
+        "nmcli connection del ovs-port-br-ex br-ex ovs-if-br-ex".split(),
+        check=True,
+    )
+
+
+@pytest.fixture
+def ovs_interface_first_and_ovs_bridge_with_same_name_ipv4():
```
tyll, line 458: It would probably be nicer to have a more generic nmcli fixture or context.

tests/integration/nm/ovs_test.py
```diff
@@ -406,3 +502,35 @@ def test_create_vlan_over_ovs_system_interface_bond_with_same_name(
...
+
+
+@pytest.mark.tier1
+def test_modify_state_with_ovs_dup_name_ovs_bridge_first_with_ipv4_dns(
+    ovs_bridge_first_and_ovs_interface_with_same_name_ipv4,
+):
+    desired_state = {
+        Interface.KEY: [
+            {
+                Interface.NAME: "eth1",
```
tyll: Is there a constant for eth1 that we can/should use nowadays?

tests/integration/nm/ovs_test.py
ffmancera: Nope, that is not accepted by black. Black requires all items individually

tests/integration/nm/ovs_test.py
ffmancera: Well, this is a very specific case. I do not think we can have a more generic fixture or context..

tests/integration/nm/ovs_test.py
ffmancera: Done, thanks!

tests/integration/nm/ovs_test.py
ffmancera: The ovs interface is now a function. I think for the other it is not necessary.

tests/integration/nm/ovs_test.py
ffmancera: Done, thank you!

tests/integration/nm/ovs_test.py
tyll: Something like this would be possible:

```
with NMCliConnection() as nmc:
    nmc.add("ovs-port-br-ex", ["type", "ovs-port", "conn.interface", "br-ex", ...])
    nmc.add("br-ex", ["type", "ovs-bridge", ...])
    nmc.add("ovs-if-br-ex", ["type", "ovs-interface", ...])
    libnmstate.apply(...)
    assert ...
```

Then NMCliConnection can run "nmcli connection add" for each ".add()" invocation and "nmcli connection delete" for each added connection when the context is left.

tests/integration/nm/ovs_test.py
```diff
@@ -406,3 +499,35 @@ def test_create_vlan_over_ovs_system_interface_bond_with_same_name(
...
+    desired_state = {
+        Interface.KEY: [
+            {
+                Interface.NAME: ETH1,
+                Interface.TYPE: InterfaceType.ETHERNET,
+                Interface.STATE: InterfaceState.UP,
+            }
+        ]
+    }
+    libnmstate.apply(desired_state)
```
tyll, line 527: Technically, this is missing an assertion to assert that the apply did what it was expected to do.

libnmstate/ifaces/base_iface.py
```diff
@@ -295,8 +295,15 @@ def master_type(self):
...
+                    slave_iface = ifaces[slave_name]
+                    slave_iface.set_master(self.name, self.type)
+                else:
+                    logging.warning(
+                        f"The interface {self.name} is setting {slave_name} as"
+                        " port. Interface duplicate names are not supported "
```
tyll:
```suggestion
                        " port. Multiple interfaces with the same names are not supported "
```

ffmancera: Done, thanks!

tests/integration/nm/ovs_test.py
ffmancera: Done, I've added the assert. Thanks!

cathay4t (approved):

PR #1544, opened 2021-03-23
packit: use actions instead of custom commands

TomasTomecek:
This should fix the build problems, sorry about the issues.

also:
```
packit: centos-stream target is now called centos-stream-8
```


TomasTomecek:
Indeed, the stream build was not submitted :/

https://github.com/packit/packit/issues/1173

ffmancera (approved):
LGTM, thanks!

PR #1549, opened 2021-03-24
[nmstate-0.3] Fix packit issue with el8

ffmancera:


ffmancera:
I am merging this without approval because is a fix from packit and already merged on base branch.

PR #1553, opened 2021-03-29
tests, bond: check bond takes the mac from the new port

ffmancera:
Due to commit [1], NetworkManager is not keeping the bond mac address
when replacing ports. This change affects NM 1.30 and greater.

[1] https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/190fd9aa9f3fbf5705c2b80b9fc64c89d22b7593

### DIFF
--- a/tests/integration/bond_test.py
+++ b/tests/integration/bond_test.py
@@ -41,6 +41,7 @@ from .testlib import cmdlib
 from .testlib import statelib
 from .testlib.assertlib import assert_mac_address
 from .testlib.bondlib import bond_interface
+from .testlib.env import nm_major_minor_version
 from .testlib.ifacelib import get_mac_address
 from .testlib.ifacelib import ifaces_init
 from .testlib.vlan import vlan_interface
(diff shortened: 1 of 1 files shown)


tests/integration/bond_test.py
```diff
@@ -1012,3 +1013,21 @@ def test_create_bond_with_copy_mac_from(eth1_up, eth2_up):
     ):
         current_state = statelib.show_only((BOND99,))
         assert_mac_address(current_state, eth2_mac)
+
+
+@pytest.mark.xfail(
+    nm_major_minor_version() < 1.30,
+    reason="Before 1.30 the bond was not keeping the mac",
+)
+def test_replacing_port_do_not_keep_mac_of_bond(bond99_with_eth2, eth1_up):
```
tyll: The test name should describe what is expected here. The current name contradicts the reason, too.

Also, IMHO, it would be good to link the test to a BZ and make somehow else sure that everyone can easily understand in a year why we have this test.

ffmancera: The reason was incorrect. Fixed. I've included the BZ in the reason. I think that is enough to understand in a year why we have this test as the BZ contains all the details of the issue.

tyll:
works for me

tests/integration/bond_test.py
```diff
@@ -1012,3 +1013,24 @@ def test_create_bond_with_copy_mac_from(eth1_up, eth2_up):
     ):
         current_state = statelib.show_only((BOND99,))
         assert_mac_address(current_state, eth2_mac)
+
+
+@pytest.mark.xfail(
```
cathay4t, line 1018: you need strict also.

cathay4t (changes requested):

tests/integration/bond_test.py
ffmancera: Done, thanks!

cathay4t (approved):

Issue #1557, opened 2021-03-30
Clean up and improve unit and integration tests


ffmancera:
The integration and unit tests require some rework and clean up. This issue is tracking the progress of the tasks needed.

* [ ] Replace all the "sleep()" calls with 'retry_till_true_or_timeout()`
* [ ] Create a `exec_line()` function on cmdlib. It should handle the "split()" of the command string inside.
* [ ] Create a general nmcli fixture or context like https://github.com/nmstate/nmstate/pull/1540#discussion_r603956821
* [ ] Add support for duplicate names in `assert_state` and `assert_state_match` functions.

tyll:
The first item is an optimization to avoid race conditions and speed up the CI. The other two items are ideas to make the tests more readable but it is not a full solution just some ideas.

PR #1558, opened 2021-03-31
[nmstate-1.0] tests, ovs: check ovs duplicate names with ipv4 dns is supported

ffmancera:
Ref: https://bugzilla.redhat.com/1939557

### DIFF
--- a/tests/integration/nm/ovs_test.py
+++ b/tests/integration/nm/ovs_test.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2019-2020 Red Hat, Inc.
+# Copyright (c) 2019-2021 Red Hat, Inc.
 #
 # This file is part of nmstate
 #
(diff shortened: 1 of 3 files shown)


tests/integration/nm/ovs_test.py
```diff
@@ -260,3 +264,136 @@ def test_ovs_bridge_remoev_external_ids(self, disable_ovsdb_plugin):
...
+    desired_state = {
+        Interface.KEY: [
+            {
+                Interface.NAME: ETH1,
+                Interface.TYPE: InterfaceType.ETHERNET,
+                Interface.STATE: InterfaceState.UP,
+            }
+        ]
+    }
+    libnmstate.apply(desired_state)
```
cathay4t, line 382: Since we are supposed to work well on supporting this duplicate name use case, can you use `assert_state_match` on old OVS interfaces.

cathay4t (changes requested):

tests/integration/nm/ovs_test.py
ffmancera: `assert_state_match` does not support duplicate interfaces. It requires some changes, I am adding this to the issue #1557. I hope this could be merged without it for now.

cathay4t (approved):

PR #1563, opened 2021-04-12
test, veth: remove peer connection on veth removal

smyle10:
When taking a veth absent, if not present its peer in desired state, then peer's connection will remain.
That causes NM recreate veth pair by automatically activating this connection.
https://bugzilla.redhat.com/show_bug.cgi?id=1943525

### DIFF
--- a/tests/integration/nm/veth_test.py
+++ b/tests/integration/nm/veth_test.py
@@ -1,0 +1,53 @@
+#
+# Copyright (c) 2021 Red Hat, Inc.
+#
+# This file is part of nmstate
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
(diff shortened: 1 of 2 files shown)


tests/integration/veth_test.py
```diff
@@ -316,3 +317,18 @@ def veth_interface(ifname, peer):
...
+    nm_major_minor_version() <= 1.28,
+    reason="Modifying veth interfaces is not supported on NetworkManager.",
+)
+@pytest.mark.xfail(reason="https://bugzilla.redhat.com/1943525", strict=True)
+def test_remove_peer_connection():
+    with veth_interface(VETH1, VETH1PEER) as desired_state:
+        desired_state[Interface.KEY][0][Interface.STATE] = InterfaceState.ABSENT
+        libnmstate.apply(desired_state)
+        assert (
+            cmdlib.exec_cmd(f"nmcli connection show {VETH1PEER}".split())[0]
```
cathay4t: If you are using `nmcli` to check whether `VETH1PEER` still exists, you should place this test in to `tests/integration/nm` folder.

You may also change to `assert len(statelib.show_only((VETH1PEER, ))[Interface.NAME]) == 0` which does not require the nm specific command.

cathay4t (changes requested):

tests/integration/veth_test.py
smyle10: After taking `VETH1` absent, this veth pair was removed from kernel, including `VETH1PEER`, so `len(statelib.show_only((VETH1PEER, ))[Interface.NAME])` is 0, there's no problem here.
But the NM profile of `VETH1PEER` remained, that's what I need to check(expecting the NM profile was deleted too)

cathay4t (approved):

PR #1567, opened 2021-04-19
specfile: fix installation of nmstate-varlink.service

ffmancera:
On the nmstate.spec file the installation was done using `cp -a` which
is incorrect. In order to install the nmstate-varlink.service it should
use the `install -p -m 644`.

### DIFF
--- a/packaging/nmstate.spec
+++ b/packaging/nmstate.spec
@@ -66,7 +66,9 @@ This package contains the nmstate plugin for OVS database manipulation.
 
 %install
 %py3_install
-mkdir -p %{buildroot}%{_unitdir} && cp -a %{buildroot}%{python3_sitelib}/nmstatectl/nmstate-varlink.service %{buildroot}%{_unitdir}/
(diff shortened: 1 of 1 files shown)


cathay4t (approved):

packaging/nmstate.spec
```diff
@@ -66,7 +66,9 @@ This package contains the nmstate plugin for OVS database manipulation.
 
 %install
 %py3_install
-mkdir -p %{buildroot}%{_unitdir} && cp -a %{buildroot}%{python3_sitelib}/nmstatectl/nmstate-varlink.service %{buildroot}%{_unitdir}/
+mkdir -p %{buildroot}%{_unitdir}
+install -p -m 644 %{buildroot}%{python3_sitelib}/nmstatectl/nmstate-varlink.service \
```
tyll, line 70: Why is it installed to the python3_sitelib at all? Seems like it should only be part of MANIFEST.in to make it part of the source tarball.

PR #1568, opened 2021-04-19
[nmstate-1.0] specfile: fix installation of nmstate-varlink.service

ffmancera:
On the nmstate.spec file the installation was done using `cp -a` which
is incorrect. In order to install the nmstate-varlink.service it should
use the `install -p -m 644`.

### DIFF
--- a/packaging/nmstate.spec
+++ b/packaging/nmstate.spec
@@ -66,7 +66,9 @@ This package contains the nmstate plugin for OVS database manipulation.
 
 %install
 %py3_install
-mkdir -p %{buildroot}%{_unitdir} && cp -a %{buildroot}%{python3_sitelib}/nmstatectl/nmstate-varlink.service %{buildroot}%{_unitdir}/
+mkdir -p %{buildroot}%{_unitdir}
+install -p -m 644 %{buildroot}%{python3_sitelib}/nmstatectl/nmstate-varlink.service \
+         %{buildroot}%{_unitdir}/nmstate-varlink.service
 
 %post
 %systemd_post nmstate-varlink.service


packaging/nmstate.spec
```diff
@@ -66,7 +66,9 @@ This package contains the nmstate plugin for OVS database manipulation.
 
 %install
 %py3_install
-mkdir -p %{buildroot}%{_unitdir} && cp -a %{buildroot}%{python3_sitelib}/nmstatectl/nmstate-varlink.service %{buildroot}%{_unitdir}/
+mkdir -p %{buildroot}%{_unitdir}
+install -p -m 644 %{buildroot}%{python3_sitelib}/nmstatectl/nmstate-varlink.service \
```
tyll, line 70: Why is it installed to the python3_sitelib at all? Seems like it should only be part of MANIFEST.in to make it part of the source tarball.

ffmancera: Well, I think both options are valid. This file is already on `package_data` section so it is being installed in the `python3_sitelib` library and therefore included in the tarball.

Anyway, do you agree on this should be installed by `install`? If you think adding it to MANIFEST.in is better, we can do it in little a follow up patch.

Thanks for pointing out this, I have been reading about this and I think the discussion is worth to have.

tyll: Why is it a good idea to include it as `package_data` and install it in the sitelib directory? It is not used there. Also, now the file is installed twice on the system.

Install sounds fine but I doubt that it is the reason for the Fedora problem. There the problem seems to be that the systemd unit is not present in the correct directory - whether it is copied there by `cp` or `install` should not matter.

ffmancera: Yes, you are right about it is not being used there so including it in `MANIFEST.in` sounds good.

I've contacted the rpm community and asked about this topic. According to them, the correct way of installing systemd unit is by using `install`. Other projects as SVN is doing it that way [1]. I don't think including it on  `MANIFEST.in` will fix the issue because the systemd unitdir is needed and the file must be moved there.

I think using `cp` was the issue because after replacing it with `install`, the problem is fixed. I created a build and tested it. I think the problem was on the file permission. Now it is possible to update and upgrade the package. 

I am sending a follow up patch. Thank you!

[1] https://src.fedoraproject.org/rpms/subversion/blob/rawhide/f/subversion.spec#_407

cathay4t (approved):
I will approve this trivial fix. I would like to see it actually be used in Fedora/RHEL packaging.

PR #1570, opened 2021-04-21
Xfail some tests on k8s

qinqon:
To make the k8s lane the test cases that fail can be marked as xfail and the is_ovs_running has to be removed (don't know if it's really needed).

### DIFF
--- a/tests/integration/bond_test.py
+++ b/tests/integration/bond_test.py
@@ -42,6 +42,7 @@ from .testlib import statelib
 from .testlib.assertlib import assert_mac_address
 from .testlib.bondlib import bond_interface
 from .testlib.env import nm_major_minor_version
+from .testlib.env import is_k8s
 from .testlib.ifacelib import get_mac_address
 from .testlib.ifacelib import ifaces_init
 from .testlib.vlan import vlan_interface
@@ -201,6 +202,15 @@ def test_add_bond_with_port_and_ipv4(eth1_up, eth2_up, setup_remove_bond99):
 
 
 @pytest.mark.tier1
+@pytest.mark.xfail(
+    is_k8s(),
+    reason=(
+        "Requires adjusts for k8s. Ref:"
+        "https://github.com/nmstate/nmstate/issues/1579"
+    ),
+    raises=AssertionError,
+    strict=False,
+)
 def test_rollback_for_bond(eth1_up, eth2_up):
     current_state = libnmstate.show()
     desired_state = {
--- a/tests/integration/nm/profile_test.py
+++ b/tests/integration/nm/profile_test.py
@@ -23,6 +23,7 @@ import os
 import pytest
 
 import libnmstate
+from libnmstate.error import NmstateValueError
 from libnmstate.schema import Interface
 from libnmstate.schema import InterfaceIPv4
 from libnmstate.schema import InterfaceIPv6
@@ -33,6 +34,7 @@ from libnmstate.schema import LinuxBridge
 from ..testlib import assertlib
 from ..testlib import cmdlib
 from ..testlib import statelib
+from ..testlib.env import is_k8s
 from ..testlib.ovslib import Bridge as OvsBridge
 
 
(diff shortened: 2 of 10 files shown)


cathay4t:
@qinqon If you are changing test python code, you need to use `python-black` to format the code and `pylink/flake8` to do lint checking also.

qinqon:
> @qinqon If you are changing test python code, you need to use `python-black` to format the code and `pylink/flake8` to do lint checking also.

Done

qinqon:
@ffmancera did you have some comments on the PR ?

qinqon:
@cathay4t @ffmancera k8s lane is green now, we can merge this, and maybe mark the lane as required.

tests/integration/bond_test.py
```diff
@@ -201,6 +202,12 @@ def test_add_bond_with_port_and_ipv4(eth1_up, eth2_up, setup_remove_bond99):
 
 
 @pytest.mark.tier1
+@pytest.mark.xfail(
+    is_k8s(),
+    reason=("Test not fixed for k8s yet"),
```
ffmancera: Please specify the following in all reasons: `Requires adjusts for k8s. Ref: https://github.com/nmstate/nmstate/issues/1579`

ffmancera (changes requested):
Just a nit, other than that it looks good to me. Please could you describe what are you doing on the commit message? This way we can track why this was done in the future. Thanks! Also, the linter is complaining, please check it out.

libnmstate/nm/plugin.py
```diff
@@ -103,7 +102,7 @@ def context(self):
     @property
     def capabilities(self):
         capabilities = []
-        if has_ovs_capability(self.client) and is_ovs_running():
+        if has_ovs_capability(self.client):
```
ffmancera: This cannot be done I think. If so, the following test will be failing  
`tests/integration/ovs_test.py::test_ovs_service_missing FAILED           [ 56%]`

qinqon: But do we really need it? I see other parts of the code that they just do the has_ovs_capability but not is_ovs_running.

tests/integration/bond_test.py
qinqon: Done

libnmstate/nm/plugin.py
ffmancera: @qinqon This is the only issue left, what do you think @cathay4t? IMO, we don't need to use `is_ovs_running()`... I didn't find any issue related to this. Thanks!

cathay4t: Currently, NM is hard coded to use socket `/var/run/openvswitch/db.sock`, can we just check the existence and type of that file as replacement of `systemctl show openvswitch`.

qinqon: Well is file does not exist the k8s pod is not going to even start so we are implicitly doing it, anyhow I am going to remove this part of the PR so we merge the xfails and the other fixes, then we can tackle ovs at separate PR.

tests/integration/nm/profile_test.py
```diff
@@ -399,6 +410,15 @@ def test_linux_bridge_with_port_holding_two_profiles(eth1_with_two_profiles):
 
 
 @pytest.mark.tier1
+@pytest.mark.xfail(
+    is_k8s(),
+    reason=(
+        "Requires adjusts for k8s. Ref:"
+        "https://github.com/nmstate/nmstate/issues/1579"
+    ),
+    raises=libnmstate.error.NmstateValueError,
```
ffmancera: Use `NmstateValueError` directly, it is imported.

ffmancera (changes requested):
Just a nit, the linter is complaining. Thanks!

tests/integration/nm/profile_test.py
qinqon: Done

cathay4t (approved):

PR #1578, opened 2021-04-27
nm.wired: add support to accept-all-mac-addresses

ffmancera:
This patch is introducing the `accept-all-mac-addresses` interface
property in Nmstate. The new property is a boolean, when True the NIC
will accept all packets.

It is supported only with NetworkManager 1.31 or greater.

Example:

```
---
interfaces:
- name: eth1
  type: ethernet
  state: up
  accept-all-mac-addresses: true
```

### DIFF
--- a/libnmstate/nispor/base_iface.py
+++ b/libnmstate/nispor/base_iface.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2020 Red Hat, Inc.
+# Copyright (c) 2020-2021 Red Hat, Inc.
 #
 # This file is part of nmstate
 #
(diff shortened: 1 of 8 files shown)


libnmstate/nispor/base_iface.py
```diff
@@ -67,6 +68,14 @@ def state(self):
             )
             return InterfaceState.DOWN
 
+    @property
+    def accept_all_mac_addresses(self):
+        np_flags = self._np_iface.flags
+        if PROMISC_FLAG in np_flags:
+            return True
+
+        return False
```
cathay4t: How about

```python
    def accept_all_mac_addresses(self):
        return  PROMISC_FLAG in self._np_iface.flags
```

cathay4t (approved):

libnmstate/nispor/base_iface.py
ffmancera: Yes, much better. Done thanks!

PR #1580, opened 2021-04-29
net_state: mark ifaces as ignored earlier

ffmancera:
Nmstate is not touching unmanaged interfaces if they are not being
included in the desired state. If an unmanaged interface has a default
gateway configured and we try to add a dns server, Nmstate will pick the
unmanaged interface because at that point it is not marked as ignored.

As the interface is being marked as ignored later, the changes are not
being applied and is failing on verification. In order to avoid this,
the unmanaged interfaces should be marked as ignored earlier so they are
not considered for DNS/Routes changes.

Ref: https://bugzilla.redhat.com/1944582

### DIFF
--- a/libnmstate/dns.py
+++ b/libnmstate/dns.py
@@ -133,7 +133,7 @@ class DnsState:
         Return tuple: (ipv4_iface, ipv6_iface)
         """
         ipv4_iface, ipv6_iface = self._find_ifaces_with_static_gateways(
-            route_state
+            ifaces, route_state
         )
         if not (ipv4_iface and ipv6_iface):
             (
@@ -147,7 +147,7 @@ class DnsState:
 
         return ipv4_iface, ipv6_iface
 
-    def _find_ifaces_with_static_gateways(self, route_state):
+    def _find_ifaces_with_static_gateways(self, ifaces, route_state):
         """
         Return tuple of interfaces with IPv4 and IPv6 static gateways.
         """
@@ -158,6 +158,11 @@ class DnsState:
                 if ipv4_iface and ipv6_iface:
                     return (ipv4_iface, ipv6_iface)
                 if route.is_gateway:
+                    # Only interfaces not ignored can be considered as valid
+                    # static gateway for nameservers.
+                    iface = ifaces.all_kernel_ifaces.get(iface_name)
+                    if iface and iface.is_ignore:
+                        continue
                     if route.is_ipv6:
                         ipv6_iface = iface_name
                     else:
@@ -168,6 +173,8 @@ class DnsState:
         ipv4_iface = None
         ipv6_iface = None
         for iface in ifaces.all_kernel_ifaces.values():
+            if iface.is_ignore:
+                continue
             if ipv4_iface and ipv6_iface:
                 return (ipv4_iface, ipv6_iface)
             for family in (Interface.IPV4, Interface.IPV6):
--- a/libnmstate/netapplier.py
+++ b/libnmstate/netapplier.py
@@ -73,7 +73,10 @@ def apply(
         validator.validate_capabilities(
             desired_state, plugins_capabilities(plugins)
         )
-        net_state = NetState(desired_state, current_state, save_to_disk)
+        ignored_ifnames = _get_ignored_interface_names(plugins)
+        net_state = NetState(
+            desired_state, ignored_ifnames, current_state, save_to_disk
+        )
         checkpoints = create_checkpoints(plugins, rollback_timeout)
         _apply_ifaces_state(plugins, net_state, verify_change, save_to_disk)
         if commit:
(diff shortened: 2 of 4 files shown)


libnmstate/dns.py
```diff
@@ -158,7 +158,15 @@ def _find_ifaces_with_static_gateways(self, route_state):
                 if ipv4_iface and ipv6_iface:
                     return (ipv4_iface, ipv6_iface)
                 if route.is_gateway:
-                    if route.is_ipv6:
+                    # Only interfaces not ignored can be considered as valid
+                    # static gateway for nameservers.
+                    iface_ignored = False
+                    for iface in ifaces.all_ifaces():
+                        if iface.name == iface_name and iface.is_ignore:
+                            iface_ignored = True
```
cathay4t: Using `ifaces.all_kernel_ifaces.get(iface_name)` can save you from searching it.

libnmstate/dns.py
```diff
@@ -147,7 +147,7 @@ def _find_ifaces_for_name_servers(self, ifaces, route_state):
 
         return ipv4_iface, ipv6_iface
 
-    def _find_ifaces_with_static_gateways(self, route_state):
+    def _find_ifaces_with_static_gateways(self, ifaces, route_state):
```
cathay4t, line 150: At the time function `_find_ifaces_with_static_gateways()` been invoked, we don't know the ignored interface yet.

The metadata generate happens at `NetState.__init__()`, but `_apply_ifaces_state()` who mark ignored interfaces is after that.

cathay4t (changes requested):
From the logic on code, you seems just hide the verification error by marking interface as ignored. When user make changes on DNS, the metadata is still selecting the old unmanged interface.

libnmstate/dns.py
ffmancera: But this could apply for OVS interfaces so we need to use `all_ifaces()`.

libnmstate/dns.py
ffmancera: That is correct, this is why I added this on the `NetState.__init__()`.

```
        if not gen_conf_mode:
            self._mark_ignored_kernel_ifaces(plugins)
```

Please, check how we are marking the interface as ignored earlier on this patch.

libnmstate/dns.py
cathay4t: OVS interface is a kernel interface.

ffmancera:
As requested, the plugins are not being passed to the NetState. Now the NetState is getting a list of ignored interface names. I hope you like more this approach. Thanks!

libnmstate/dns.py
```diff
@@ -158,7 +158,15 @@ def _find_ifaces_with_static_gateways(self, route_state):
                 if ipv4_iface and ipv6_iface:
                     return (ipv4_iface, ipv6_iface)
                 if route.is_gateway:
-                    if route.is_ipv6:
+                    # Only interfaces not ignored can be considered as valid
+                    # static gateway for nameservers.
+                    iface_ignored = False
+                    for iface in ifaces.all_ifaces():
```
cathay4t: Should be `all_kernel_ifaces()` as only kernel interface can hold gateway or having auto-ip.

And `all_kernel_ifaces()` is a hashmap, so you don't need the flag `iface_ignored`.

libnmstate/netapplier.py
```diff
@@ -145,3 +143,12 @@ def _net_state_contains_sriov_interface(net_state):
 def _verify_change(plugins, net_state):
     current_state = show_with_plugins(plugins)
     net_state.verify(current_state)
+
+
+def _get_ignored_interface_names(plugins):
+    ifaces = []
+    for plugin in plugins:
+        for iface_name in plugin.get_ignored_kernel_interface_names():
+            ifaces.append(iface_name)
```
cathay4t: Please use a `set()` in case plugins are returning the same interface name.

libnmstate/nmstate.py
```diff
@@ -392,9 +392,9 @@ def generate_configurations(desire_state):
     you may run this function on different system.
     """
     configs = {}
-    net_state = NetState(desire_state, gen_conf_mode=True)
 
     with plugin_context() as plugins:
+        net_state = NetState(desire_state, gen_conf_mode=True)
```
cathay4t: This is not required.

cathay4t (changes requested):

libnmstate/dns.py
ffmancera: Done, thanks!

libnmstate/nmstate.py
ffmancera: Thanks, this was a leftover.

libnmstate/netapplier.py

ffmancera:
Changes done, thanks!

cathay4t (approved):

Issue #1513, opened 2021-02-10
Nmstate is crashing on integ_slow with nm-master


ffmancera:
## Describe the bug
Nmstate is crashing when using NetworkManager-master on integ_slow on CI.

## Nmstate version
- Nmstate base branch

## NetworkManager version
- NM-master

## Platform
- Platform: centOS 8

## How reproducible
- 50%~~

## Steps to Reproduce
Run the integ_slow tests on CI.

## Nmstate output
- Expected output:
No errors.

- Current output:
```
________________________ test_lot_of_vlans_with_bridges ________________________

eth1_up = {'interfaces': [{'ipv4': {'address': [], 'dhcp': False, 'enabled': False}, 'ipv6': {'address': [{'ip': 'fe80::d0da:6ff...toconf': False, 'dhcp': False, 'enabled': True}, 'lldp': {'enabled': False}, 'mac-address': 'D2:DA:06:18:0A:25', ...}]}

    @pytest.mark.tier1
    @pytest.mark.slow
    def test_lot_of_vlans_with_bridges(eth1_up):
        interfaces = []
        for i in range(100, 600):
            interfaces.extend(
                [
                    {
                        Interface.NAME: "vlan." + str(i),
                        Interface.TYPE: InterfaceType.VLAN,
                        Interface.STATE: InterfaceState.UP,
                        VLAN.CONFIG_SUBTREE: {VLAN.BASE_IFACE: "eth1", VLAN.ID: i},
                    },
                    {
                        Interface.NAME: "linux-br" + str(i),
                        Interface.TYPE: InterfaceType.LINUX_BRIDGE,
                        Interface.STATE: InterfaceState.UP,
                        LinuxBridge.CONFIG_SUBTREE: {
                            LinuxBridge.PORT_SUBTREE: [
                                {LinuxBridge.Port.NAME: "vlan." + str(i)}
                            ]
                        },
                    },
                ]
            )
        try:
>           libnmstate.apply({Interface.KEY: interfaces})

tests/integration/timeout_test.py:56: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

[...]

PR #1584, opened 2021-05-06
bond: normalize bond options for warnings

ffmancera:
Nmstate raises warning when bond options do not match i.e kernel option
does not match current option. When setting mac addresses values in
lowercase and then setting the same one in uppercase, Nmstate does not
modify kernel value and raises a warning.

This is solving the following issue for all string options like mac
addresses.

```
2020-11-09 17:30:48,647 root         WARNING  Desire iface bond1 bond
option ad_actor_system=d4:ee:07:25:42:5a does not match with kernel
value: D4:EE:07:25:42:5A
```

Ref: https://bugzilla.redhat.com/1895851

### DIFF
--- a/libnmstate/ifaces/bond.py
+++ b/libnmstate/ifaces/bond.py
@@ -202,19 +202,37 @@ class BondIface(BaseIface):
             Bond.OPTIONS_SUBTREE, {}
         )
         for key, value in self_bond_options.items():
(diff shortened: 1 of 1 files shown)


cathay4t (approved):

cathay4t:
The CI failure is tracking at #1513

PR #1590, opened 2021-05-12
tests, integ: extend accept-all-mac-addresses testing

ffmancera:
In NetworkManager `accept-all-mac-addresses` works in different ways
for virtual interfaces. These tests is making sure we are covering the
feature in the virtual interfaces too.


cathay4t (changes requested):
All the test case are sharing the same name. Can we somehow differentiate them?

ffmancera:
> All the test case are sharing the same name. Can we somehow differentiate them?

Done, I've added the interface type to the test name. Thank you!

cathay4t (dismissed):

cathay4t (approved):

PR #1603, opened 2021-05-25
CI: install libndp as new dependency of NetworkManager

ffmancera:
In addition, this patch is fixing the Copr installation of
NetworkManager-main in centos-stream.


cathay4t (approved):
LGTM. Thanks.

PR #1610, opened 2021-05-26
README: Fix link to installation guide

varesa:
The README has a broken link to www.nmstate.io, which gives a certificate error since github returns a default certificate.

Fix the link to point to https://nmstate.io/user/install.html instead

Signed-off-by: Esa Varemo <git@esav.fi>


ffmancera (approved):
Even if it does not fail, all the other links are using `nmstate.io` instead of `www.nmstate.io`. It looks good to me. Thank you!

PR #1620, opened 2021-06-01
Add VXLAN properties required to support an external control plane (BGP-EVPN)

varesa:
A work in progress state and request for feedback. 
Will squash, etc. once any TODOs/issues have been fixed.

Add support for a configuration where a VXLAN interface is created in a non-learning mode without predefined remote VTEPs. MAC learning and VTEP configuration are handled by an external process (e.g. a BGP daemon like FRR for BGP-EVPN setups). 

In this configuration the VXLAN interface might also not have an explicit parent interface, and instead traffic could go over any number of physical interfaces based on host routes (ECMP over equal links).

I've briefly tested these modifications on Fedora and been running them on CentOS 8 Stream for a while now (https://copr.fedorainfracloud.org/coprs/varesa/nmstate-git/package/nmstate/). I'll look at the project internal test suite to see how the VXLAN side is currently tested there and if it needs adding. And how to run the test suite, of course :)

### DIFF
--- a/libnmstate/ifaces/ifaces.py
+++ b/libnmstate/ifaces/ifaces.py
@@ -321,6 +321,7 @@ class Ifaces:
             if (
                 iface.type in [InterfaceType.VLAN, InterfaceType.VXLAN]
                 and iface.is_up
+                and iface.parent
             ):
                 if (
                     self._kernel_ifaces[iface.parent].type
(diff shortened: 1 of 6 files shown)


ffmancera:
Not stale :-)

varesa:
@ffmancera could you comment on the few comments marked TODO in libnmstate/ifaces/vxlan.py ?

ffmancera:
> libnmstate

Sure, let me do it. Thanks!

PR #1637, opened 2021-06-14
libnmstate: do not report perm_mac_address when showing

ffmancera:
Nmstate is currently showing _perm_mac_address when showing. This is not
correct because perm_mac_address is only for metadata.


cathay4t (changes requested):
I tested in my VM. The `_permanent_mac_address` still exists in `libnmstate.show()`.

cathay4t (approved):

PR #1645, opened 2021-06-22
libnmstate: introduce show_secrets argument

ffmancera:
Nmstate needs to handle properly the secrets from the interfaces when
reporting, logging, crashing and verifying. In order to do that, this
patch is introducing the support for `show_secrets` argument for
`libnmstate.show()` and `--show-secrets` option for `nmstatectl show`.

Nmstate will not report, log, show when failing on verification or show
when crashing the sensitive data by default. When the new option is
present, Nmstate will show them on these situations. This can be used
for debugging or applying the current configuration in another host.

For `apply` and `edit` functions, Nmstate will only report, log or show
sensitive data if the desired state contains sensitive data. If the
desired state does not contain sensitive data, Nmstate will not do it.

Nmstatectl manpage documentation has been added.

Integration test case added.

### DIFF
--- a/doc/nmstatectl.8.in
+++ b/doc/nmstatectl.8.in
@@ -3,7 +3,7 @@
 .SH NAME
 nmstatectl \- A nmstate command line tool
 .SH SYNOPSIS
-.B nmstatectl show \fR[\fIINTERFACE_NAME\fR] [\fB--json\fR]
+.B nmstatectl show \fR[\fIINTERFACE_NAME\fR] [\fB--json\fR] [\fB--show-secrets\fR]
 .br
 .B nmstatectl show [\fB-r, --running-config\fR]
 .br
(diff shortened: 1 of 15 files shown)


cathay4t (changes requested):
Instead of asking plugin to handle the secrets, I prefer:
 * `BaseIface` with function `to_dict_without_secrets`.
 * Plugins still show full data. Top level code decide whether to show/verify it.

And the top level code `netapplier.py` should not hold code for 802.1x specific code, that's bad on code design.

ffmancera:
> Instead of asking plugin to handle the secrets, I prefer:
> 
>     * `BaseIface` with function `to_dict_without_secrets`.
> 

That is not possible as BaseIface is not being used when doing a show operation.

>     * Plugins still show full data. Top level code decide whether to show/verify it.
> 

That is possible but we must consider:
1. That create a risk, in case the application crash at somepoint the information will be there. My idea with this patch was to do not collect it if not requested. Is that risk fine? IMO, yes as it is very uncommon this will happen.

> 
> And the top level code `netapplier.py` should not hold code for 802.1x specific code, that's bad on code design.

This is not specific for 802.1X, is about any secrets we could introduce. In addition, I am fine with this but then we would need to move the function `_desired_state_contains_secrets` to another place.. anyway, we will need to handle if we need to keep secrets or not during the verification on `netapplier.py`..

Let me see what can I do. Thanks!

ffmancera:
> Instead of asking plugin to handle the secrets, I prefer:
> 
>     * `BaseIface` with function `to_dict_without_secrets`.
> 
>     * Plugins still show full data. Top level code decide whether to show/verify it.
> 
> 
> And the top level code `netapplier.py` should not hold code for 802.1x specific code, that's bad on code design.

I think I can introduce BaseIface to 'show_with_plugins'.. from that I think I will be able to do what you requested. Thinking about it, you are right. Nmstate should not contain code about 802.1X in the top level code.

Thanks!

cathay4t:
@ffmancera Can we close this as dup of #1657

ffmancera:
Closing this in favor of #1657

PR #1649, opened 2021-06-22
team: Drop the support of team interfaces

ffmancera:


libnmstate/validator.py
```diff
@@ -59,7 +59,10 @@ def validate_interface_capabilities(ifaces_state, capabilities):
             raise NmstateDependencyError(
                 "Open vSwitch support not properly installed or started"
             )
-        elif iface_type == InterfaceType.TEAM and not has_team_capability:
+        elif (
+            iface_type == NmstatePlugin.TEAM_CAPABILITY
+            and not has_team_capability
+        ):
```
cathay4t: This lines should be also removed.

cathay4t (changes requested):

libnmstate/validator.py
ffmancera: Thanks! Done

cathay4t (approved):

PR #1655, opened 2021-06-25
bridge, bond: drop slave support from schema

ffmancera:
In order to follow our efforts to support inclusive language in Nmstate,
we are dropping the support of the "slave" term.


cathay4t (approved):
LGTM. Thanks

PR #1677, opened 2021-07-16
Clean up nmstatectl script and fix long arguments

ffmancera:


ffmancera:
@cathay4t I think we should backport the first commit to 1.1. It is affecting nmstatectl.

cathay4t (approved):

PR #1688, opened 2021-07-30
k8s, ci: bump centos stream at nodes

qinqon:


cathay4t:
@qinqon The upgrade seems has dependency issue of docker, any reason we don't use podman?

qinqon:
> @qinqon The upgrade seems has dependency issue of docker, any reason we don't use podman?

The kubevirtci is able to use podman https://github.com/kubevirt/kubevirtci/blob/main/PODMAN.md, let's give it a try.

qinqon:
@cathay4t after upgrading the kubevirtci version (it's no longer using docker at nodes) k8s lane is working fine.

qinqon:
@cathay4t is this ok now ?

cathay4t (approved):

PR #1692, opened 2021-08-09
veth: throw better error when peer is missing

ffmancera:
When creating a veth interface if the peer is missing, Nmstate is now
throwing a better error.

### DIFF
--- a/libnmstate/ifaces/ifaces.py
+++ b/libnmstate/ifaces/ifaces.py
@@ -282,6 +282,7 @@ class Ifaces:
         self._validate_ovs_patch_peers()
         self._remove_unknown_type_interfaces()
         self._validate_vrf_table_id_changes()
+        self._validate_veth_peers()
 
     def _bring_port_up_if_not_in_desire(self):
         """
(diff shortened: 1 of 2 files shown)


libnmstate/nm/veth.py
```diff
@@ -48,6 +49,13 @@ def create_setting(iface, base_con_profile):
 
     if not veth_setting:
         veth_setting = getattr(NM, SETTING_VETH).new()
+
+    if not veth_setting.props.peer and not iface.peer:
```
cathay4t: This kind of check should done at nmstate level instead of nm level.

cathay4t (changes requested):

libnmstate/nm/veth.py
ffmancera: Done, the checking is being done at ifaces.

libnmstate/ifaces/ifaces.py
```diff
@@ -440,6 +441,19 @@ def _validate_vrf_table_id_changes(self):
...
+        for ifname, iface in self._kernel_ifaces.items():
+            if (
+                iface.type == InterfaceType.VETH
+                and iface.is_up
+                and not iface.peer
+            ):
+                if not self._cur_kernel_ifaces.get(iface.name):
+                    raise NmstateValueError(
+                        f"Veth interface {iface.name} must have a peer, "
+                        f"current value is {iface.peer}"
```
cathay4t: the `iface.peer` is None.
How about `Veth interface {iface.name} does not exists, peer name is required for creating it`.

cathay4t (changes requested):

libnmstate/ifaces/ifaces.py
ffmancera: Sure, let me do it.

cathay4t (approved):

PR #1694, opened 2021-08-26
yaml: drop pyyaml dependency in favor of ruamel.yaml

ffmancera:
Currently pyyaml implements YAML 1.1 which is causing us problems.

When an interface has a float64 name, nmstate does not quote the name.
That is causing the state reported cannot be applied back.

```yaml
interfaces:
- name: '60e+02'
  type: dummy
  state: down
```

This is solved by using ruamel.yaml because it is implementing YAML 1.2.


cathay4t:
The ruamel yaml is in EPEL8. So I will not approve this change.

PR #1697, opened 2021-08-30
ovsdb: drop NM.connection.uuid entry

ffmancera:
Currently, when trying to add or modify the NM created external ids it
is not possible because Nmstate is overwritting the desired value with
the existing one.

This is fixing the following verification error:

```
difference
==========
--- desired
+++ current
@@ -51,4 +51,4 @@
 mtu: 1500
 ovs-db:
   external_ids:
-    NM.connection.uuid: 930aa309-99b1-4fe9-a300-ceec3efd9907
+    NM.connection.uuid: 2fe14d2b-3465-43bb-8a49-f65ded332238
```

Ref: https://bugzilla.redhat.com/1660250

### DIFF
--- a/libnmstate/plugins/nmstate_plugin_ovsdb.py
+++ b/libnmstate/plugins/nmstate_plugin_ovsdb.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2020 Red Hat, Inc.
+# Copyright (c) 2020-2021 Red Hat, Inc.
 #
 # This file is part of nmstate
 #
@@ -17,6 +17,7 @@
 # along with this program. If not, see <https://www.gnu.org/licenses/>.
 #
 
+import copy
 import os
 import time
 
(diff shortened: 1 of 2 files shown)


cathay4t (dismissed):

cathay4t:
I am OK with the patch. But wondering why openstack need to alter the `NM.connection.uuid` which is supposed by maintained by NetworkManager only.

libnmstate/plugins/nmstate_plugin_ovsdb.py
```diff
@@ -169,28 +170,42 @@ def get_interfaces(self):
             else:
                 continue
 
+            # should remove external_id created by NetworkManager
+            modified_external_ids = self._drop_nm_external_ids(
```
tyll: The comment seems to be unnecessary since it contains the same information as the method name that is called. Also at the other line.

ffmancera: Done, thank you!

cathay4t (approved):
Patch looks good. Please provide more info in the commit message: root cause, how to fix in this patch.

PR #1699, opened 2021-09-02
nm.connection: add prefixes to OVS bridges and interfaces connections

ffmancera:
When configuring an OVS bridge and interface with the same using
Nmstate, it will generate two connection with the same name. If the user
is willing to use nmcli to manage this profiles, it is not user friendly
because they will need to use the connection UUID instead.

In order to solve this, Nmstate is adding a prefix on the connection
name. "-if" for OVS interface and "-br" for OVS bridge.

Please, note that this issue does not affect kernel interfaces because
it is not possible to have duplicated interfaces names in kernel.

Integration test case added.

### DIFF
--- a/libnmstate/nm/connection.py
+++ b/libnmstate/nm/connection.py
@@ -113,8 +113,20 @@ def create_new_nm_simple_conn(iface, nm_profile):
     if nm_profile and not is_multiconnect_profile(nm_profile):
         con_setting.import_by_profile(nm_profile, iface.is_controller)
     else:
+        # OVS bridge and interfaces could sharing the same interface name, to
+        # distinguish them at NM connection level, instead of using interface
+        # name as connection name, we append a postfix.
+        con_name = iface.name
+        if iface.type == InterfaceType.OVS_BRIDGE:
+            con_name = con_name + "-br"
+        elif iface.type == InterfaceType.OVS_INTERFACE:
+            con_name = con_name + "-if"
+
         con_setting.create(
-            iface.name, iface.name, nm_iface_type, iface.is_controller
+            con_name,
+            iface.name,
+            nm_iface_type,
+            iface.is_controller,
         )
 
     apply_lldp_setting(con_setting, iface_info)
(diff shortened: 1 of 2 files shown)


libnmstate/nm/connection.py
```diff
@@ -61,12 +61,21 @@ class _ConnectionSetting:
     def __init__(self, con_setting=None):
         self._setting = con_setting
 
-    def create(self, con_name, iface_name, iface_type, is_controller):
+    def create(
+        self, con_name, iface_name, nm_iface_type, is_controller, iface_type
+    ):
+        # OVS bridges and interfaces can have duplicated names, this way we
+        # avoid the connection names will be duplicated too.
```
cathay4t: The `aviod ... will` seems confusing. How about

```
OVS bridge and interfaces could sharing the same interface name, to distinguish them at NM connection level, instead of using interface name as connection name, we append a prefix
```

cathay4t: And instead of changing `_ConnectionSetting.create()`, you should change the caller of it.
The `_ConnectionSetting.create()` already took `con_name` as argument, we should not alter that internally.

The caller `create_new_nm_simple_conn()` could decide which connection name should be used base on `iface` it got. This isolate the `_ConnectionSetting` worry about OVS stuff.

cathay4t (changes requested):

libnmstate/nm/connection.py
ffmancera: Done, thank you!

libnmstate/nm/connection.py
```diff
@@ -113,8 +113,20 @@ def create_new_nm_simple_conn(iface, nm_profile):
     if nm_profile and not is_multiconnect_profile(nm_profile):
         con_setting.import_by_profile(nm_profile, iface.is_controller)
     else:
+        # OVS bridge and interfaces could sharing the same interface name, to
+        # distinguish them at NM connection level, instead of using interface
+        # name as connection name, we append a prefix
```
cathay4t: Actually, your code is adding a postfix. :)

cathay4t (changes requested):

libnmstate/nm/connection.py
ffmancera: I changed all the references to "postfix", sorry about that. Thanks!

cathay4t (approved):

PR #1703, opened 2021-09-17
Add copy_mac_from to ovs_interface

rhrazdil:
Allow using copy_mac_from for OVS insterface.

Signed-off-by: Radim Hrazdil <rhrazdil@redhat.com>

### DIFF
--- a/libnmstate/ifaces/ifaces.py
+++ b/libnmstate/ifaces/ifaces.py
@@ -172,12 +172,13 @@ class Ifaces:
             if iface.type not in (
                 InterfaceType.LINUX_BRIDGE,
                 InterfaceType.BOND,
+                InterfaceType.OVS_INTERFACE,
             ):
                 continue
             if not iface.copy_mac_from:
                 continue
 
-            if iface.copy_mac_from not in iface.port:
+            if iface.type != InterfaceType.OVS_INTERFACE and iface.copy_mac_from not in iface.port:
                 raise NmstateValueError(
                     f"The interface {iface.name} is holding invalid "
                     f"{Interface.COPY_MAC_FROM} property "
--- a/libnmstate/ifaces/ifaces.py
+++ b/libnmstate/ifaces/ifaces.py
@@ -178,13 +178,7 @@ class Ifaces:
             if not iface.copy_mac_from:
                 continue
 
-            if iface.type != InterfaceType.OVS_INTERFACE and iface.copy_mac_from not in iface.port:
-                raise NmstateValueError(
-                    f"The interface {iface.name} is holding invalid "
-                    f"{Interface.COPY_MAC_FROM} property "
-                    f"as {iface.copy_mac_from} is not in the port "
-                    f"list: {iface.port}"
-                )
+            self._validate_copy_mac_from_iface_is_in_port(iface)
             port_iface = self.all_kernel_ifaces.get(iface.copy_mac_from)
             # TODO: bridge/bond might refering the mac from new veth in the
             #       same desire state, it too complex to support that.
@@ -198,6 +192,32 @@ class Ifaces:
                 iface.apply_copy_mac_from(port_iface.permanent_mac_address)
             else:
                 iface.apply_copy_mac_from(port_iface.mac)
+
+    def _validate_copy_mac_from_iface_is_in_port(self, iface):
+        iface_port = self._port_for_interface(iface)
+        if iface.copy_mac_from not in iface_port:
+            raise NmstateValueError(
+                f"The interface {iface.name} is holding invalid "
+                f"{Interface.COPY_MAC_FROM} property "
+                f"as {iface.copy_mac_from} is not in the port "
+                f"list: {iface.port}"
+            )
+
+    def _port_for_interface(self, iface):
+        if iface.type == InterfaceType.OVS_INTERFACE:
+            ovs_bridges = (
+                x
+                for x in self.all_user_space_ifaces
+                if x.type == InterfaceType.OVS_BRIDGE and iface.name in x.port
+            )
+            ovs_bridge = next(ovs_bridges, None)
+            if ovs_bridge is None:
+                raise NmstateValueError(
+                    f"The ovs-interface {iface.name} is not attached "
+                    f"to any bridge"
+                )
+            return ovs_bridge.port
+        return iface.port
 
     def _create_virtual_port(self):
         """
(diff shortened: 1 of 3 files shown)


libnmstate/ifaces/ifaces.py
```diff
@@ -172,6 +172,7 @@ def _apply_copy_mac_from(self):
             if iface.type not in (
                 InterfaceType.LINUX_BRIDGE,
                 InterfaceType.BOND,
+                InterfaceType.OVS_INTERFACE,
```
phoracek, line 175: I think this will not work. We have to copy the MAC onto an internal interface which is modeled as a separate object in the "interfaces" list (unlike with bondings and bridges where the MAC is defined directly on the bridge/bond). Due to that, this will fail on line 182.

phoracek (changes requested):

libnmstate/ifaces/ifaces.py
cathay4t: Correct. The code should expand the check for OVS usage.

cathay4t (changes requested):

libnmstate/ifaces/ifaces.py
rhrazdil: Thanks! My intention (not sure it's the right one) was to add the copy-mac-from attribute to an ovs-interface, not to the ovs-bridge, so the usage would be for example:

```
    interfaces:
    - name: ovs0
      type: ovs-interface
      state: up
      ipv4:
        enabled: true
        dhcp: true
      copy-mac-from: eth0
    - name: ovs-br10
      type: ovs-bridge
      state: up
      bridge:
        options:
          stp: true
        port:
        - name: eth0
        - name: ovs0
```
Does this make sense or it should be added to the ovs-bridge? If that's the case, how do we know which internal interface to copy the mac to? IIUC? there may be many

phoracek: It should be added to ovs-interface. But to do that, you must extend the validation code of nmstate that makes sure that copy-mac-from references a port of the interface it is defined on, which is not the case with ovs-interface.

libnmstate/ifaces/ifaces.py
```diff
@@ -198,6 +193,28 @@ def _apply_copy_mac_from(self):
             else:
                 iface.apply_copy_mac_from(port_iface.mac)
 
+    def _validate_copy_mac_from_iface_is_in_port(self, iface):
+        iface_port = iface.port
+        if iface.type == InterfaceType.OVS_INTERFACE:
+            ovs_bridges = [
```
phoracek: You could use lazy evaluation here.

Instead of list, construct a generator `ovs_bridges = ( .... )` and then use `if next(ovs_bridges, None):`. I think something like this should do it.

I will not fight for this, but it may be a little more efficient (even though I guess it won't save much in the whole scope of things, especially if all_kernel_ifaces.values() is not a generator itself).

libnmstate/ifaces/ifaces.py
```diff
@@ -198,6 +193,28 @@ def _apply_copy_mac_from(self):
             else:
                 iface.apply_copy_mac_from(port_iface.mac)
 
+    def _validate_copy_mac_from_iface_is_in_port(self, iface):
+        iface_port = iface.port
```
phoracek: Would you create a helper function for the code below that finds the iface_port? It would make the function more readable and you would not need to pre-initialize this variable.

phoracek (changes requested):
Very minor nit, feel free to ignore that,

libnmstate/ifaces/ifaces.py
rhrazdil: I think your suggestion reads better, apart from performance. Done

libnmstate/ifaces/ifaces.py
rhrazdil: Done

libnmstate/ifaces/ifaces.py
```diff
@@ -198,6 +193,31 @@ def _apply_copy_mac_from(self):
...
+        if iface.copy_mac_from not in iface_port:
+            raise NmstateValueError(
+                f"The interface {iface.name} is holding invalid "
+                f"{Interface.COPY_MAC_FROM} property "
+                f"as {iface.copy_mac_from} is not in the port "
+                f"list: {iface.port}"
+            )
+
+    def _port_for_interface(self, iface):
+        if iface.type == InterfaceType.OVS_INTERFACE:
```
phoracek, line 207: This does not seem correct. It returns the bridge itself, not the list of its ports, no?

So while it would perform the check as expected, it does not do what the function name suggests.

libnmstate/ifaces/ifaces.py
```diff
@@ -198,6 +193,32 @@ def _apply_copy_mac_from(self):
...
+            ovs_bridges = (
+                x for x in self.all_kernel_ifaces.values()
+                if x.type == InterfaceType.OVS_BRIDGE
+                   and iface.name in x.port)
+            ovs_bridge = next(ovs_bridges, None)
+            if ovs_bridge is None:
+                raise NmstateValueError(
+                    f"The interface {iface.name} is holding invalid "
```
phoracek: This is not a valid exception message for a function that has a sole purpose of looking up a `port` list for interface. It should raise that this ovs-interface does not belong to any bridge

libnmstate/ifaces/ifaces.py
phoracek: Resolved

phoracek (changes requested):

phoracek (approved):
Looks good to me, thanks o/

libnmstate/ifaces/ifaces.py
rhrazdil: Thanks!

libnmstate/ifaces/ifaces.py
```diff
@@ -198,6 +193,32 @@ def _apply_copy_mac_from(self):
...
+                f"{Interface.COPY_MAC_FROM} property "
+                f"as {iface.copy_mac_from} is not in the port "
+                f"list: {iface.port}"
+            )
+
+    def _port_for_interface(self, iface):
+        if iface.type == InterfaceType.OVS_INTERFACE:
+            ovs_bridges = (
+                x
+                for x in self.all_ifaces()
```
cathay4t: Please use `self.all_user_space_ifaces` to search OVS bridge only.

cathay4t (changes requested):

libnmstate/ifaces/ifaces.py
rhrazdil: Thanks! [Updated](https://github.com/nmstate/nmstate/compare/79e191b615bd5aa69cdc4e7a30d866e6b055d074..d8396b8a764a10da9cfec0f61b58b624e6fdce3f)

cathay4t (approved):

PR #1710, opened 2021-10-18
gitignore: add clib test binary to .gitignore

ffmancera:
The binaries should not be tracked by git.


ffmancera:
All the tests are failing because dockerhub does not allow automated builds for free accounts anymore. I am doing a manual build tomorrow to fix this but we should try a different solution.

PR #1711, opened 2021-10-18
Makefile: use the correct path for clib tests

ffmancera:
The path test/clib is incorrect. The correct path for clib tests is
src/clib/test.


ffmancera:
All the tests are failing because dockerhub does not allow automated builds for free accounts anymore. I am doing a manual build tomorrow to fix this but we should try a different solution.

cathay4t (approved):

PR #1713, opened 2021-10-19
rust: add golang bindings

ffmancera:
This patch is introducing golang bindings for the C bindings of
nmstate. The following functions are being introduced:

 - RetrieveNetState(flags int)
   * Returns (state, error) where state is a string containing the
     current state.

 - ApplyNetState(flags int, state string)
   * Returns (state, error) where state is a string containing the
     desired state applied.

This needs to handle the logs better as currently they are being shown
only when Nmstate fails on the operation.

### DIFF
--- a/rust/Makefile
+++ b/rust/Makefile
@@ -58,10 +58,12 @@ check:
 		cargo test -- --test-threads=1 --show-output --ignored; \
 	fi
 	make check -C src/clib/test
+	make check -C $(GO_MODULE_SRC)
 
 clean:
 	cargo clean
 	make clean -C src/clib/test
+	make clean -C $(GO_MODULE_SRC)
 
 install: $(CLI_EXEC_RELEASE)
 	install -p -v -D -m755 $(CLI_EXEC_RELEASE) \
--- a/rust/src/go/nmstate/Makefile
+++ b/rust/src/go/nmstate/Makefile
@@ -1,0 +1,13 @@
+# This Makefile is handling the go build, install, test and clean. It is needed
+# to avoid isues with the GOROOT and GOPATH.
+
+all: build check install
+
+build:
+	go build;
+
+check:
+	go test;
+
+clean:
+	go clean;
--- a/rust/src/go/nmstate/go.mod
+++ b/rust/src/go/nmstate/go.mod
@@ -1,0 +1,5 @@
+module github.com/nmstate/nmstate/rust/src/go/nmstate
+
+go 1.16
+
+require github.com/stretchr/testify v1.7.0
--- a/rust/src/go/nmstate/go.sum
+++ b/rust/src/go/nmstate/go.sum
@@ -1,0 +1,10 @@
+github.com/davecgh/go-spew v1.1.0 h1:ZDRjVQ15GmhC3fiQ8ni8+OwkZQO4DARzQgrnXU1Liz8=
+github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
+github.com/pmezard/go-difflib v1.0.0 h1:4DBwDE0NGyQoBHbLQYPwSUPoCMWR5BEzIk/f1lZbAQM=
+github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
+github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
+github.com/stretchr/testify v1.7.0 h1:nwc3DEeHmmLAfoZucVR881uASk0Mfjw8xYJ99tb5CcY=
+github.com/stretchr/testify v1.7.0/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
+gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
+gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c h1:dUUwHk2QECo/6vqA44rthZ8ie2QXMNeKRTHCNY2nXvo=
+gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c/go.mod h1:K4uyk7z7BCEPqu6E+C64Yfv1cQ7kz7rIZviUmN+EgEM=
--- a/rust/src/go/nmstate/nmstate.go
+++ b/rust/src/go/nmstate/nmstate.go
@@ -1,0 +1,182 @@
+package nmstate
+
+// #cgo CFLAGS: -g -Wall
+// #cgo LDFLAGS: -lnmstate
+// #include <nmstate.h>
+// #include <stdlib.h>
+import "C"
+import (
+	"fmt"
+	"io"
+	"time"
+)
+
+type Nmstate struct {
+	timeout	    uint
+	logsWriter  io.Writer
+	flags	    byte
+}
+
+const (
+	kernelOnly = 1 << iota
+	noVerify
+	includeStatusData
+	includeSecrets
+	noCommit
+)
+
+func New(options ...func(*Nmstate)) *Nmstate {
+	return &Nmstate{}
+}
+
+func WithTimeout(timeout time.Duration) func(*Nmstate) {
+	return func(n *Nmstate) {
+		n.timeout = uint(timeout.Seconds())
+	}
+}
+
+func WithLogsWritter(log_writter io.Writer) func(*Nmstate) {
+	return func(n *Nmstate) {
+		n.logsWriter = log_writter
+	}
+}
+
+func WithKernelOnly() func(*Nmstate) {
+	return func(n *Nmstate) {
+		n.flags = n.flags | kernelOnly
+	}
+}
+
+func WithNoVerify() func(*Nmstate) {
+	return func(n *Nmstate) {
+		n.flags = n.flags | noVerify
+	}
+}
+
+func WithIncludeStatusData() func(*Nmstate) {
+	return func(n *Nmstate) {
+		n.flags = n.flags | includeStatusData
+	}
+}
+
+func WithIncludeSecrets() func(*Nmstate) {
+	return func(n *Nmstate) {
+		n.flags = n.flags | includeSecrets
+	}
+}
+
+func WithNoCommit() func(*Nmstate) {
+	return func(n *Nmstate) {
+		n.flags = n.flags | noCommit
+	}
+}
+
+// Retrieve the network state in json format. This function returns the current
+// network state or an error.
+func (n *Nmstate) RetrieveNetState() (string, error) {
+	var (
+		state    *C.char
+		log      *C.char
+		err_kind *C.char
+		err_msg  *C.char
+	)
+	rc := C.nmstate_net_state_retrieve(C.uint(n.flags), &state, &log, &err_kind, &err_msg)
+	defer func() {
+		C.nmstate_net_state_free(state)
+		C.nmstate_err_msg_free(err_msg)
+		C.nmstate_err_kind_free(err_kind)
+		C.nmstate_log_free(log)
+	}()
+	_, err := io.WriteString(n.logsWriter, C.GoString(log))
+	if err != nil {
+		return "", fmt.Errorf("failed writting logs: %v", err)
+	}
+	if rc != 0 {
+		return "", fmt.Errorf("failed retrieving nmstate net state with rc: %d, err_msg: %s, err_kind: %s", rc, C.GoString(err_msg), C.GoString(err_kind))
+	}
+	return C.GoString(state), nil
+}
+
+// Apply the network state in json format. This function returns the applied
+// network state or an error.
+func (n *Nmstate) ApplyNetState(state string) (string, error) {
+	var (
+		c_state  *C.char
+		log      *C.char
+		err_kind *C.char
+		err_msg  *C.char
+	)
+	c_state = C.CString(state)
+	rc := C.nmstate_net_state_apply(C.uint(n.flags), c_state, C.uint(n.timeout), &log, &err_kind, &err_msg)
+
+	defer func() {
+		C.nmstate_net_state_free(c_state)
+		C.nmstate_err_msg_free(err_msg)
+		C.nmstate_err_kind_free(err_kind)
+		C.nmstate_log_free(log)
+	}()
+	_, err := io.WriteString(n.logsWriter, C.GoString(log))
+	if err != nil {
+		return "", fmt.Errorf("failed writting logs: %v", err)
+	}
+	if rc != 0 {
+		return "", fmt.Errorf("failed applying nmstate net state %s with rc: %d, err_msg: %s, err_kind: %s", state, rc, C.GoString(err_msg), C.GoString(err_kind))
+	}
+	return state, nil
+}
+
+// Commit the checkpoint path provided. This function returns the committed
+// checkpoint path or an error.
+func (n *Nmstate) CommitCheckpoint(checkpoint string) (string, error) {
+	var (
+		c_checkpoint *C.char
+		log	     *C.char
+		err_kind     *C.char
+		err_msg      *C.char
+	)
+	c_checkpoint = C.CString(checkpoint)
+	rc := C.nmstate_checkpoint_commit(c_checkpoint, &log, &err_kind, &err_msg)
+
+	defer func() {
+		C.nmstate_checkpoint_free(c_checkpoint)
+		C.nmstate_err_msg_free(err_msg)
+		C.nmstate_err_kind_free(err_kind)
+		C.nmstate_log_free(log)
+	}()
+	_, err := io.WriteString(n.logsWriter, C.GoString(log))
+	if err != nil {
+		return "", fmt.Errorf("failed writting logs: %v", err)
+	}
+	if rc != 0 {
+		return "", fmt.Errorf("failed commiting checkpoint %s with rc: %d, err_msg: %s, err_kind: %s", checkpoint, rc, C.GoString(err_msg), C.GoString(err_kind))
+	}
+	return checkpoint, nil
+}
+
+// Rollback to the checkpoint provided. This function returns the checkpoint
+// path used for rollback or an error.
+func (n *Nmstate) RollbackCheckpoint(checkpoint string) (string, error) {
+	var (
+		c_checkpoint *C.char
+		log	     *C.char
+		err_kind     *C.char
+		err_msg      *C.char
+	)
+	c_checkpoint = C.CString(checkpoint)
+	rc := C.nmstate_checkpoint_rollback(c_checkpoint, &log, &err_kind, &err_msg)
+
+	defer func() {
+		C.nmstate_checkpoint_free(c_checkpoint)
+		C.nmstate_err_msg_free(err_msg)
+		C.nmstate_err_kind_free(err_kind)
+		C.nmstate_log_free(log)
+	}()
+	_, err := io.WriteString(n.logsWriter, C.GoString(log))
+	if err != nil {
+		return "", fmt.Errorf("failed writting logs: %v", err)
+	}
+	if rc != 0 {
+		return "", fmt.Errorf("failed when doing rollback checkpoint %s with rc: %d, err_msg: %s, err_kind: %s", checkpoint, rc, C.GoString(err_msg), C.GoString(err_kind))
+	}
+	return checkpoint, nil
+}
--- a/rust/src/go/nmstate/nmstate_test.go
+++ b/rust/src/go/nmstate/nmstate_test.go
@@ -1,0 +1,29 @@
+package nmstate
+
+import "testing"
+import "os"
+import "github.com/stretchr/testify/assert"
+
+func TestRetrieveNetState(t *testing.T) {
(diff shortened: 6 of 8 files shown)


ffmancera:
@qinqon Could you take a look? Thanks!

rust/src/go/pkg/nmstate/Makefile
```diff
@@ -0,0 +1,16 @@
+# This Makefile is handling the go build, install, test and clean. It is needed
+# to avoid isues with the GOROOT and GOPATH.
+
+all: build check install
+
+build:
+	go build;
+
+install: build
+	go install;
```
cathay4t: You need to honor `DESTDIR` and maybe other variable for rpmbuild.

rust/src/go/pkg/nmstate/nmstate.go
```diff
@@ -0,0 +1,53 @@
...
+
+	defer func() {
+		C.nmstate_net_state_free(c_state)
+		C.nmstate_err_msg_free(err_msg)
+		C.nmstate_err_kind_free(err_kind)
+		C.nmstate_log_free(log)
+	}()
+	if rc != 0 {
+		//TODO: Handle the logs properly
+		return "", fmt.Errorf("failed applying nmstate net state %s with rc: %d and logs: %s", state, rc, C.GoString(log))
```
cathay4t: The error kind is not exposed, does go have error type stuff?

rust/src/go/pkg/nmstate/nmstate_test.go
```diff
@@ -0,0 +1,19 @@
+package nmstate
+
+import "testing"
+import "github.com/stretchr/testify/assert"
+
+const NMSTATE_FLAG_NONE = 0
+const NMSTATE_FLAG_KERNEL_ONLY = 1 << 1
```
cathay4t: There are 3 constants I forgot to expose in C header:

```
const NMSTATE_FLAG_NO_VERIFY: u32 = 1 << 2;
const NMSTATE_FLAG_INCLUDE_STATUS_DATA: u32 = 1 << 3;
const NMSTATE_FLAG_INCLUDE_SECRETS: u32 = 1 << 4;

```

cathay4t (changes requested):

rust/src/go/pkg/nmstate/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/nmstate/nmstate
```
qinqon: I am trying to tes to integrate this PR at kubernetes-nmstate but I think the module has to be `module github.com/nmstate/nmstate/rust/src/go/pkg/nmstate/` since it's the place where the go.mod is going to be living.

qinqon (changes requested):

rust/src/go/pkg/nmstate/go.mod
ffmancera: Done, thanks!

.gitignore
```diff
@@ -27,3 +27,4 @@ _kubevirtci
 rust/Cargo.lock
 rust/target
 rust/src/clib/test/nmstate_test
+rust/src/go/pkg/nmstate/*.sum
```
qinqon: The go.sum cannot be excluded, it's the one that pin the dependencies.

rust/Makefile
```diff
@@ -11,6 +11,7 @@ CLIB_PKG_CONFIG_FILE=nmstate.pc
 CLIB_PKG_CONFIG_FILE_IN=src/clib/$(CLIB_PKG_CONFIG_FILE).in
 PYTHON_MODULE_NAME=nmstate
 PYTHON_MODULE_SRC=src/python/nmstate
+GO_MODULE_SRC=src/go/pkg/nmstate
```
qinqon: We don't need the "pkg" part we can go directly with `src/go/nmstate` also having it under rust directory is weird too maybe it has to live next to the rust directory at `nmstate/go/nmstate`

rust/Makefile
```diff
@@ -76,6 +79,9 @@ install: $(CLI_EXEC_RELEASE)
 	    install -p -v -D -m644 $(PYTHON_MODULE_SRC)/*.py \
 		    $(DESTDIR)$(PYTHON3_SITE_DIR)/$(PYTHON_MODULE_NAME)/; \
 	fi
+	if [ $(SKIP_GO_INSTALL) != 1 ];then \
+	    make install -C $(GO_MODULE_SRC); \
+	fi
```
qinqon: There is no need to install go src, the user points directly to github.com name.

rust/src/go/pkg/nmstate/Makefile
qinqon: We don't have a binary to install this is just a golang lib that users will compile before use it, there is no need for an RPM.

rust/src/go/pkg/nmstate/Makefile
```diff
@@ -0,0 +1,16 @@
...
+
+check:
+	go test;
```
qinqon: There is no unit test here, we can check linters though.

rust/src/go/pkg/nmstate/nmstate.go
```diff
@@ -0,0 +1,53 @@
+package nmstate
+
+// #cgo CFLAGS: -g -Wall
+// #cgo LDFLAGS: -L/usr/local/lib64 -lnmstate -Wl,-rpath=/usr/local/lib64
```
qinqon: Normally `/usr/local/lib64` is implicit you don't need to specify it, -lnmstate should be enough.

rust/src/go/pkg/nmstate/nmstate.go
```diff
@@ -0,0 +1,53 @@
...
+// #include <nmstate.h>
+// #include <stdlib.h>
+import "C"
+import "fmt"
+
+func RetrieveNetState(flags uint) (string, error) {
```
qinqon: We can use bitmask at golang, https://play.golang.org/p/_bTfJBuonzs

```suggestion
type Flags uint

const None Flags = 0

const (
    KernelOnly = 1 << iota
)
func RetrieveNetStateWithFlags(flags Flags) (string, error) {
...
}

func RetrieveNetState() (string, error) {
    return RetrieveNetStateWithFlags(None)
}
```

rust/src/go/pkg/nmstate/nmstate.go
```diff
@@ -0,0 +1,53 @@
...
+	rc := C.nmstate_net_state_retrieve(C.uint(flags), &state, &log, &err_kind, &err_msg)
+
+	defer func() {
+		C.nmstate_net_state_free(state)
+		C.nmstate_err_msg_free(err_msg)
+		C.nmstate_err_kind_free(err_kind)
+		C.nmstate_log_free(log)
+	}()
+	if rc != 0 {
+		//TODO: Handle the logs properly
```
qinqon: My two cents is to add a "verbose" bitmask to print the errors or pass an new `io.Writer` argument to print the logs where the user want to, maybe that's the best option.

qinqon (changes requested):
What do you think about wrapping all with a struct that will have the flags and a writer to print logs ? 

```golang
type Controller struct {
    Flags Flags
    Logs  io.Writer
}

func (c *Controller) RetrieveNetState() (string, error){...}
func (c *Controller) ApplyNetState(state string) (string, error){...}

```

Users normally want to retrieve and apply the same way (at least our case), so they will use this like

```golang
ncl := nmstate.Controller{Flags: nmstate.KernelOnly, Output: os.Stdout}
state, err := ncl.RetrieveNetState()...
if err != nil {...}
err := ncl.ApplyNetState(desiredState)
if err != nil {...}
```

rust/Makefile
qinqon: Maybe we can setup sone "vanity import" at `nmstate.io` so users will reference it and nmstate can change the path where the dep is living without affecting them.

Placing a html somewhere at nmstate.io should be enough and github pages will serve it.

UPDATE: blog post about doing it with github pages https://alexejk.io/article/github-pages-go-vanity-urls/

rust/src/go/pkg/nmstate/Makefile
ffmancera: I've dropped the install step. Thanks!

.gitignore
ffmancera: Done, it is not excluded anymore.

rust/Makefile
ffmancera: Done, I have dropped it. Thank you!

rust/src/go/pkg/nmstate/Makefile
ffmancera: Wouldn't this run `rust/src/go/pkg/nmstate/nmstate_test.go`? If not, I will drop it too. Thanks!

rust/src/go/pkg/nmstate/nmstate.go
ffmancera: Done, I have dropped it. Currently it is: `// #cgo LDFLAGS: -lnmstate`. Thanks!

rust/src/go/pkg/nmstate/Makefile
qinqon: Now that there are unit tests that should run it, although is better to do `go test ./...` so it runs all the unit tests tha are present event a subdirs.

rust/src/go/pkg/nmstate/nmstate.go
ffmancera: Done, I have introduced bitmask following the pattern you described. Thank you!

rust/src/go/pkg/nmstate/Makefile
ffmancera: Changed to `go test ./`, thanks!

rust/src/go/pkg/nmstate/nmstate.go
ffmancera: According to @qinqon the error kind is not used in golang. So we don't need it, if any customer is willing to use it they can request it later. But the logs needs to be written properly.

rust/src/go/pkg/nmstate/nmstate_test.go
ffmancera: I exposed all the flags now.

rust/src/go/pkg/nmstate/nmstate.go
ffmancera: Done, the user will need to pass a io.Writer. So this way they can use the logs as they prefer.

rust/src/go/nmstate/Makefile
```diff
@@ -0,0 +1,13 @@
+# This Makefile is handling the go build, install, test and clean. It is needed
```
qinqon, line 1: Let's move this to go top level  `rust/src/go/Makefile`

rust/src/go/nmstate/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/nmstate/nmstate/rust/src/go/nmstate
```
qinqon, line 1: Move to `rust/src/go/go.mod`

rust/src/go/nmstate/go.sum
```diff
@@ -0,0 +1,10 @@
+github.com/davecgh/go-spew v1.1.0 h1:ZDRjVQ15GmhC3fiQ8ni8+OwkZQO4DARzQgrnXU1Liz8=
```
qinqon, line 1: ditto

rust/src/go/nmstate/nmstate.go
```diff
@@ -0,0 +1,111 @@
+package nmstate
```
qinqon, line 1: Add the CGO build tag so this is compiled only whent CGO is enabled.

```
// +build cgo
```

rust/src/go/nmstate/nmstate.go
```diff
@@ -0,0 +1,111 @@
...
+
+const None Flags = 0
+
+const (
+	KernelOnly Flags = 1 << iota
+	NoVerify
+	IncludeStatusData
+	IncludeSecrets
+	NoCommit
+)
```
qinqon: You can do 

```golang
const (
	None = 0 << iota
        KernelOnly Flags
	NoVerify
	IncludeStatusData
	IncludeSecrets
	NoCommit
)
```

rust/src/go/nmstate/nmstate.go
```diff
@@ -0,0 +1,111 @@
...
+	)
+	rc := C.nmstate_net_state_retrieve(C.uint(flags), &state, &log, &err_kind, &err_msg)
+
+	defer func() {
+		C.nmstate_net_state_free(state)
+		C.nmstate_err_msg_free(err_msg)
+		C.nmstate_err_kind_free(err_kind)
+		C.nmstate_log_free(log)
+	}()
+	io.WriteString(logs, C.GoString(log))
```
qinqon: You have handler `WriteString` error https://pkg.go.dev/io#WriteString

```golang
if err := io.WriteString(logs, C.GoString(log)); err != nil {
    return "", fmt.Errorf("failed wiring logs: %v", err)
}
```

rust/src/go/nmstate/nmstate.go
```diff
@@ -0,0 +1,111 @@
...
+	c_state = C.CString(state)
+	rc := C.nmstate_net_state_apply(C.uint(flags), c_state, C.uint(timeout), &log, &err_kind, &err_msg)
+
+	defer func() {
+		C.nmstate_net_state_free(c_state)
+		C.nmstate_err_msg_free(err_msg)
+		C.nmstate_err_kind_free(err_kind)
+		C.nmstate_log_free(log)
+	}()
+	io.WriteString(logs, C.GoString(log))
```
qinqon: ditto

rust/src/go/nmstate/nmstate.go
```diff
@@ -0,0 +1,111 @@
...
+	c_checkpoint = C.CString(checkpoint)
+	rc := C.nmstate_checkpoint_commit(c_checkpoint, &log, &err_kind, &err_msg)
+
+	defer func() {
+		C.nmstate_checkpoint_free(c_checkpoint)
+		C.nmstate_err_msg_free(err_msg)
+		C.nmstate_err_kind_free(err_kind)
+		C.nmstate_log_free(log)
+	}()
+	io.WriterString(logs, C.GoString(log))
```
qinqon: ditto

rust/src/go/nmstate/nmstate.go
```diff
@@ -0,0 +1,111 @@
...
+	c_checkpoint = C.CString(checkpoint)
+	rc := C.nmstate_checkpoint_rollback(c_checkpoint, &log, &err_kind, &err_msg)
+
+	defer func() {
+		C.nmstate_checkpoint_free(c_checkpoint)
+		C.nmstate_err_msg_free(err_msg)
+		C.nmstate_err_kind_free(err_kind)
+		C.nmstate_log_free(log)
+	}()
+	io.WriterString(logs, C.GoString(log))
```
qinqon: ditto

rust/src/go/nmstate/nmstate_test.go
```diff
@@ -0,0 +1,16 @@
...
+	netState, err := RetrieveNetState(None)
+	assert.NoError(t, err, "must succeed calling retrieve_net_state c binding")
+	assert.NotEmpty(t, netState, "net state should not be empty")
+}
+
+func TestRetrieveNetStateKernelOnly(t *testing.T) {
+	netState, err := RetrieveNetState(KernelOnly)
+	assert.NoError(t, err, "must succeed calling retrieve_net_state c binding")
+	assert.NotEmpty(t, netState, "net state should not be empty")
+}
```
qinqon: Don't we want to test the rest of them ?

qinqon (changes requested):
Add some go docs to the Exported functions.

rust/src/go/nmstate/nmstate_test.go
ffmancera: Yes, I will create test for the other functions in a follow up PR. I just wanted to introduce very basic test coverage

rust/src/go/nmstate/Makefile
ffmancera: We are going to keep using `rust/src/go/nmstate/Makefile` as we discussed. Later we can use the vanity imports to simplify the import. Thanks!

rust/src/go/nmstate/nmstate.go
ffmancera: Done, thanks!

rust/src/go/nmstate/nmstate.go
ffmancera: Done, thanks!

rust/src/go/nmstate/nmstate.go
ffmancera: Done, thank you! :-)

filanov:
@ffmancera Is it possible to add `Generate configurations` as part of this effort?

rust/src/go/nmstate/nmstate.go
```diff
@@ -0,0 +1,199 @@
...
+	"io"
+	"time"
+)
+
+type Nmstate struct {
+	Timeout	    uint
+	LogsWritter io.Writer
+	Flags	    byte
+	State	    string
+	Checkpoint  string
```
qinqon: These two has to be arguments of their functions.

rust/src/go/nmstate/nmstate.go
```diff
@@ -0,0 +1,199 @@
...
+import (
+	"fmt"
+	"io"
+	"time"
+)
+
+type Nmstate struct {
+	Timeout	    uint
+	LogsWritter io.Writer
+	Flags	    byte
```
qinqon: We don't hvae to expose this since they will be set with the `With` function.

rust/src/go/nmstate/nmstate.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+const (
+	none byte = 0 << iota
+	kernel_only
+	no_verify
+	include_status_data
+	include_secrets
+	no_commit
+)
```
qinqon:
```suggestion
const none = 0
const (
	kernelOnly = 1 << iota
	noVerify
	includeStatusData
	includeSecrets
	noCommit
)
```

rust/src/go/nmstate/nmstate.go
```diff
@@ -0,0 +1,199 @@
...
+		C.nmstate_err_msg_free(err_msg)
+		C.nmstate_err_kind_free(err_kind)
+		C.nmstate_log_free(log)
+	}()
+	_, err := io.WriteString(n.LogsWritter, C.GoString(log))
+	if err != nil {
+		return "", fmt.Errorf("failed writting logs: %v", err)
+	}
+	if rc != 0 {
+		return "", fmt.Errorf("failed retrieving nmstate net state with rc: %d", rc)
```
qinqon: We should add `err_msg`and `err_kind` here right ?

rust/src/go/nmstate/nmstate_test.go
```diff
@@ -0,0 +1,20 @@
...
+
+func TestRetrieveNetStateKernelOnly(t *testing.T) {
+	nms := nmstate.New(
+		nms.WithKernelOnly(),
+	)
+	netState, err := nms.RetrieveNetState()
+	assert.NoError(t, err, "must succeed calling retrieve_net_state c binding")
+	assert.NotEmpty(t, netState, "net state should not be empty")
+}
```
qinqon, line 29: Don't we want to test apply ?

qinqon (changes requested):

ffmancera:
> @ffmancera Is it possible to add `Generate configurations` as part of this effort?

Hi! What do you mind by `Generate configuration`, the `gc`  command? It is not implemented on Rust yet so it is not possible. If done later we can add the Golang binding, that won't be a problem. But adding the gc command here is too much.

filanov:
> > @ffmancera Is it possible to add `Generate configurations` as part of this effort?
> 
> Hi! What do you mind by `Generate configuration`, the `gc` command? It is not implemented on Rust yet so it is not possible. If done later we can add the Golang binding, that won't be a problem. But adding the gc command here is too much.

Yes the `gc` command. Great later is fine for me as well. 
Is there an issue  that i can follow? and what is the timeline for that? 
I'm using in a another project and it will really help us to have it as well.

rust/src/go/nmstate/nmstate.go
ffmancera: Done!

rust/src/go/nmstate/nmstate.go
ffmancera: Done!

rust/src/go/nmstate/nmstate.go
ffmancera: I noticed that `none` is not needed. I changed the other flags names to use the correct format.

rust/src/go/nmstate/nmstate.go
ffmancera: Done, thank you!

rust/src/go/nmstate/nmstate_test.go
ffmancera: Yes, but we are going to integrate this with the CI later. Let's merge a simple support and basic coverage. Thanks!

qinqon (approved):

ffmancera:
> > > @ffmancera Is it possible to add `Generate configurations` as part of this effort?
> > 
> > 
> > Hi! What do you mind by `Generate configuration`, the `gc` command? It is not implemented on Rust yet so it is not possible. If done later we can add the Golang binding, that won't be a problem. But adding the gc command here is too much.
> 
> Yes the `gc` command. Great later is fine for me as well. Is there an issue that i can follow? and what is the timeline for that? I'm using in a another project and it will really help us to have it as well.

I am sorry but there is not a timeline for this. We expect to do it in the following months but we cannot promise that. We are working on providing all the features in nmstate-rust. Binding will implement the features ASAP. Thank you!

PR #1720, opened 2021-10-28
rust, nm: support reapply operation when possible

ffmancera:
When activating a connection, Nmstate should try first to reapply and in
case it is not possible, it will fallback on the activation.


ffmancera:
I've tested this setting a static ip address and doing `ip monitor link dev eth1`. I couldn't see any `state UP` event.

cathay4t (approved):

PR #1721, opened 2021-10-29
rust, ip: Fix deserialization/serialization of ipv4/ipv6

ffmancera:
Nmstate is using a custom serializer for ipv4 and ipv6. We are relying
in `prop_list` to serialize the values but when creating a NetState
struct and doing `serde::to_string(net_state)` the `prop_list` is not
populated. Therefore it will not show the properties.

For example if we generate a NetState from the following file:

```yaml
---
interfaces:
  - name: eth1
    type: ethernet
    state: up
    ipv4:
      enabled: true
      address:
        - ip: 192.168.100.200
          prefix-length: 24
```

And we use the `serde::to_string()` method, it will show:

```yaml
---
interfaces:
  - name: eth1
    type: ethernet
    state: up
    ipv4:
      enabled: true
```

In order to fix this, we must implement our deserializer in which we
populate `prop_list`. This must be done on each struct that relies on
`prop_list` to serialize.

### DIFF
--- a/rust/src/lib/ip.rs
+++ b/rust/src/lib/ip.rs
@@ -1,19 +1,17 @@
+use std::fmt;
 use std::str::FromStr;
 
 use log::{debug, warn};
+use serde::de::{self, Deserializer, MapAccess, Visitor};
 use serde::{ser::SerializeStruct, Deserialize, Serialize, Serializer};
 
 use crate::NmstateError;
 
-#[derive(Debug, Clone, PartialEq, Deserialize, Default)]
+#[derive(Debug, Clone, PartialEq, Default)]
 pub struct InterfaceIpv4 {
-    #[serde(default)]
     pub enabled: bool,
-    #[serde(skip)]
     pub prop_list: Vec<&'static str>,
-    #[serde(default)]
     pub dhcp: bool,
-    #[serde(rename = "address", default)]
     pub addresses: Vec<InterfaceIpAddr>,
 }
 
@@ -25,9 +23,9 @@ impl Serialize for InterfaceIpv4 {
         let mut serial_struct = serializer.serialize_struct(
             "ipv4",
             if self.enabled {
-                1
-            } else {
                 self.prop_list.len()
+            } else {
+                1
             },
         )?;
         serial_struct.serialize_field("enabled", &self.enabled)?;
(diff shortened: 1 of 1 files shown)


ffmancera:
@cathay4t IMO, we could move all the serializer/deserializer implementation to its own folder. I think using `prop_list` method is worth because the NetState much clear. But I will do that in a different PR. What do you think?

cathay4t (changes requested):
The `el8-nm_stable-integ_rust` CI is failed. The failure indicate your patch still have problems.

From my view, the `Vistor` seems overkill to me, could you try something like `enum Interface` did to simply the code?

Meanwhile, just FYI. the `prop_list` is designed for state merging from multiple plugins(and in future, merging from current to desire).

cathay4t:
The reapply failure is caused by race problem. When we use `self.dbus.get_connection_by_uuid()` we might still get the outdated `NmConnection`.

This patch should fix it:

```patch
diff --git a/rust/src/libnm_dbus/nm_api.rs b/rust/src/libnm_dbus/nm_api.rs
index c1a1160..80a579e 100644
--- a/rust/src/libnm_dbus/nm_api.rs
+++ b/rust/src/libnm_dbus/nm_api.rs
@@ -21,7 +21,7 @@ use crate::{
     active_connection::{
         get_nm_ac_by_obj_path, nm_ac_obj_path_uuid_get, NmActiveConnection,
     },
-    connection::{nm_con_get_from_obj_path, NmConnection, NmSettingConnection},
+    connection::{nm_con_get_from_obj_path, NmConnection},
     dbus::NmDbus,
     device::{
         nm_dev_delete, nm_dev_from_obj_path, NmDevice, NmDeviceState,
@@ -134,15 +134,7 @@ impl<'a> NmApi<'a> {
         nm_conn: &NmConnection,
     ) -> Result<(), NmError> {
         debug!("connection_add: {:?}", nm_conn);
-        if let NmConnection {
-            connection:
-                Some(NmSettingConnection {
-                    uuid: Some(ref uuid),
-                    ..
-                }),
-            ..
-        } = nm_conn
-        {
+        if let Some(uuid) = nm_conn.uuid() {
             if let Ok(con_obj_path) = self.dbus.get_connection_by_uuid(uuid) {
                 return self.dbus.connection_update(&con_obj_path, nm_conn);
             }
@@ -161,28 +153,20 @@ impl<'a> NmApi<'a> {
         }
     }
 
-    pub fn connection_reapply(&self, uuid: &str) -> Result<(), NmError> {
-        debug!("connection_reapply: {}", uuid);
-        let obj_path = &self.dbus.get_connection_by_uuid(uuid)?;
-        let nm_conn =
-            nm_con_get_from_obj_path(&self.dbus.connection, obj_path)?;
-        if let NmConnection {
-            connection:
-                Some(NmSettingConnection {
-                    iface_name: Some(ref iface_name),
-                    ..
-                }),
-            ..
-        } = nm_conn
-        {
+    pub fn connection_reapply(
+        &self,
+        nm_conn: &NmConnection,
+    ) -> Result<(), NmError> {
+        debug!("connection_reapply: {:?}", nm_conn);
+        if let Some(iface_name) = nm_conn.iface_name() {
             let nm_dev_obj_path = self.dbus.nm_dev_obj_path_get(iface_name)?;
-            self.dbus.nm_dev_reapply(&nm_dev_obj_path, &nm_conn)
+            self.dbus.nm_dev_reapply(&nm_dev_obj_path, nm_conn)
         } else {
             Err(NmError::new(
                 ErrorKind::InvalidArgument,
                 format!(
-                    "Failed to extract interface name from connection {}",
-                    uuid
+                    "Failed to extract interface name from connection {:?}",
+                    nm_conn
                 ),
             ))
         }
```

cathay4t:
The reapply issue mention above should be fixed by https://github.com/nmstate/nmstate/pull/1723

ffmancera:
> The `el8-nm_stable-integ_rust` CI is failed. The failure indicate your patch still have problems.
> 

Fixed! Thanks!

> From my view, the `Vistor` seems overkill to me, could you try something like `enum Interface` did to simply the code?

I agree that the visitor is too much, but we cannot do something similar to what we did for `enum Interface` because for types it works differents. Please refer to https://serde.rs/impl-deserialize.html

> 
> Meanwhile, just FYI. the `prop_list` is designed for state merging from multiple plugins(and in future, merging from current to desire).

cathay4t (approved):

PR #1724, opened 2021-11-03
ovs: remove ovs-port- prefix for OVS bridge LAG ports

rhrazdil:
The ovnkube-node pods perform health checks, and they seem to expect the port to be called the same
as the underlying interface:
https://github.com/ovn-org/ovn-kubernetes/blob/acfc09e8a7ffae73918bece102d0c520e4eb48fe/go-controller/pkg/node/healthcheck.go#L298

Renaming of the Ports causes these pods to crash.

Signed-off-by: Radim Hrazdil <rhrazdil@redhat.com>

### DIFF
--- a/libnmstate/nm/connection.py
+++ b/libnmstate/nm/connection.py
@@ -121,6 +121,8 @@ def create_new_nm_simple_conn(iface, nm_profile):
             con_name = con_name + "-br"
         elif iface.type == InterfaceType.OVS_INTERFACE:
             con_name = con_name + "-if"
+        elif iface.type == InterfaceType.OVS_PORT:
+            con_name = con_name + "-port"
 
         con_setting.create(
             con_name,
--- a/libnmstate/nm/ovs.py
+++ b/libnmstate/nm/ovs.py
@@ -32,8 +32,6 @@ from libnmstate.schema import OvsDB
 from .common import NM
 
 
-PORT_PROFILE_PREFIX = "ovs-port-"
-
 CONTROLLER_TYPE_METADATA = "_controller_type"
 CONTROLLER_METADATA = "_controller"
 SETTING_OVS_EXTERNALIDS = "SettingOvsExternalIDs"
@@ -311,7 +309,7 @@ def create_iface_for_nm_ovs_port(iface):
     if ovs.is_ovs_lag_port(port_options):
         port_name = port_options[OB.Port.NAME]
     else:
-        port_name = PORT_PROFILE_PREFIX + iface_name
+        port_name = iface_name
     return OvsPortIface(
         {
             Interface.NAME: port_name,
--- a/tests/integration/nm/ovs_test.py
+++ b/tests/integration/nm/ovs_test.py
@@ -59,6 +59,14 @@ def bridge_with_ports(eth1_up):
 
 
 @pytest.fixture
+def bridge_port_types():
+    yield {
+        ETH1: InterfaceType.ETHERNET,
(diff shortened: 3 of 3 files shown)


cathay4t:
@rhrazdil Is `ovnkube` checking the NetworkManager `connection.id` or `connection.interface-name`.

Can we use `ovs-port-` prefix for `connection.id` and user desired interface name as `connection.interface-name`?

rhrazdil:
Hello @cathay4t 
By the error message in the ovnkube pod it's checking with ovs-vsctl:
```
I1101 14:47:29.362491 2671168 ovs.go:213] exec(16): stderr: "ovs-vsctl: no row \"ovs-port-eno1\" in table Interface\n"
I1101 14:47:29.362495 2671168 ovs.go:215] exec(16): err: exit status 1
F1101 14:47:29.362522 2671168 ovnkube.go:130] Bridge for interface failed for br-ex: Failed to find intfName for br-ex: failed to get ofport of ovs-port-eno1, stderr: "ovs-vsctl: no row \"ovs-port-eno1\" in table Interface\n", error: exit status 1
```

This part specifically: `"ovs-vsctl: no row \"ovs-port-eno1\" in table Interface\n` brings me to the conclusion that they expect the Port name to be the same as the Interface.


> Can we use ovs-port- prefix for connection.id and user desired interface name as connection.interface-name?
Pardon me, but I don't follow, could you elaborate?

EDIT:
> Can we use ovs-port- prefix for connection.id and user desired interface name as connection.interface-name?

If this means that the Port names in ovsdb would be preserved, then I think yes

cathay4t:
In that case, please change `nm/connection.py`: `create_new_nm_simple_conn()` to use `-port` postfix for OVS port as connection name, and use user defined name as interface name.

rhrazdil:
> In that case, please change nm/connection.py: create_new_nm_simple_conn() to use -port postfix for OVS port as connection name, and use user defined name as interface name.

Thanks! I'm hitting an issue with, when I add the `-port` postfix in `create_new_nm_simple_conn()`,
in this test: `test_remove_ovs_internal_iface_got_port_profile_removed`

The connection has the postfix
```
DEVICE    TYPE           STATE         CONNECTION    
brtest0   ovs-bridge     connected     brtest0-br    
ovstest0  ovs-interface  connected     ovstest0-if   
ovstest0  ovs-port       connected     ovstest0-port 
```
And the bridge has the interfaces unchanged
```
d26c2cae-4823-42b7-8ca9-deb75b276631
    Bridge "brtest0"
        Port "ovstest0"
            Interface "ovstest0"
                type: internal
    ovs_version: "2.11.0"
```

But the test takes the interface names from the bridge connections: https://github.com/nmstate/nmstate/blob/base/tests/integration/nm/ovs_test.py#L122
which contains `['eth1', 'ovstest0']`, so the test fails on this [assert](https://github.com/nmstate/nmstate/blob/base/tests/integration/nm/ovs_test.py#L124)
because of th missing postfix.

> and use user defined name as interface name.

Do you mean setting the name of the interface (not the ovs-port) to be called with the `-port`
postfix as well? Where would be the right place to make this change please?

rhrazdil:
@cathay4t Hey, updated the PR according to your suggestion.

Failure in pull-nmstate-integ_tier1-k8s seems to be unrelated as it fails on `base` branch as well

ffmancera (approved):
LGTM! Could you take a look @cathay4t?

cathay4t:
Looks good to me.

cathay4t (approved):

PR #1733, opened 2021-11-17
rust, base: support accept-all-mac-addresses mode

ffmancera:
Please, notice that `accept-all-mac-addresses` DBus property is a i32
because it uses `NM_TERNARY` values. In Nmstate we consider it as
boolean so we need to convert the value.


ffmancera:
We need to check on NM version before trying to set the property.

bengal:
FWIW, looks good to me.

PR #1735, opened 2021-11-18
rust: Add initial VLAN support

almusil:
Add initial support of vlan so creating and removing
vlans should work.

### DIFF
--- a/rust/src/libnm_dbus/connection/vlan.rs
+++ b/rust/src/libnm_dbus/connection/vlan.rs
@@ -1,0 +1,52 @@
+use crate::dbus_value::{own_value_to_string, own_value_to_u32};
+use crate::NmError;
+use std::collections::HashMap;
+use std::convert::TryFrom;
+
+#[derive(Debug, Clone, PartialEq, Default)]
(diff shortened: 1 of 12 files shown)


rust/src/libnm_dbus/connection/vlan.rs
```diff
@@ -0,0 +1,52 @@
+use crate::dbus_value::{own_value_to_string, own_value_to_u32};
+use crate::NmError;
+use std::collections::HashMap;
+use std::convert::TryFrom;
+
+#[derive(Debug, Clone, PartialEq, Default)]
+pub struct NmSettingVlan {
+    pub base: Option<String>,
```
cathay4t: Please use `parent` to be consistent with NM dbus interface.

cathay4t (changes requested):

rust/src/libnm_dbus/connection/vlan.rs
almusil: Done

cathay4t (approved):

PR #1739, opened 2021-11-19
rust, libnm_dbus: Make some improvements the libnm_dbus library

almusil:
- Use serde by proxy going through TryFrom when we are getting answer from dbus.  
- Remove some clones that are not needed
- Add macro for more convenient way to extract values from dbus `HashMap` responses


almusil:
@ffmancera @cathay4t  
Please take a look

rust/src/libnm_dbus/connection/bridge.rs
```diff
@@ -58,19 +58,20 @@ impl NmSettingBridge {
...
+#[serde(try_from = "DbusDictionary")]
 pub struct NmSettingBridgePort {
     _other: HashMap<String, zvariant::OwnedValue>,
 }
 
-impl TryFrom<HashMap<String, zvariant::OwnedValue>> for NmSettingBridgePort {
+impl TryFrom<DbusDictionary> for NmSettingBridgePort {
     type Error = NmError;
     fn try_from(
         setting_value: HashMap<String, zvariant::OwnedValue>,
```
cathay4t: The TryFrom is using `DbusDictionary`, so the `fn try_from` should be also using `DbusDictionary`

cathay4t (changes requested):

rust/src/libnm_dbus/connection/bridge.rs
almusil: Good catch, thanks, done

cathay4t (approved):
LGTM.

PR #1740, opened 2021-11-19
rust,cli: Add manual commit/rollback

qinqon:
As an effort to brindge parity to nmstate rust from python version this PR add the manual commit/rollback feature and also the `--timeout` argument to `apply` default it to 60 seconds.


rust/src/lib/checkpoint.rs
```diff
@@ -0,0 +1,11 @@
+use crate::nm::{nm_checkpoint_destroy, nm_checkpoint_rollback};
+
+use crate::NmstateError;
+
+pub fn checkpoint_rollback(checkpoint: &str) -> Result<(), NmstateError> {
+    return nm_checkpoint_rollback(checkpoint);
```
cathay4t: Simply put `nm_checkpoint_rollback(checkpoint)` is enough.

qinqon:
@cathay4t @ffmancera this is unrelated right ? https://github.com/nmstate/nmstate/runs/4396045140?check_suite_focus=true

rust/src/cli/ncl.rs
```diff
@@ -108,12 +147,33 @@ fn main() {
     } else if let Some(matches) = matches.subcommand_matches(SUB_CMD_APPLY) {
         let is_kernel = matches.is_present("KERNEL");
         let no_verify = matches.is_present("NO_VERIFY");
+        let no_commit = matches.is_present("NO_COMMIT");
+        let timeout = matches
+            .value_of("TIMEOUT")
+            .unwrap()
```
cathay4t: Instead of unwrapping , please use convert the error to `CliError`.

cathay4t (changes requested):
Please also patch the C binding and python API, then you can enable test of `nmstatectl_test.py` (patching it to allow using `ncl` instead of `nmstatectl` from a system variable).

cathay4t:
@qinqon If you think the C binding and Python binding it too complex, we can merge it and let me do it in another PR.

qinqon:
> Please also patch the C binding and python API, then you can enable test of `nmstatectl_test.py` (patching it to allow using `ncl` instead of `nmstatectl` from a system variable).

Cool, will do that.

rust/src/lib/checkpoint.rs
qinqon: It was already done at newer version, snappy complains about it.

qinqon:
@cathay4t does it make sense to change tox to also lint rust python code ?

cathay4t:
@qinqon I dislike tox.

qinqon:
> @qinqon I dislike tox.

Well we have python code there and the linting is integrated with tox, is better tox than nothing.

cathay4t:
> > @qinqon I dislike tox.
> 
> Well we have python code there and the linting is integrated with tox, is better tox than nothing.

We may expand current tox when we move all python code to rust one. That would 2022 Q3.

cathay4t:
@qinqon The CI is failing on your tox changes. Please take a look.

qinqon:
Looks like rust code is not copied into the container 

```
pylint run-test-pre: PYTHONHASHSEED='2263479695'
pylint runtests: commands[0] | pylint --errors-only libnmstate nmstatectl tests/lib tests/ctl tests/integration setup.py ./rust/src/python
************* Module python
rust/src/python/__init__.py:1:0: F0010: error while code parsing: Unable to load file rust/src/python/__init__.py:
[Errno 2] No such file or directory: 'rust/src/python/__init__.py' (parse-error)
```

qinqon:
@cathay4t Have removed the tox bits, since rust changes cannot break the python nmstate product.

ffmancera (changes requested):
It seems the CI is failing. https://github.com/nmstate/nmstate/runs/4483366656?check_suite_focus=true#step:5:1633 please, check it out! Thank you!

rust/src/lib/ifaces/inter_ifaces.rs
```diff
@@ -284,16 +284,6 @@ impl Interfaces {
         ))
     }
 
-    pub(crate) fn has_sriov_enabled(&self) -> bool {
```
cathay4t: Please restore this function.

cathay4t (changes requested):

rust/src/lib/ifaces/inter_ifaces.rs
qinqon: There is a warning if I added since it's no longer used at this PR, the behaviour with it is the same as with preivous nmstatectl python version, there is a timeout argument with a default value of 60s (the SRIOV timeout), so it's not needed to detect if we are at SRIOV or not.

cathay4t: That function is used to determine the timeout of verification retry, not the timeout of checkpoint.

qinqon: Right, fixing.

rust/src/libnm_dbus/nm_api.rs
```diff
@@ -57,13 +57,26 @@ impl<'a> NmApi<'a> {
...
+        if checkpoint_to_rollback.is_empty() {
+            checkpoint_to_rollback = self.last_active_checkpoint()?
+        }
+        debug!("checkpoint_rollback: {}", checkpoint_to_rollback);
+        self.dbus
+            .checkpoint_rollback(checkpoint_to_rollback.as_str())
+    }
+
+    fn last_active_checkpoint(&self) -> Result<String, NmError> {
+        Ok(self.dbus.checkpoints()?.remove(0))
```
cathay4t: The `remove()` function will panic which is not allowed as a library.

rust/src/lib/checkpoint.rs
```diff
@@ -0,0 +1,11 @@
+use crate::nm::{nm_checkpoint_destroy, nm_checkpoint_rollback};
+
+use crate::NmstateError;
+
+pub fn checkpoint_rollback(checkpoint: &str) -> Result<(), NmstateError> {
```
cathay4t: Please place them as function of `NetworkState`. For example: `NetworkState::checkpoint_rollback()`.

rust/src/python/libnmstate/clib_wrapper.py
```diff
@@ -134,3 +139,77 @@ def apply_net_state(
...
+        elif err_kind == "Bug":
+            raise NmstateInternalError(err_msg)
+        elif err_kind == "PluginFailure":
+            raise NmstatePluginError(err_msg)
+        elif err_kind == "NotImplementedError":
+            raise NmstateNotImplementedError(err_msg)
+        elif err_kind == "KernelIntegerRoundedError":
+            raise NmstateKernelIntegerRoundedError(err_msg)
+        else:
+            raise NmstateError(f"{err_kind}: {err_msg}")
```
cathay4t: These are duplicate code. Please place them in a function instead.

cathay4t (changes requested):

rust/src/lib/checkpoint.rs
qinqon: But this functions depends only on checkpoing not on NetworkState, they are free functions.

cathay4t: They are standalone. Just would like them in the namespace of NetworkState. Very few crates shipping top level functions, I like the trend of that.

qinqon:
Some typos at the python binding fixed still have some segmentation fault at the rust test
```
tests/integration/vlan_test.py::test_two_vlans_on_eth1_change_mtu_rollback + open_shell                                                                                                                                                      
+ res=139                                                                                                                                                                                                                                    
+ '[' 139 -ne 0 ']'                                                                                                                                                                                                                          
+ echo '*** ERROR: 139'                                                                                                                                                                                                                     
*** ERROR: 139              
```

```
tests/integration/vlan_test.py::test_two_vlans_on_eth1_change_mtu_rollback run.sh: line 3:  8820 Segmentation fault      pytest --verbose --verbose --log-level=DEBUG --log-date-format='%Y-%m-%d %H:%M:%S' --log-format='%(asctime)s %(filename)s:%(lineno)d %(levelname)s %(message)s' --durations=5 --cov /usr/lib/python*/site-packages/libnmstate --cov /usr/lib/python*/site-packages/nmstatectl --cov-report=term --cov-report=xml --log-file=pytest-run.log tests/integration/vlan_test.py -k test_two_vlans_on_eth1_change_mtu_rollback
```

qinqon:
@cathay4t rust lane is fixed, rollback timeout has to be 60 by default like python version.

cathay4t (approved):

PR #1743, opened 2021-11-26
rust: Check overbooked port and orphaned interface

almusil:
This PR adds check for overbooked port and orphaned interface.


rust/src/lib/error.rs
```diff
@@ -38,6 +38,22 @@ impl NmstateError {
     pub fn msg(&self) -> &str {
         self.msg.as_str()
     }
+
+    pub fn port_overbook(
```
cathay4t: Put it as `pub(crate) `or move to `ifaces/inter_ifaces_controller.rs`

rust/src/lib/iface.rs
```diff
@@ -460,6 +460,13 @@ impl Interface {
             br_iface.remove_port(port_name);
         }
     }
+
+    pub(crate) fn parent(&self) -> Option<&str> {
+        match self {
+            Interface::Vlan(vlan) => vlan.parent(),
```
cathay4t, line 466: OVS interface also has parent.

rust/src/lib/ifaces/inter_ifaces.rs
```diff
@@ -281,6 +281,74 @@ impl Interfaces {
...
+            })
+            // Replace interfaces that are updated
+            .map(|iface| {
+                update
+                    .get_iface(iface.name(), iface.iface_type())
+                    .unwrap_or(iface)
+            })
+            // Append all new interfaces
+            .chain(add.to_vec().into_iter())
+            .collect();
```
cathay4t: The code with these levels are impossible to read. Please create variable for some of steps, for use to know the workflow.

I can only tell the propose of it by guessing.

rust/src/lib/ifaces/inter_ifaces.rs
```diff
@@ -281,6 +281,74 @@ impl Interfaces {
     }
 }
 
+pub struct DesiredInterfaces<'a> {
```
cathay4t: This will confuse other developers. We call `desire` as user inputs. But this struct is holding both current and desire interfaces.

rust/src/lib/lib.rs
```diff
@@ -13,8 +13,8 @@ pub use crate::iface::{
     Interface, InterfaceState, InterfaceType, UnknownInterface,
 };
 pub use crate::ifaces::{
-    BaseInterface, DummyInterface, EthernetConfig, EthernetInterface,
-    Interfaces, LinuxBridgeConfig, LinuxBridgeInterface,
+    BaseInterface, DesiredInterfaces, DummyInterface, EthernetConfig,
```
cathay4t: DesiredInterfaces should not be public API.

cathay4t (changes requested):
The code is more like show off instead of maintainable code.

Meanwhile, please place all function as `pub(crate)`  unless clear requirement as public API.

rust/src/lib/iface.rs
almusil: Right, but ovs interface is in user_space, I don't mind adding it here if you want

rust/src/lib/lib.rs
almusil: Done

rust/src/lib/ifaces/inter_ifaces.rs
almusil: Renamed to `MergedInterfaces`

rust/src/lib/iface.rs
almusil: I have added parent for ovs interface as well

rust/src/lib/error.rs
almusil: Done

rust/src/lib/ifaces/inter_ifaces.rs
almusil: Done

rust/src/lib/error.rs
```diff
@@ -38,6 +38,22 @@ impl NmstateError {
     pub fn msg(&self) -> &str {
         self.msg.as_str()
     }
+
+    pub(crate) fn port_overbook(
```
cathay4t: This function is only used once. And other code of this project is using `NmstateError::new()`.
Please be consistent with other codes.

rust/src/lib/ifaces/inter_ifaces.rs
```diff
@@ -281,6 +281,76 @@ impl Interfaces {
...
+                .is_none()
+            {
+                merged.push(current);
+            }
+        }
+        for iface in merged.iter_mut() {
+            if let Some(updated) =
+                update.get_iface(iface.name(), iface.iface_type())
+            {
+                *iface = updated;
```
cathay4t, line 397: When a bridge is only mentioned by its name without defining a port list, your code are treating it like losing all ports.

cathay4t (changes requested):

rust/src/lib/ifaces/inter_ifaces.rs
cathay4t: And please place the code in `ifaces/inter_ifaces_controller.rs`

rust/src/lib/error.rs
almusil: Done

rust/src/lib/ifaces/inter_ifaces.rs
almusil: In the merged state nothing is mentioned only by name. It keeps list of all interfaces so the port information should still be there, unless I am missing something. Also this code is not specific only for interfaces with controllers so I am not sure if the `ifaces/inter_ifaces_controller.rs` is the right place where to put it.

cathay4t: The `update` is holding the interface mentioned in desire state and found in current state. It only contain information on user provides.

For example, you code does not fail on this yml. (Before this yml, br0 is holding eth1 as its port).

```yml
---
interfaces:
  - name: br0
  - name: br1
    type: linux-bridge
    state: up
    bridge:
      options:
        stp:
          enabled: false
      port:
        - name: eth1
```

almusil: True but I am not sure if this can solved by the new code that is checking that. This issue is the same for other interfaces e.g. vlan. So we need to probably solve proper merging of current state and desired state somewhere in the `gen_state_for_apply`. Do you think it should be part of this PR?

cathay4t: The vlan does not have overbook problem. You can just use current interface port list as fallback when desire is None.

For orphan vlan, iterate all interface to make sure vlan parent is found in current and not in absent list.

There is no other code need top level state merging. So please keep it dedicate for this purpose.

almusil: Maybe I am still missing something, the merged state does not contain interfaces marked as absent. So the vlan cannot find it's parent that will be absent.

cathay4t: Let me merge this PR and continue the work later.

PR #1747, opened 2021-12-03
rust, bond: add support to bond interfaces

ffmancera:
This is adding support to bond interfaces in rust.

### DIFF
--- a/rust/src/lib/ifaces/bond.rs
+++ b/rust/src/lib/ifaces/bond.rs
@@ -1,0 +1,571 @@
+use crate::{BaseInterface, ErrorKind, InterfaceType, NmstateError};
+use serde::{de::Error, Deserialize, Deserializer, Serialize};
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+#[serde(rename_all = "kebab-case")]
+pub struct BondInterface {
+    #[serde(flatten)]
+    pub base: BaseInterface,
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        rename = "link-aggregation"
+    )]
+    pub bond: Option<BondConfig>,
+}
+
+impl Default for BondInterface {
+    fn default() -> Self {
+        let mut base = BaseInterface::new();
+        base.iface_type = InterfaceType::Bond;
+        Self { base, bond: None }
+    }
+}
+
+impl BondInterface {
+    pub(crate) fn update_bond(&mut self, other: &BondInterface) {
+        if let Some(bond_conf) = &mut self.bond {
+            bond_conf.update(other.bond.as_ref());
+        } else {
+            self.bond = other.bond.clone();
+        }
+    }
+
+    // Return None when desire state does not mention ports
+    pub(crate) fn ports(&self) -> Option<Vec<&str>> {
+        self.bond
+            .as_ref()
+            .and_then(|bond_conf| bond_conf.port.as_ref())
+            .map(|ports| ports.as_slice().iter().map(|p| p.as_str()).collect())
+    }
+
+    pub(crate) fn pre_verify_cleanup(&mut self) {
+        self.base.pre_verify_cleanup();
+        self.drop_empty_arp_ip_target();
+        self.sort_ports();
+    }
+
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    fn sort_ports(&mut self) {
+        if let Some(ref mut bond_conf) = self.bond {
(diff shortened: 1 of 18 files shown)


rust/src/lib/ifaces/bond.rs
```diff
@@ -0,0 +1,475 @@
+use crate::{BaseInterface, ErrorKind, InterfaceType, NmstateError};
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+#[serde(rename_all = "kebab-case")]
+pub struct BondInterface {
+    #[serde(flatten)]
+    pub base: BaseInterface,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub link_aggregation: Option<BondConfig>,
```
cathay4t: This name is really annoying when typing. Could you change it to `bond` and use:

`#[serde(skip_serializing_if = "Option::is_none", rename = "link_aggregation")]`

rust/src/lib/ifaces/bond.rs
```diff
@@ -0,0 +1,475 @@
...
+const BOND_ARP_VALIDATE_ACTIVE: u32 = 1 << BOND_STATE_ACTIVE as u32;
+const BOND_ARP_VALIDATE_BACKUP: u32 = 1 << BOND_STATE_BACKUP as u32;
+const BOND_ARP_VALIDATE_ALL: u32 =
+    BOND_ARP_VALIDATE_ACTIVE | BOND_ARP_VALIDATE_BACKUP;
+const BOND_ARP_FILTER: u32 = BOND_ARP_VALIDATE_ALL + 1;
+const BOND_ARP_FILTER_ACTIVE: u32 = BOND_ARP_VALIDATE_ACTIVE | BOND_ARP_FILTER;
+const BOND_ARP_FILTER_BACKUP: u32 = BOND_ARP_VALIDATE_BACKUP | BOND_ARP_FILTER;
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
+#[serde(rename_all = "lowercase")]
```
cathay4t: `#[serde(rename_all = "kebab-case")]` is better.

rust/src/lib/ifaces/bond.rs
```diff
@@ -0,0 +1,475 @@
...
+        match self {
+            Self::Always => 0,
+            Self::Better => 1,
+            Self::Failure => 2,
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
+#[serde(rename_all = "lowercase")]
```
cathay4t: Every member has a `serde(rename)`. So there is no need for this `rename_all()`.

tests/integration/bond_test.py
```diff
@@ -190,7 +190,7 @@ def test_add_bond_with_port_and_ipv4(eth1_up, eth2_up, setup_remove_bond99):
                         eth1_up[Interface.KEY][0][Interface.NAME],
                         eth2_up[Interface.KEY][0][Interface.NAME],
                     ],
-                    Bond.OPTIONS_SUBTREE: {"miimon": "140"},
+                    Bond.OPTIONS_SUBTREE: {"miimon": 140},
```
cathay4t: This is incorrect. We should change what we already supported.

A better approach might be using `HashMap<String, serde_json::Value>` instead of `struct BondOptions`.

cathay4t (changes requested):
Good work!

Besides change requests, please also ignore bond option miss match during verification as we did for python function `match_ignore_bond_options`. To do that, in the `Interface.verify()`, ignore the mismatch when the reference contains `link-aggregation.options`

rust/src/lib/ifaces/bond.rs
ffmancera: Sure, done! Thanks.

rust/src/lib/ifaces/bond.rs
ffmancera: Done!

rust/src/lib/ifaces/bond.rs
ffmancera: Thanks!

tests/integration/bond_test.py
ffmancera: I have followed a different approach. I think it is important to be able to control all the supported properties. I have introduced a custom deserializing function `json_to_u32`, `json_to_u16` and `json_to_u8`. This way, the bond options will accept `"140"` and `140`. Thanks!

PR #1763, opened 2021-12-23
ovs, dpdk: add support to OVS-DPDK

ffmancera:


ffmancera:
The support works well but I am trying to make OVS DPDK work on CI.

cathay4t (approved):

PR #1768, opened 2021-12-29
rust, bond: drop ValueError and use InvalidArgument instead

ffmancera:
The integration test for nmstate-rust have been adjusted.


cathay4t (changes requested):
I think a better fix is fix bond code to use existing `ErrorKind::InvalidArgument`

cathay4t (approved):

cathay4t:
@ffmancera You have failure on rust CI `test_create_bond_without_mode`

PR #1795, opened 2022-01-28
rust: unify clib free functions

qinqon:
All the "free" functions are doing exactly the same and can be unify
into one.


cathay4t:
Could you explain why we should break C API in this PR?

qinqon:
> Could you explain why we should break C API in this PR?

We don't have to keep retrocompatibility yet right ? if this is the case I can put some wrapper functions to keep it.

The idea was that all the `free` functions do exactly the same, so one `free` functions is enough.

cathay4t (approved):

cathay4t:
Yes. Remove duplicate code is one benefit.  Approved. Please rebase.

qinqon:
> Yes. Remove duplicate code is one benefit. Approved. Please rebase.

Done

PR #1816, opened 2022-03-02
test,k8s: Ignore calico interfaces

qinqon:
The integration tests should ignore interfaces related to the CNI used
by the k8s cluster calico. This change introduces an env var to pass
inteface substring to ignore.


ffmancera (approved):
LGTM!

PR #1820, opened 2022-03-07
rust: Implement logger for dynamic lib

qinqon:
Implement an in memory logger for the rust clib that can dump the logs
until a certain point to a CString.

Based on https://docs.rs/memory_logger/latest/memory_logger/


qinqon:
The  `log`  function from Loggger is not called, something is wrong at `setup`.

cathay4t:
@qinqon I would like to take over this week as it is important for us to get logging in python CI test.

qinqon:
> @qinqon I would like to take over this week as it is important for us to get logging in python CI test.

The main problem we have is that the retrieved logs are the logs from all the functions, we need a feature to mark the scope of the logs to be the function we are calling, I remember the tokio logger library is able to do it.

https://github.com/tokio-rs/tracing

cathay4t:
Closing as duplicate of https://github.com/nmstate/nmstate/pull/1910

PR #1821, opened 2022-03-08
rust: Use From and Display to ser/des type

qinqon:
Rust NMState shows types that are unexpected with the following format:

```yaml
type:
  other: Other("IpTun")
```

So what it's suppose to be a scalar string value it's a dictionary.

This change use the already implemented for InterfaceType From and
Display traits to serialize and deserialize them so the value is always
a string.

### DIFF
--- a/rust/src/lib/iface.rs
+++ b/rust/src/lib/iface.rs
@@ -1,5 +1,5 @@
 use log::{error, warn};
-use serde::{Deserialize, Deserializer, Serialize};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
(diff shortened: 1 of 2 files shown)


cathay4t (dismissed):

qinqon:
Is serializing as an string with the double quotes, fixing it...

cathay4t:
The CI is failing. Nmstate is using `mac-vlan` while NetworkManager is using "macvlan".

qinqon:
@cathay4t inte_rust CI job is fine now.

cathay4t (approved):

PR #1833, opened 2022-03-18
rust: add support to ethtool

ffmancera:


automation/run-tests.sh
```diff
@@ -259,6 +259,15 @@ function run_tests {
...
             not test_explicitly_ignore_a_bridge_port' \
             ${nmstate_pytest_extra_args}"
+        exec_cmd "
+          env  \
+          PYTHONPATH=$CONTAINER_WORKSPACE/rust/src/python \
+          pytest \
+            $PYTEST_OPTIONS \
+            tests/integration/ethtool_test.py \
+            -k '\
+            not test_ethtool_invalid_feature' \
```
cathay4t: Maybe we should change this test case to expect `NmstateValueError` as we should not allowing arbitrary feture string. The failure should happen at validation stage.

The `#[serde(deny_unknown_fields)]` can do the magic.

rust/src/lib/ifaces/ethtool.rs
```diff
@@ -0,0 +1,160 @@
...
+    pub tx_generic_segmentation: Option<bool>,
+}
+
+impl EthtoolFeatureConfig {
+    pub fn new() -> Self {
+        Self::default()
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, Eq, PartialEq, Clone, Default)]
```
cathay4t: This is struct only bool and u32, you may enable `Copy` trait which could make the life easier as you don't need to worry about ownership of it.

cathay4t (approved):
Looks good to me!

automation/run-tests.sh
ffmancera: Done, thanks!

rust/src/lib/ifaces/ethtool.rs
ffmancera: Done, thank you!

PR #1839, opened 2022-03-31
rust: Disable at the same time ipv4 and remove a route with an static interface

qinqon:
When at the same state ipv4 is disabled and a route is removed, the routes are never removed and verify fails.

Use case:

current state:
```yaml
---                                                                             
interfaces:                                                                     
  - name: veth1                                                                 
    type: ethernet                                                              
    state: up                                                                   
    ipv4:                                                                       
      address:                                                                  
        - ip: 192.0.2.251                                                       
          prefix-length: 24                                                     
      dhcp: false                                                               
      enabled: true                                                             
                                                                                
routes:                                                                         
  config:                                                                       
      metric: 150                                                               
      next-hop-address: 192.0.2.1                                               
      next-hop-interface: veth1                                                 
      table-id: 254                       
```
desired state:
```yaml
---                                                                             
interfaces:                                                                     
  - name: veth1                                                                 
    ipv4:                                                                       
      enabled: false                                                            
routes:                                                                         
  config:                                                                       
    - next-hop-interface: veth1                                                 
      state: absent                       
```

Failure:
```
NmstateError: VerificationError: Desired absent route RouteEntry { state: Some(Absent), destination: None, next_hop_iface: Some("veth1"), next_hop_addr: None, metric: None, table_id: None } still found after apply: RouteEntry { state: None, destination: Some("192.0.2.1/32"), next_hop_iface: Some("veth1"), next_hop_addr: Some("0.0.0.0"), metric: Some(150), table_id: Some(254) }
```

### DIFF
--- a/rust/src/lib/nm/ip.rs
+++ b/rust/src/lib/nm/ip.rs
@@ -70,6 +70,10 @@ fn gen_nm_ipv4_setting(
             nm_setting.gateway = None;
         }
     }
+    if !iface_ip.enabled {
+        // Clean up static routes if ip is disabled
+        nm_setting.routes = Vec::new();
+    }
     if let Some(rules) = rules {
         nm_setting.route_rules = gen_nm_ip_rules(rules, false)?;
     }
--- a/tests/integration/route_test.py
+++ b/tests/integration/route_test.py
@@ -468,6 +468,34 @@ def test_disable_ipv4_with_routes_in_current(eth1_up):
 
 
 @pytest.mark.tier1
+def test_disable_ipv4_and_remove_wildcard_route(eth1_up):
+    libnmstate.apply(
+        {
+            Interface.KEY: [ETH1_INTERFACE_STATE],
+            Route.KEY: {Route.CONFIG: _get_ipv4_test_routes()},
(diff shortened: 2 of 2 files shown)


rust/src/lib/nm/ip.rs
```diff
@@ -62,7 +62,7 @@ fn gen_nm_ipv4_setting(
         // enabled.
         nm_setting.routes = Vec::new();
     }
-    if iface_ip.enabled && !iface_ip.dhcp {
+    if !iface_ip.enabled || !iface_ip.dhcp {
```
cathay4t: This is hard to read for mortal like me.

Please use this one and keep the old code.
```rust
if !iface_ip.enabled {
    nm_setting.routes = Vec::new();
}
```

cathay4t (changes requested):

rust/src/lib/nm/ip.rs
qinqon: It's even hard for me after the weekend :-) I will change it.

cathay4t (approved):

PR #1852, opened 2022-04-11
rust: Add support of OVS vlan filtering

ffmancera:
Renamed `LinuxBridgePortVlanConfig` to `BridgePortVlanConfig`, so
that it could be shared by both linux bridge and OVS bridge.

Integration test case enabled.

Signed-off-by: Gris Ge <fge@redhat.com>


ffmancera:
This PR is #1842 but I needed to create it again.

PR #1866, opened 2022-04-19
lib: Use base to serialize unknown interfaces

qinqon:
When there is a UnknownInterface the state is serialized as "- {}" since
the "other" field is not being set. This change convert "base" from serde skip
to flatten and remove the field other.

### DIFF
--- a/rust/src/lib/iface.rs
+++ b/rust/src/lib/iface.rs
@@ -162,10 +162,8 @@ impl From<&str> for InterfaceState {
 #[derive(Debug, Clone, PartialEq, Serialize, Default)]
 #[non_exhaustive]
 pub struct UnknownInterface {
-    #[serde(skip)]
-    pub base: BaseInterface,
     #[serde(flatten)]
-    other: serde_json::Value,
+    pub base: BaseInterface,
 }
 
 impl UnknownInterface {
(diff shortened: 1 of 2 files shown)


rust/src/lib/iface.rs
```diff
@@ -162,10 +162,8 @@ impl From<&str> for InterfaceState {
 #[derive(Debug, Clone, PartialEq, Serialize, Default)]
 #[non_exhaustive]
 pub struct UnknownInterface {
-    #[serde(skip)]
-    pub base: BaseInterface,
     #[serde(flatten)]
-    other: serde_json::Value,
```
cathay4t, line 168: Removing other will cause problem as the BaseInterface does not allows extra property.
Anyway,  It is fixed by in PR https://github.com/nmstate/nmstate/pull/1857

```rust
#[derive(Debug, Clone, PartialEq, Serialize, Default)]
#[non_exhaustive]
pub struct UnknownInterface {
    #[serde(skip_deserializing, flatten)]
    pub base: BaseInterface,
    #[serde(flatten)]
    other: serde_json::Value,
}
```

cathay4t (changes requested):

rust/src/lib/iface.rs
qinqon: Cool, closing this, thanks Gris!

PR #1878, opened 2022-04-22
cli: Show plugins version

qinqon:
Add a nmstatectl "plugins" cmd to show plugins information like version.

### DIFF
--- a/rust/src/lib/plugin.rs
+++ b/rust/src/lib/plugin.rs
@@ -1,0 +1,13 @@
+use crate::{nm::version, NmstateError};
+
+pub struct Plugin {
+    name: String,
+    version: String,
+}
+
+pub(crate) fn list_plugins() -> Result<Vec<Plugin>, NmstateError> {
+    Ok(vec![Plugin {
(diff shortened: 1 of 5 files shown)


rust/src/lib/plugin.rs
```diff
@@ -0,0 +1,16 @@
...
+    name: String,
+    version: String,
+}
+
+pub(crate) fn list_plugins() -> Result<Vec<Plugin>, NmstateError> {
+    Ok(vec![Plugin {
+        name: "NetworkManager".to_string(),
+        version: version::nm_version().unwrap(),
+    }])
+}
```
ffmancera, line 16: I am not sure if this makes sense.. nispor and ovsdb should be there too.. what do you think @cathay4t?

ffmancera (changes requested):

rust/src/lib/plugin.rs
cathay4t: nispor and ovsdb version is not required as rust is doing pull-in dependency, not runtime.

PR #1883, opened 2022-04-28
checkpoint: Extend rollback with timeout

qinqon:
The checkpoint rollback timeout has to be extended so it start to count
after apply is executed and some ovs operations are done, the extended
NM dbus operation is adding that "now" not to the checkpoint timeout.
This change set the extended time to timeout instead of choosing min
with will make checkpoint timeout be always the default 60s.


cathay4t (approved):

cathay4t:
Please fix the CI for rust lint.

PR #1887, opened 2022-04-28
rpm: Install nmpolicyctl

qinqon:


cathay4t:
@qinqon Why nmpolicyctl should shipped in this RPM? Any context?

qinqon:
> @qinqon Why nmpolicyctl should shipped in this RPM? Any context?

The idea is to have both nmstatectl and nmpolicyctl at RHCOS so we can do proper network setup before kubernetes-nmstate is ready, and zipping both at the same RPM facilitate that

qinqon:
The nmpolicy project is going to own the RPM

PR #1892, opened 2022-05-04
ip: Add NM may-fail option

qinqon:
At some environments [1] is needed to set the may-fail attribute from
NetworkManager [2], this change add that attribute to the ip settings.

Closes https://bugzilla.redhat.com/show_bug.cgi?id=2082042

[1] https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/files/configure-ovs-network.yaml#L304
[2] https://developer-old.gnome.org/NetworkManager/stable/settings-ipv4.html

### DIFF
--- a/rust/src/lib/ip.rs
+++ b/rust/src/lib/ip.rs
@@ -50,6 +50,13 @@ pub struct InterfaceIpv4 {
         deserialize_with = "crate::deserializer::option_u32_or_string"
     )]
     pub auto_table_id: Option<u32>,
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        rename = "may-fail",
+        default,
+        deserialize_with = "crate::deserializer::option_bool_or_string"
+    )]
+    pub may_fail: Option<bool>,
 }
 
 impl InterfaceIpv4 {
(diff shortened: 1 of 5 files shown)


cathay4t:
@qinqon Any more detail on why this only change `nm_dbus`? Changing there does not actually fix anything in nmstate.

qinqon:
> @qinqon Any more detail on why this only change `nm_dbus`? Changing there does not actually fix anything in nmstate.

I have to add them, this is not ready for review.

qinqon:
@cathay4t now it's ready (I think).

qinqon:
Somehow it's correctly send over dbus but it not changed at NM

cathay4t:
It is unclear why we need this feature. Please discuss the use case in https://bugzilla.redhat.com/show_bug.cgi?id=2082042#c2 first before spending more time on this.

I personally dislike introduce NM specific options to nmstate schema, it might make our life terrible if we migrate away from NetworkManager some day in the future.

qinqon:
> It is unclear why we need this feature. Please discuss the use case in https://bugzilla.redhat.com/show_bug.cgi?id=2082042#c2 first before spending more time on this.
> 
> I personally dislike introduce NM specific options to nmstate schema, it might make our life terrible if we migrate away from NetworkManager some day in the future.

Agree, answered in the bz, let's see if we can make the behaviour implicit and make sense for other projects.

PR #1897, opened 2022-05-16
[nmstate-2.0] ovs, dpdk: add support to rx_queue parameter

ffmancera:
```
---
interfaces:
  - name: ovs-br0
    type: ovs-bridge
    state: up
    bridge:
      options:
        datapath: netdev
      port:
        - name: ovs0
  - name: ovs0
    type: ovs-interface
    state: up
    dpdk:
      devargs: "000:18:00.2"
      rx_queue: 1000
```

Integration test case added.

### DIFF
--- a/automation/run-tests.sh
+++ b/automation/run-tests.sh
@@ -30,10 +30,6 @@ PYTEST_OPTIONS="--verbose --verbose \
         --log-date-format='%Y-%m-%d %H:%M:%S' \
         --log-format='%(asctime)s %(filename)s:%(lineno)d %(levelname)s %(message)s' \
         --durations=5 \
-        --cov /usr/lib/python*/site-packages/libnmstate \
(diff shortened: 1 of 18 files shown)


cathay4t (approved):
Code looks good to me. Waiting server to try on real hardware.

cathay4t (changes requested):
Please also bring the rust code in for this feature.

ffmancera:
> Please also bring the rust code in for this feature.

Hi, considering this is Nmstate-2.0 I think the support in rust is not needed. Users would be able to use 2.1.0 or greater. In order to backport this patch to rust, the OVS DPDK implementation and many other patches would be required.

What do you think?

cathay4t:
No need to touch rust code in nmstate-2.0 branch.

PR #1915, opened 2022-05-31
gc: Add flag to skip con UUID generation

qinqon:
At certain scenarios is important to generate the same configuration
from the same nmstate when running "gc" multiple times, also the UUID is
not mandatory since NetworkManager can generate one for you. This change
add the flag "--skip-uuid-gen" to be able to skip the UUID generation.

### DIFF
--- a/rust/src/cli/ncl.rs
+++ b/rust/src/cli/ncl.rs
@@ -154,6 +154,12 @@ fn main() {
             clap::Command::new(SUB_CMD_GEN_CONF)
                 .about("Generate network configuration for specified state")
                 .arg(
+                    clap::Arg::new("SKIP_UUID_GEN")
+                        .long("skip-uuid-gen")
+                        .takes_value(false)
+                        .help("Do not generate the UUID for the connection"),
+                )
+                .arg(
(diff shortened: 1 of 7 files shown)


cathay4t:
@qinqon This UUID stuff is NM backend specific. I think we should that through PR #1902 

I dislike empty UUID as our NM code is referring the controller using UUID in order to support OVS interface holding the same name as OVS bridge.

Via #1902, I would expect user to passing to have a consist/same configuration output of `gc` command.

```yaml
---
interfaces:
  - name: eth1
    type: ethernet
    backend-specific:
      networkmanager:
        connection.uuid: f5e185f9-137e-3e72-866f-246271bf9b84
    ipv4:
      enabled: true
      dhcp: true
```

cathay4t:
@qinqon Is above yaml meet your use case?

cathay4t:
PR https://github.com/nmstate/nmstate/pull/1916 should fix this use case without modification on schema or API.

qinqon:
Closing in favor of https://github.com/nmstate/nmstate/pull/1916

PR #1920, opened 2022-06-06
gc: Use nm connection id as key

qinqon:
At some scenarios is convenient to save nm keyfiles with the same name
as the one expected from nmstate, this change convert Vec<String> as
HashMap<String, String> for other backends it can be still used.

### DIFF
--- a/rust/src/lib/net_state.rs
+++ b/rust/src/lib/net_state.rs
@@ -377,7 +377,7 @@ impl NetworkState {
 
     pub fn gen_conf(
         &self,
-    ) -> Result<HashMap<String, Vec<String>>, NmstateError> {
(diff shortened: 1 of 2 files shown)


cathay4t:
~~Closing as this is fixed by https://github.com/nmstate/nmstate/pull/1916~~

cathay4t:
This is breaking API for no good reason.

You should never assume the `connection.id` is unique. And never assume network backend generate a unique key.

qinqon:
> This is breaking API for no good reason.
> 
> You should never assume the `connection.id` is unique. And never assume network backend generate a unique key.

The idea was to return to users the name of the .nmconnection that nmstatectl apply would have generated, so they generate exacly nm connections with same name.

PR #1934, opened 2022-06-08
cargo: Add gc, apply and show features

qinqon:
When consuming nmstate as a library is good to be able to compile only
the needed bits. This change split lib/net_state into features:
- show: show func
- apply: retrieve func
- gc: gen_conf func

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -56,6 +56,27 @@ jobs:
     - name: Unit test
       run: cd rust && cargo test -- --test-threads=1 --show-output
 
+  rust_features:
+    strategy:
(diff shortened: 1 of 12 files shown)


qinqon:
This open a pandora box of dependencies and warnings on code not used, we may need first a refactor to put gc and non gc code under different modules so it's easy to skip re-exporting at mod.rs.

I would add just the "gc" so we have the following:
- Common code don't need `cfg`
- Code needed only for gc  `cfg(feature = "gc")`
- Code needed only for apply, show and rollback `cfg(not(feature = "gc"))`

At the end we just one to be able to compile nmstate with just the gc code.

PR #1939, opened 2022-06-21
RPM package nmstate static lib

celebdor:
To enable applications to embed the nmstate clib, we want to build and
RPM package the static library so that in order to build those
applications, it suffices with just installing nmstate-static (which
also pulls the headers for a successful build).

Signed-off-by: Antoni Segura Puimedon <antoni@redhat.com>


cathay4t (approved):
LGTM

PR #1978, opened 2022-07-22
Introduce a logo of the nmstate project

phoracek:
These are the logos we use on kubernetes-nmstate for a while. Since nmstate is getting distributed in several ways now (standalone, OpenShift IPI, kubernetes nmstate, ...), I think it would be a good idea to represent all these efforts with a single logo. This would bring us one step closer to have nmstate shirts and stickers one day.

Huge kudos to Mary Shakshober who was kind enough to create this great logo for us.

Signed-off-by: Petr Horáček <phoracek@redhat.com>


cathay4t:
Thanks @phoracek !

cathay4t (approved):

PR #1990, opened 2022-08-01
[nmstate-1.3] libnmstate: expose the 'state_match()' for different states comparison

liangwen12year:
The network role needs such a api for determining if the network state
changes between the desired state and the current state.


cathay4t:
After discussed with Wen, we decided to take another approach on this.

PR #1999, opened 2022-08-03
[base] libnmstate: expose the 'state_match()' for different states comparison

liangwen12year:
The network role needs such a api for determining if the network state
changes between the desired state and the current state.


cathay4t (changes requested):
We are not doing this for python in 2.x branch. The correct approach should be in rust and python just wrapping it out.

cathay4t:
After discussed with Wen, we decided to take another approach on this.

PR #2052, opened 2022-10-03
gen_conf: add support to route rules

ffmancera:
Integration test case added.

### DIFF
--- a/rust/src/lib/nm/nm_dbus/gen_conf/route_rule.rs
+++ b/rust/src/lib/nm/nm_dbus/gen_conf/route_rule.rs
@@ -1,3 +1,46 @@
 // SPDX-License-Identifier: Apache-2.0
 
-// TODO
+use std::collections::HashMap;
+
+use super::super::NmIpRouteRule;
+
+const DEFAULT_ROUTE_TABLE: u32 = 254;
+
+impl NmIpRouteRule {
+    pub(crate) fn to_keyfile(&self) -> HashMap<String, String> {
+        let mut ret = HashMap::new();
+        if let Some(priority) = self.priority.as_ref() {
+            let mut keys = Vec::new();
+            let prio_str = format!("priority {}", priority);
+            keys.push(prio_str);
+
+            let to_str = match (self.to.as_ref(), self.to_len.as_ref()) {
+                (Some(t), Some(t_len)) => format!("to {}/{}", t, t_len),
+                (Some(t), None) => format!("to {}", t),
+                _ => "".to_string(),
+            };
+            if !to_str.is_empty() {
+                keys.push(to_str);
+            }
+
+            let from_str = match (self.from.as_ref(), self.from_len.as_ref()) {
+                (Some(f), Some(f_len)) => format!("from {}/{}", f, f_len),
+                (Some(f), None) => format!("from {}", f),
+                _ => "".to_string(),
+            };
+            if !from_str.is_empty() {
+                keys.push(from_str);
+            }
+
+            let mut table_str = format!("table {}", DEFAULT_ROUTE_TABLE);
+            if let Some(table) = self.table {
+                table_str = format!("table {}", table);
+            }
+            keys.push(table_str);
+
+            let rl_line = keys.join(" ");
+            ret.insert("".to_string(), rl_line);
+        }
+        ret
+    }
+}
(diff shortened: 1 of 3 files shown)


tests/integration/nm/gen_conf_test.py
```diff
@@ -82,6 +84,51 @@ def test_gen_conf_ovs_same_name(eth1_up, cleanup_ovs_same_name):
...
+    - ip-to: 192.0.2.0/24
+      ip-from: 192.168.2.0/24
+      priority: 1
+      route-table: 254
+"""
+    )
+    result = libnmstate.generate_configurations(desired_state)[
+        "NetworkManager"
+    ]
+    with nm_service_restart():
```
cathay4t: Please use `testlib.gen_conf.gen_conf_apply()`

rust/src/lib/nm/nm_dbus/gen_conf/route_rule.rs
```diff
@@ -1,3 +1,47 @@
...
+                _ => format!(
+                    "from {}/{}",
+                    if self.family.unwrap_or_default() == AF_INET {
+                        DEFAULT_IPV4
+                    } else {
+                        DEFAULT_IPV6
+                    },
+                    DEFAULT_IP_PREFIX
+                ),
+            };
```
cathay4t: When `ip_from` not defined, I got this in keyfile:

```
[ipv4]
address1=192.168.1.1/24
method=manual
route1=192.168.2.0/24,192.168.1.3,108
route1_options=table=200
routing-rule1=priority 30000 to 192.168.3.2 table 200

[ipv6]
addr-gen-mode=eui64
address1=2001:db8:1::1/64
method=manual
route1=2001:db8:a::/64,2001:db8:1::2,108
route1_options=table=200
routing-rule1=priority 30000 to 2001:db8:b::/64 table 200
```

cathay4t (changes requested):

rust/src/lib/nm/nm_dbus/gen_conf/route_rule.rs
ffmancera: I noticed that the default `ip_from` is only set when the `ip_to` is not set. Let me fix that

tests/integration/nm/gen_conf_test.py
ffmancera: Done, thank you!

rust/src/lib/nm/nm_dbus/gen_conf/route_rule.rs
ffmancera: Done, now the default `ip_from` is only set when `ip_to` is also not set. Thanks!

rust/src/lib/nm/nm_dbus/gen_conf/route_rule.rs
```diff
@@ -1,3 +1,54 @@
...
+                    }
+                }
+            };
+            let mut table_str = format!("table {}", DEFAULT_ROUTE_TABLE);
+            if let Some(table) = self.table {
+                table_str = format!("table {}", table);
+            }
+
+            let rl_line =
+                format!("{} {} {} {}", prio_str, from_str, to_str, table_str);
```
cathay4t: This will generate two spaces when `from_str` or `to_str` is empty.
Not sure it works or not. But better to be consistent with NetworkManager generated file.

You may try:

```
let mut keys = Vec::new();
keys.push(prio_str);
keys.push(from_str);

let rl_line = keys.join(" ").to_string();
```

rust/src/lib/nm/nm_dbus/gen_conf/route_rule.rs
```diff
@@ -1,3 +1,54 @@
...
+                    // we need to check if ip_to is empty or not
+                    if to_str.is_empty() {
+                        format!(
+                            "from {}/{}",
+                            if self.family.unwrap_or_default() == AF_INET {
+                                DEFAULT_IPV4
+                            } else {
+                                DEFAULT_IPV6
+                            },
+                            DEFAULT_IP_PREFIX
```
cathay4t: Just FYI. Nmstate will raise error if both `ip-to` and `ip-from` is empty.

cathay4t (changes requested):

rust/src/lib/nm/nm_dbus/gen_conf/route_rule.rs
ffmancera: In that case, we don't need to handle that case. Let me drop it. Thanks!

cathay4t (approved):

PR #2072, opened 2022-10-21
Add support to fwmark and fwmask in routing policy

ffmancera:
This is related to https://bugzilla.redhat.com/show_bug.cgi?id=2136820

### DIFF
--- a/rust/src/lib/route_rule.rs
+++ b/rust/src/lib/route_rule.rs
@@ -224,6 +224,20 @@ pub struct RouteRuleEntry {
         deserialize_with = "crate::deserializer::option_u32_or_string"
     )]
     pub table_id: Option<u32>,
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        default,
+        deserialize_with = "crate::deserializer::option_u32_or_string",
+        serialize_with = "crate::serializer::option_u32_as_hex"
+    )]
+    pub fwmark: Option<u32>,
+    #[serde(
+        skip_serializing_if = "Option::is_none",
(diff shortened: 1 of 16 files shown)


ffmancera:
I need to fix example integration test, trivial, will look at it later.

rust/src/lib/route_rule.rs
```diff
@@ -224,6 +224,20 @@ pub struct RouteRuleEntry {
...
+        default,
+        deserialize_with = "crate::deserializer::option_u32_or_string",
+        serialize_with = "crate::serializer::option_u32_as_hex"
+    )]
+    pub fwmark: Option<u32>,
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        default,
+        deserialize_with = "crate::deserializer::option_u32_or_string",
+        serialize_with = "crate::serializer::option_u32_as_hex"
```
cathay4t, line 238: The CI failure https://github.com/nmstate/nmstate/actions/runs/3298609230/jobs/5441042094#step:6:851 indicate this `serialize_with` is not working as expected.

cathay4t (changes requested):
Please fix the CI failure.

rust/src/lib/route_rule.rs
ffmancera: The `serialize_with` is working properly. In this case it is not being used because it was using `state` property directly. The issues was that the ip-from on the example file was not formatted properly. Fixed it, thanks!

cathay4t (approved):

PR #2086, opened 2022-11-07
Makefile: add `-Z namespaced-features` to fix compiler error

ffmancera:
When compiling nmstatectl in Copr infraestructure, it is failing with an error related to the `Cargo.toml`. This option has been already stabilized but Copr is using an older version of rustc.

```
namespaced features with the `dep:` prefix are only allowed on the
nightly channel and requires the `-Z namespaced-features` flag on the
command-line
```


ffmancera:
Ugh, this won't work. -Z is only allowed for nightly. This means Copr people will need to update the images they are using for their chroots. I will send an email about this.

PR #2087, opened 2022-11-07
route_rule: support `from all to all` routing policy

### LINKED ISSUE #1417: route rule: Support `from all to all look <table_id>`
Currently, nmstate only support route rule with destination or from defined. Please remove that restriction by supporting
`from all to all` route rule.

This could open a door where we could support `oif` or `iif` in route rule in the future.

ffmancera:
Add support to `from all to all` routing policy. In static config
    generator it will add `from 0.0.0.0/0` or `from ::/0` to the keyfile.
    
    A new parameter `family` have been introduced to the route-rule config
    field. This parameter is used to specify the address family. If `family`
    is not specified it will fail on validation, if not specified from
    Nispor it will assume IPv4 as it was done in the past.
    
    Example:
    
    ```yaml
    route-rules:
      config:
        - route-table: 254
          priority: 100
          family: ipv4
    ```
    
    Integration test cases added.
    
    Fixes #1417
    
Fixes https://github.com/nmstate/nmstate/issues/1417

### DIFF
--- a/rust/src/lib/ip.rs
+++ b/rust/src/lib/ip.rs
@@ -964,7 +964,7 @@ fn default_allow_extra_address() -> bool {
     true
 }
 
-#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone)]
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone, Copy)]
 #[serde(rename_all = "lowercase")]
 #[non_exhaustive]
 pub enum AddressFamily {
--- a/rust/src/lib/ip.rs
+++ b/rust/src/lib/ip.rs
@@ -7,6 +7,8 @@ use crate::{
     BaseInterface, DnsClientState, ErrorKind, MptcpAddressFlag, NmstateError,
 };
 
+const AF_INET: u8 = 2;
+const AF_INET6: u8 = 10;
 const IPV4_ADDR_LEN: usize = 32;
 const IPV6_ADDR_LEN: usize = 128;
 
@@ -960,4 +962,29 @@ fn is_none_or_empty_mptcp_flags(v: &Option<Vec<MptcpAddressFlag>>) -> bool {
 // Allow extra IP by default
 fn default_allow_extra_address() -> bool {
     true
+}
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone)]
+#[serde(rename_all = "lowercase")]
+#[non_exhaustive]
+pub enum AddressFamily {
+    IPv4,
+    IPv6,
+    Unknown,
+}
+
+impl From<u8> for AddressFamily {
+    fn from(d: u8) -> Self {
+        match d {
+            AF_INET => AddressFamily::IPv4,
+            AF_INET6 => AddressFamily::IPv6,
+            _ => AddressFamily::Unknown,
+        }
+    }
+}
+
+impl Default for AddressFamily {
+    fn default() -> Self {
+        Self::IPv4
+    }
 }
--- a/rust/src/lib/nispor/route_rule.rs
+++ b/rust/src/lib/nispor/route_rule.rs
@@ -1,14 +1,51 @@
-use crate::{RouteRuleEntry, RouteRules};
+use log::warn;
 
-pub(crate) fn get_route_rules(np_rules: &[nispor::RouteRule]) -> RouteRules {
+use crate::{AddressFamily, RouteRuleEntry, RouteRules};
+
+// Due to a bug in NetworkManager all route rules added using NetworkManager are
+// using RTM_PROTOCOL UnSpec. Therefore, we need to support it until it is
+// fixed.
+const SUPPORTED_STATIC_ROUTE_PROTOCOL: [nispor::RouteProtocol; 3] = [
+    nispor::RouteProtocol::Boot,
+    nispor::RouteProtocol::Static,
+    nispor::RouteProtocol::UnSpec,
+];
+
+const SUPPORTED_ROUTE_PROTOCOL: [nispor::RouteProtocol; 8] = [
+    nispor::RouteProtocol::Boot,
+    nispor::RouteProtocol::Static,
+    nispor::RouteProtocol::Ra,
+    nispor::RouteProtocol::Dhcp,
+    nispor::RouteProtocol::Mrouted,
+    nispor::RouteProtocol::KeepAlived,
+    nispor::RouteProtocol::Babel,
+    nispor::RouteProtocol::UnSpec,
+];
+
+pub(crate) fn get_route_rules(
+    np_rules: &[nispor::RouteRule],
+    running_config_only: bool,
+) -> RouteRules {
     let mut ret = RouteRules::new();
 
     let mut rules = Vec::new();
+    let protocols = if running_config_only {
+        SUPPORTED_STATIC_ROUTE_PROTOCOL.as_slice()
+    } else {
+        SUPPORTED_ROUTE_PROTOCOL.as_slice()
+    };
+
     for np_rule in np_rules {
         let mut rule = RouteRuleEntry::new();
         // We only support route rules with 'table' action
         if np_rule.action != nispor::RuleAction::Table {
             continue;
+        }
+        // Filter out the routes with protocols that we do not support
+        if let Some(rule_protocol) = np_rule.protocol.as_ref() {
+            if !protocols.contains(rule_protocol) {
+                continue;
+            }
         }
         rule.ip_to = np_rule.dst.clone();
         rule.ip_from = np_rule.src.clone();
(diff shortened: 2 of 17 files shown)


rust/src/lib/ip.rs
```diff
@@ -961,3 +963,27 @@ fn is_none_or_empty_mptcp_flags(v: &Option<Vec<MptcpAddressFlag>>) -> bool {
 fn default_allow_extra_address() -> bool {
     true
 }
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone)]
+#[serde(rename_all = "lowercase")]
```
cathay4t, line 968: Please add `#[non_exhaustive]`

rust/src/lib/ip.rs
```diff
@@ -961,3 +963,27 @@ fn is_none_or_empty_mptcp_flags(v: &Option<Vec<MptcpAddressFlag>>) -> bool {
...
+            AF_INET => AddressFamily::IPv4,
+            AF_INET6 => AddressFamily::IPv6,
+            _ => AddressFamily::Unknown,
+        }
+    }
+}
+
+impl Default for AddressFamily {
+    fn default() -> Self {
+        Self::Unknown
```
cathay4t: As you mentioned in commit message, the default should be `IPv4`

rust/src/lib/nm/nm_dbus/gen_conf/route_rule.rs
```diff
@@ -23,10 +25,26 @@ impl NmIpRouteRule {
...
-                (Some(f), Some(f_len)) => format!("from {f}/{f_len}"),
-                (Some(f), None) => format!("from {f}"),
-                _ => "".to_string(),
+            let mut from_str =
+                match (self.from.as_ref(), self.from_len.as_ref()) {
+                    (Some(f), Some(f_len)) => format!("from {f}/{f_len}"),
+                    (Some(f), None) => format!("from {f}"),
+                    _ => "".to_string(),
+                };
+            from_str = match (self.from.as_ref(), self.to.as_ref()) {
```
cathay4t: Please use `if self.from.as_ref().is_none() && self.to.as_ref().is_none()` instead of match when you only handle single condition.

rust/src/lib/nm/nm_dbus/gen_conf/route_rule.rs
```diff
@@ -23,10 +25,26 @@ impl NmIpRouteRule {
...
+                (None, None) => {
+                    if let Some(family) = self.family {
+                        if family == AF_INET {
+                            "from 0.0.0.0/0".to_string()
+                        } else {
+                            "from ::/0".to_string()
+                        }
+                    } else {
```
cathay4t: How about do this in `Deserialize` of `RouteRuleEntry` or `pre_edit_cleanup()`? This is not NM specific.

The top level code just tidy everything out, NM level just store them without any check or modification(unless NM specific).

cathay4t (changes requested):

ffmancera:
I need to fix a small bug! Working on it

rust/src/lib/ip.rs
ffmancera: Done, thank you!

rust/src/lib/ip.rs
ffmancera: Done, thanks!

rust/src/lib/nm/nm_dbus/gen_conf/route_rule.rs
ffmancera: Done, thank you!

rust/src/lib/nm/nm_dbus/gen_conf/route_rule.rs
ffmancera: This is NM keyfile specific. In kernel this is not needed at all but nmcli adds it to the keyfile this is why we are doing this only for keyfile. We do not want to add the default ip always because it is not necessary and I would prefer to follow iproute2 format which is a standard somehow, that is:

```yaml
route-rules:
  config:
    - route-table: 254
      priority: 100
      family: ipv4
```

We do not want `ip-from: 0.0.0.0/0` there. I will ask NM people while are they adding the default ip address to keyfile if it is not needed at all.

ffmancera:
The integration test failing is not related to the changes.

cathay4t:
With this desire state:

```
route-rules:
  config:
  - family: ipv4
    route-table: 200
```

I got route rule set to IPv6 also:

```
[fge@c9s rule]$ ip -6 rule
0:	from all lookup local
30000:	from all lookup 200
32766:	from all lookup main
```

Please fix the code and also add test case to make sure it does not happen again.

ffmancera:
> With this desire state:
> 
> ```
> route-rules:
>   config:
>   - family: ipv4
>     route-table: 200
> ```
> 
> I got route rule set to IPv6 also:
> 
> ```
> [fge@c9s rule]$ ip -6 rule
> 0:	from all lookup local
> 30000:	from all lookup 200
> 32766:	from all lookup main
> ```
> 
> Please fix the code and also add test case to make sure it does not happen again.

Fixes thanks, I added to tests to cover this.

ffmancera:
Also this is related to: https://bugzilla.redhat.com/show_bug.cgi?id=2141064

rust/src/lib/nispor/route_rule.rs
```diff
@@ -1,30 +1,66 @@
-use crate::{RouteRuleEntry, RouteRules};
+use log::warn;
 
-pub(crate) fn get_route_rules(np_rules: &[nispor::RouteRule]) -> RouteRules {
+use crate::{AddressFamily, RouteRuleEntry, RouteRules};
+
+const SUPPORTED_STATIC_ROUTE_PROTOCOL: [nispor::RouteProtocol; 3] = [
+    nispor::RouteProtocol::Boot,
+    nispor::RouteProtocol::Static,
+    nispor::RouteProtocol::UnSpec,
```
cathay4t, line 11: Any reason why we have additional `UnSpec` route? 
And please  remove duplicate `SUPPORTED_ROUTE_PROTOCOL` and `SUPPORTED_STATIC_ROUTE_PROTOCOL` against `nispor::route`.

cathay4t (changes requested):
* At the top `RouteRules::pre_edit_clean()`, we always set `RouteRuleEntry.family` to correct value and raise error if family is incorrect according to `ip_from` or `ip_to`. And also convert `any` to `0.0.0.0/0` or `::/0`.
* In the `gen_nm_ip_setting()`, we can only pass the rules for certain family. like:
```rust
    gen_nm_ipv6_setting(
        base_iface.ipv6.as_ref(),
        routes,
        rules.filter(|r| r.family == Some(AddressFamily::IPv6)),
        nm_conn,
    )?;
```

* Instruct nispor plugin to always include address family.

cathay4t:
Meanwhile, we have problem when multiple route rule are using the same priority(default to 30000), I am fixing it.

rust/src/lib/nispor/route_rule.rs
ffmancera: > * At the top `RouteRules::pre_edit_clean()`, we always set `RouteRuleEntry.family` to correct value and raise error if family is incorrect according to `ip_from` or `ip_to`. And also convert `any` to `0.0.0.0/0` or `::/0`.

I do not agree with this. Nmstate should not convert `any` to `0.0.0.0/0` or `::/0`. It is redundant and could confuse users as the standard way to set routing policy `from all to all` is not setting the ip-from/ip-to fields. If we auto-complete the missing fields with `0.0.0.0/0` or `::/0` it will also be shown in the current state. IMO, this shouldn't be like that. Why do you think Nmstate should add `0.0.0.0/0`?

> 
>     * In the `gen_nm_ip_setting()`, we can only pass the rules for certain family. like:
> 
> 
> ```rust
>     gen_nm_ipv6_setting(
>         base_iface.ipv6.as_ref(),
>         routes,
>         rules.filter(|r| r.family == Some(AddressFamily::IPv6)),
>         nm_conn,
>     )?;
> ```

I agree here, let me work on this. Thanks!

> 
>     * Instruct nispor plugin to always include address family.

This is already done, please check `rust/src/lib/nispor/route_rule.rs:53`. Thanks!

ffmancera: > Any reason why we have additional `UnSpec` route? And please remove duplicate `SUPPORTED_ROUTE_PROTOCOL` and `SUPPORTED_STATIC_ROUTE_PROTOCOL` against `nispor::route`.

This due to a bug in NetworkManager that I plan to fix. All route rules added using NetworkManager they have UnSpec protocol set. It shouldn't be like that. I already discussed this with Thomas and will send a patch but meanwhile we should allow UnSpec protocol rules.

Therefore, `SUPPORTED_ROUTE_PROTOCOL` and `SUPPORTED_STATIC_ROUTE_PROTOCOL` are not duplicated from the ones at `route.rs`

cathay4t (approved):

PR #2113, opened 2022-11-23
rust/lib: Replace the unsafe 'libc::sysconf' to 'nix::unistd::sysconf'

liangwen12year:
The `libc::sysconf` is unsafe function of rust, use its safe replacement `nix::unistd::sysconf` with proper error handling instead.


ffmancera:
The erros are not related to your patches, they are related to Loopback support in NM. We need to adapt Nmstate code.

ffmancera (approved):

PR #2115, opened 2022-11-23
ovs: add support for port trunks

bengal:
This commit adds support for the "trunk-tags" property in ovs ports.
    
Example:
   
```yml
  - name: ovs0
    type: ovs-interface
    state: up
  - name: ovs-br0
    type: ovs-bridge
    state: up
    bridge:
      port:
        - name: ovs0
          vlan:
            mode: trunk
            trunk-tags:
              - id-range:
                  min: 20
                  max: 22
              - id-range:
                  min: 30
                  max: 32
              - id: 40
```

The property is implemented in NM via "ovs-port.trunks" and corresponds to the "trunks" property of ports in ovs:

```
    Bridge ovs-br0
        Port ovs0
            trunks: [20, 21, 22, 30, 31, 32, 40]
            Interface ovs0
                type: internal
```
The NM part is at:

https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/1467


ffmancera:
https://github.com/nmstate/nmstate/pull/2117 this needs to be merged before :-)

tests/integration/ovs_test.py
```diff
@@ -387,6 +387,36 @@ def test_ovs_vlan_access_tag():
     assertlib.assert_absent(PORT1)
 
 
+@pytest.mark.tier1
+def test_ovs_vlan_trunks():
```
ffmancera: This test won't be available in all NM version. We need to skip it when NM version is 1.41 (we do not care about the micro version as stable won't use 1.41 at all) or greater.

```
@pytest.mark.skipif(nm_major_minor_version() < 1.41, reason="OVS VLAN trunks was not supported in NM")
```

Add this after the `@pytest.mark.tier1` decorator so CI will skip it when using NM stable :-)

ffmancera (changes requested):
Thank you! It looks quite good, I made a comment about CI, other than that LGTM

cathay4t (changes requested):
Please include test case proving the trunk vlan tags can be purged.

bengal:
> Please include test case proving the trunk vlan tags can be purged.

Sure, what do you mean exactly? That other trunk vlans should already exist before applying the state (and then, check they are removed), or that I should apply first a state with trunks and then one without? 

Any example of that in the existing code?

ffmancera:
> > Please include test case proving the trunk vlan tags can be purged.
> 
> Sure, what do you mean exactly? That other trunk vlans should already exist before applying the state (and then, check they are removed), or that I should apply first a state with trunks and then one without?
> 

Usually we test that the user can configure the trunk vlans and then drop them. I think it is fine if you apply a state with the trunk vlans and then one with them empty. Of course the test must check they have been dropped.

> Any example of that in the existing code?

Check `test/integration/bond_test.py` line `119` should contain `test_add_and_remove_bond_with_two_port` test. Of course, this is testing something else but the structure is similar. Thank you!

bengal:
@ffmancera: thanks, I changed the test to also check for trunks removal. How does it look?

tests/integration/ovs_test.py
```diff
@@ -387,6 +387,60 @@ def test_ovs_vlan_access_tag():
...
+    bridge_port = bridge_config_subtree[OVSBridge.PORT_SUBTREE][0]
+    bridge_port.update(
+        {
+            OVSBridge.Port.VLAN_SUBTREE: {
+                OVSBridge.Port.Vlan.MODE: OVSBridge.Port.Vlan.Mode.TRUNK
+            }
+        }
+    )
+    libnmstate.apply(desired_state)
+    assertlib.assert_state_match(desired_state)
```
ffmancera: I think this was not going to work. The trunks must be set as empty because Nmstate supports partial editing.

ffmancera: In addition the test should query the current state see `show_only()` helper and assert explicitly that the trunks are empty.

ffmancera (dismissed):
It looks good and in the right direction but there are some nits. In addition, could you split it in two tests? One focused on configuring trunk vlans and then another test focusing on removing them.

ffmancera:
In addition, please modify `test_add_remove_ovs_bridge_vlan` to use `assert_state_match` instead of `assert_state`, the difference is that `assert_state` will do a plain `==` but `assert_state_match` passes if the desired state is a subset of the current one.

As now trunks is supported for ovs bridges, the current state in that example will contain `trunks: []` which is correct but that property is not present in the desired state.. so it should use `assert_state_match`.

Thank you!

bengal:
I have split the test and added explicit checks for the the number of trunks.

tests/integration/ovs_test.py
```diff
@@ -387,6 +387,102 @@ def test_ovs_vlan_access_tag():
...
+    with bridge.create() as state:
+        assertlib.assert_state_match(state)
+
+    assertlib.assert_absent(BRIDGE1)
+    assertlib.assert_absent(PORT1)
+
+
+@pytest.mark.tier1
+@pytest.mark.skipif(
+    nm_major_minor_version() < 1.41,
```
cathay4t: The `1.5` is bigger than `1.41`. Please use `nm_minor_version() < 41`

tests/integration/ovs_test.py
```diff
@@ -387,6 +387,102 @@ def test_ovs_vlan_access_tag():
...
+                OVSBridge.Port.Vlan.TRUNK_TAGS: [
+                    {OVSBridge.Port.Vlan.TrunkTags.ID: 100},
+                    {OVSBridge.Port.Vlan.TrunkTags.ID: 300},
+                ],
+            },
+        },
+    )
+
+    desired_state = bridge.state
+    libnmstate.apply(desired_state)
```
cathay4t: Before this line, they should be stored as `pytest.fixture`, so that any error happening in the test case will trigger clean up in this fixture. Your current code will have leftover if any failure happens.

cathay4t (changes requested):

tests/integration/ovs_test.py
bengal: Done

tests/integration/ovs_test.py
bengal: Done

cathay4t (approved):
LGTM. Thanks!

PR #2130, opened 2022-11-29
route: Validate the route destination address

liangwen12year:
'0.0.0.0/8' and its subnet should never be used as the destination address for the route, thus raise the invalid argument error during the route validation.

### DIFF
--- a/rust/src/lib/ifaces/sriov.rs
+++ b/rust/src/lib/ifaces/sriov.rs
@@ -99,11 +99,11 @@ impl SrIovConfig {
             vfs.sort_unstable_by(|a, b| a.id.cmp(&b.id));
 
             if !vfs.is_empty() {
-                let total_vfs = self.total_vfs.unwrap_or(
+                let total_vfs = self.total_vfs.unwrap_or_else(|| {
                     current.and_then(|c| c.total_vfs).unwrap_or(
                         vfs.iter().map(|v| v.id).max().unwrap_or_default() + 1,
-                    ),
-                );
+                    )
+                });
                 self.total_vfs = Some(total_vfs);
                 // Auto fill the missing
                 if total_vfs as usize != vfs.len() {
--- a/rust/src/lib/route.rs
+++ b/rust/src/lib/route.rs
@@ -330,6 +330,7 @@ impl MergedRoutes {
         current: Routes,
         merged_ifaces: &MergedInterfaces,
     ) -> Result<Self, NmstateError> {
+        desired.validate()?;
         let mut desired_routes = Vec::new();
         if let Some(rts) = desired.config.as_ref() {
             for rt in rts {
--- a/rust/src/lib/route.rs
+++ b/rust/src/lib/route.rs
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::collections::{hash_map::Entry, HashMap, HashSet};
+use std::net::Ipv4Addr;
+use std::str::FromStr;
 
 use serde::{Deserialize, Serialize};
 
@@ -73,6 +75,9 @@ impl Routes {
                             is not supported: {route:?}"
                         ),
                     ));
+                }
+                if let Some(dst) = route.destination.as_deref() {
+                    validate_route_dst(dst)?;
                 }
             }
         }
@@ -509,4 +514,52 @@ impl MergedRoutes {
     pub(crate) fn is_changed(&self) -> bool {
         !self.route_changed_ifaces.is_empty()
     }
+}
+
+// Validating if the route destination network is valid,
+// 0.0.0.0/8 and its subnet cannot be used as the route destination network
+fn validate_route_dst(dst: &str) -> Result<(), NmstateError> {
+    if !is_ipv6_addr(dst) {
+        let ip_net: Vec<&str> = dst.split('/').collect();
+        let ip_addr = Ipv4Addr::from_str(ip_net[0])?;
+        if ip_addr.octets()[0] == 0 {
+            if dst.contains('/') {
+                let prefix = match ip_net[1].parse::<i32>() {
+                    Ok(p) => p,
+                    Err(_) => {
+                        return Err(NmstateError::new(
+                            ErrorKind::InvalidArgument,
+                            format!(
+                                "The prefix of the route destination network \
+                                '{dst}' is invalid"
+                            ),
+                        ));
+                    }
+                };
+                if prefix >= 8 {
+                    let e = NmstateError::new(
+                        ErrorKind::InvalidArgument,
+                        "0.0.0.0/8 and its subnet cannot be used as \
+                        the route destination, please use the default \
+                        gateway 0.0.0.0/0 instead"
+                            .to_string(),
+                    );
+                    log::error!("{}", e);
+                    return Err(e);
+                }
+            } else {
+                let e = NmstateError::new(
+                    ErrorKind::InvalidArgument,
+                    "0.0.0.0/8 and its subnet cannot be used as \
+                    the route destination, please use the default \
+                    gateway 0.0.0.0/0 instead"
+                        .to_string(),
+                );
+                log::error!("{}", e);
+                return Err(e);
+            }
+        }
+        return Ok(());
+    }
+    Ok(())
 }
--- a/rust/src/lib/unit_tests/route.rs
+++ b/rust/src/lib/unit_tests/route.rs
@@ -326,3 +326,85 @@ fn test_route_ipv4_ecmp_is_match() {
     .unwrap();
     assert!(absent_route.is_match(&route));
 }
+
+#[test]
+fn test_route_valid_default_gateway() {
+    let routes: Routes = serde_yaml::from_str(
+        r#"
+config:
+- destination: 0.0.0.0/0
+  next-hop-address: 192.0.2.1
+  next-hop-interface: eth1
+"#,
+    )
+    .unwrap();
+    routes.validate().unwrap();
+}
+
+#[test]
+fn test_route_invalid_destination() {
+    let routes1: Routes = serde_yaml::from_str(
+        r#"
+config:
+- destination: 0.0.0.0/8
+  next-hop-address: 192.0.2.1
+  next-hop-interface: eth1
+"#,
+    )
+    .unwrap();
+    let result = routes1.validate();
+    assert!(result.is_err());
+    assert_eq!(result.err().unwrap().kind(), ErrorKind::InvalidArgument);
+
+    let routes2: Routes = serde_yaml::from_str(
+        r#"
+config:
+- destination: 0.0.0.0/f
+  next-hop-address: 192.0.2.1
+  next-hop-interface: eth1
+"#,
+    )
+    .unwrap();
+    let result = routes2.validate();
+    assert!(result.is_err());
+    assert_eq!(result.err().unwrap().kind(), ErrorKind::InvalidArgument);
+
+    let routes3: Routes = serde_yaml::from_str(
+        r#"
+config:
+- destination: 0.0.0.0
+  next-hop-address: 192.0.2.1
+  next-hop-interface: eth1
+"#,
+    )
+    .unwrap();
+    let result = routes3.validate();
+    assert!(result.is_err());
+    assert_eq!(result.err().unwrap().kind(), ErrorKind::InvalidArgument);
+
+    let routes4: Routes = serde_yaml::from_str(
+        r#"
+config:
+- destination: 0.0.0.0.0/0
+  next-hop-address: 192.0.2.1
+  next-hop-interface: eth1
+"#,
+    )
+    .unwrap();
+    let result = routes4.validate();
+    assert!(result.is_err());
+    assert_eq!(result.err().unwrap().kind(), ErrorKind::InvalidArgument);
+
+    let routes5: Routes = serde_yaml::from_str(
+        r#"
+config:
+- destination: 0.0.0.0.0/7
+  next-hop-address: 192.0.2.1
+  next-hop-interface: eth1
+"#,
+    )
+    .unwrap();
+    let result = routes5.validate();
+    assert!(result.is_err());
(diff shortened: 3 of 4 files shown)


rust/src/lib/route.rs
```diff
@@ -78,6 +78,22 @@ impl Routes {
                     error!("{}", e);
                     return Err(e);
                 }
+                if !route.destination.as_deref().unwrap_or_default().is_empty() {
```
cathay4t: Please use `if let Some(dst) = route.destination.as_ref()` to save the extra check on `String::new().is_empty()`.

rust/src/lib/route.rs
```diff
@@ -78,6 +78,22 @@ impl Routes {
...
+                    let destination_address = route.destination.as_deref().unwrap_or_default();
+                    if destination_address.ip.to_str() == "0.0.0.0"
+                        && destination_address.prefix_length >= 8 {
+                        let e = NmstateError::new(
+                            ErrorKind::InvalidArgument,
+                            format!(
+                                "0.0.0.0/8 and its subnet cannot be used as \
+                            the route destination: {:?}",
```
cathay4t: Normally we do not show debug output to user as error message.

Meanwhile, provide a hint to user that they might want to use default gateway `0.0.0.0/0`.

cathay4t (changes requested):
Besides change request, please fix CI.

cathay4t:
And test case is also required.

rust/src/lib/route.rs
liangwen12year: I updated. Thanks.

liangwen12year:
> And test case is also required.

Test case is added.

rust/src/lib/route.rs
```diff
@@ -77,6 +78,24 @@ impl Routes {
                     error!("{}", e);
                     return Err(e);
                 }
+                if let Some(dst) = route.destination.as_ref() {
+                    if !is_ipv6_addr(dst.as_str()) {
+                        let net: Ipv4Net = dst.parse()?;
```
ffmancera: There is a problem here. This works only for addresses with a prefix. That means it will parse `192.0.2.0/24` correctly but it will fail on `192.0.2.0`. Checking ipnet docs you will find `Ipv4Net provides a [FromStr](https://doc.rust-lang.org/std/str/trait.FromStr.html) implementation for parsing network addresses represented in CIDR notation`

It needs to be fixed, thank you!

ffmancera (changes requested):

rust/src/lib/route.rs
```diff
@@ -77,6 +78,33 @@ impl Routes {
...
+                            let e = NmstateError::new(
+                                ErrorKind::InvalidArgument,
+                                "0.0.0.0/8 and its subnet cannot be used as \
+                            the route destination, please use the default \
+                            gateway 0.0.0.0/0 instead"
+                                    .to_string(),
+                            );
+                            error!("{}", e);
+                            return Err(e);
+                        }
```
ffmancera, line 105: It seems a little bit complex IMO. I think you can skip the validation if the address is not in CIDR notation. In order to do that you can combine it with the `is_ipv6_addr` if statement. That means it will look something like that `if !is_ipv6_addr() && dst_str.contains('/')`. I think that is the right approach instead of having an extra variable to handle if the dst is invalid or not.

ffmancera (changes requested):

ffmancera (approved):

rust/src/lib/route.rs
```diff
@@ -77,6 +78,33 @@ impl Routes {
...
+                                && net.prefix_len() >= 8
+                            {
+                                is_invalid_dst = true;
+                            }
+                        }
+                        if is_invalid_dst {
+                            let e = NmstateError::new(
+                                ErrorKind::InvalidArgument,
+                                "0.0.0.0/8 and its subnet cannot be used as \
+                            the route destination, please use the default \
```
cathay4t: Please fix the indent.

cathay4t (changes requested):
Using `ipnet::Ipv4Net` is overkill. Simple `split()` and compare first entry for `0.*` and second entry for `>=8`.

When we introduce something from crate, we are simplify the code, not make it worse.

cathay4t:
More detail:
 * `let ip_net: Vec<&str> = addr.split('/').collect()`
 * Use `Ipv4Addr::from_str(ip_net[0])` and `u8::parse(ip_net[1]).unwrap_or_default()`.
 * Compare `IPv4Addr.octets()[0] == 0` and `prefix >= 8`.

With this, we do not depend on ipnet which I intend to remove its depency from nmstate.

rust/src/lib/route.rs
liangwen12year: fixed.

rust/src/lib/route.rs
```diff
@@ -77,6 +79,34 @@ impl Routes {
...
+                    if !is_ipv6_addr(dst_str) {
+                        let mut is_invalid_dst = false;
+                        if dst_str == "0.0.0.0" {
+                            is_invalid_dst = true;
+                        } else if dst_str.contains('/') {
+                            let ip_net: Vec<&str> =
+                                dst_str.split('/').collect();
+                            let ip_addr = Ipv4Addr::from_str(ip_net[0]);
+                            let prefix = ip_net[1].parse::<i32>().unwrap_or(32);
+                            if ip_addr?.octets()[0] == 0 && prefix >= 8 {
```
cathay4t: Normally, we do not place `?` in the middle of statement. Please try `let ip_addr = Ipv4Addr::from_str(ip_net[0])?;"

cathay4t (changes requested):
Please place these check code into a dedicate function like `is_dst_valid() -> bool` with comment to this function explaining what you are validating.

In general, using `mut is_invalid_dst` as flag is a bad practice, considering this flag been modified or overrided after 100+ lines, developer could easily forgot the exact original purpose.

rust/src/lib/route.rs
```diff
@@ -77,6 +79,34 @@ impl Routes {
...
+                    let dst_str = dst.as_str();
+                    if !is_ipv6_addr(dst_str) {
+                        let mut is_invalid_dst = false;
+                        if dst_str == "0.0.0.0" {
+                            is_invalid_dst = true;
+                        } else if dst_str.contains('/') {
+                            let ip_net: Vec<&str> =
+                                dst_str.split('/').collect();
+                            let ip_addr = Ipv4Addr::from_str(ip_net[0]);
+                            let prefix = ip_net[1].parse::<i32>().unwrap_or(32);
```
cathay4t: The `unwrap_or(32)` is wrong. What if I use `0.0.0.0/a`.

cathay4t: BTW. The `ip_net[1]` will panic if the dst is `1.1.1.1`.

cathay4t: Using `[]` is dangerous unless you really sure the index exists. I think `ip_net[0]` might cause problem on empty string `dst`, as a library, we should __never__ panic!

liangwen12year: > BTW. The `ip_net[1]` will panic if the dst is `1.1.1.1`.

I think that the dst will not be `1.1.1.1`, because we the code will be called under the if condition `else if dst_str.contains('/') `.

liangwen12year: > Using `[]` is dangerous unless you really sure the index exists. I think `ip_net[0]` might cause problem on empty string `dst`, as a library, we should **never** panic!

I am pretty sure that the index exists base on the if condition `else if dst_str.contains('/')`. The dst can not be empty string under the if condition.

rust/src/lib/route.rs
liangwen12year: I updated. thanks.

rust/src/lib/route.rs
```diff
@@ -402,3 +420,22 @@ fn flat_absent_route(
     }
     ret
 }
+
+// Validating if the route destination address is valid,
+// 0.0.0.0/8 and its subnet cannot be used as the route destination
+fn is_route_dst_valid(dst_str: &str) -> Result<bool, AddrParseError> {
+    if dst_str.contains('/') {
+        let ip_net: Vec<&str> = dst_str.split('/').collect();
+        let ip_addr = Ipv4Addr::from_str(ip_net[0])?;
```
cathay4t, line 524: This will panic if `dst_str` is empty string.

cathay4t (changes requested):

rust/src/lib/route.rs
```diff
@@ -402,3 +420,22 @@ fn flat_absent_route(
...
+        let prefix = ip_net[1].parse::<i32>().unwrap_or_default();
```
cathay4t: This means `0.0.0.0/f` is equal to `0.0.0.0/0` with your parser.

rust/src/lib/unit_tests/route.rs
```diff
@@ -436,3 +436,33 @@ next-hop-address: "2001:db8:a:0000:000::1"
...
+- destination: 0.0.0.0/8
+  next-hop-address: 192.0.2.1
+  next-hop-interface: eth1
+"#,
+    )
+    .unwrap();
+    let result = routes.validate();
+    assert!(result.is_err());
+    assert_eq!(result.err().unwrap().kind(), ErrorKind::InvalidArgument);
+}
```
cathay4t: Please include more test on string like:

 * `0.0.0.0/f`
 * `0.0.0.0.0/0`
 * `0.0.0.0.0/7`
 * `0.0.0.0`

cathay4t (changes requested):

rust/src/lib/ifaces/sriov.rs
```diff
@@ -79,11 +79,11 @@ impl SrIovConfig {
             vfs.sort_unstable_by(|a, b| a.id.cmp(&b.id));
 
             if !vfs.is_empty() {
-                let total_vfs = self.total_vfs.unwrap_or(
+                let total_vfs = self.total_vfs.unwrap_or_else(|| {
```
cathay4t, line 102: This should in another patch

rust/src/lib/route.rs
```diff
@@ -77,6 +79,22 @@ impl Routes {
                     error!("{}", e);
                     return Err(e);
                 }
+                if let Some(dst) = route.destination.as_ref() {
+                    let dst_str = dst.as_str();
+                    if !is_ipv6_addr(dst_str)
+                        && !is_route_dst_valid(dst_str).unwrap_or_default()
```
cathay4t: You have two layer of raising errors. And this `!foo().unwrap_or_default()` is hurting human eyes.

Try change to something like:


```rust
if let Some(dst) = route.destination.as_deref() {
    validate_route_dst(dst)?;
}


fn validate_route_dst(dst: &str) -> Result<(), NmstateError> {
    todo!()
}
```

cathay4t (changes requested):

rust/src/lib/route.rs
```diff
@@ -468,3 +473,72 @@ impl MergedRoutes {
...
+                    separated by three periods"
+                        .to_string(),
+                );
+                log::error!("{}", e);
+                return Err(e);
+            } else if dst.contains('/') {
+                let ip_net: Vec<&str> = dst.split('/').collect();
+                let ip_addr = Ipv4Addr::from_str(ip_net[0])?;
+                let prefix = ip_net[1].parse::<i32>().unwrap_or_default();
+                if prefix == 0 && ip_net[1] != "0" {
```
cathay4t: Why not just:

```
let prefix = match ip_net[1].parse::<i32>() {
    Ok(p) => p,
    Err(e) => ...
}
```

rust/src/lib/route.rs
```diff
@@ -468,3 +473,72 @@ impl MergedRoutes {
...
+                    let e = NmstateError::new(
+                        ErrorKind::InvalidArgument,
+                        "The prefix of the route destination network is \
```
cathay4t: When you raise errors, it is better to include the original invalid value, so user know which line to fix.

Something like `format!("The prefix of the route destination '{dst}' is invalid")`.

rust/src/lib/route.rs
```diff
@@ -468,3 +473,72 @@ impl MergedRoutes {
...
     }
 }
+
+// Validating if the route destination network is valid,
+// 0.0.0.0/8 and its subnet cannot be used as the route destination network
+fn validate_route_dst(dst: &str) -> Result<(), NmstateError> {
+    if !is_ipv6_addr(dst) {
+        if dst.contains('.') {
+            let dot_count = dst.matches('.').count();
+            if dot_count != 3 {
```
cathay4t: This is not required. The `Ipv4Addr::from_str(ip_net[0])?;` can do the magic.

cathay4t (changes requested):

rust/src/lib/route.rs
```diff
@@ -473,3 +478,64 @@ impl MergedRoutes {
...
+                }
+            }
+
+            return Ok(());
+        }
+        let e = NmstateError::new(
+            ErrorKind::InvalidArgument,
+            format!(
+                "The route destination network '{dst}' is invalid and not \
+            in dotted decimal format"
```
cathay4t: The `Ipv4Addr::from_str()` handle this. You do not need to handle it.

rust/src/lib/route.rs
```diff
@@ -473,3 +478,64 @@ impl MergedRoutes {
         !self.route_changed_ifaces.is_empty()
     }
 }
+
+// Validating if the route destination network is valid,
+// 0.0.0.0/8 and its subnet cannot be used as the route destination network
+fn validate_route_dst(dst: &str) -> Result<(), NmstateError> {
+    if !is_ipv6_addr(dst) {
+        if dst.contains('.') {
+            if dst.contains('/') {
```
cathay4t, line 526: You can do split regardless it contains `/` or not, you could save some lines of codes.

cathay4t (changes requested):

ffmancera (approved):

PR #2137, opened 2022-11-30
[nmstate-1.3] tests, ovs: add updelay and downdelay ovs-bond options test

ffmancera:
Add missing integration tests for OVS Link Aggregation support. We support downdelay and updelay so they need to be tested properly.

(cherry picked from commit 7ddd51853dfb7c34d2d173a068d39063d6bcb0c3)


cathay4t:
@ffmancera This patch does not exist in nmstate 1.4. Meanwhile, can we only mark OVS port as changed when it has changed OVS bond config?

Of course, if there is no bug report on this, feel free to close this.

PR #2253, opened 2023-02-23
Support dropping local route rule with priority 0

ffmancera:


ffmancera:
@cathay4t integration test is missing here, I will add it tomorrow.

ffmancera:
This PR depends on https://github.com/nmstate/nmstate/pull/2262

ffmancera:
@cathay4t This is ready for review. Please notice that I am not showing all kernel route rules anymore. Nmstate will only report local route rule with priority 0. This is needed to make MergedState work properly when modifying this rule.

Let me know if you have any questions. Thx!

ffmancera:
There are tier 2 tests that are failing, will look at it tomorrow. I know the root cause.

rust/src/lib/nm/nm_dbus/connection/ip.rs
```diff
@@ -272,6 +298,21 @@ impl ToDbusValue for NmSettingIp {
     }
 }
 
+impl NmSettingIp {
+    fn i32_to_replace_local_rule(val: Option<i32>) -> Option<bool> {
```
liangwen12year, line 321: Why not make the return value as `Result<Option<bool> , Nmerror>` ?

rust/src/lib/nm/settings/ip.rs
```diff
@@ -27,6 +27,19 @@ fn gen_nm_ipv4_setting(
                 let mut nm_setting = NmSettingIp::default();
                 nm_setting.method = Some(NmSettingIpMethod::Disabled);
                 nm_conn.ipv4 = Some(nm_setting);
+            } else if let Some(setting) = nm_conn.connection.as_ref() {
+                // if the network state only contains local route rule absent,
+                // we need to specify the ip setting for
+                // loopback with replace_local_rule set a false
```
liangwen12year: The code below sets `replace_local_rule` to true. Is the comment wrong?

ffmancera: Nice catch, the comment is wrong. It should say true. Thx!

rust/src/lib/nm/nm_dbus/connection/ip.rs
ffmancera: I don't think so. Here we are reading the NmSettingIp, in NM this property is Ternanry which means the valid values are -1, 0, 1. It is very unlikely that we are going to get something different and if that happens we should not fail to the user.

The user will get confused which such strange error, instead we just raise a warning so if this happens and we need to debug we can notice it quickly.

ffmancera:
@cathay4t We need the default local rule in MergedNetworkState as well, not only for verification but also in order to configure the NmSettingIp in loopback interface.

The way we have to set replace-local-rule properly in NmSettingIp for loopback is by looking at the rules that loopback is holding. If we filter out the default local rule when retrieving the information for building the MergedNetworkState, we do not have a method to actually check if the user marked the default local rule as absent or it is being omitted by nispor.

If that is a problem, let me work more on this as I would need to rework how we detect if the rule is present or not to set replace-local-rule setting properly.

Thanks!

rust/src/lib/nm/settings/ip.rs
```diff
@@ -27,6 +27,19 @@ fn gen_nm_ipv4_setting(
...
+                if let Some(iface_name) = setting.iface_name.clone() {
+                    if iface_name.eq(&"lo") {
+                        let mut nm_setting = NmSettingIp::default();
+                        nm_setting.method = Some(NmSettingIpMethod::Manual);
+                        nm_setting.addresses = vec!["127.0.0.1/8".to_string()];
+                        nm_setting.replace_local_rule = Some(true);
+                        nm_conn.ipv4 = Some(nm_setting);
+                    }
+                }
```
cathay4t: I think `append_route_rule()` and `append_route_rule()` will copy current ip stack to desire state and mark loopback as changed. Why we need extra code here.

rust/src/lib/nispor/route_rule.rs
```diff
@@ -61,6 +63,13 @@ pub(crate) fn get_route_rules(
             if !protocols.contains(rule_protocol) {
                 continue;
             }
+            // We only support modifying local rule from kernel, the others
+            // should be ignored.
+            if *rule_protocol == nispor::RouteProtocol::Kernel
+                && np_rule.table != Some(255)
+            {
+                continue;
+            }
```
cathay4t, line 72: I still think we should not showing `RouteProtocol::Kernel` rules even it is for 255(local) route table.
We have `retrieve_full()` for internal usage of `MergeNetworkState` where you can only use those for verification like we did for `controller` property.

rust/src/lib/nm/settings/ip.rs
```diff
@@ -86,6 +99,19 @@ fn gen_nm_ipv4_setting(
     if let Some(rules) = iface_ip.rules.as_ref() {
         nm_setting.route_rules = gen_nm_ip_rules(rules, false)?;
     }
+    if let Some(setting) = nm_conn.connection.as_ref() {
+        if let Some(iface_name) = setting.iface_name.clone() {
+            if iface_name.eq(&"lo") {
+                if let Some(rules) = iface_ip.rules.as_ref() {
+                    nm_setting.replace_local_rule =
+                        Some(!rules_contains_local_rule(rules, false));
```
cathay4t, line 115: This is wrong. Changes to local route rule should not set `replace_local_rule`. We should only set `nm_setting.replace_local_rule` when user desire state has

```
priority: 0
state: absent
```

Other cases, we are expecting user to append route rule and do not touch current value of `nm_setting.replace_local_rule`.

cathay4t (changes requested):

rust/src/lib/nispor/route_rule.rs
ffmancera: Done, now we are filtering out the kernel rules except for retrieve_full().

rust/src/lib/nm/settings/ip.rs
ffmancera: There can be a situation where the user only specified the following:

```
route-rules:
  config:
  - priority: 0
    route-table: 255
    state: absent
```

Here `iface_ip` will be `None` and thefore it would skip the generation of ipv4_setting which is wrong. This why the extra code is needed.

rust/src/lib/nm/settings/ip.rs
ffmancera: Oh that is different to what I was thinking. My idea was that:
1. When the user explictly define the default local rule as absent Nmstate will se `replace_local_rule` to True.
2. If the user then want to restore the default local rule they must explicitly set it without absent and therefore it will set `replace_local_rule` to False
3. If the user does not set the default local rule explictly Nmstate will check if it is present in the MergedState, if so, `replace_local_rule` will be set to False if not to True.

The decision about the last step is because if the user apply and state that requires the usage of `loopback` profile but doesn't mention default local rule it will modify the replace_local_rule setting.. which in my opinion is incorrect.

ffmancera: By doing this, we will be requiring the user to explicitly set the default local rule as absent all the time to avoid possible situations where the `replace_local_rule` setting is set back to False. I do not think that is a great idea.

rust/src/lib/nm/settings/ip.rs
cathay4t: Then you should fix `append_route_rule()` and `apply_absent_rule()` instead of doing similar work in other place.

rust/src/lib/nm/settings/ip.rs
cathay4t: You are mixing the `local route` and `priority 0`. This PR is focusing on `priority 0`.

* To delete priority 0 route rule, user can do `priority: 0` and `state: absent`. No matter user add another local route or not, we set `NM.replace_local_rule`  to true.

* To add it back, user can just do `prioirity: 0` with whatever detail they want, if the desire happen to be matching with `priority 0, route: 255`, we use kernel autogenerate one by set `NM.replace_local_rule` to false, if not, we still do `replace_local_rule: true` and manual rule.

ffmancera: No I do not think so. `local route rule` and `priority 0` is not a problem. We already support them as expected.

There is an special `rule` created by kernel which is placed on the local route table (255) with priority 0. This specific rule requires special treatment. IMHO we should drop it only when the user explicitly specify it e.g

```
priority: 0
route-table: 255
state: absent
family: ipv4
```

also for family ipv6. Please, notice that we shouldn't do the usual matching we do here. If an user specify only `priority: 0` or `route-table: 255` as absent we shouldn't drop this special rule. In most of the cases without this rule the user will lose their connectivity.

This is what was mentioned in BZ https://bugzilla.redhat.com/show_bug.cgi?id=2166630

rust/src/lib/nm/settings/ip.rs
```diff
@@ -299,3 +337,18 @@ fn apply_nmstate_wait_ip(
...
+
+fn rules_contains_local_rule(rules: &[RouteRuleEntry], is_ipv6: bool) -> bool {
+    for rule in rules {
+        if let Some(family) = rule.family {
+            if is_ipv6 != matches!(family, AddressFamily::IPv6) {
+                continue;
+            }
+        }
+        if rule.table_id == Some(255) && rule.priority == Some(0) {
+            return true;
```
cathay4t: This indicate we set `NM.replace_local_rule = false` when user state

```
route-rules:
  config:
    - family: ipv4
      route-table: 255
      priority: 0
      from: 192.168.1.1
```

ffmancera: That is a nice catch. Thanks, I will fix that

cathay4t (changes requested):
Two changes please:

 * Please use YAML in https://bugzilla.redhat.com/show_bug.cgi?id=2166630 for integration test case.
 *  This YAML should have `replace-local-rule` set to `false`:
```yml
---
route-rules:
  config:
    - route-table: 255
      state: absent
```

cathay4t:
With your latest patch [1d9e833](https://github.com/nmstate/nmstate/pull/2253/commits/1d9e833cd4d62ed7d25b5742959011247db1fa71)


I noticed even ipv6 route rule not mentioned in YAML:

```yml
---
route-rules:
  config:
    - route-table: 255
      family: ipv4
      state: absent
    - route-table: 255
      family: ipv4
      priority: 0
```

 After applied, the ip -6 rule indicate ipv6 rule for local is removed also.

cathay4t:
Meanwhile, the test cases which merely remove rule for route table 255 is not actual use case.

rust/src/lib/nispor/route_rule.rs
```diff
@@ -61,6 +63,13 @@ pub(crate) fn get_route_rules(
             if !protocols.contains(rule_protocol) {
                 continue;
             }
+            // We only support modifying local rule from kernel, the others
+            // should be ignored.
```
liangwen12year, line 67: Just to be sure, should we silently ignore other rules from kernel or raise the exception ?

ffmancera: This is the nispor code, we cannot raise an exception if we get a rule with RouteProtocol::Kernel. If we do that, we will always fail as we cannot control rules that are set/handled by kernel.

As we did in the past, we silently ignore them.

ffmancera:
> With your latest patch [1d9e833](https://github.com/nmstate/nmstate/pull/2253/commits/1d9e833cd4d62ed7d25b5742959011247db1fa71)
> 
> I noticed even ipv6 route rule not mentioned in YAML:
> 
> ```yaml
> ---
> route-rules:
>   config:
>     - route-table: 255
>       family: ipv4
>       state: absent
>     - route-table: 255
>       family: ipv4
>       priority: 0
> ```
> 
> After applied, the ip -6 rule indicate ipv6 rule for local is removed also.

Hi Gris, I have been trying to reproduce this and won't be able. I applied this state several times using [1d9e833](https://github.com/nmstate/nmstate/pull/2253/commits/1d9e833cd4d62ed7d25b5742959011247db1fa71)

Could you provide more steps or retry it? I did it on a loop..

rust/src/lib/nm/version.rs
```diff
@@ -37,3 +37,35 @@ pub(crate) fn nm_supports_accept_all_mac_addresses_mode(
...
+    let version = {
+        let nm_api = crate::nm::nm_dbus::NmApi::new()
+            .map_err(crate::nm::error::nm_error_to_nmstate)?;
+        nm_api
+            .version()
+            .map_err(crate::nm::error::nm_error_to_nmstate)?
+    };
+    let version_split = version.split('.');
+    let supported_version = Vec::<u32>::from([1, 42]);
+    let mut supported_elem = supported_version.iter();
```
cgwalters, line 52: Doesn't seem like the heap allocated `Vec` here is necessary, this is just:
```
let mut supported_elem = [1u32, 42].into_iter();
```
or so.

rust/src/lib/route_rule.rs
```diff
@@ -216,6 +218,18 @@ impl RouteRuleEntry {
     }
 
     pub(crate) fn is_match(&self, other: &Self) -> bool {
+        // For local route rule with priority 0 we force the user to specify it
+        // as absent explicitly. Otherwise, it won't match.
+        if (self.table_id.is_none()
+            || self.priority.is_none()
+            || self.family.is_none())
+            && (other.table_id == Some(255)
```
liangwen12year, line 226: `other.table_id == Some(Self::LOCAL_ROUTE_TABLE_ID)`  ?

tests/integration/route_test.py
```diff
@@ -1047,6 +1048,87 @@ def test_route_rule_add_and_remove_using_loopback():
...
+    )
+    try:
+        libnmstate.apply(desired_state)
+        _check_ip_rules(desired_state[RouteRule.KEY][RouteRule.CONFIG])
+    finally:
+        desired_state[RouteRule.KEY][RouteRule.CONFIG][0].pop(RouteRule.STATE)
+        desired_state[RouteRule.KEY][RouteRule.CONFIG][1][
+            RouteRule.STATE
+        ] = RouteRule.STATE_ABSENT
+        libnmstate.apply(desired_state)
```
liangwen12year, line 1129: Should we also add another `_check_ip_rules()` in the end to check the rules?

cathay4t (changes requested):
Still have ipv6 route rule impacted even not desired. Reproduce step:

```yml
---
route-rules:
  config:
    - route-table: 255
      priority: 0
      family: ipv4
      state: absent
    - route-table: 255
      priority: 32765
      family: ipv4
```

```yml
---
route-rules:
  config:
    - route-table: 255
      family: ipv4
      state: absent
    - route-table: 255
      family: ipv4
      priority: 0
```

After these two YMLs applied, I got 
```
[fge@c9s local]$ ip -6 rule
32766:	from all lookup main
```

PR #2257, opened 2023-02-25
vxlan: Add VXLAN properties for EVPN use

varesa:
In order to be able to use VXLAN interfaces as part of an EVPN fabric (separate control plane for VTEP and MAC learning), some expansions are done to the configurable properties:

- Make `base-if` non-mandatory

  EVPN networks are often run on top of a L3 underlay, which means that there are several L3 links to the host where ECMP routing is desired. The VXLAN interface should not be tied to any physical interface (and it's not required since multicast is not used) and the egress interface is determinted through standard routing.

- Allow specifying `local` IP

  Since the VXLAN interface is not tied to any physical interface, it is often desired to source the tunnel from a loopback IP.

- Disable data plane MAC `learning`

  In EVPN MAC learning is handled through a separate control plane like MP-BGP (e.g. FRR as daemon) writing entries to the bridge FDB.

### DIFF
--- a/rust/src/lib/ifaces/vxlan.rs
+++ b/rust/src/lib/ifaces/vxlan.rs
@@ -48,7 +48,13 @@ impl VxlanInterface {
     }
 
     pub(crate) fn parent(&self) -> Option<&str> {
-        self.vxlan.as_ref().map(|cfg| cfg.base_iface.as_str())
+        self.vxlan.as_ref().and_then(|cfg| {
+            if cfg.base_iface.is_empty() {
+                None
+            } else {
+                Some(cfg.base_iface.as_str())
+            }
+        })
     }
 }
 
@@ -56,9 +62,14 @@ impl VxlanInterface {
 #[serde(rename_all = "kebab-case", deny_unknown_fields)]
 #[non_exhaustive]
 pub struct VxlanConfig {
+    #[serde(default, skip_serializing_if = "String::is_empty")]
     pub base_iface: String,
     #[serde(deserialize_with = "crate::deserializer::u32_or_string")]
     pub id: u32,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub learning: Option<bool>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub local: Option<std::net::IpAddr>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub remote: Option<std::net::IpAddr>,
     #[serde(
--- a/rust/src/lib/nispor/vxlan.rs
+++ b/rust/src/lib/nispor/vxlan.rs
@@ -1,3 +1,5 @@
+// SPDX-License-Identifier: Apache-2.0
+
 use std::str::FromStr;
 
 use crate::{BaseInterface, VxlanConfig, VxlanInterface};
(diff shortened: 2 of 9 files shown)


rust/src/lib/ifaces/vxlan.rs
```diff
@@ -48,18 +48,26 @@ impl VxlanInterface {
     }
 
     pub(crate) fn parent(&self) -> Option<&str> {
-        self.vxlan.as_ref().map(|cfg| cfg.base_iface.as_str())
+        self.vxlan
+            .as_ref()
+            .and_then(|cfg| cfg.base_iface.as_ref())
+            .map(|base_iface| base_iface.as_str())
```
cathay4t: How about `as_deref()`?

rust/src/lib/ifaces/vxlan.rs
```diff
@@ -48,18 +48,26 @@ impl VxlanInterface {
...
     }
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Default)]
 #[serde(rename_all = "kebab-case", deny_unknown_fields)]
 #[non_exhaustive]
 pub struct VxlanConfig {
-    pub base_iface: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub base_iface: Option<String>,
```
cathay4t: This is breaking rust API which we have no plan on releasing nmstate 3.x.

Please use `serde(default)` to store the undefined `base_iface` as emptry.

rust/src/lib/ifaces/vxlan.rs
```diff
@@ -48,18 +48,26 @@ impl VxlanInterface {
...
     #[serde(deserialize_with = "crate::deserializer::u32_or_string")]
     pub id: u32,
     #[serde(skip_serializing_if = "Option::is_none")]
+    pub learning: Option<bool>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub local: Option<std::net::IpAddr>,
+    #[serde(skip_serializing_if = "Option::is_none")]
```
cathay4t, line 73: Please add `///` document here.

cathay4t (changes requested):
Breaking rust API is not allowed yet. Please use `serde(default)`.

cathay4t:
@varesa Can you amend the patch base on above feedback?

varesa:
> @varesa Can you amend the patch base on above feedback?

That's my intention, but I've been rather busy the past week (and will be for a while more). Sorry about that

rust/src/lib/ifaces/vxlan.rs
varesa: Can you clarify what sort of documentation you want? 

Explaining what these options do in the context of VXLANs in general? Or more about the technical implementation in context of `nmstate`?

There's not much documentation for the existing fields either to align myself with

cathay4t: It is for crate user who look at the document like https://docs.rs/nmstate/latest/nmstate/struct.MacVlanConfig.html#structfield.accept_all_mac .

Yes, I forgot to add document for VXLAN, you may skip this.

cathay4t:
Let me see whether I can allocate sometime to amend your patches to get it merged.

cathay4t:
I have amend the patch to preserve API stability. Patch manual tested. Merging.

cathay4t (approved):

PR #2262, opened 2023-02-27
nm, rules: fix fetching loopback information from NM plugin

ffmancera:


rust/src/lib/route_rule.rs
```diff
@@ -258,9 +258,17 @@ impl RouteRuleEntry {
...
+            && other.family.is_some()
+            && self.family != other.family
+        {
+            return false;
+        }
+
         if self.priority.is_some()
             && self.priority != Some(RouteRuleEntry::USE_DEFAULT_PRIORITY)
             && self.priority != other.priority
+            && !(self.priority == Some(0) && other.priority.is_none())
```
liangwen12year, line 271: Maybe this should be considered as another commit ?  Or the subject line should be adjusted to include this line of change.

ffmancera: I will update the commit title, thanks!

tests/integration/route_test.py
```diff
@@ -1020,6 +1020,31 @@ def test_route_rule_from_all_to_all_ipv6(route_rule_test_env):
         assert _check_ip_rules(rules)
 
 
+@pytest.mark.tier1
+@pytest.mark.skipif(
+    nm_minor_version() < 42, reason="Loopback is only support on NM 1.42+"
+)
+def test_route_rule_add_and_remove_using_loopback():
+    desired_state = {
```
liangwen12year, line 1028: I noticed that you did not specify any device in the `desired_state`, does that mean `RouteRule.IIF` is the loopback device by default ?

cathay4t: It is allowed to have empty `RouteRule.iif`. The loopback interface is the final fallback on choosing interface to store route rule. You make check `store_route_rule_config()` for detail.

cathay4t (approved):
LGTM.

PR #2276, opened 2023-03-16
show: Don't hide controller property

qinqon:
For use cases where we want to filter by interfaces not owned by a bridge showing the controller is quite helpful.

### DIFF
--- a/rust/src/lib/query_apply/net_state.rs
+++ b/rust/src/lib/query_apply/net_state.rs
@@ -38,12 +38,6 @@ impl NetworkState {
     /// Retrieve the `NetworkState`.
     /// Only available for feature `query_apply`.
     pub fn retrieve(&mut self) -> Result<&mut Self, NmstateError> {
-        self.retrieve_full()?;
-        self.interfaces.hide_controller_prop();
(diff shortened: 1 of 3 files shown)


rust/src/lib/query_apply/net_state.rs
```diff
@@ -39,7 +39,6 @@ impl NetworkState {
     /// Only available for feature `query_apply`.
     pub fn retrieve(&mut self) -> Result<&mut Self, NmstateError> {
         self.retrieve_full()?;
```
cathay4t: This function only exists because of `hide_controller_prop()`, if we are remove it, we should remove `retrieve_full()` also.

cathay4t (changes requested):

rust/src/lib/query_apply/net_state.rs
ffmancera: Well, I was planning to use it from for the default local route rule with priority 0. I can recreate if you drop it but it is fine if you leave it IMO.

ffmancera (approved):

cathay4t (approved):

PR #2316, opened 2023-04-19
cli: Delete `persist-nic-names` stamp file after cleanup

jlebon:
This makes the `--cleanup` operation idempotent since it alreayd knows to no-op if the stamp file is missing. But also, the MCO can use the presence of the stamp file to know to even call `--cleanup`. Finally, this "resets" the state so that we could in the future do this whole dance again for another RHEL transition.

The MCO logs will remember whether whether this logic was invoked, and the link files are prefixed by `98-nmstate`, so there's no mistaking that it was nmstate that wrote them.


cathay4t (approved):
LGTM

PR #2324, opened 2023-04-20
cli: make `.nmstate-persist.stamp` contain JSON

jlebon:
This JSON object maps MAC addresses to interface names for the interfaces that were pinned. This will be read by the MCO to easily know the list of interface names involved and do some additional steps necessary in some cases.


jlebon:
I ended up just parsing the link files since it cleaned up other things too so I think we can drop this.

PR #2333, opened 2023-04-25
README: Remove IRC channel

tyll:
Most questions on IRC are ignored, therefore it does not make sense to mention it in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,8 +68,6 @@ interfaces:
 
 *Nmstate* [GitHub Issues pages][github_issue_url] for discussion.
 
-You may find us in `#nmstate` on [Libera IRC](https://libera.chat/) also.
-
 ## Contributing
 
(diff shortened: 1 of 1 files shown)


ffmancera:
The questions are not ignored. There can be situations when no one is present when the question is asked.

I do not see the value of removing this from the README.

tyll:
This year, I see 7 topics raised by contributors, only two were responded to. There is no development discussion otherwise, for example between you, Gris and Wen so pointing people in the README to a mostly non-functional IRC channel is not very helpful. Removing the IRC channel (and decommissioning it) will avoid that people join there unnecessarily.

ffmancera:
@cathay4t what is your opinion?

cathay4t:
I prefer removing IRC from README.

ffmancera (approved):

PR #2338, opened 2023-05-08
go, api: Create a structure API

qinqon:
At some scenarios is useful to have a "structured" version of the
network_state, this change creates a golang api tested toward examples
and stated used at integration tests it also add the kubebuilder tags to
be k8s CRD friendly. The golang code is generated with a rust tool that traverse the AST of specific rust source files.
- Depends on: 
  -  https://github.com/nmstate/nmstate/pull/2479
  - https://github.com/nmstate/nmstate/pull/2513
  - https://github.com/nmstate/nmstate/pull/2521
- PoC integrating with k-nmstate
  - https://github.com/nmstate/kubernetes-nmstate/pull/1203
- [x] Add validation annotations https://book.kubebuilder.io/reference/markers/crd-validation.html
  - [x] Some enums contains an "other" variant to pass unknown values, we should not validate those types
  - [x] IntOrString remove enum validations, we should use pattern validation.
- [x] Add linter golangci-lint
- [x] Add some CI job to check that `make generate` has being call.
- [x] Add descriptions from rust into golang comments
  - [x] This breaks yaml inside comments https://github.com/kubernetes-sigs/controller-tools/pull/517
  - [x] https://github.com/kubernetes-sigs/controller-tools/pull/870
- [x] Resolve `stp` problem, field with same name at same level with different types, for ovs is directly at boolean.
- [x] False possitive for dhcp-client-id


tests/integration/conftest.py
```diff
@@ -219,3 +227,36 @@ def restore_old_state(old_state):
...
     if len(desire_state["interfaces"]):
         libnmstate.apply(desire_state, verify_change=False)
+
+
+def _setup_state_dump():
+    shutil.rmtree(STATES_DUMP_DIR, ignore_errors=True)
+    libnmstate.apply = _libnmstate_dump_and_apply
+
+
+def _libnmstate_dump_and_apply(
```
qinqon: Add an option to run the checker directly instead of dumping files 
```python
def pytest_collection_modifyitems(config, items):
    if config.getoption("--go-checker"):
        libnmstate.apply = _custom_apply_with_go_checker
```

qinqon:
issue 
```
error unmarshaling JSON: while decoding JSON: json: cannot unmarshal object into Go struct field LldpConfig.interfaces.lldp.neighbors of type 
```

rust/src/go/crd/boilerplate.go.txt
```diff
@@ -0,0 +1,16 @@
+/*
+Copyright The NMState Authors.
```
tyll: IANAL, this probably needs to have a year and real names as shown at https://www.apache.org/licenses/LICENSE-2.0

cathay4t: maintaining the year number is time consuming. Can we just add a `// SPDX-License-Identifier: Apache-2.0` and create a dedicate file AUTHORS to contain names of contributors

tyll: I am not a lawyer. I can only refer you to https://www.apache.org/licenses/LICENSE-2.0#apply which provides the instructions.

qinqon:
/hold

We are going to move this to `https://github.com/nmstate/kubernetes-api`

PR #2344, opened 2023-05-15
bridge: add support to default-vlan-pvid

ffmancera:
Add support to default-vlan-pvid, currently the nm_dbus part was already done, so it didn't require much work.

Add integration test.

### DIFF
--- a/rust/src/lib/ifaces/linux_bridge.rs
+++ b/rust/src/lib/ifaces/linux_bridge.rs
@@ -105,6 +105,14 @@ impl LinuxBridgeInterface {
         is_desired: bool,
     ) -> Result<(), NmstateError> {
         if let Some(opts) =
+            self.bridge.as_ref().and_then(|b| b.options.as_ref())
+        {
+            if is_desired {
+                opts.validate_vlan_default_pvid(self)?;
+            }
+        }
+
+        if let Some(opts) =
             self.bridge.as_mut().and_then(|b| b.options.as_mut())
         {
             opts.sanitize_group_fwd_mask(&self.base)?;
@@ -526,6 +534,8 @@ pub struct LinuxBridgeOptions {
     pub stp: Option<LinuxBridgeStpOptions>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub vlan_protocol: Option<VlanProtocol>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub vlan_default_pvid: Option<u16>,
 }
 
 impl LinuxBridgeOptions {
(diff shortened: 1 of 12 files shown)


cathay4t (dismissed):
LGTM

cathay4t (changes requested):
Changes required:
 1. the commit message title is wrong, should be `vlan-default-pvid`.
 2. When desire state contains `vlan-default-pvid` without VLAN filtering enabled, we should return ValueError instead of VerificationError.
 3. Got verification error when using `vlan-default-pvid` with VLAN trunk:

```yml
---
interfaces:
  - name: br0
    type: linux-bridge
    state: up
    bridge:
      options:
        vlan-default-pvid: 1000
      ports:
        - name: eth1
          vlan:
            mode: trunk
            trunk-tags:
            - id: 101
            - id-range:
                min: 500
                max: 599
            tag: 100
            enable-native: true
        - name: eth2
          vlan:
            mode: trunk
            trunk-tags:
            - id: 500
```

ffmancera:
I fixed the problems you mentioned and all the tests must be passing now. Thanks!

rust/src/lib/ifaces/linux_bridge.rs
```diff
@@ -104,6 +104,12 @@ impl LinuxBridgeInterface {
         &mut self,
         is_desired: bool,
     ) -> Result<(), NmstateError> {
+        if let Some(opts) =
+            self.bridge.as_ref().and_then(|b| b.options.as_ref())
+        {
+            opts.validate_vlan_default_pvid(self)?;
```
cathay4t: This function also run against current state, there is no need to validate there. Please check `is_desired` also.

rust/src/lib/ifaces/linux_bridge.rs
```diff
@@ -567,6 +575,25 @@ impl LinuxBridgeOptions {
...
+    pub(crate) fn validate_vlan_default_pvid(
+        &self,
+        linux_bridge: &LinuxBridgeInterface,
+    ) -> Result<(), NmstateError> {
+        if let Some(pvid) = self.vlan_default_pvid {
+            if pvid != 1 && !linux_bridge.vlan_filtering_is_enabled() {
+                return Err(NmstateError::new(
+                        ErrorKind::InvalidArgument,
+                        format!(
+                            "Linux bridge {} has vlan-default-pvid different than 1 but VLAN filtering is not enabled.",
```
cathay4t: This line is too long. Break it like:

```rust
"Linux bridge {} has vlan-default-pvid different \
than 1 but VLAN filtering is not enabled.",
```

cathay4t (approved):
Just two trivial issue. Please fix it then we can merge it.

PR #2370, opened 2023-06-20
Ovn bridge mappings api

maiqueb:
This PR introduces syntactic sugar for the OVN API to configure the bridge mappings through which OVN packets enter the underlay.

It addresses [this RFE](https://bugzilla.redhat.com/show_bug.cgi?id=2209690), which hopes to simplify (and make less error prone...) the configuration of the underlay for OVN-Kubernetes multi-networks.
More information about the OVN bridge mappings can be found [here](https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/16.0/html/networking_guide/bridge-mappings).

### DIFF
--- a/rust/src/lib/net_state.rs
+++ b/rust/src/lib/net_state.rs
@@ -5,6 +5,9 @@ use std::collections::HashMap;
 
 use serde::{Deserialize, Deserializer, Serialize};
 
+use crate::ovn::{
+    MergedOvnConfiguration, OvnConfiguration, OVN_BRIDGE_MAPPINGS,
+};
 use crate::{
     DnsState, ErrorKind, HostNameState, Interface, Interfaces, MergedDnsState,
     MergedHostNameState, MergedInterfaces, MergedOvsDbGlobalConfig,
@@ -99,6 +102,9 @@ pub struct NetworkState {
     )]
     /// The global configurations of OpenvSwitach daemon
     pub ovsdb: OvsDbGlobalConfig,
+    #[serde(default, skip_serializing_if = "OvnConfiguration::is_none")]
+    /// The OVN configuration in the system
+    pub ovn: OvnConfiguration,
     #[serde(skip)]
     // Contain a list of struct member name which is defined explicitly in
     // desire state instead of generated.
@@ -163,6 +169,11 @@ impl<'de> Deserialize<'de> for NetworkState {
             net_state.ovsdb = OvsDbGlobalConfig::deserialize(ovsdb_value)
                 .map_err(serde::de::Error::custom)?;
         }
+        if let Some(ovn_value) = v.remove("ovn") {
+            net_state.prop_list.push("ovn");
+            net_state.ovn = OvnConfiguration::deserialize(ovn_value)
+                .map_err(serde::de::Error::custom)?;
+        }
         if let Some(hostname_value) = v.remove("hostname") {
             net_state.prop_list.push("hostname");
             net_state.hostname = Some(
@@ -189,6 +200,7 @@ impl NetworkState {
             && self.routes.is_empty()
             && self.interfaces.is_empty()
             && self.ovsdb.is_none()
+            && self.ovn.is_none()
     }
 
     pub(crate) const PASSWORD_HID_BY_NMSTATE: &'static str =
@@ -355,6 +367,7 @@ pub(crate) struct MergedNetworkState {
     pub(crate) dns: MergedDnsState,
     pub(crate) interfaces: MergedInterfaces,
     pub(crate) ovsdb: MergedOvsDbGlobalConfig,
+    pub(crate) ovn: MergedOvnConfiguration,
     pub(crate) routes: MergedRoutes,
     pub(crate) rules: MergedRouteRules,
     pub(crate) memory_only: bool,
@@ -386,12 +399,29 @@ impl MergedNetworkState {
         let hostname =
             MergedHostNameState::new(desired.hostname, current.hostname);
 
+        let mut ovsdb =
+            MergedOvsDbGlobalConfig::new(desired.ovsdb, current.ovsdb);
+        let ovn_config = MergedOvnConfiguration::new(desired.ovn, current.ovn);
+        if let Some(updated_mapping_value) =
+            ovn_config.clone().mappings_ext_id_value
+        {
+            if updated_mapping_value.is_empty() {
+                ovsdb.external_ids.remove(OVN_BRIDGE_MAPPINGS);
+            } else {
+                ovsdb
+                    .external_ids
+                    .entry(OVN_BRIDGE_MAPPINGS.to_string())
+                    .and_modify(|m| *m = Some(updated_mapping_value.clone()))
+                    .or_insert(Some(updated_mapping_value.clone()));
+            }
+        }
         let ret = Self {
             interfaces,
             routes,
             rules,
             dns: MergedDnsState::new(desired.dns, current.dns)?,
-            ovsdb: MergedOvsDbGlobalConfig::new(desired.ovsdb, current.ovsdb),
+            ovsdb,
+            ovn: ovn_config,
             hostname,
             memory_only,
             prop_list: desired.prop_list,
--- a/rust/src/lib/net_state.rs
+++ b/rust/src/lib/net_state.rs
@@ -404,7 +404,7 @@ impl MergedNetworkState {
             current.ovsdb,
             current.ovn.clone().bridge_mappings,
         )?;
-        let ovn_config = MergedOvnConfiguration::new(desired.ovn, current.ovn);
+        let ovn_config = MergedOvnConfiguration::new(desired.ovn, current.ovn)?;
         if let Some(updated_mapping_value) =
             ovn_config.clone().mappings_ext_id_value
         {
--- a/rust/src/lib/net_state.rs
+++ b/rust/src/lib/net_state.rs
@@ -400,7 +400,7 @@ impl MergedNetworkState {
             MergedHostNameState::new(desired.hostname, current.hostname);
 
         let mut ovsdb =
-            MergedOvsDbGlobalConfig::new(desired.ovsdb, current.ovsdb);
+            MergedOvsDbGlobalConfig::new(desired.ovsdb, current.ovsdb)?;
         let ovn_config = MergedOvnConfiguration::new(desired.ovn, current.ovn);
         if let Some(updated_mapping_value) =
             ovn_config.clone().mappings_ext_id_value
--- a/rust/src/lib/net_state.rs
+++ b/rust/src/lib/net_state.rs
@@ -399,8 +399,11 @@ impl MergedNetworkState {
         let hostname =
             MergedHostNameState::new(desired.hostname, current.hostname);
 
-        let mut ovsdb =
-            MergedOvsDbGlobalConfig::new(desired.ovsdb, current.ovsdb)?;
+        let mut ovsdb = MergedOvsDbGlobalConfig::new(
+            desired.ovsdb,
+            current.ovsdb,
+            current.ovn.clone().bridge_mappings,
+        )?;
         let ovn_config = MergedOvnConfiguration::new(desired.ovn, current.ovn);
         if let Some(updated_mapping_value) =
             ovn_config.clone().mappings_ext_id_value
--- a/rust/src/lib/ovn.rs
+++ b/rust/src/lib/ovn.rs
@@ -39,8 +39,16 @@ impl OvnConfiguration {
     }
 
     pub fn sanitize(&self) -> Result<(), NmstateError> {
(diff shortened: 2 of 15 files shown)


rust/src/lib/ovs.rs
```diff
@@ -201,12 +221,166 @@ impl MergedOvsDbGlobalConfig {
...
+            physnet,
+            bridge: Some(bridge),
+            state: OvnBridgeMappingState::Present,
+        })
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase", deny_unknown_fields)]
+pub enum OvnBridgeMappingState {
```
cathay4t: Please include `#[non_exhaustive]` to ensure we don't break rust API when adding new state.

rust/src/lib/ovsdb/db.rs
```diff
@@ -421,3 +473,99 @@ impl OvsDbUpdate {
...
+    }
+    if mappings.is_empty() {
+        Ok(None)
+    } else {
+        Ok(Some(mappings))
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
cathay4t: Please move test code to folder `rust/src/lib/unit_tests`

cathay4t (changes requested):
If this syntax sugar, the plugins should not required to make any changes, all the wrapping should be done at top level `lib/ovs.rs` file. The `ovsdb/db.rs` do not need to handle this.

rust/src/lib/ovsdb/db.rs
maiqueb: I had it there, but I honestly couldn't get this to pass the linters because of the rust modules. I'll need some help with that.

maiqueb:
> If this syntax sugar, the plugins should not required to make any changes, all the wrapping should be done at top level `lib/ovs.rs` file. The `ovsdb/db.rs` do not need to handle this.

OK, I'll take a look into that.

rust/src/lib/ovs.rs
maiqueb: Done.

rust/src/lib/ovsdb/db.rs
maiqueb: Done.

rust/src/lib/ovs.rs
```diff
@@ -176,7 +182,25 @@ impl MergedOvsDbGlobalConfig {
...
+                                OVN_BRIDGE_MAPPINGS.to_string(),
+                                Some(ovn_bridge_mappings_to_string(
+                                    current_bridge_mappings,
+                                )),
+                            );
+                        }
+                    };
                 } else {
+                    if ex_ids.get(OVN_BRIDGE_MAPPINGS).is_some() {
+                        const INVALID_EXTERNAL_IDS_KEY: &str = "Cannot use the `ovn-bridge-mappings` external_ids key directly. Please use ovs_db.ovn-bridge-mappings API instead";
```
cathay4t: Please change `ovs_db.ovn-bridge-mappings` to `ovn.bridge-mappings`.

Please also break long line with `\`

rust/src/lib/ovn.rs
```diff
@@ -0,0 +1,268 @@
...
+impl<'de> Deserialize<'de> for OvnConfiguration {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let mut ovn_conf = Self::default();
+        let v = serde_json::Value::deserialize(deserializer)?;
+
+        if let Some(v) = v.as_object() {
+            if let Some(v) = v.get("bridge-mappings") {
```
cathay4t: If you are doing `Deserialize` by yourself, you need to handle invalid key like:

```
ovn:
  invalid-set:
    abc: 1
  bridge-mappings:
    - localnet: yellow
      abc: ab
      bridge: ovsbr3
```

The `#[serde(deny_unknown_fields)]` can do the magic if using auto deserializer `#[derive(Deserialize)]`. From my view, you do not need special deserializer here.

rust/src/lib/ovn.rs
```diff
@@ -0,0 +1,268 @@
...
+                ovn_conf.prop_list.push("bridge-mappings");
+                match value_to_ovn_bridge_mapping(v) {
+                    Ok(bridge_mappings) => {
+                        ovn_conf.bridge_mappings = bridge_mappings
+                    }
+                    Err(err) => {
+                        return Err(serde::de::Error::custom(format!(
+                        "Error unmarshalling `OvnBridgeMapping` {v:?}: {err}"
```
cathay4t: Normally we do not scare people with debug output `{:?}`.

rust/src/lib/ovn.rs
```diff
@@ -0,0 +1,268 @@
...
+use serde::{Deserialize, Deserializer, Serialize};
+use std::collections::HashMap;
+use std::fmt;
+use std::str::FromStr;
+
+pub const OVN_BRIDGE_MAPPINGS: &str = "ovn-bridge-mappings";
+
+#[derive(Debug, Clone, PartialEq, Eq, Default, Serialize)]
+#[non_exhaustive]
+pub struct OvnConfiguration {
```
cathay4t, line 28: Public document required. Try `/// `

rust/src/lib/ovn.rs
```diff
@@ -0,0 +1,268 @@
...
+        )
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase", deny_unknown_fields)]
+#[non_exhaustive]
+pub enum OvnBridgeMappingState {
+    Present,
+    Absent,
```
cathay4t, line 237: In my test, this YAML is sufficient enough to delete a map:

```yml
ovn:
  bridge-mappings:
    - localnet: red
      state: absent
```

If is on purpose, please add public document for this usage.

tests/integration/ovs_test.py
```diff
@@ -638,6 +645,69 @@ def test_remove_all_ovsdb_global_config():
     }
 
 
+def test_ovsdb_global_config_add_delete_mapping():
```
cathay4t, line 648: Please add two more test cases:
 1. Delete one map out of two+ maps.
 2. Replace/Modify a map via:
```yml
ovn:
  bridge-mappings:
    - localnet: yellow
      state: absent
    - localnet: yellow
      bridge: ovsbr5
```

cathay4t (changes requested):
In general, this patch looks very good and also passed all my manual tests.

Just some lints to fix on test coverage, error message, unknown key and documentation.

Thank you!

tests/integration/ovs_test.py
maiqueb: I think I see; your objective is to assert we can
1. remove without naming the bridge
2. state is optional when adding / preserving

Right ?

rust/src/lib/ovn.rs
maiqueb: It is (enough to delete, and on purpose).

I'll add the public doc.

rust/src/lib/ovn.rs
maiqueb: OK, I'll throw away the custom deserializer and run with derived.

rust/src/lib/ovs.rs
maiqueb: Good catch.

tests/integration/ovs_test.py
maiqueb: Wait, I just now noticed your second example features the **same** localnet key for both listed mappings.

That's weird... Is that intentional ? I really don't know what the expected behavior should be.
@cathay4t

cathay4t: > 1. Delete one map out of two+ maps.

This is to make sure nmstate will not touch undesired ovn bridge map.

> 2. Replace/Modify a map via

This is for changing OVN mapping in single YML.

maiqueb: We've decided offline overriding the bridge would provide a good user experience, and thus we'll implement that.

To make the usage less error prone, we've also decided to reject desired states where the bridge mappings feature duplicate `localnet` keys.

Provided an e2e + unit test for this scenario.

I chose to just reject, instead of printing the exact duplicated localnet key, since that would require more code.

Let me know if you prefer me to take that route (i.e. point at the duplicate error key in the error msg).

rust/src/lib/ovn.rs
maiqueb: Done.

rust/src/lib/ovn.rs
maiqueb: Done - I pretty much just moved it from the `OvnBridgeMapping` struct to here. 

Let me know if we need more docs.

rust/src/lib/ovs.rs
maiqueb: Done.

rust/src/lib/ovn.rs
maiqueb: Still trying to do this ... Facing some issues.

maiqueb:
failing on one lane on an unrelated test: `test_new_bond_uses_mac_of_first_port_by_name`.

Can we somehow request that particular lane to be re-executed ? @cathay4t / @ffmancera

cathay4t:
> failing on one lane on an unrelated test: `test_new_bond_uses_mac_of_first_port_by_name`.

It is a known random failure never able to fix or even reproduced in my VM.

> 
> Can we somehow request that particular lane to be re-executed ? @cathay4t / @ffmancera

Restarted

rust/src/lib/ovn.rs
maiqueb: Hm, I had just copied this from https://github.com/nmstate/nmstate/blob/f0f7c1edd8152dc03e6287c2222b524167c6f128/rust/src/lib/ovs.rs#L114

Well, doesn't matter, I'm trying to see if I can avoid using the custom deserializer - if so, all this code will be gone.

rust/src/lib/ovn.rs
maiqueb: I'm thinking about this some more and there may be a reason to use a custom deserializer - although not a very good one.

If I use a custom deserializer I can provide a better error msg to the user if they try to provision the following bridge mapping:
```yaml
ovn:
  bridge-mappings:
  - localnet: net1
```

The above example is wrong because the user **must** also provide the bridge.

Right now, all I get is:
```
[2023-07-07T07:37:51Z INFO  nmstate::query_apply::net_state] Retrying on: VerificationError: Verification failure: ovn.bridge-mappings desire '[{"localnet":"net1"}]', current '[]'
```

Thoughts @cathay4t ?

cathay4t: The `sanitize()` function is designed post-deserialize check and clean up.

maiqueb: Added a `sanitize` func to ensure a mappings being added / modified has the required info.

Also moved asserting there aren't duplicate `localnet` keys in the desired state to a separate sanitize func.

Finally, removed the custom deserializer.

rust/src/lib/ovn.rs
maiqueb: This was removed, since the deserializer code is now gone.

rust/src/lib/net_state.rs
```diff
@@ -99,6 +102,13 @@ pub struct NetworkState {
     )]
     /// The global configurations of OpenvSwitach daemon
     pub ovsdb: OvsDbGlobalConfig,
+    #[serde(
+        default,
+        rename = "ovn",
```
cathay4t: No need to rename.

rust/src/lib/query_apply/net_state.rs
```diff
@@ -309,6 +312,41 @@ impl NetworkState {
...
+                                let mut updated_external_ids =
+                                    other.ovsdb.external_ids.clone();
+                                updated_external_ids
+                                    .as_mut()
+                                    .unwrap()
+                                    .remove(OVN_BRIDGE_MAPPINGS);
+
+                                let mut sorted_mappings =
+                                    updated_mappings.clone();
+                                sorted_mappings.sort_by(|v1, v2| {
```
cathay4t: `sort_unstable_by()` is quicker.

cathay4t (approved):
Good enough to merge. Just two lints:
 * No need to rename from `ovn` to `ovn` in serializer.
 * Use `sort_unstable_by()` which is quicker(according to rust doc).

rust/src/lib/query_apply/net_state.rs
maiqueb: Done.

rust/src/lib/net_state.rs
maiqueb: Wow of course :sweat_smile: 

Done.

PR #2397, opened 2023-08-17
tests: check ovs-bridge and ovs-interface same name can modify

ffmancera:
When ovs-bridge and ovs-interface share names, Nmstate should be able to modify ovs-interface without any error.

https://bugzilla.redhat.com/show_bug.cgi?id=2231843


ffmancera:
Please, notice that this is not failing in nmstate 2.x but this is a scenario that we should cover as we found a bug in 1.x

.github/workflows/main.yml
```diff
@@ -146,7 +146,7 @@ jobs:
       - name: Install extra kernel modules(e.g. vrf)
         run: |
           sudo apt-get update;
-          sudo apt-get -y install "linux-modules-extra-$(uname -r)"
+          sudo apt-get -y install "linux-modules-extra-azure"
```
cathay4t: The reason we use `uname -r` here is because repo might hold newer version than running kernel.

The CI failure should be fixed by https://github.com/actions/runner-images/issues/8080

cathay4t (changes requested):
Let's wait github CI fix the repo.

.github/workflows/main.yml
ffmancera: Sure, let's wait then. I am dropping the commit. Thanks for sharing the link.

PR #2418, opened 2023-09-25
bond: Add support for the ports-config

liangwen12year:
The bond port priority is only supported to set when bonding mode is active-backup(1) or balance-tlb (5) or balance-alb (6).


liangwen12year:
This PR is still in draft status.

examples/bond_port_up.yml
```diff
@@ -0,0 +1,27 @@
...
+  - name: bond0
+    type: bond
+    state: up
+    link-aggregation:
+      mode: active-backup
+      options:
+        miimon: 140
+      port:
+        - eth1
+        - eth2
```
cathay4t, line 14: We should not suggest user to define port names in two places.

And in commit message, we should explain an error will happen if `port` does not agree with `ports-config`.

rust/src/lib/ifaces/bond.rs
```diff
@@ -104,8 +105,17 @@ impl BondInterface {
         }
     }
 
+    fn sort_ports_config(&mut self) {
+        if let Some(ref mut bond_conf) = self.bond {
+            if let Some(ref mut port_conf) = &mut bond_conf.ports_config {
+                port_conf.sort_unstable_by_key(|p| p.name.clone())
```
cathay4t, line 111: Just a rust quiz: could you explain why we need `clone()` here?

rust/src/python/libnmstate/schema.py
```diff
@@ -169,6 +169,12 @@ class Bond:
     PORT = "port"
     PORTS = "ports"
     OPTIONS_SUBTREE = "options"
+    PORTS_CONFIG_SUBTREE = "ports-config"
+
+    class PortsConfig:
+        NAME = "name"
+        PRIO = "prio"
```
cathay4t: Please use `PRIORITY = "priority"`.

rust/src/lib/ifaces/bond.rs
```diff
@@ -1282,3 +1346,46 @@ impl MergedInterface {
...
+#[non_exhaustive]
+pub struct BondPortConfig {
+    /// name is mandatory when specifying the ports configuration.
+    pub name: String,
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        default,
+        deserialize_with = "crate::deserializer::option_i32_or_string"
+    )]
+    pub prio: Option<i32>,
```
cathay4t: Please use `priority` instead of `prio`. This user facing, abbreviation is bad idea.

cathay4t: And document please.

cathay4t (changes requested):
* Change property name from `prio` to `priority`
* Add unit test case for conflict YAML between `port` and `ports-config`.

rust/src/lib/ifaces/bond.rs
liangwen12year: It relates to Rust's ownership, which prevents moving the element in a collection during sorting unless the elements are cloned ?

examples/bond_port_up.yml
liangwen12year: I revised the commit message and added the unit case to assert that an error will be raised.

liangwen12year: Maybe I need to add some validation function to raise specific errors.

liangwen12year:
@cathay4t ,thanks for reviewing, I think that all your comments are addressed now.

cathay4t (approved):
LGTM

PR #2421, opened 2023-09-29
[nmstate-1.4] ip: Support treating string as int for `prefix-length`

liangwen12year:
When the network role user is using the `network_state` variable to configure the network, and if they are using Jinja2 template to define the `prefix-length`, the type conversion
`prefix-length: "{{ __str_val | int }}"` does not work as expected, the type for `prefix-length` in the end is still string. Therefore, nmstate need to support treating string as int for `prefix-length` in order to make the apply succeed.

### DIFF
--- a/libnmstate/ifaces/base_iface.py
+++ b/libnmstate/ifaces/base_iface.py
@@ -48,6 +48,7 @@ class IPState:
         self._info = info
         self._remove_stack_if_disabled()
         self._canonicalize_ip_addr()
+        self._canonicalize_ip_prefix()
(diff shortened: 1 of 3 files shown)


cathay4t:
I have fixed the CI. Please rebase and try again.

liangwen12year:
> I have fixed the CI. Please rebase and try again.

I rebased, thanks.

cathay4t (approved):

cathay4t:
Merged. Thanks!

PR #2429, opened 2023-10-11
Validate prefix length for IP addresses

ihuguet:
Resolves: RHEL-1670

Add validation for the prefix_length field so the configuration is rejected if it's too high.

Without this, the invalid IP address is sent to NetworkManager to be configured. Then NetworkManager rejects to use it, but due to a bug it doesn't report an error back to nmstate.

Additionally, fix an existent minor bug that allowed an error to be reported in `sanitize` with `is_desired=false`.

### DIFF
--- a/rust/src/lib/ip.rs
+++ b/rust/src/lib/ip.rs
@@ -338,6 +338,19 @@ impl InterfaceIpv4 {
                         ),
                     ));
                 }
+                if let Some(addr) = addrs
+                    .iter()
+                    .find(|a| a.prefix_length as usize > IPV4_ADDR_LEN)
+                {
+                    return Err(NmstateError::new(
+                        ErrorKind::InvalidArgument,
+                        format!(
+                            "Invalid IPv4 network prefix length '{}', \
+                            should be in the range of 0 to {IPV4_ADDR_LEN}",
+                            addr.prefix_length
+                        ),
+                    ));
+                }
             }
             addrs.retain(|a| !a.is_auto());
             addrs.iter_mut().for_each(|a| {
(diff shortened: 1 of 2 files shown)


rust/src/lib/ip.rs
```diff
@@ -338,6 +338,18 @@ impl InterfaceIpv4 {
...
                     ));
                 }
+                if let Some(addr) = addrs
+                    .iter()
+                    .find(|a| a.prefix_length as usize > IPV4_ADDR_LEN)
+                {
+                    return Err(NmstateError::new(
+                        ErrorKind::InvalidArgument,
+                        format!(
+                            "Invalid IPv4 prefix_length {}",
```
cathay4t: `Invalid IPv4 prefix_length {}, should be in the range of 0 to 32`

rust/src/lib/ip.rs
```diff
@@ -618,13 +630,27 @@ impl InterfaceIpv6 {
...
+                    ));
+                }
+                if let Some(addr) = addrs
+                    .iter()
+                    .find(|a| a.prefix_length as usize > IPV6_ADDR_LEN)
+                {
+                    return Err(NmstateError::new(
+                        ErrorKind::InvalidArgument,
+                        format!(
+                            "Invalid IPv6 prefix_length {}",
```
cathay4t: `Invalid IPv4 prefix_length {}, should be in the range of 0 to 128`

cathay4t (approved):
Besides small lints, looks good to me.

ihuguet:
Changed the error mesages and rebased `base` branch.

PR #2450, opened 2023-11-07
go: Configure GOPROXY and GOSUMDB

qinqon:
Latest go 1.21 installed at dev c9s image need a pair of go env to be set. This change configure them with default values.

https://github.com/golang/go/issues/62274
https://github.com/golang/go/issues/31964


cathay4t (approved):
LGTM

PR #2453, opened 2023-11-09
route: Add support for the `route-type`

liangwen12year:
When users have BGP routing setups, it is common practice to blackhole some less-specific routes in order to avoid routing loops, and the BGP router might insert a more specific route dynamically afterwards.

Examples:

```
interfaces:
  - name: eth1
    type: ethernet
    state: up
    ipv4:
      address:
        - ip: 192.0.2.251
          prefix-length: 24
      dhcp: false
      enabled: true

routes:
  config:
      metric: 150
      next-hop-address: 192.0.2.1
      next-hop-interface: eth1
      table-id: 254
    - destination: 198.51.200.0/24
      route-type: blackhole
```

### DIFF
--- a/rust/src/lib/nm/nm_dbus/connection/route.rs
+++ b/rust/src/lib/nm/nm_dbus/connection/route.rs
@@ -16,6 +16,7 @@ pub struct NmIpRoute {
     pub table: Option<u32>,
     pub metric: Option<u32>,
     pub weight: Option<u32>,
+    pub route_type: Option<String>,
     _other: DbusDictionary,
 }
 
@@ -35,6 +36,7 @@ impl TryFrom<DbusDictionary> for NmIpRoute {
             table: _from_map!(v, "table", u32::try_from)?,
             metric: _from_map!(v, "metric", u32::try_from)?,
             weight,
+            route_type: _from_map!(v, "type", String::try_from)?,
             _other: v,
         })
     }
@@ -82,7 +84,12 @@ impl NmIpRoute {
                 zvariant::Value::new(zvariant::Value::new(v)),
             )?;
         }
-
+        if let Some(v) = &self.route_type {
+            ret.append(
+                zvariant::Value::new("type"),
+                zvariant::Value::new(zvariant::Value::new(v)),
+            )?;
+        }
         for (key, value) in self._other.iter() {
             ret.append(
                 zvariant::Value::new(key.as_str()),
--- a/rust/src/lib/route.rs
+++ b/rust/src/lib/route.rs
@@ -11,6 +11,9 @@ use crate::{
     ErrorKind, InterfaceType, MergedInterfaces, NmstateError,
 };
 
+const DEFAULT_TABLE_ID: u32 = 254; // main route table ID
+const LOOPBACK_IFACE_NAME: &str = "lo";
+
 #[derive(Debug, Clone, PartialEq, Eq, Default, Serialize, Deserialize)]
 #[non_exhaustive]
 #[serde(deny_unknown_fields)]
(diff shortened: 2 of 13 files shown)


cathay4t (changes requested):
Blackhole route should not hold `next-hop-interface`, we should raise error when user desire so.

For NetworkManager, we always store blackhole route into loopback interface.

liangwen12year:
> Blackhole route should not hold `next-hop-interface`, we should raise error when user desire so.
> 
> For NetworkManager, we always store blackhole route into loopback interface.

I dropped the `next-hop-interface` from the blackhole route and added the validation code.

rust/src/lib/route.rs
```diff
@@ -64,17 +64,43 @@ impl Routes {
...
+                ];
+                if !route.is_absent() {
+                    if route.route_type.is_some()
+                        && no_nexthop_route_type
+                            .contains(&route.route_type.unwrap())
+                        && (route.next_hop_iface.is_some()
+                            || route.next_hop_addr.is_some())
+                    {
+                        return Err(NmstateError::new(
+                            ErrorKind::NotImplementedError,
```
liangwen12year: ahh, Maybe I should add test case testing the validation code here.. Will do.

liangwen12year: Tests added.

rust/src/lib/nm/nm_dbus/connection/route.rs
```diff
@@ -113,3 +120,52 @@ pub(crate) fn nm_ip_routes_to_value(
...
+const RTN_UNREACHABLE: u8 = 7;
+const RTN_PROHIBIT: u8 = 8;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum NmIpRouteType {
+    Blackhole,
+    Unreachable,
+    Prohibit,
+    Other(u8),
```
cathay4t: We do not need this as `nm_dbus` is not a public API.

tests/integration/route_test.py
```diff
@@ -140,6 +141,67 @@ def test_add_static_route_without_next_hop_address(eth1_up):
...
+            Route.NEXT_HOP_INTERFACE: "eth1",
+            Route.ROUTETYPE: Route.ROUTETYPE_BLACKHOLE,
+        },
+    ]
+    state = {
+        Interface.KEY: [ETH1_INTERFACE_STATE],
+        Route.KEY: {Route.CONFIG: route},
+    }
+
+    with pytest.raises(NmstateNotImplementedError):
```
cathay4t: Should be `NmstateValueError` as this is incorrect desired state.

cathay4t (changes requested):
Also need test to make sure verification passed regardless user defined `next-hop-interface: lo` or not.

tests/integration/route_test.py
liangwen12year: I changed into `NmstateValueError`.

liangwen12year:
> Also need test to make sure verification passed regardless user defined `next-hop-interface: lo` or not.

I added the test which defines  `next-hop-interface: lo` .

cathay4t (changes requested):
1. I see duplicate route entry in `nmcli c show lo` if I apply the same YAML mulitple times.
2. The kernel and iproute command never show blackhole route as `next-hop-interface: lo`. We should do the same by hiding it for ipv4 route.
3. The `state: absent` remove other routes:

```yml
routes:
  config:
  - state: absent
    route-type: blackhole
```

4. Please convert above as test case(unit or integration).

liangwen12year:
> 1. I see duplicate route entry in `nmcli c show lo` if I apply the same YAML mulitple times.
> 2. The kernel and iproute command never show blackhole route as `next-hop-interface: lo`. We should do the same by hiding it for ipv4 route.
> 3. The `state: absent` remove other routes:
> 
> ```yaml
> routes:
>   config:
>   - state: absent
>     route-type: blackhole
> ```
> 
> 4. Please convert above as test case(unit or integration).

Will do, thanks for reviewing it.

liangwen12year:
> 1. I see duplicate route entry in `nmcli c show lo` if I apply the same YAML mulitple times.

Actually, regular route has the same issues when applying the same YAML mulitple times (caused by retrieving the routes and showing route table by default ).  e.g. 
```
[root@fcvm rust]# nmcli c show eth1 | grep routes
ipv4.routes:                            { ip = 198.51.100.0/24, nh = 192.0.2.1, mt = 150 }; { ip = 198.51.100.0/24, nh = 192.0.2.1, mt = 150 table=254 }
```

cathay4t:
You patch never modify `RouteEntry.sort_key()`, hence dedup never works.

For problem you mentioned, please also fix.

liangwen12year:
> You patch never modify `RouteEntry.sort_key()`, hence dedup never works.
> 
> For problem you mentioned, please also fix.

I modified the  `RouteEntry.sort_key()`, dedup should work now, I also added the proper tests for checking dedup.

cathay4t (approved):
LGTM. Thanks!

PR #2465, opened 2023-11-20
Skip activation of port as its controller desired to be down/absent

liangwen12year:
The test `test_create_down_linux_bridge` fails randomly, the root cause is that when the controller desires to be down or absent, we should not activate the port connection again, because the activation of the port will risk making its controller activate again.

### DIFF
--- a/rust/src/lib/nm/profile.rs
+++ b/rust/src/lib/nm/profile.rs
@@ -148,6 +148,32 @@ fn can_skip_activation(
     nm_conn: &NmConnection,
     exist_nm_conns: &[NmConnection],
 ) -> bool {
+    // if the controller is desired to be down or absent, activating the
+    // connection on the port will risk making the controller activate again,
+    // therefore skip the activation on the port
+    if let Some(desired_iface) = merged_iface.for_apply.as_ref() {
+        if let (Some(ctrl_iface), Some(ctrl_type)) = (
+            desired_iface.base_iface().controller.as_deref(),
+            desired_iface.base_iface().controller_type.as_ref(),
+        ) {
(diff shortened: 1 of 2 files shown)


rust/src/lib/nm/profile.rs
```diff
@@ -148,6 +148,33 @@ fn can_skip_activation(
...
+        if let (Some(ctrl_iface), Some(ctrl_type)) = (
+            desired_iface.base_iface().controller.as_deref(),
+            desired_iface.base_iface().controller_type.as_ref(),
+        ) {
+            if let Some(merged_ctrl_iface) =
+                merged_ifaces.get_iface(ctrl_iface, ctrl_type.clone())
+            {
+                if merged_ctrl_iface.for_apply.is_some()
+                    && (merged_ctrl_iface.merged.is_absent()
+                        || merged_ctrl_iface.merged.is_down())
```
liangwen12year, line 164: Not sure if we should call `merged_ctrl_iface.desired.as_ref().unwrap.is_absent()` or `merged_ctrl_iface.desired.as_ref().unwrap.is_down()` here ?

liangwen12year:
I run the test `test_create_down_linux_bridge` 300 times, but not found any random failure.

rust/src/lib/nm/profile.rs
cathay4t: unwrap() is never allowed in library code. Current code is OK as you already checked `for_apply()`, then the `merged` will hold `desired` state.

cathay4t (approved):
Nicely done.

I have confirmed the random failure is caused by nmstate modify bridge port and reapply the bridge port.

PR #2469, opened 2023-11-27
hsr: add support to HSR/PRP interface

ffmancera:
Please notice that HSR and PRP interface are represented by the same driver [1]. In order to decide the protocol to be used by the HSR interface the protocol field was added. By default, the protocol used is HSR.

Example:

```
---
interfaces:
  - name: hsr0
    type: hsr
    state: up
    hsr:
      port1: eth1
      port2: eth2
      supervision-address: 70:FF:76:1C:0E:8D
      multicast-spec: 40
      protocol: prp
```

[1] https://lwn.net/Articles/826386/


ffmancera:
This cannot be merged until the nispor PR is merged and there is a release

cathay4t (changes requested):
I have rebased the patch with small fixes to python schema and CI test code typo.

But the test all fails because `supervision_addr` failed on verification, the config has sent to NetworkManager, but NM seems failed to apply so.

tests/integration/hsr_test.py
```diff
@@ -0,0 +1,77 @@
...
+from .testlib import assertlib
+
+
+ETH1 = "eth1"
+ETH2 = "eth2"
+HSR0 = "hsr0"
+
+
+@pytest.mark.tier1
+def test_add_hsr_and_remove(eth1_up, eth2_up):
```
liangwen12year, line 25: Does NM always support the hsr/prp interface? maybe the tests need to add certain condition about which NM version the test should be based on.

liangwen12year:
And I see the random failure on `ERROR at setup of test_auto_manage_linux_ignored_ports` recently on the CI, maybe it is worthy to take a look at it as a separate task (I already created the jira issue to track this).    

```
input = None, capture_output = False, timeout = None, check = True
popenargs = (['ip', 'link', 'add', 'dummy2', 'type', 'dummy'],)
kwargs = {'close_fds': True, 'env': None, 'stderr': -1, 'stdout': -1}
process = <Popen: returncode: 2 args: ['ip', 'link', 'add', 'dummy2', 'type', 'dummy']>
stdout = b'', stderr = b'RTNETLINK answers: File exists\n', retcode = 2
```
It seems that the cleanup of some previous tests fails to remove dummy2 interface.

tests/integration/hsr_test.py
liangwen12year: I also find similar failures when I apply the example manually:
```
[root@fcvm rust]# ./target/debug/nmstatectl apply ../examples/hsr0_up.yml 
[2023-12-14T21:58:08Z INFO  nmstate::query_apply::net_state] Created checkpoint /org/freedesktop/NetworkManager/Checkpoint/1
[2023-12-14T21:58:08Z INFO  nmstate::nm::query_apply::profile] Creating connection UUID Some("419ff1b8-138b-4cd3-a9d1-888e9869cdbe"), ID Some("hsr0"), type Some("hsr") name Some("hsr0")
[2023-12-14T21:58:08Z INFO  nmstate::query_apply::net_state] Rollbacked to checkpoint /org/freedesktop/NetworkManager/Checkpoint/1
NmstateError: DependencyError: Please upgrade NetworkManager for specified interface type: Connection(InvalidSetting):hsr: unknown setting name

```

ffmancera: Nice catch. Fixed!

cathay4t:
> And I see the random failure on `ERROR at setup of test_auto_manage_linux_ignored_ports` recently on the CI, maybe it is worthy to take a look at it as a separate task (I already created the jira issue to track this).
> 
> ```
> input = None, capture_output = False, timeout = None, check = True
> popenargs = (['ip', 'link', 'add', 'dummy2', 'type', 'dummy'],)
> kwargs = {'close_fds': True, 'env': None, 'stderr': -1, 'stdout': -1}
> process = <Popen: returncode: 2 args: ['ip', 'link', 'add', 'dummy2', 'type', 'dummy']>
> stdout = b'', stderr = b'RTNETLINK answers: File exists\n', retcode = 2
> ```
> 
> It seems that the cleanup of some previous tests fails to remove dummy2 interface.

Should be fixed now.

rust/src/lib/ifaces/hsr.rs
```diff
@@ -0,0 +1,100 @@
...
+#[non_exhaustive]
+#[derive(Default)]
+pub struct HsrConfig {
+    /// The port1 interface name.
+    pub port1: String,
+    /// The port2 interface name.
+    pub port2: String,
+    /// The MAC address used for the supervision frames. This property is
+    /// read-only.
+    pub supervision_address: Option<String>,
```
cathay4t, line 77: If this property is read only, we should raise an info message indicate we are ignoring desired supervision_address when user applying back the current state.

cathay4t: Also need `#[serde(skip_serializing_if = "Option::is_none"`.

cathay4t (changes requested):

rust/src/lib/ifaces/hsr.rs
ffmancera: Done, thank you!

ffmancera:
There is a bug on kernel side and nispor is reporting multicast_spec always zero. The multicast spec is being set correctly on supervision_address so it is working fine. The bug is just on the reporting of the property.

I am creating a patch for the kernel but it is something that will take time. In order to workaround it, we parse the value directly from supervison_address.

ffmancera:
The test failing is because the image needs to be rebuild.

cathay4t (approved):

PR #2487, opened 2023-12-12
vlan: Implement show and apply for "flags" fields

qinqon:
The vlan NM field "flags" is present at nispor but it was missing at nmstate to show and apply, this change implement it as individual bool flags and as enum for registration protocols.

Closes https://issues.redhat.com/browse/RHEL-19142

### DIFF
--- a/rust/src/lib/nm/nm_dbus/connection/vlan.rs
+++ b/rust/src/lib/nm/nm_dbus/connection/vlan.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::convert::TryFrom;
 
 use log::error;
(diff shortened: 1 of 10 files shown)


rust/src/lib/nm/nm_dbus/connection/vlan.rs
```diff
@@ -15,6 +15,7 @@ pub struct NmSettingVlan {
     pub parent: Option<String>,
     pub id: Option<u32>,
     pub protocol: Option<String>,
+    pub flags: Option<u32>,
```
cathay4t: Please use `Vec<NmSettingVlanFlag>` here. You may take `Vec<NmSettingsConnectionFlag>` as example.

cathay4t (changes requested):

cathay4t (approved):

PR #2513, opened 2024-01-03
lib: Use serde serialize at enums for Display

qinqon:
There are some enums that has a custom function to convert to string, this can be done using serde "rename" annotation and using serde to convert them to string.

### DIFF
--- a/rust/src/lib/iface.rs
+++ b/rust/src/lib/iface.rs
@@ -1,9 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
-
 use std::collections::HashSet;
 use std::iter::FromIterator;
 
-use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use serde::{Deserialize, Deserializer, Serialize};
 
 use crate::{
     BaseInterface, BondInterface, DummyInterface, ErrorKind, EthernetInterface,
@@ -15,8 +14,11 @@ use crate::{
 
 use crate::state::merge_json_value;
 
-#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
+#[derive(
+    Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize,
+)]
 #[non_exhaustive]
+#[serde(rename_all = "kebab-case")]
 /// Interface type
 pub enum InterfaceType {
     /// [Bond interface](https://www.kernel.org/doc/Documentation/networking/bonding.txt)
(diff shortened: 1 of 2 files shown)


rust/src/lib/iface.rs
```diff
@@ -83,82 +104,15 @@ impl Default for InterfaceType {
...
-        let v = serde_json::Value::deserialize(deserializer)?;
-        match v.as_str() {
-            Some(s) => Ok(InterfaceType::from(s)),
-            None => Ok(InterfaceType::Unknown),
+impl fmt::Display for InterfaceType {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        if let Ok(interface_type) = serde_yaml::to_string(&self) {
+            let mut interface_type_chars = interface_type.chars();
+            // Remove yaml last new line char
+            interface_type_chars.next_back();
```
cathay4t: We are assuming `serde_yaml` not appending `---` at the beginning and always add new line at the end, this is never a API promise of  `serde_yaml`.

This is hack. If ` #[serde(rename` can help your golang binding, then keep both serde rename and `impl std::fmt::Display` with a comment.

cathay4t (changes requested):

rust/src/lib/iface.rs
cathay4t: And I don't like `std::fmt::Display` depend on `serde::Serialize`. The dependency direction is wrong.

qinqon: ok, I will keep both.

rust/src/lib/ifaces/base.rs
```diff
@@ -77,12 +77,12 @@ pub struct BaseInterface {
     /// network interface activated.
     /// Serialize and deserialize to/from `wait-ip`.
     pub wait_ip: Option<WaitIp>,
-    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(skip_serializing_if = "Option::is_none", rename = "ipv4")]
```
cathay4t: This rename has no effect. Why we need it?

cathay4t (approved):

rust/src/lib/ifaces/base.rs
qinqon: I think this is wrong  from my part at golang code the field is "IPv4" so my kebab-case convertions conert it to ipv-4, let me check if fixing that make this irrelevant.

qinqon: Fixed

rust/src/lib/ip.rs
```diff
@@ -1035,133 +1035,109 @@ impl std::convert::TryFrom<&str> for InterfaceIpAddr {
...
+                Ipv6AddrGenMode::StablePrivacy => "stable-privacy",
+                Ipv6AddrGenMode::Other(s) => s,
+            }
+        )
     }
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
-#[serde(rename_all = "kebab-case", deny_unknown_fields)]
+#[serde(rename_all = "lowercase", deny_unknown_fields)]
```
qinqon: we don't need this change either, it was a problem at the golang generator.

cathay4t (approved):

PR #2520, opened 2024-01-09
connection: drop "controller" from the extra values

ffmancera:
NetworkManager is introducing a new alias for "master" which is
"controller". Nmstate is taking unknown DBus parameters and sending it
back unmodified. NetworkManager is prioritising "master" over
"controller" if both are present but when removing the port only
"controller" is present in Nmstate code.

In order to fix this temporarily, drop "controller" from the other
values dictionary. This is going to be dropped when using "controller"
everywhere.

### DIFF
--- a/rust/src/lib/nm/nm_dbus/connection/conn.rs
+++ b/rust/src/lib/nm/nm_dbus/connection/conn.rs
@@ -366,6 +366,7 @@ pub struct NmSettingConnection {
 impl TryFrom<DbusDictionary> for NmSettingConnection {
     type Error = NmError;
     fn try_from(mut v: DbusDictionary) -> Result<Self, Self::Error> {
+        v.remove("controller");
         Ok(Self {
             id: _from_map!(v, "id", String::try_from)?,
             uuid: _from_map!(v, "uuid", String::try_from)?,


ffmancera:
This is related to: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/1809

thom311:
> In order to fix this temporarily, set "master" to empty string.

This comment seems not correct? I think the patch does not set master to an empty string.

cathay4t:
In thread of removing the `controller` property, can we just override it with current `master` value.

ffmancera:
> In thread of removing the `controller` property, can we just override it with current `master` value.

That would be also fine!

ffmancera:
> > In order to fix this temporarily, set "master" to empty string.
> 
> This comment seems not correct? I think the patch does not set master to an empty string.

I updated the commit but the PR description didn't change, sorry. Changing it now.

cathay4t (changes requested):
Instead of removing `controller` as workaround. I prefer we do:
 * If existing NmConnection has `controller` property, use that and discard `master` property.
 * Optionally: For new NmConnection, we check `nm_api.version()` and use `controller` when possible.

With this, we do not need to visit this topic again in the future.

ihuguet:
Gris is right.
- If the daemon sends "controller", we know that the daemon is new enough to understand "controller":
    1. If there is also a "master" property, we delete it.
    2. We only send back "controller".
- If the daemon does NOT send "controller":
    - It might be an old daemon: we have to send "master" back.
    - Or it is a transition version that understands "controller" but is still only serializing "master": it will accept either "master" or "controller".
    - Only checking the NM version we can know what case it is, but I don't see the benefit of that. We still have to send "master" for old versions, so let's just send "master" in both cases.

We'll remove the support for "master" when enough time has passed to support only NM >= 1.46.

liangwen12year:
> Instead of removing `controller` as workaround. I prefer we do:
> 
> * If existing NmConnection has `controller` property, use that and discard `master` property.
> * Optionally: For new NmConnection, we check `nm_api.version()` and use `controller` when possible.
> 
> With this, we do not need to visit this topic again in the future.

+1

ffmancera:
I am going to work on this also for all the deprecated properties not only "controller".

PR #2532, opened 2024-01-22
Makefile: fix python install path for Python >= 3.12

ihuguet:
The module distutils has been removed in Python 3.12. This causes that the command to get the install path for Python files fails.

Replace distutil's get_python_lib with sysconfig.get_path('purelib').


liangwen12year:
I have checked https://docs.python.org/3/library/sysconfig.html#installation-paths and https://peps.python.org/pep-0632/#backwards-compatibility, the changes lgtm.

liangwen12year (approved):

PR #2535, opened 2024-01-24
ip: Support assigning static routes to interface without addresses

liangwen12year:
OpenShift MetalLB team requests to configure additional routes whenever the nodes does not have a configured IP address or route for the subnet in which MetalLB issues addresses.

Note in linux network stack, it does not matter what interface you add the address on a node (for example, loopback), the kernel is always processing arp-requests and sending arp-replies to any of them, this behavior is considered correct and, moreover, it is widely used in the a dynamic environment as Kubernetes.

https://issues.redhat.com/browse/RHEL-4908

### DIFF
--- a/rust/src/lib/nm/settings/ip.rs
+++ b/rust/src/lib/nm/settings/ip.rs
@@ -42,11 +42,16 @@ fn gen_nm_ipv4_setting(
         .cloned()
         .collect();
 
+    let mut nm_setting = nm_conn.ipv4.as_ref().cloned().unwrap_or_default();
     let mut addresses: Vec<String> = Vec::new();
     let method = if iface_ip.enabled {
+        let ipv4_routes = gen_nm_ip_routes(routes.unwrap_or_default(), false)?;
         if iface_ip.dhcp == Some(true) {
             NmSettingIpMethod::Auto
-        } else if !nmstate_ip_addrs.is_empty() {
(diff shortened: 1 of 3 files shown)


liangwen12year:
Since the corresponding NM updates landed in NM version 1.45.11, I have manually tested the patch in the container, it works as expected.

liangwen12year:
Hi @cathay4t , I would like to hear your opinion about two solutions around my implementation on assigning IPv4 static route to interface without ip:
1. when ipv4/6 is enabled, just preserve all the routes on that device by default, when ipv4/6 is disabled, just clear all the routes on that device by default
 2. introduce `allow_routes_on_default` flag to the ipv4/6 setting which is false by default, but this can allow users to explicitly control if they want to preserve the  routes  on that device
 
 Which method is better in your opinion ?

cathay4t (changes requested):
Please remove the comment regarding treating empty IP address as disabled in `test_remove_all_ip_address` test case.

Besides that, the patch looks good to me.

qinqon (approved):
lgtm

liangwen12year:
> Please remove the comment regarding treating empty IP address as disabled in `test_remove_all_ip_address` test case.
> 
> Besides that, the patch looks good to me.

I removed it, thanks.

PR #2537, opened 2024-01-26
route: Allow route destination `0.0.0.0/8` when route type is specified

liangwen12year:
According to RFC6890, `0.0.0.0/8` and its subnet cannot be used as the route destination except for the route type is unreachable, blackhole or prohibit.

### DIFF
--- a/rust/src/lib/route.rs
+++ b/rust/src/lib/route.rs
@@ -71,15 +71,8 @@ impl Routes {
         // for route with route type `Blackhole`, `Unreachable`, `Prohibit`.
         if let Some(config_routes) = self.config.as_ref() {
             for route in config_routes.iter() {
-                let no_nexthop_route_type = [
-                    RouteType::Blackhole,
-                    RouteType::Unreachable,
-                    RouteType::Prohibit,
-                ];
                 if !route.is_absent() {
-                    if route.route_type.is_some()
-                        && no_nexthop_route_type
-                            .contains(&route.route_type.unwrap())
+                    if !route.is_unicast()
                         && (route.next_hop_iface.is_some()
                             && route.next_hop_iface
                                 != Some(LOOPBACK_IFACE_NAME.to_string())
@@ -94,9 +87,7 @@ impl Routes {
                             ),
                         ));
                     } else if route.next_hop_iface.is_none()
-                        && (route.route_type.is_none()
-                            || !no_nexthop_route_type
-                                .contains(&route.route_type.unwrap()))
+                        && route.is_unicast()
                     {
                         return Err(NmstateError::new(
                             ErrorKind::NotImplementedError,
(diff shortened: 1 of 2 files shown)


rust/src/lib/route.rs
```diff
@@ -612,7 +611,11 @@ impl MergedRoutes {
 
 // Validating if the route destination network is valid,
 // 0.0.0.0/8 and its subnet cannot be used as the route destination network
-fn validate_route_dst(dst: &str) -> Result<(), NmstateError> {
+// except for the route type is unreachable, blackhole or prohibit
+fn validate_route_dst(
+    dst: &str,
+    is_no_nexthop_route_type: bool,
```
cathay4t: Please do not use `negative` for boolean variable, it will be hard for people(as least for mortal like me) to understand `is_no_nexthop_route_type: false`.

Meanwhile, instead of spreading decision making code in two place, please just pass `route_type: Option<RouteType>` to `validate_route_dst()`.

cathay4t (changes requested):

rust/src/lib/route.rs
cathay4t: You may do

```rust
impl RouteType {
  pub(crate) fn is_unicast(&self) -> bool;
}
```

So you may state that only unicast route to 0.0.0.0/8 is not allowed.

liangwen12year:
@cathay4t it seems that nispor should update the netlink-packet-route version into 0.19.0 ?

cathay4t:
> @cathay4t it seems that nispor should update the netlink-packet-route version into 0.19.0 ?

Done

rust/src/lib/route.rs
```diff
@@ -349,6 +338,16 @@ impl RouteEntry {
         self.destination.as_ref().map(|d| is_ipv6_addr(d.as_str()))
             == Some(true)
     }
+
+    pub(crate) fn is_unicast(&self) -> bool {
+        let route_type = [
+            RouteType::Blackhole,
+            RouteType::Unreachable,
+            RouteType::Prohibit,
+        ];
```
cathay4t: No need to generate this array at runtime whenever this function been invoked. Please place it as `const` to `RouteType`.

cathay4t (changes requested):

rust/src/lib/route.rs
cathay4t: on a second through, why doing deny list here? We should do allowlist. Just check it by `self.route_type == RouteType::Unicast`

liangwen12year: the current enum of Routetype does not have `RouteType::Unicast`, but we can add support for this.

cathay4t: no need to add new. Please use struct level `const`.

liangwen12year: I updated it into:

```
    pub(crate) fn is_unicast(&self) -> bool {
        self.route_type.is_none()
            || u8::from(self.route_type.unwrap()) == RTN_UNICAST
    }
```

cathay4t (approved):

PR #2545, opened 2024-02-07
service: Use content instead of checksum

### LINKED ISSUE #2539: New behavior of `nmstatectl service` breaks backwards compatibility
## Describe the bug
As of the merge of  https://github.com/nmstate/nmstate/commit/dbb98e1d30c727b5e47ea7daacc8f2936ab5a0e0 the behavior of `nmstatectl service` is to store a checksum of the applied configs and not reapply configs that have previously been applied.

This is a breaking change as you can no longer evaluate the contents of the `.applied` file to determine what the current running state should be. The semantics of the behavior of the command have fundamentally changed as well as the contents/format of the files involved in the service. This should not have happened in a patch release (behavior is changed between 2.2.21 and 2.2.22), typically a change of this nature would be implemented in a manner that preserves backwards compatibility and would be in a minor release and removal of backwards compatibility would be noted as a coming deprecation/breaking change to be subsequently included in a major release.

## Nmstate version
- Version 2.2.23

## NetworkManager version
- Version 1.44.0

## Platform
- Platform: rhel 9.3

## How reproducible
- 100%

## Steps to Reproduce
1. create a file `/etc/nmstate/somechanges.yml` with some changes to be made
2. run `sudo nmstatectl service`
3. observe that:
    1. `/etc/nmstate/somechanges.yml` is still present unlike in previous versions
    2. `/etc/nmstate/somechanges.applied` file contains a sha256 digest instead of being a full copy of the applied change as in previous versions
4. any software integration that depended on the previous documented behavior is now broken in unanticipated ways

### Additional context
The motivation for the change is documented as `At some envs like openshift machine-config-operator moving configuration
files can break in place upgrades` which acknowledges that the existing behavior exists and the change fundamentally changes how this command behaves. It is inappropriate and unexpected that a downstream consumer of the nmstate failing to work correctly within the documented behavior of the software would be enough to motivate a breaking change in a patch release that is subsequently pushed out to a stable release of an enterprise OS distribution. @qinqon as an employee of Redhat  I'm surprised to see that you signed off on this change

Also, it should be noted that the accepted change does not include any update to the manpage for `nmstatectl` which still says:
```
       service
              Apply all network state files ending with .yml in specified( default: /etc/nmstate) folder.   The  applied  network
              state file will be renamed with postfix .applied to prevent repeated applied on next run.
```
and [the changelog ](https://github.com/nmstate/nmstate/blob/base/CHANGELOG) fails to not that any breaking changes were included in the release instead declaring the changed behavior as a "bugfix" even though the original behavior was documented and behaved exactly as documented. Typically a "bug" is a behavior that varies from the expected and documented behavior.

qinqon:
Using checksum for the .applied file breaks backward compatibility and introduce at new crypto library with possible CVEs.

This changes replace that by copying the content instead of calculating the checksum.

Closes https://github.com/nmstate/nmstate/issues/2539

### DIFF
--- a/doc/nmstate.service.8.in
+++ b/doc/nmstate.service.8.in
@@ -9,8 +9,8 @@ nmstate\&.service
 nmstate\&.service invokes \fBnmstatectl service\fR command which
 apply all network state files ending with \fB.yml\fR in
 \fB/etc/nmstate\fR folder.
-The applied network state file sha256 digest will be stored at \fB.applied\fR
-file to prevent repeated applied on next service start.
+The applied network state file will be copied to new file with \fB.applied\fR
+postfix to prevent repeated applied on next service start.
 .SH BUG REPORTS
 Report bugs on nmstate GitHub issues <https://github.com/nmstate/nmstate>.
 .SH COPYRIGHT
--- a/rust/src/cli/Cargo.toml
+++ b/rust/src/cli/Cargo.toml
@@ -28,7 +28,6 @@ ctrlc = { version = "3.2.1", optional = true }
 uuid = { version = "1.1", features = ["v4"] }
 chrono = "0.4"
 nispor = { version = "1.2", optional = true }
-ring = "0.17.7"
 
 [features]
 default = ["query_apply", "gen_conf", "gen_revert"]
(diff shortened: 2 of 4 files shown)


tyll:
This is currently missing the update to the man page.

cathay4t:
I have amended the patch to include manpage updates.

cathay4t (approved):

PR #2549, opened 2024-02-13
controller: Fix incorrect ports name matching when validating ports

liangwen12year:
Resolves: https://issues.redhat.com/browse/RHEL-1414


cathay4t (changes requested):
Please do not modify existing test case, they might be for other use case.

Meanwhile, commit message should explains:
 * What's the problem
 * Root cause
 * How did you fix it.

liangwen12year:
> Please do not modify existing test case, they might be for other use case.
> 
> Meanwhile, commit message should explains:
> 
> * What's the problem
> * Root cause
> * How did you fix it.

I added the unit test case and modified the commit message, thanks.

cathay4t (approved):
LGTM. Thanks!

PR #2550, opened 2024-02-15
Revert bad words reintroduced in the project

ffmancera:
Master and slave were deprecated in nmstate-0.4.1 for first time. They were supported during the whole nmstate-1 lifetime. When we branched out for nmstate-2 we decided to drop the support of these words.

Now they are back. If we must enforce our own decisions.

### DIFF
--- a/rust/src/lib/iface.rs
+++ b/rust/src/lib/iface.rs
@@ -736,15 +736,6 @@ impl Interface {
             iface.change_port_name(org_port_name, new_port_name);
         }
     }
-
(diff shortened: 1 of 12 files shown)


cathay4t:
I am OK on removing them. Please also get consent from OpenShift team(OVN, installer, bare-metal network)

cathay4t (approved):
LGTM. I will use downstream patch for RHEL 9.2 for OCP upgrade compatibility.

PR #2557, opened 2024-02-21
Static routes auto ip

ihuguet:
NetworkManager doesn't add the routes until the interface gets at least
one IP. If static routes and auto IP are configured, nmstate fails the
verification if the DHCP request hasn't completed because NM hasn't
added the route yet.

However, it's very difficult to understand the reason of the failure for a user,
let's add a warning message explaining it.

Resolves: https://issues.redhat.com/browse/RHEL-15086

### DIFF
--- a/rust/src/lib/query_apply/route.rs
+++ b/rust/src/lib/query_apply/route.rs
@@ -1,6 +1,9 @@
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::{ErrorKind, MergedRoutes, NmstateError, RouteEntry, Routes};
+use crate::{
+    ErrorKind, InterfaceType, Interfaces, MergedRoutes, NmstateError,
+    RouteEntry, Routes,
+};
 
 impl MergedRoutes {
     fn routes_for_verify(&self) -> Vec<RouteEntry> {
@@ -39,6 +42,7 @@ impl MergedRoutes {
         &self,
         current: &Routes,
         ignored_ifaces: &[&str],
+        current_ifaces: &Interfaces,
     ) -> Result<(), NmstateError> {
         let mut cur_routes: Vec<&RouteEntry> = Vec::new();
         if let Some(cur_rts) = current.config.as_ref() {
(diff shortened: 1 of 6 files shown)


rust/src/lib/unit_tests/route.rs
```diff
@@ -432,3 +440,81 @@ fn test_route_matching_empty_via_with_none() {
     assert!(!absent_route.is_match(&not_match_route));
     assert!(!absent_route.is_match(&match_route));
 }
+
+#[test]
+fn test_routes_delayed_by_nm() {
+    let route4: RouteEntry = serde_yaml::from_str(
+        r"
+        destination: 191.0.2.1
```
cathay4t: `192.0.2.1`

rust/src/lib/query_apply/route.rs
```diff
@@ -80,36 +84,84 @@ impl MergedRoutes {
...
+                    rt = &rt2
                 }
-                if !cur_routes
-                    .as_slice()
-                    .iter()
-                    .any(|cur_rt| route_type_rt.is_match(cur_rt))
-                {
+
+                if !cur_routes.iter().any(|cur_rt| rt.is_match(cur_rt)) {
+                    if is_route_delayed_by_nm(rt, current_ifaces) {
```
cathay4t, line 99: Instead of iterate the current interfaces for every miss match route, please store the interfaces with auto ip enabled into a `HashSet` to save the repeat search.

rust/src/lib/unit_tests/route.rs
```diff
@@ -432,3 +440,81 @@ fn test_route_matching_empty_via_with_none() {
...
+
+    assert!(is_route_delayed_by_nm(&route4, &current_ifaces));
+    assert!(is_route_delayed_by_nm(&route6, &current_ifaces));
+}
+
+#[test]
+fn test_routes_not_delayed_by_nm() {
+    let route4: RouteEntry = serde_yaml::from_str(
+        r"
+        destination: 191.0.2.1
```
cathay4t: `192.0.2.1`

cathay4t (changes requested):
Also need integration test case to make sure use facing API behave as expected(still have verification error).

cathay4t (dismissed):
LGTM

cathay4t (approved):

PR #2572, opened 2024-02-29
base_iface: Add driver

mkowalski:
This commits adds a `driver` field to the base interface. It is meant to provide an information about the driver used by the specific interface if available. As some interfaces do not have any driver assigned (e.g. loopback), this field will not be always present.

As per https://docs.kernel.org/admin-guide/sysfs-rules.html, we are using sysfs and read the "driver"-link to get what's needed.


mkowalski:
/close

Implementation in nispor via https://github.com/nispor/nispor/pull/265

PR #2577, opened 2024-03-04
Route: add support for cwnd attribute

ihuguet:
Add support to define the route's clamp for the TCP congestion window (route's attribute name: cwnd).

According to `man ip-route`: It is ignored if the lock flag is not used. We have checked in the kernel that, indeed, setting cwnd without the lock flag is useless. Because of this, nmstate will set the lock flag automatically because the user's intention is clear enough just defining 'cwnd'.

Resolves: https://issues.redhat.com/browse/RHEL-19409

### DIFF
--- a/rust/src/lib/route.rs
+++ b/rust/src/lib/route.rs
@@ -180,6 +180,9 @@ pub struct RouteEntry {
     /// Serialize and deserialize to/from `route-type`.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub route_type: Option<RouteType>,
+    /// Congestion window clamp
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub cwnd: Option<u32>,
 }
 
 #[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq)]
@@ -264,12 +267,15 @@ impl RouteEntry {
         if self.route_type.is_some() && self.route_type != other.route_type {
             return false;
         }
+        if self.cwnd.is_some() && self.cwnd != other.cwnd {
+            return false;
+        }
         true
     }
 
     // Return tuple of (no_absent, is_ipv4, table_id, next_hop_iface,
-    // destination, next_hop_addr, weight)
-    fn sort_key(&self) -> (bool, bool, u32, &str, &str, &str, u16) {
+    // destination, next_hop_addr, weight, cwnd)
+    fn sort_key(&self) -> (bool, bool, u32, &str, &str, &str, u16, u32) {
         (
             !matches!(self.state, Some(RouteState::Absent)),
             !self
@@ -284,6 +290,7 @@ impl RouteEntry {
             self.destination.as_deref().unwrap_or(""),
             self.next_hop_addr.as_deref().unwrap_or(""),
             self.weight.unwrap_or_default(),
+            self.cwnd.unwrap_or_default(),
         )
     }
 
@@ -332,6 +339,14 @@ impl RouteEntry {
                             .to_string(),
                     ));
                 }
+            }
+        }
+        if let Some(cwnd) = self.cwnd {
+            if cwnd == 0 {
+                return Err(NmstateError::new(
+                    ErrorKind::InvalidArgument,
+                    "The value of 'cwnd' cannot be 0".to_string(),
+                ));
             }
         }
         Ok(())
@@ -401,6 +416,9 @@ impl std::fmt::Display for RouteEntry {
         }
         if let Some(v) = self.weight {
             props.push(format!("weight: {v}"));
+        }
+        if let Some(v) = self.cwnd {
+            props.push(format!("cwnd: {v}"));
         }
 
         write!(f, "{}", props.join(" "))
--- a/rust/src/lib/route.rs
+++ b/rust/src/lib/route.rs
@@ -233,6 +233,7 @@ impl RouteEntry {
         matches!(self.state, Some(RouteState::Absent))
     }
 
+    /// Whether the desired route (self) matches with another
     pub(crate) fn is_match(&self, other: &Self) -> bool {
         if self.destination.as_ref().is_some()
             && self.destination.as_deref() != Some("")
--- a/rust/src/lib/unit_tests/route.rs
+++ b/rust/src/lib/unit_tests/route.rs
@@ -580,3 +580,35 @@ fn test_route_cwnd_is_match() {
     assert!(!desired_route.is_match(&not_match_route_2));
     assert!(desired_route.is_match(&match_route));
 }
+
+#[test]
+fn test_route_without_options_is_match_with_any() {
+    let desired_route: RouteEntry = serde_yaml::from_str(
+        r#"
+        destination: "192.0.2.1"
+        "#,
+    )
+    .unwrap();
+
+    let not_match_route: RouteEntry = serde_yaml::from_str(
+        r#"
+        destination: "192.0.2.100"
+        metric: 1
+        table-id: 2
+        cwnd: 3
+        "#,
+    )
+    .unwrap();
+    let match_route: RouteEntry = serde_yaml::from_str(
+        r#"
+        destination: "192.0.2.1"
+        metric: 1
+        table-id: 2
+        cwnd: 3
+        "#,
+    )
+    .unwrap();
+
+    assert!(!desired_route.is_match(&not_match_route));
+    assert!(desired_route.is_match(&match_route));
+}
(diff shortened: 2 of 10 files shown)


ihuguet:
@cathay4t I have maintained the `tcp_metric_lock_flags` in `RouteEntry`. Otherwise we would need to consider "cwnd=20 lock-cwnd=false" as "cwnd=None", thus having inaccurate information about the route in nmstate. I fear that this can lead to future bugs.

Other option might be to use `cwnd: Option<(u32, bool)>` or a new `enum TcpMetric { None, Locked(u32), Unlocked(u32) }` that can be reused for other TCP metrics values that we might add in the future (we already have some requests).

rust/src/lib/route.rs
```diff
@@ -232,42 +241,56 @@ impl RouteEntry {
         matches!(self.state, Some(RouteState::Absent))
     }
 
-    pub(crate) fn is_match(&self, other: &Self) -> bool {
+    /// Whether the desired route (self) matches with an existing route (current)
+    pub(crate) fn is_match(&self, current: &Self) -> bool {
```
cathay4t: The `is_match` is not only for verification. It is also used for applying route with `state: absent`. So the second argument is not current, it should be `other`.

rust/src/lib/route.rs
```diff
@@ -14,6 +14,9 @@ use crate::{
 const DEFAULT_TABLE_ID: u32 = 254; // main route table ID
 const LOOPBACK_IFACE_NAME: &str = "lo";
 
+// TCP metric lock flags
+pub const TCP_METRICS_LOCK_CWND: u32 = 1 << 7;
```
cathay4t: Please change 7 to `RTAX_CWND`, so future developer know what kernel constant to look for.

rust/src/lib/route.rs
```diff
@@ -179,6 +182,12 @@ pub struct RouteEntry {
     /// Serialize and deserialize to/from `route-type`.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub route_type: Option<RouteType>,
+    /// Congestion window clamp
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub cwnd: Option<u32>,
+    /// TCP metric lock flags. At the moment, they are for internal use only.
+    #[serde(skip)]
+    pub(crate) tcp_metric_lock_flags: Option<u32>,
```
cathay4t: Even though I think this lock flags is kernel internal implementation on RCU. But since it require userspace to set it, I am OK to mark it as `pub(crate)` and expose or remove in the future.

rust/src/lib/unit_tests/route.rs
```diff
@@ -432,3 +433,146 @@ fn test_route_matching_empty_via_with_none() {
...
+fn test_route_cwnd_zero_invalid() {
+    let route = serde_yaml::from_str::<RouteEntry>(
+        r"
+        cwnd: 0
+        ",
+    );
+
+    let result = route.expect("Deserialization failed").sanitize();
+    assert!(result.is_err());
+    assert_eq!(result.unwrap_err().msg(), "The value of 'cwnd' cannot be 0");
```
cathay4t: Error message content is not API. Not need to assert it.

cathay4t (changes requested):
Need integration test case indicate we can:
 1. Remove `cwnd` route only.
 2. Modify existing route's `cwnd` by `absent` and `append`.

tests/integration/route_test.py
```diff
@@ -807,6 +807,32 @@ def test_add_and_remove_ipv4_link_local_route(eth1_static_gateway_dns):
...
+            Route.NEXT_HOP_INTERFACE: "eth1",
+            Route.DESTINATION: IPV4_TEST_NET1,
+            Route.NEXT_HOP_ADDRESS: IPV4_ADDRESS1,
+            Route.CWND: 20,
+        },
+        {
+            Route.NEXT_HOP_INTERFACE: "eth1",
+            Route.DESTINATION: IPV6_TEST_NET1,
+            Route.NEXT_HOP_ADDRESS: IPV6_GATEWAY1,
+            Route.CWND: 20,
```
liangwen12year, line 824: We may need integration tests changing cwnd value and remove routes with matching cwnd value.
And also testing setting the cwnd value with route type defined. Because  I think that cwnd is usually applied to the unicast route, maybe we can raise warning when setting the cwnd value with non-unicast route (blackhole, unreachable etc).

rust/src/lib/unit_tests/route.rs
```diff
@@ -544,3 +544,35 @@ fn test_route_lock_cwnd_is_match() {
...
+        cwnd: 3
+        "#,
+    )
+    .unwrap();
+    let match_route: RouteEntry = serde_yaml::from_str(
+        r#"
+        destination: "192.0.2.1"
+        metric: 1
+        table-id: 2
+        cwnd: 3
```
liangwen12year, line 571: `is_match` is called when deleting the routes, can you also add a unit test matching the cwnd value ? 

```
    let desired_route: RouteEntry = serde_yaml::from_str(
        r#"
        cwnd: 3
        "#,
    )
    .unwrap();
    ```
    assert!(desired_route.is_match(&match_route));

File: rust/src/lib/route.rs (hunk 3 above)
Comment (liangwen12year) [reply rust/src/lib/route.rs]: I actually would like to expose it in `nmstatectl  show` as `lock-cwnd: true` to be consistent with the behavior of NM and ip route. What do you think?


```
[root@fcvm rust]# nmcli c show eth1 | grep lock
ipv4.routes:                            { ip = 198.51.100.0/24, nh = 192.0.2.1, mt = 150 cwnd=20 lock-cwnd=true table=254 }
[root@fcvm rust]# ip -4 route
198.51.100.0/24 via 192.0.2.1 dev eth1 proto static metric 150 cwnd lock 20 
 ```

ihuguet: I don't think it's a good idea. In order to set `cwnd` and not being useless, you need to set `lock-cwnd` too. I see it as a useless step that some users might not be aware of, so automatically setting `lock-cwnd` when the user sets `cwnd` in the yaml file seems a good idea to me.

Showing it when executing `nmstatectl show` doesn't sound bad, at first, but I think it might be confusing to show a field that you cannot set with `apply`.

tests/integration/route_test.py
liangwen12year: Since kernel and NM do not complain on setting cwnd value for non-unitcast route, maybe just ingore my comment on raise warning when setting the cwnd value with non-unicast route.

ihuguet:
Added tests for the cases "remove" and "remove and add".

ffmancera (approved):
LGTM!

liangwen12year:
> Added tests for the cases "remove" and "remove and add".

Thanks, the integration tests look good to me.

liangwen12year (approved):

rust/src/lib/route.rs
cathay4t: 1. Nmstate schema is user facing or story driven. There is no point of hiding `tcp_metric_lock_flags` to user, but need backend to set it. The nispor and nm backend should set `cwnd` value property and set it as `None` if lock flag missing.
2. In kernel, this is a bit filed of u32, if we have `tcp_metric_lock_flags`, we might have 10+ properties for this single flags in the future.

cathay4t (changes requested):
Please remove `tcp_metric_lock_flags` from top level code.

tests/integration/route_test.py
```diff
@@ -807,6 +807,130 @@ def test_add_and_remove_ipv4_link_local_route(eth1_static_gateway_dns):
...
+            },
+        }
+    )
+
+    absent_routes = [{Route.CWND: 20, Route.STATE: Route.STATE_ABSENT}]
+    libnmstate.apply({Route.KEY: {Route.CONFIG: absent_routes}})
+
+    expected_routes = []
+    cur_state = libnmstate.show()
+    assert_routes(expected_routes, cur_state)
```
cathay4t: The `assert_routes()` does nothing against an empty array. You need to check the routes are removed manually.

cathay4t (changes requested):

tests/integration/route_test.py
```diff
@@ -807,6 +808,128 @@ def test_add_and_remove_ipv4_link_local_route(eth1_static_gateway_dns):
...
+    ]
+    libnmstate.apply({Route.KEY: {Route.CONFIG: routes}})
+
+    expected_routes = routes[1:]
+    cur_state = libnmstate.show()
+    assert_routes(expected_routes, cur_state)
+
+
+@pytest.mark.tier1
+def test_route_cwnd_without_lock_not_match(eth1_up):
```
cathay4t: `test_route_cwnd_without_lock_means_cwnd_none`

cathay4t (changes requested):
Just a nit on test case naming.

Please squash into single commit.

cathay4t (approved):

PR #2586, opened 2024-03-07
Added alias for apply and show commands

saleha-zf:
following lines of code were added to create aliases for "apply" and "show" on cli

        //setting alias a=> apply, s => show
        .subcommand(clap::SubCommand::with_name("a").alias("apply"))
        .subcommand(clap::SubCommand::with_name("s").alias("show"));


cathay4t (changes requested):
This is incorrect. We already has `apply` and `show` command defined.

And your patch cannot even pass the compile!

Please try harder.

Issue #2588, opened 2024-03-08
Deprecate `PrettyState` in python API


cathay4t:
The `PrettyState` is only used for test codes.

1. Mark `PrettyState` as deprecated in Python API, providing message indicate `libnmstate.show()` output is already sorted.
2. Remove test cases regarding `PrettyState`

jona42-ui:
hi @cathay4t  kindly 
 assign me this to avoid duplication

ffmancera:
Fixed!

jona42-ui:
thanks too 
i need another task @cathay4t @ffmancera

PR #2591, opened 2024-03-08
(Fix)Deprecate PrettyState in python API

jona42-ui:
Issue: https://github.com/nmstate/nmstate/issues/2588

Description;

The PrettyState is only used for test codes.
    

- [x] Mark PrettyState as deprecated in Python API, providing message indicate libnmstate.show() output is already sorted.
- [x] Remove test cases regarding PrettyState

### DIFF
--- a/rust/src/python/libnmstate/prettystate.py
+++ b/rust/src/python/libnmstate/prettystate.py
@@ -22,6 +22,7 @@ from collections.abc import Sequence
 from copy import deepcopy
 import difflib
 import json
+import warnings
 
 import yaml
 
@@ -72,6 +73,11 @@ def format_desired_current_state_diff(desired_state, current_state):
 class PrettyState:
     def __init__(self, state):
         yaml.add_representer(dict, represent_dict)
+        warnings.warn(
+            "PrettyState class is deprecated;to be removed in future release."
+            "The output of libnmstate.show() is already sorted.",
+            DeprecationWarning,
+        )
         self.state = _sort_with_priority(state)
 
     @property
--- a/tests/integration/linux_bridge_test.py
+++ b/tests/integration/linux_bridge_test.py
@@ -13,7 +13,6 @@ import libnmstate
 from libnmstate.error import NmstateKernelIntegerRoundedError
 from libnmstate.error import NmstateValueError
 from libnmstate.error import NmstateVerificationError
-from libnmstate.prettystate import PrettyState
 from libnmstate.schema import Interface
 from libnmstate.schema import InterfaceIP
 from libnmstate.schema import InterfaceIPv4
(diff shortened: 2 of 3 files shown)


rust/src/python/libnmstate/prettystate.py
```diff
@@ -68,6 +69,9 @@ def format_desired_current_state_diff(desired_state, current_state):
         "{}\n".format(pretty_desired_state, pretty_current_state, diff)
     )
 
+warnings.warn("The PrettyState class is deprecated and will be removed in a future release. "
+              "The output of libnmstate.show() is already sorted.", DeprecationWarning)
+
```
ffmancera: Instead of introducing the warning here, could you place it under the `__init__()` function in `PrettyState` class definition? This way the warning will be displayed when creating an instance of the class.

tests/integration/linux_bridge_test.py
```diff
@@ -614,8 +613,8 @@ def test_pretty_state_port_name_first(
         self, bridge_with_trunk_port_and_native_config
     ):
         current_state = show_only((TEST_BRIDGE0,))
-        pretty_state = PrettyState(current_state)
-        assert VLAN_FILTER_PORT_YAML in pretty_state.yaml
+        pretty_state = show_only((TEST_BRIDGE0,))
+        assert VLAN_FILTER_PORT_YAML in pretty_state
```
ffmancera: The whole test case needs to be droppped. The test is specific to pretty_state that we do not want to test anymore. Please drop the whole `test_pretty_state_port_name_first` test.

ffmancera (changes requested):
There is also a missing place where `PrettyState` needs to be dropped. `tests/integration/ovs_test.py`

ffmancera:
Hi! Instead of creating multiple commits could you unify all the changes into a single one? In addition, instead of doing merges to update the branch, please do rebases. The git history will be more clear

jona42-ui:
cleaning this up, sorry!

ffmancera:
It looks much better but the commit message it reflects messages from other commits. I guess this is because you squashed the commits. Please, adjust the commit message to reflect only the relevant information of this task :)

jona42-ui:
updated, thanks

ffmancera (approved):

ffmancera:
LGTM! Let's wait for CI results before merging it

ffmancera:
Whoops, the linter is failing. Could you follow the formatting suggestion? https://github.com/nmstate/nmstate/actions/runs/8202211120/job/22432495995?pr=2591#step:3:436

```
 class PrettyState:
     def __init__(self, state):
         yaml.add_representer(dict, represent_dict)
-        warnings.warn("The PrettyState class is deprecated and will be removed in a future release. "
-                      "The output of libnmstate.show() is already sorted.", DeprecationWarning)
+        warnings.warn(
+            "The PrettyState class is deprecated and will be removed in a future release. "
+            "The output of libnmstate.show() is already sorted.",
+            DeprecationWarning,
+        )
         self.state = _sort_with_priority(state)
```

jona42-ui:
the linter is quite annoying!

ffmancera:
> the linter is quite annoying!

It might be but it is important to take care of the style used when coding. An unified and clean style leads to a codebase easier to read and understand.

There is one more lint error, you are almost there!

```
 +
 class PrettyState:
     def __init__(self, state):
         yaml.add_representer(dict, represent_dict)
         warnings.warn(
             "The PrettyState class is deprecated and will be removed in a future release. "
```

ffmancera:
All good, merging!

cathay4t:
According to https://peps.python.org/pep-0702/  we should use `@warnings.deprecated()`. 

@jona42-ui Can you investigate and fix it in next PR?

jona42-ui:
thanks makes sense

jona42-ui:
can i make a followup PR with only the decorator fix?

PR #2602, opened 2024-03-13
(fix)Use @warnings.deprecated()

jona42-ui:
Issue: https://github.com/nmstate/nmstate/issues/2588

Description;

The PrettyState is only used for test codes.

Mark PrettyState as deprecated in Python API, providing message indicate libnmstate.show() output is already sorted.

According to https://peps.python.org/pep-0702/ we should use @warnings.deprecated()

### DIFF
--- a/rust/src/python/libnmstate/prettystate.py
+++ b/rust/src/python/libnmstate/prettystate.py
@@ -70,6 +70,11 @@ def format_desired_current_state_diff(desired_state, current_state):
     )
 
 
+# pylint: disable=E1101
+@warnings.deprecated(
+    message="PrettyState class is deprecated; libnmstate.show() is sorted.",
+    category=DeprecationWarning,
+)
 class PrettyState:
     def __init__(self, state):
         yaml.add_representer(dict, represent_dict)


cathay4t:
please fix the CI failure on pylint

jona42-ui:
what could cause the failure of the rest of the 
 checks

jona42-ui:
https://github.com/nmstate/nmstate/actions/runs/8260958541/job/22597548460?pr=2602#step:6:461

>  /usr/lib/python3.9/site-packages/libnmstate/prettystate.py:74: in <module>
    ???
      E   AttributeError: module 'warnings' has no attribute 'deprecated'
      + run_exit
      + write_separator TEARDOWN
      + set +x

from here https://peps.python.org/pep-0702/#backwards-compatibility its says there no backward compatibility issues

cathay4t:
It is because RHEL 9 ships python 3.9. And PEP 702 is for python 3.13.

You can either close this PR or use `try except` to use old `warnings.warn`.

This is PR is also wrong as you only add the new decorator without removing old warning meseage

jona42-ui:
> It is because RHEL 9 ships python 3.9. And PEP 702 is for python 3.13.
> 
> You can either close this PR or use `try except` to use old `warnings.warn`.
> 
> This is PR is also wrong as you only add the new decorator without removing old warning meseage

my bad.

jona42-ui:
let me use an exception with the old warnings.warn()

PR #2609, opened 2024-03-19
Implement std::fmt::Debug for NetworkState and its subtypes

beekaz:
Implemented the fmt::Debug trait for the NetworkState struct, 
Created a new impl fmt::Debug block for the NetworkState struct.
Defined the fmt function within the impl block std::fmt::Formatter and returns a std::fmt::Result.
Constructed a debug representation of the NetworkState using the debug_struct method from std::fmt::Formatter.


rust/src/lib/net_state.rs
```diff
@@ -121,6 +126,40 @@ pub struct NetworkState {
...
+        debug_struct.field("hostname", &self.hostname);
+        debug_struct.field("dns", &self.dns);
+        debug_struct.field("rules", &self.rules);
+        debug_struct.field("routes", &self.routes);
+        debug_struct.field("interfaces", &self.interfaces);
+        debug_struct.field("ovsdb", &self.ovsdb);
+        debug_struct.field("ovn", &self.ovn);
+        debug_struct.field("kernel_only", &self.kernel_only);
+    
+        if !self.include_secrets {
```
cathay4t: We will never show secrets in debug print, no need to check this flag.

rust/src/lib/net_state.rs
```diff
@@ -2,9 +2,12 @@
 
 #[cfg(not(feature = "gen_conf"))]
 use std::collections::HashMap;
+use std::fmt;
 
 use serde::{Deserialize, Serialize};
-
+// use core::fmt::Debug;
+// use serde::__private::fmt::Debug;
+// // use std::fmt::Debug;
```
cathay4t, line 9: Remove these.

rust/src/lib/net_state.rs
```diff
@@ -2,9 +2,12 @@
 
 #[cfg(not(feature = "gen_conf"))]
 use std::collections::HashMap;
+use std::fmt;
```
cathay4t: Since this `std::fmt` only been used once. There is no need to import it.

rust/src/lib/net_state.rs
```diff
@@ -121,6 +126,40 @@ pub struct NetworkState {
     pub(crate) memory_only: bool,
 }
 
+impl fmt::Debug for NetworkState {
```
cathay4t: Just use `impl std::fmt::Debug for NetworkState {`

cathay4t (changes requested):
* Please fix the warnings of `cargo fmt` and `cargo clippy` , they are in our CI.
 * Please include unit test case indicate this actual works. The folder is `rust/src/lib/unit_tests`

rust/src/lib/net_state.rs
beekaz: Thanks fixed

cathay4t:
Please fix you code by using `cargo fmt` and `cargo clippy`, that's what `CI / rust_lint` is doing.

beekaz:
Please my actual code passed the cargo fmt and cargo clippy in my local. Its the previous two unwanted commits thats troubling me, which i have tried  reverting but it s not working. Please how can  fix it.

Issue #2589, opened 2024-03-08
Use `SPDX-License-Identifier` to all rust and python API code


cathay4t:
All rust code should have `// SPDX-License-Identifier: Apache-2.0` as first line.

All python code should have `# SPDX-License-Identifier: LGPL-2.1-or-later` as first line.

Meanwhile, please introduce a CI task for this, you may take a look on https://github.com/apache/skywalking-eyes for github action.

saleha-zf:
I'd like to work on this, can i be assigned this? thankyou

ffmancera:
Assigned!

PR #2617, opened 2024-03-20
Create test to check for license header

elomscansio:
This commit adds a continuous integration (CI) task that checks for SPDX
    license headers in source files. The CI task is configured to run
    automatically on every code push or pull request.
    
    The CI task uses a Python script to scan the source code files and ensure that
    they contain the required SPDX license headers. If any files are found without
    the proper license headers, the CI task will fail, indicating a licensing
    compliance issue.
    
    See: #2589


.github/workflows/license_header_check.py
```diff
@@ -0,0 +1,77 @@
...
+import re
+import sys
+import yaml
+from pathlib import Path
+
+
+def recursive_glob(directories, pattern):
+    directory_paths = (Path(directory) for directory in directories)
+    files = (
+        directory_path.glob("**/" + pattern)
```
ihuguet: This `recursive_glob` function is only called once, with pattern="*". That indicates that the `pattern` argument is not needed, just call ` directory_path.glob("**/*")` in this line and remove the `pattern` argument.
Also, rename `recursive_glob` to something more representative after the change like `get_files_recursive` or similar.

.github/workflows/license_header_check.py
```diff
@@ -0,0 +1,77 @@
...
+    for dir in recursive_glob(paths, "*"):
+        for file_path in dir:
+            if file_path.is_dir():
+                continue
+
+            scanned_files += 1
+
+            if any(file_path.match(pattern) for pattern in excludes):
+                excluded_files += 1
+                continue
```
ihuguet: I would move this to the recursive_glob/get_files_recursive function.

.github/workflows/license_header_check.py
```diff
@@ -0,0 +1,77 @@
...
+    has_files_without_header = False
+    files_without_header = 0
+    excluded_files = 0
+    scanned_files = 0
+    files_with_header = 0
+
+    for dir in recursive_glob(paths, "*"):
+        for file_path in dir:
+            if file_path.is_dir():
+                continue
```
ihuguet: Same, I would move this to the recursive_glob/get_files_recursive function.

.github/workflows/license_header_check.py
```diff
@@ -0,0 +1,77 @@
...
+
+            scanned_files += 1
+
+            if any(file_path.match(pattern) for pattern in excludes):
+                excluded_files += 1
+                continue
+
+            with open(file_path, "r") as file:
+                content = file.read()
+                matches = [re.search(pattern, content) for pattern in patterns]
```
ihuguet: The SPDX text must be in the first line of the file, you need to check only the first line.

.github/workflows/license_header_check.py
```diff
@@ -0,0 +1,77 @@
...
+                if not any(matches):
+                    if not has_files_without_header:
```
ihuguet: This variable is not needed. You can do `if files_without_header == 0` here.

ihuguet (changes requested):
Instead of adding a separate .license_header_check.yml, I would put that information inside the Python script, as Python variables, directly. Not sure what others think about that.

Also, this is a test that can be run locally too, there is no need to limit it to be run on Github CI. What I would do: 
- Move the python script to the `tests` folder.
- Delete the file `.github/workflows/license_header_check.yml`
- Create a new target `license_check` in the Makefile
- Also in the Makefile, add license_check to the line `check: rust_check clib_check go_check`
- Modify the file `automation/run_tests.sh` so it can run `make license_check`.

Doing that, the license will be tested both locally and on Github CI.

This is a good work! Well done!

elomscansio:
I have made the suggested changes

tests/license_header_check.py
```diff
@@ -0,0 +1,105 @@
...
+)
+patterns = [re.compile(header) for header in headers]
+
+for file_path in get_included_files():
+
+    with open(file_path, "r") as file:
+        content = file.readline()
+
+        # Read second line if the file is a shell script
+        if f"{file_path}".endswith(".sh"):
```
ihuguet: Maybe this is not enough because not all shell files ends with ".sh". Some doesn't have extension, or a different one. And now that I think about it, Python scripts can also have a shebang line in the first line.

So, maybe, check if the first line starts with "#!", and if so, read the second line.

ihuguet (changes requested):

automation/run-tests.sh
```diff
@@ -364,3 +364,4 @@ if [ $TEST_TYPE != $TEST_TYPE_ALL ] && \
     install_nmstate
 fi
 run_tests
+make license_header_check
```
ihuguet: `automation/run-tests.sh` does not run all the tests always, it allows to select what tests to run. The same way, you don't have to run the licenses check always. Study how run-tests.sh works and put this line in an appropriate place. Maybe we can run it when `$TEST_TYPE == $TEST_TYPE_FORMAT`.

elomscansio:
I have made the changes

tests/license_header_check.py
```diff
@@ -0,0 +1,105 @@
...
+    "------------------------- License Header Check Start -------------------------"
+)
+patterns = [re.compile(header) for header in headers]
+
+for file_path in get_included_files():
+
+    with open(file_path, "r") as file:
+        content = file.readline()
+
+        # Read second line if the file is a shell script
```
ihuguet: Fix this comment: Read second line if the first is a shebang

automation/run-tests.sh
```diff
@@ -76,6 +76,8 @@ function run_tests {
     if [ $TEST_TYPE == $TEST_TYPE_ALL ] || \
        [ $TEST_TYPE == $TEST_TYPE_FORMAT ];then
         exec_cmd "tox -e black"
+        make license_header_check
+
```
ihuguet: Please remove this blank line.

ihuguet:
The script is great, I think.
However, if we add a new test, we have to ensure that it passes because it makes no sense to have a tests constantly failing. Can you add **in a separate commit** all the SPDX lines that are missing in the project?

elomscansio:
I have made the necessary changes

tests/license_header_check.py
```diff
@@ -0,0 +1,105 @@
...
+
+import re
+import sys
+from pathlib import Path
+
+paths = ["rust/src", "tests"]
+
+headers = [
+    "#\\s|\\SSPDX-License-Identifier: LGPL-2.1-or-later",
+    "//\\s|\\SSPDX-License-Identifier: Apache-2.0",
```
ihuguet: The regex are wrong.
```
regex = "#\\s|\\SSPDX-License-Identifier: LGPL-2.1-or-later"
re.match(regex, "# ")  # this will match
re.match(regex, "ASPDX-License-Identifier: LGPL-2.1-or-later")  # this too
```

You need to remove the `|\\S` from both of them.

tests/license_header_check.py
```diff
@@ -0,0 +1,105 @@
...
+for file_path in get_included_files():
+
+    with open(file_path, "r") as file:
+        content = file.readline()
+
+        # Read second line if the first is a shebang
+        if content.startswith("#!"):
+            content = file.readline()
+
+        matches = [re.search(pattern, content) for pattern in patterns]
```
ihuguet: Here you should us `re.match`, not `re.search`, because we want to match it at the beginning of the line, not at any point of the line. Or better, use `re.fullmatch`.

ihuguet (changes requested):
Thanks for the changes.
Regarding "adding all the missing SPDX lines in a separate commit", what I mean is to add the line `# SPDX-License-Identifier: LGPL-2.1-or-later` or `// SPDX-License-Identifier: Apache-2.0` to all the files that doesn't have them. Otherwise, the test will be failing all the time until someone does it, so it is a good moment to do it.

elomscansio:
> Thanks for the changes. Regarding "adding all the missing SPDX lines in a separate commit", what I mean is to add the line `# SPDX-License-Identifier: LGPL-2.1-or-later` or `// SPDX-License-Identifier: Apache-2.0` to all the files that doesn't have them. Otherwise, the test will be failing all the time until someone does it, so it is a good moment to do it.

ok, according to https://github.com/nmstate/nmstate/issues/2589#issue-2175050374 right

elomscansio:
i have added the license header fixing but i need help here https://github.com/nmstate/nmstate/actions/runs/8452081881/job/23151720812?pr=2617#step:3:431 https://github.com/elomscansio/nmstate/blob/930c58db6d41e22dc2990cf25d81f47120d26e30/tests/license_header_check.py#L144

ihuguet:
Oh, I didn't mean to add them automatically! I hope you can revert to the previous version because it was ready!

What I meant is that there are files without the SPDX license comment. We have to **MANUALLY** add them so the CI passes once your new test is added to the main branch!

ihuguet:
Now it's great, contratulations for the good work!
The CI is failing because it tries to merge the `base` branch with yours and find a conflict. Please rebase your changes on top of the `base` branch, solve the conflicts and push again.

elomscansio:
> Now it's great, contratulations for the good work! The CI is failing because it tries to merge the `base` branch with yours and find a conflict. Please rebase your changes on top of the `base` branch, solve the conflicts and push again.

Thanks. 
I've rebase on top of base

ihuguet (approved):

elomscansio (commented):
Ci test are now passing

tests/license_header_check.py
```diff
@@ -0,0 +1,101 @@
...
+
+total_files = 0
+files_with_header = 0
+files_without_header = 0
+excluded_files = 0
+
+
+def get_included_files():
+    global total_files
+    global excluded_files
```
liangwen12year, line 47: Nit, maybe instead of using the global variables, we should return values from functions and pass them as parameters where needed. This will make functions more self-contained and easier to understand. But it more sounds like personal taste.

liangwen12year (approved):

tests/license_header_check.py
elomscansio: or should the test be wrapped with class

liangwen12year:
@elomscansio , there is a conflict when merging it, can you resolve the conflict?

tests/license_header_check.py
cathay4t, line 1: Instead of self-cooked test case, can we use similar stuff like https://github.com/rust-netlink/netlink-packet-route/blob/main/.github/workflows/license.yml ?

cathay4t (changes requested):

tests/license_header_check.py
elomscansio: That was actually the first case here https://github.com/nmstate/nmstate/issues/2589#issue-2175050374, but it doesn't work for us because we are checking for two different types of files, each with a different header

elomscansio:
> @elomscansio , there is a conflict when merging it, can you resolve the conflict?

I have addressed the conflict.

tests/license_header_check.py
elomscansio: I have rechecked and encountered the same issue. See:

https://github.com/elomscansio/nmstate/blob/98f91556b0f85db964628984ccee983e3c5df4bf/rust/src/cli/apply.rs#L1

The file has the correct header, but the `license-eye` did not validate it in the action:

https://github.com/elomscansio/nmstate/actions/runs/8886394723/job/24399751393#step:4:138

I used the same example as https://github.com/elomscansio/netlink-packet-route/actions/runs/8886146058/job/24399050097

Could it be that I'm missing something, which causes the `license-eye` not to match any file, or what?

cathay4t: `apache/skywalking-eyes` support different license for different paths, please check their document and try again.

cathay4t:
I am still in the favor of `apache/skywalking-eyes` over self-cooked license checker because:
 1. Regex is hard to be correct, we might has false alarm or miss validation.
 2. Using `apache/skywalking-eyes` allows to adopt new SPDX format in the future with minimum efforts.

Please provide feedback why you think self-cooked license checker with complex regex is better.

ihuguet:
The only reasons I have in favor of the self-cooked checker are:
1. It works
2. It's done

About the Regex, I don't think they are going to fail in this case because the SPDX text is always the same. They are more risky if you have to parse many different texts, then you can get a corner case that you didn't think of. About adopting new SPDX licences: that will happen only once or twice, or maybe zero times. And it's not hard to add a new line with a Regex for it.

I would be really fine of using skywalking-eyes, but nobody is putting effort on it and, as I say, this solution already works. Probably it's better to merge it, and it's easy to replace it in the future if someone works on it.

It's up to you.

tests/license_header_check.py
elomscansio: `apache/skywalking-eyes` matches when i change the header to `MIT` not so with `Apache-2.0` or `LGPL-2.1-or-later`

elomscansio:
I have been able to use skywalker see https://github.com/nmstate/nmstate/pull/2704.
But i have to create two config file for each header check

elomscansio:
> I have been able to use skywalker see #2704. But i have to create two config file for each header check

The CI now succeed https://github.com/nmstate/nmstate/actions/runs/9549291820/job/26318515086

cathay4t:
Close because author has created new PR #2704 to replace this one.

PR #2618, opened 2024-03-20
 Implement  std::fmt::Debug for NetworkState and its subtypes

beekaz:
Implemented the fmt::Debug trait for the NetworkState struct,
Created a new impl fmt::Debug block for the NetworkState struct.
Defined the fmt function within the impl block std::fmt::Formatter and returns a std::fmt::Result.
Constructed a debug representation of the NetworkState using the debug_struct method from std::fmt::Formatter.
Created Tests to verify the debug output of the NetworkState struct includes the expected fields and that the sensitive information is appropriately hidden.


cathay4t (changes requested):
The unit case you wrote cannot pass. Please fix it before asking for review.

rust/src/lib/unit_tests/net_state.rs
```diff
@@ -23,3 +28,46 @@ fn test_invalid_top_type() {
...
+        no_commit: false,
+        ovn: OvnConfiguration::default(),
+        include_secrets: true,
+        include_status_data: false,
+        running_config_only: true,
+        memory_only: false,
+        kernel_only: false,
+    };
+
+    network_state.hide_secrets();
```
cathay4t: Please remove this line. Even without explicitly call of `hide_secrets()`, debug output of NetworkState should never reveal secrets.

Issue #2452, opened 2023-11-09
Nmstate should ignore interface with `state: unknown` also


cathay4t:
This state should NO_OP:

```yml
---
interfaces:
  - name: eth1
    type: ethernet
    state: unknown
    mtu: 9000
```

divyakelaskar:
Can I work on this? (_the state should `NO_OP` right?_)

cathay4t:
Sure. NO_OP with a warning messge

divyakelaskar:
I read the repo contents thoroughly and realized that currently I don't have a Linux OS on my PC. Thus, I wont be able to contribute to the issue. Sorry for the inconvenience.

EjiroLaurelD:
Hello, my name is Ejiro an outreachy applicant, can i work on this issue?

abigailajohn:
Hello, can I proceed with this issue?

cathay4t:
Both of you can work on it. Even it will cause overlap, but I will review your patch separately and help you to form a good patch.

More detail on this `NO_OP`.

When you run

```bash
echo '
---
interfaces:
  - name: eth1
    type: ethernet
    state: unknown
    mtu: 9000' | sudo nmstatectl apply -
```

nothing should happen, nmstate just ignore the change request on eth1 because its state is unknown, with a warning message printed to STDERR also.

EjiroLaurelD:
Hello @cathay4t , I'm sorry I haven't responded to this project since. I had some issues with my laptop and have thankfully fixed it now.
I am currently encountering this error when I try running the code above. I am not sure how to go from here a
<img width="641" alt="Screenshot 2024-03-12 121513" src="https://github.com/nmstate/nmstate/assets/99752279/d9c00608-f646-4a15-be48-3e3581a42b4c">

EjiroLaurelD:
[...]

PR #2626, opened 2024-03-25
Nmstate should ignore  change request on eth1 because `state: unknown ` 

EjiroLaurelD:
Hello @ihuguet 
Please review my changes for issue #2452
I made some changes to base_iface file and also iface.rs, but the info message didn't show in the output as expected. I have gone through the code base looking out for how to make this work but I am not sure how to go from here. pls help
 am doing wrong. 
Screenshot 2024-03-25 165446

### DIFF
--- a/rust/src/lib/iface.rs
+++ b/rust/src/lib/iface.rs
@@ -188,7 +188,7 @@ impl From<&str> for InterfaceState {
             "up" => Self::Up,
             "down" => Self::Down,
             "absent" => Self::Absent,
-            "ignore" => Self::Ignore,
+            "unknown" => Self::Ignore,
             _ => Self::Unknown,
         }
     }
@@ -548,9 +548,10 @@ impl Interface {
     }
 
     /// Whether interface is marked as ignore.
-    pub fn is_ignore(&self) -> bool {
+    pub fn is_unknown(&self) -> bool {
         self.base_iface().state == InterfaceState::Ignore
     }
+
 
     // Whether desire state only has `name, type, state`.
     pub(crate) fn is_up_exist_config(&self) -> bool {
--- a/rust/test.yml
+++ b/rust/test.yml
@@ -1,0 +1,6 @@
+---
+interfaces:
+  - name: eth1
+    type: ethernet
+    state: unknown
+    mtu: 9000


jona42-ui (changes requested):
there are some commits that creeped in, they are affecting  the readability.
can you clean the commit please

jona42-ui:
and about the patch I would think we are changing the  deserialization logic of the `BaseInterface` rust struct to accept `unknown` state so that it can   handle it appropriately

cc @cathay4t @ihuguet @ffmancera

jona42-ui:
Also @EjiroLaurelD  it would be good practise  for the PR title to be a little more descriptive more less like the issue title.

EjiroLaurelD:
> there are some commits that creeped in, they are affecting  the readability.
> can you clean the commit please

Oh my!... I will clean the commit. Thank you

ihuguet:
The reason why it doesn't work is that you have to map "unknown" to Ignore, not to Unknown (yes, it is quite confusing but the term unknown means something different internally).
And yes, please clean up and leave only the related commits so we can review more.

EjiroLaurelD:
> The reason why it doesn't work is that you have to map "unknown" to Ignore, not to Unknown (yes, it is quite confusing but the term unknown means something different internally). And yes, please clean up and leave only the related commits so we can review more.

ohhh I see... I was confused and couldn't figure out what i was doing wrong. I have cleaned up the commits after struggling with the process a bit. It is exciting that i am also learning so much more about git through the process. 

I will make the new changes and squash the commits into one once I am done. Thank you so much

ihuguet:
It doesn't build because of a typographic error. Even if you fix that, it won't do what we have requested: you have to map the string "unknown" to InterfaceState::Ignore, not the string "ignore" to InterfaceState::Unknown.

Please, it is OK if the PRs are not perfect and it is OK to ask if you have doubts if you are stalled. But you MUST try to compile and test that it does what it's supposed to do before requesting a review.

rust/src/lib/iface.rs
```diff
@@ -188,7 +188,7 @@ impl From<&str> for InterfaceState {
             "up" => Self::Up,
             "down" => Self::Down,
             "absent" => Self::Absent,
-            "ignore" => Self::Ignore,
+            "ignore" => Self::unknown,
             _ => Self::Unknown,
         }
     }
```
jona42-ui, line 194: @EjiroLaurelD can you take a look at how the `BaseInterface` struct is defined here
https://github.com/nmstate/nmstate/blob/base/rust/src/lib/ifaces/base.rs

jona42-ui: @ihuguet  made a very important input here:
https://github.com/nmstate/nmstate/issues/2452#issuecomment-2010133918

you can take a look at it. it should make an ultimate solution to this

EjiroLaurelD:
> It doesn't build because of a typographic error. Even if you fix that, it won't do what we have requested: you have to map the string "unknown" to InterfaceState::Ignore, not the string "ignore" to InterfaceState::Unknown.
> 
> Please, it is OK if the PRs are not perfect and it is OK to ask if you have doubts if you are stalled. But you MUST try to compile and test that it does what it's supposed to do before requesting a review.

Thank you so much for your patience and guidance.

I would carefully review all the comments on this issue again and work it out from there.

rust/src/lib/iface.rs
cathay4t: The `state: ignore` is very important for many use cases. Please add new entry instead of change it to unknown.

rust/test.yml
```diff
@@ -0,0 +1,6 @@
+---
+interfaces:
+  - name: eth1
+    type: ethernet
+    state: unknown
+    mtu: 9000
```
cathay4t, line 6: This is not how we test stuff. There is no magic tool using this file.

cathay4t:
This has been stale for a while. If you still want to work on it, please reply with comment indicate plan.

cathay4t:
Closing due to stale and proposed patch is wrong(not possible to amend).

PR #2636, opened 2024-04-01
integ: Dump applied and showed states

qinqon:
This change dumps the network configuration involved at integration test so other projects can check their implementations with them.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -3,6 +3,10 @@
 import logging
 import os
 import subprocess
+import yaml
+import tempfile
+
+from pathlib import Path
 
 import pytest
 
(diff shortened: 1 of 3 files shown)


tests/integration/conftest.py
```diff
@@ -169,6 +186,51 @@ def _get_osname():
...
+        yaml.dump(state, outfile)
+
+
+def _custom_apply_with_dump_state(
+    desired_state,
+    kernel_only=False,
+    verify_change=True,
+    save_to_disk=True,
+    commit=True,
+    rollback_timeout=60,
```
cathay4t:
```python
def _custom_apply_with_dump_state(desired_state, *args, **kwargs):
    result = LIBNMSTATE_APPLY(desired_state, *args, **kargs)
    _dump_state(desired_state)
    return result
```

cathay4t (changes requested):

cathay4t (approved):

PR #2637, opened 2024-04-01
Nmstate should ignore interface with state: unknown #2452

### LINKED ISSUE #2452: Nmstate should ignore interface with `state: unknown` also

abigailajohn:
This PR adds a custom Serde deserializer in the Nmstate 
Rust project to treat interface states labeled as **'unknown'** 
as **'InterfaceState::Ignore'** and logs a warning. 

This ensures the application remains stable and avoids 
unnecessary errors when encountering an unexpected 
interface states.

**Changes:**
- Implemented custom deserialization for **'unknown'** 
interface states.
- I added log warnings to notify users that the state is 
unknown and would be ignored. 

**Testing:**
- Compiled the project successfully with cargo build.
- Verified through manual test that the 'unknown' 
state is logged correctly.

Resolves [#2452](https://github.com/nmstate/nmstate/issues/2452)

### DIFF
--- a/rust/src/lib/iface.rs
+++ b/rust/src/lib/iface.rs
@@ -3,6 +3,9 @@
 use std::collections::HashSet;
 
 use serde::{Deserialize, Deserializer, Serialize};
+use serde::de::Visitor;
+use std::fmt;
+use serde::de;
 
 use crate::{
     BaseInterface, BondInterface, DummyInterface, ErrorKind, EthernetInterface,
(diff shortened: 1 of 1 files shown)


cathay4t (changes requested):
1. Please fix the warnings in `cargo fmt` or `cargo clippy`.
2. Implement `Deserialize` is overkill, please just amend `get_ignored_ifaces()` function or just `Interface::is_ignore()` function.
3. Unit test case required.

cathay4t (changes requested):
This should not be done at `Deserialize` stage. User might forgot to mention interface type in desire state, but they are expecting nmstate to use existing interface type for it.

So this changing `state: unknown` to `state: ignore` should be handled after `resolve_unknown_ifaces()`.

cathay4t:
It has been  a while since my review, if you still have time and interest on this, please let us know.

PR #2639, opened 2024-04-01
route: Normalize route table of current route when determining route removed

liangwen12year:
Normalizing the `None` route table ID into default table ID when
creating the merged routes will cause the routes difference in new NM
conncetion and activated NM connection when applying the same desired
state the second time, as the result, the activated connection will be
dactivated first and activated again. This is troublesome, because the
routes added by iproute2 utility will be deleted by mistake. To fix
that, normalize the route table of the current routes in the activated
NM connection before comparing the difference between current routes
and new routes.

Resolves: https://issues.redhat.com/browse/RHEL-29241

### DIFF
--- a/tests/integration/route_test.py
+++ b/tests/integration/route_test.py
@@ -1834,6 +1834,47 @@ def test_absent_route_with_invalid_empty_destination(static_eth1_with_routes):
 
 
 @pytest.mark.tier1
+def test_preserve_unmanaged_routes(eth1_static_ip):
+    routes = [
+        {
+            Route.NEXT_HOP_INTERFACE: "eth1",
+            Route.DESTINATION: IPV4_DEFAULT_GATEWAY,
+            Route.NEXT_HOP_ADDRESS: IPV4_ADDRESS2,
+            Route.WEIGHT: 1,
+        },
+        {
+            Route.NEXT_HOP_INTERFACE: "eth1",
+            Route.DESTINATION: IPV4_DEFAULT_GATEWAY,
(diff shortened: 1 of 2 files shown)


tests/integration/route_test.py
```diff
@@ -1834,6 +1834,47 @@ def test_absent_route_with_invalid_empty_destination(static_eth1_with_routes):
         )
 
 
+@pytest.mark.tier1
+def test_add_ip_routes_reapply(eth1_static_ip):
```
cathay4t: Reapply is NM terminology, please use `test_preserve_unmanaged_routes()`

liangwen12year: I changed it to test_preserve_unmanaged_routes.

liangwen12year:
@cathay4t , @ffmancera , if no other objection, this PR can be merged ?

tests/integration/route_test.py
```diff
@@ -1834,6 +1834,47 @@ def test_absent_route_with_invalid_empty_destination(static_eth1_with_routes):
...
+        {
+            Route.KEY: {Route.CONFIG: routes},
+        }
+    )
+    cur_state = libnmstate.show()
+    assert_routes(routes, cur_state)
+
+    cmdlib.exec_cmd(
+        f"ip route add {IPV4_TEST_NET1} via {IPV4_ADDRESS1} "
+        "dev eth1 proto keepalived metric 50".split(),
```
cathay4t: The keepalived route is supported by nmstate, it does not count as unmanaged route

cathay4t (changes requested):

cathay4t (approved):

PR #2642, opened 2024-04-04
networking: Sort route rules retrieved from nispor 

### LINKED ISSUE #1416: Sort the route rule before showing via `nmstatectl` or `libnmstate.show()`
The route rule should be sorted in:
 * IPv6 is preferred than IPv4.
 * Smaller numbered priority is preferred than bigger number.

jona42-ui:
The` Ord` trait implementation of `RouteRuleEntry` already includes a sorting function. However, the order provided by nispor_retrieve() does not guarantee the desired sorting based on priority, which can lead to inconsistencies, particularly when the kernel prioritizes certain rules over others.

This commit addresses the task by adding explicit sorting of route rules returned by `nispor_retrieve()`. The `sort_unstable_by_key() `function is used to ensure stable sorting, with priority taking precedence.

Unit test cases have been added to verify that the sorting function works as expected. These tests validate the ordering of route rules based on priority, ensuring that the sorting implementation meets the requirements specified in the task.

Resolves: https://github.com/nmstate/nmstate/issues/1416

### DIFF
--- a/rust/src/lib/nispor/route_rule.rs
+++ b/rust/src/lib/nispor/route_rule.rs
@@ -83,6 +83,8 @@ pub(crate) fn get_route_rules(
         };
         rules.push(rule);
     }
+
+    rules.sort_unstable_by_key(|rule| rule.priority.unwrap_or(0));
     ret.config = Some(rules);
 
     ret
--- a/rust/src/lib/unit_tests/route_rule.rs
+++ b/rust/src/lib/unit_tests/route_rule.rs
@@ -320,3 +320,27 @@ fn test_route_rule_auto_priority_increasing_from_empty() {
     );
     assert_eq!(rules[2].priority, Some(30002));
 }
+
+#[test]
+fn test_route_rule_sorting() {
+    let mut rules = vec![
+        RouteRuleEntry {
+            priority: Some(200),
+            ..Default::default()
+        },
+        RouteRuleEntry {
+            priority: Some(100),
+            ..Default::default()
+        },
+        RouteRuleEntry {
+            priority: Some(300),
+            ..Default::default()
+        },
+    ];
+
+    rules.sort_unstable();
+
+    assert_eq!(rules[0].priority, Some(100));
+    assert_eq!(rules[1].priority, Some(200));
+    assert_eq!(rules[2].priority, Some(300));
+}


rust/src/lib/nispor/route_rule.rs
```diff
@@ -83,6 +83,8 @@ pub(crate) fn get_route_rules(
         };
         rules.push(rule);
     }
+
+    rules.sort_unstable_by_key(|rule| rule.priority.unwrap_or(0));
```
liangwen12year, line 87: When routing rule priority is not specified by the user, we can not just blindly assign 0 to the rule for sorting purpose, if you do `ip -4 rule show`, you would notice that rule lookup local table has priority 0.   Instead, we should prefer the rule with priority specified during sorting, and the lower the better. And those rules without the priority specified should be sorted into last row during the `nmstatectl show`. @cathay4t , correct me if my idea is wrong.     

```
$ ip -4 rule
0:	from all lookup local
32766:	from all lookup main
32767:	from all lookup default
```

rust/src/lib/unit_tests/route_rule.rs
```diff
@@ -320,3 +320,27 @@ fn test_route_rule_auto_priority_increasing_from_empty() {
...
+            priority: Some(100),
+            ..Default::default()
+        },
+        RouteRuleEntry {
+            priority: Some(300),
+            ..Default::default()
+        },
+    ];
+
+    rules.sort_unstable();
```
liangwen12year, line 341: What is the purpose of this unit test? The test does not call any internal function, I think that what is missing in this PR is to have a integration test which has the state with several route rule entries, then the test applies the desired state and retrieves the state with `libnmstate.show()`  to see if the route rules are properly sorted.

liangwen12year:
Also, the commit message is not properly formatted,  https://github.com/nmstate/nmstate/blob/base/CONTRIBUTING.md#write-a-good-commit-message, please follow the instructions in the link to limit the subject line to 50 characters and wrap the body at 72 characters.

rust/src/lib/nispor/route_rule.rs
cathay4t: This PR is for route rule from nispor where kernel always has priority set.

rust/src/lib/unit_tests/route_rule.rs
cathay4t: @jona42-ui integration test might be too complex. Once this PR is in good shape, I can write integration test case for this PR.

ffmancera:
@jona42-ui hey what is the situation for this PR? Do you need any help with them? We can talk about it on Matrix if that works better for you.

jona42-ui:
> @jona42-ui hey what is the situation for this PR? Do you need any help with them? We can talk about it on Matrix if that works better for you.

ooh sorry
I was taken up by another job
I happy to be unassigned as I have not yet gotten some time

cathay4t:
Closing as stale.

PR #2659, opened 2024-04-25
ipsec: support libreswan rightcert property

ihuguet:
This property allows to specify a locally loaded certificate to authenticate the "right" end of the IPSec connection (which is normally the remote).

Test case added.

Resolves: https://issues.redhat.com/browse/RHEL-28898

### DIFF
--- a/rust/src/lib/ifaces/ipsec.rs
+++ b/rust/src/lib/ifaces/ipsec.rs
@@ -90,6 +90,8 @@ pub struct LibreswanConfig {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub rightrsasigkey: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
+    pub rightcert: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub left: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub leftid: Option<String>,
@@ -157,6 +159,7 @@ impl std::fmt::Debug for LibreswanConfig {
             .field("right", &self.right)
             .field("rightid", &self.rightid)
             .field("rightrsasigkey", &self.rightrsasigkey)
+            .field("rightcert", &self.rightcert)
             .field("left", &self.left)
             .field("leftid", &self.leftid)
             .field("leftrsasigkey", &self.leftrsasigkey)
--- a/tests/integration/ipsec_test.py
+++ b/tests/integration/ipsec_test.py
@@ -590,6 +590,45 @@ def test_ipsec_ipv4_libreswan_cert_auth_add_and_remove(
     )
 
 
+@pytest.mark.xfail(
+    reason="NetworkManager-libreswan might be too old",
+)
+def test_ipsec_ipv4_libreswan_rightcert(
+    ipsec_hosta_conn_cleanup,
+):
+    desired_state = yaml.load(
+        f"""---
+        interfaces:
+        - name: hosta_conn
+          type: ipsec
+          ipv4:
+            enabled: true
+            dhcp: true
+          libreswan:
+            left: {HOSTA_IPV4_CRT}
+            leftid: '%fromcert'
+            leftcert: hosta.example.org
+            right: {HOSTB_IPV4_CRT}
+            rightid: '%fromcert'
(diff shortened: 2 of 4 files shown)


liangwen12year:
Can you add the issue link to the PR ?

tests/integration/ipsec_test.py
```diff
@@ -590,6 +590,38 @@ def test_ipsec_ipv4_libreswan_cert_auth_add_and_remove(
...
+            rightid: '%fromcert'
+            rightcert: hostb.example.org
+            ikev2: insist
+            ikelifetime: 24h
+            salifetime: 24h""",
+        Loader=yaml.SafeLoader,
+    )
+    libnmstate.apply(desired_state)
+    assert retry_till_true_or_timeout(
+        RETRY_COUNT, _check_ipsec, HOSTA_IPV4_CRT, HOSTB_IPV4_CRT
```
liangwen12year, line 621: I do not have too much experience with Libreswan. Is there a way to check the `rightcert` is configured as specified ? Otherwise, it seems to me that this test is almost the same as the test `test_ipsec_ipv4_libreswan_fromcert`.

rust/src/lib/ifaces/ipsec.rs
```diff
@@ -90,6 +90,8 @@ pub struct LibreswanConfig {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub rightrsasigkey: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
+    pub rightcert: Option<String>,
```
liangwen12year, line 93: Maybe also consider updating the documentation in line 20 to add the rightcert ? 

```
/// interfaces:
/// - name: hosta_conn
///   type: ipsec
///   ipv4:
///     enabled: true
///     dhcp: true
///   libreswan:
///     right: 192.0.2.252
///     rightid: '@hostb.example.org'
///     left: 192.0.2.251
///     leftid: '%fromcert'
///     leftcert: hosta.example.org
///     ikev2: insist
/// ```
```

tests/integration/ipsec_test.py
```diff
@@ -590,6 +590,38 @@ def test_ipsec_ipv4_libreswan_cert_auth_add_and_remove(
     )
 
 
+def test_ipsec_ipv4_libreswan_remote_cert_auth_add_and_remove(
```
cathay4t: We haven't decided to use `remote` to replace `right` yet.

So please use `test_ipsec_ipv4_libreswan_rightcert()`.

cathay4t (changes requested):
Small lint. I have also manually tested with `NetworkManager-libreswan-1.2.20-1.el9.x86_64`. Works well.

tests/integration/ipsec_test.py
cathay4t: Meanwhile, please mark this test case as xfail as the required NetworkManager-libreswan is not shipped to our test environment yet.

rust/src/lib/ifaces/ipsec.rs
ihuguet: Not sure about this. Maybe I'm wrong, but I think that using rightcert is not the most common config, so better to leave as example a more common one. Otherwise users might think that they need to add it when they don't.

tests/integration/ipsec_test.py
ihuguet: The only way I can think of is checking the NetworkManager profile to see if it's present.

rust/src/lib/ifaces/ipsec.rs
cathay4t: Using rightcert is not common. The common case is cert should match with rightid or IP address.

tests/integration/ipsec_test.py
cathay4t: A valid test is using a cert with IP/hostname mismatch with `rightid` or `right`, but that is overkill. Current test is enough to me.

cathay4t (approved):
LGTM.

cathay4t:
The CI failure on Fedora is caused by our pin on ctrlc 3.4.2 which is not available in Fedora. Let's ignore Fedora CI error till we remove the dependency of ctrlc.

liangwen12year (approved):

PR #2661, opened 2024-04-30
ipsec: add support to "leftsubnet" property

ffmancera:
This property allow you to define the private subnet of the left side expressed as network/netmask.

Fixes: https://issues.redhat.com/browse/RHEL-26755

### DIFF
--- a/tests/integration/ipsec_test.py
+++ b/tests/integration/ipsec_test.py
@@ -32,6 +32,7 @@ HOSTA_IPV4_CRT = "192.0.2.251"
 HOSTA_IPV4_PSK = "192.0.2.250"
 HOSTA_IPV4_RSA = "192.0.2.249"
 HOSTA_IPV4_CRT_P2P = "192.0.2.248"
+HOSTA_IPV4_CRT_SUBNET = "192.0.4.0/24"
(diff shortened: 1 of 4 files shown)


tests/integration/ipsec_test.py
```diff
@@ -893,6 +893,41 @@ def test_ipsec_ipv4_libreswan_p2p_cert_auth_add_and_remove(
     )
 
 
+@pytest.mark.xfail(
+    reason="NetworkManager-libreswan might be too old",
```
liangwen12year: Instead of giving a vague reason, can you mention the earliest version that NetworkManager-libreswan supports leftsubnet? 
Based on the NEWS file, https://gitlab.gnome.org/GNOME/NetworkManager-libreswan/-/blob/1d3182af8ffa04981c703095508c4e279737de11/NEWS, the version is NetworkManager-libreswan-1.2.20.

liangwen12year: Other than this, the patch looks good to me.

ffmancera: Sure thing!

ihuguet (approved):

PR #2663, opened 2024-05-05
test: Disable the NM connectivity check in integration tests

liangwen12year:
Enabling the NM connectivity check in the integration tests will risk in making the configured routes get penalty metric because of the limited connectivity.

### DIFF
--- a/automation/run-tests.sh
+++ b/automation/run-tests.sh
@@ -210,6 +210,10 @@ function upgrade_nm_from_rpm_dir {
     find $nm_rpm_dir -name \*.rpm -exec cp -v {} "${EXPORT_DIR}/nm_rpms/" \;
     exec_cmd "dnf remove --assumeyes --noautoremove NetworkManager"
     exec_cmd "dnf install -y ${CONT_EXPORT_DIR}/nm_rpms/*.rpm"
+    # It is fragile for the system to have connectivity check enabled in the
+    # integration testing, NM will add the penalty metric to the route when the
+    # machine is not connected to the Internet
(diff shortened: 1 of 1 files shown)


automation/run-tests.sh
```diff
@@ -210,6 +210,10 @@ function upgrade_nm_from_rpm_dir {
     find $nm_rpm_dir -name \*.rpm -exec cp -v {} "${EXPORT_DIR}/nm_rpms/" \;
     exec_cmd "dnf remove --assumeyes --noautoremove NetworkManager"
     exec_cmd "dnf install -y ${CONT_EXPORT_DIR}/nm_rpms/*.rpm"
+    # It is fragile for the system to have connectivity check enabled in the
+    # integration testing, NM will add the penalty metric to the route when the
+    # machine is not connected to the Internet
+    exec_cmd "dnf remove --assumeyes NetworkManager-config-connectivity"
```
cathay4t, line 216: This rpm does not exists in `quay.io/nmstate/c9s-nmstate-dev` container. Why we need to remove it at run time?

cathay4t (changes requested):

cathay4t:
Please also provide more detail than `configured routes get penalty metric`

automation/run-tests.sh
liangwen12year: I think that if you run `./automation/run-tests-in-nmci.sh --el9 --debug --rpm-dir /root/nm-build/NetworkManager/contrib/fedora/rpm/latest0/RPMS/ `, you  will notice that `NetworkManager-config-connectivity` is actually installed.

cathay4t (approved):

PR #2668, opened 2024-05-14
tests: add hostb scenario for leftsubnet tests

ffmancera:
There needs to be an agreement on the subnet values, left and right. Otherwise, the connection won't be established and the test will fail.


liangwen12year:
@ffmancera , pylint failed.

cathay4t (approved):

PR #2693, opened 2024-06-06
Validate interface in desired port list of controller

liangwen12year:
When the last port of the controller is absent, the port list of the controller becomes empty, in the situaiton of applying the controller and port settings again, before generating the port config, we should not validate if interface is in the port list of controller when port is empty.

Resolves: https://issues.redhat.com/browse/RHEL-31977

### DIFF
--- a/tests/integration/linux_bridge_test.py
+++ b/tests/integration/linux_bridge_test.py
@@ -1282,3 +1282,18 @@ def test_controller_detach_from_linux_bridge(bridge0_with_port0):
         )
         == 0
     )
+
+
+@pytest.fixture
+def empty_bridge():
+    bridge_config = {}
+    with linux_bridge(TEST_BRIDGE0, bridge_config) as state:
+        yield state
+
+
+def test_attach_bond_to_empty_bridge(empty_bridge, bond0):
+    bond0[Interface.KEY][0][Interface.CONTROLLER] = TEST_BRIDGE0
+    desired_state = bond0
+    desired_state[Interface.KEY].append(empty_bridge[Interface.KEY][0])
+    libnmstate.apply(desired_state)
+    assertlib.assert_state_match(desired_state)
--- a/rust/src/lib/ifaces/inter_ifaces_controller.rs
+++ b/rust/src/lib/ifaces/inter_ifaces_controller.rs
@@ -66,7 +66,11 @@ impl MergedInterfaces {
                     if !ctrl_iface.is_desired() {
                         continue;
                     }
-                    if let Some(ports) = ctrl_iface.merged.ports() {
+                    if let Some(ports) = ctrl_iface
+                        .desired
+                        .as_ref()
+                        .and_then(|desire| desire.ports())
+                    {
                         if !ports.contains(&merged_iface.merged.name()) {
                             return Err(NmstateError::new(
                                 ErrorKind::InvalidArgument,
--- a/rust/src/lib/unit_tests/ifaces_ctrller.rs
+++ b/rust/src/lib/unit_tests/ifaces_ctrller.rs
@@ -612,6 +612,35 @@ fn test_iface_detach_controller_been_list_in_other_port_list() {
 }
 
 #[test]
+fn test_iface_controller_conflict_with_empty_br_ports() {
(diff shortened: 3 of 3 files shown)


cathay4t (changes requested):
Instead of checking whether it is empty, you should check whether the port list is desired or not.

With your current path, you incorrectly allows this YAML:

```yml
---
interfaces:
- name: br0
  type: linux-bridge
  bridge:
    ports: []
- name: bond99
  type: bond
  state: up
  controller: br0
  link-aggregation:
    mode: balance-rr
    port:
    - eth2
    - eth1
```

liangwen12year:
> Instead of checking whether it is empty, you should check whether the port list is desired or not.
> 
> With your current path, you incorrectly allows this YAML:
> 
> ```yaml
> ---
> interfaces:
> - name: br0
>   type: linux-bridge
>   bridge:
>     ports: []
> - name: bond99
>   type: bond
>   state: up
>   controller: br0
>   link-aggregation:
>     mode: balance-rr
>     port:
>     - eth2
>     - eth1
> ```

Thanks for the comment, I updated the patch.

cathay4t (changes requested):
Please include unit test case for

```yml
---
interfaces:
- name: br0
  type: linux-bridge
  bridge:
    ports: []
- name: bond99
  type: bond
  state: up
  controller: br0
  link-aggregation:
    mode: balance-rr
    port:
    - eth2
    - eth1
```

cathay4t (changes requested):
Your integration test case `test_port_remove_and_validate_port_controller` has no clean up. Please introduce a fixture and do the clean up no matter test case passed or not.

liangwen12year:
> Please include unit test case for
> 
> ```yaml
> ---
> interfaces:
> - name: br0
>   type: linux-bridge
>   bridge:
>     ports: []
> - name: bond99
>   type: bond
>   state: up
>   controller: br0
>   link-aggregation:
>     mode: balance-rr
>     port:
>     - eth2
>     - eth1
> ```

I added the unit test.

liangwen12year:
> Your integration test case `test_port_remove_and_validate_port_controller` has no clean up. Please introduce a fixture and do the clean up no matter test case passed or not.

I updated the code to reuse the existing fixture for proper cleanup, thanks.

tests/integration/linux_bridge_test.py
```diff
@@ -1282,3 +1282,32 @@ def test_controller_detach_from_linux_bridge(bridge0_with_port0):
...
+                    {
+                        Interface.NAME: bond_name,
+                        Interface.STATE: InterfaceState.ABSENT,
+                    }
+                ]
+            }
+        )
+        assertlib.assert_absent(bond_name)
+
+        libnmstate.apply(desired_state)
```
cathay4t: The key reproducer of original problem is bond0 has controller set to br0, but br0 already exists with empty ports.

It is very hard to summarize that from your test code.

How about:


```python

@pytest.fixture
def empty_bridge():
    # create a empty bridge then yiled and clean up afterwards

def test_attach_bond_to_empty_bridge(empty_bridge, bond0):
    # apply a desire state with bond0 holding controller property and br0 mentioned with name and type only.
    # no need to clean up and assert cleanup as they are not the focus of this test.
    # the fixture should clean up everything. 

```

cathay4t (changes requested):

tests/integration/linux_bridge_test.py
liangwen12year: Thanks for the comment, 
https://issues.redhat.com/browse/RHEL-31977, 
applying the bond0 holding controller property and empty bridge can not reproduce the issue, we have to delete the bond interface in the second step. Then in the final step,  applying the bond0 holding controller property and empty bridge again can reproduce the issue.

cathay4t: The test case should reflect the key reproducer not unrelated environment setup.

liangwen12year: Thanks for reviewing it, I updated the patch again.

cathay4t (approved):

PR #2714, opened 2024-06-24
automation: install dnf5 copr module for Rawhide

vbenes:
We need to have dnf5 copr plugin installed in Rawhide as dnf5 is the default there.


ffmancera (dismissed):
Thanks for the contribution, could you check the CI? This is breaking plenty of jobs.

cathay4t (approved):

PR #2715, opened 2024-06-24
c10s: create new podman files for c10s

vbenes:
* We don't have epel repo so using files from Fedora40 where needed.
* For tox and virtualenv we need to use pip to be compatible for the same lack of epel reasons.
* It seems /etc/systemd/journald.conf is not present either so RateLimits added directly (not sure here if not too invasive)
* We don't need nispor as it should be part of nmstate now.

### DIFF
--- a/automation/run-tests.sh
+++ b/automation/run-tests.sh
@@ -20,6 +20,7 @@ FEDORA_IMAGE_DEV="quay.io/nmstate/fed-nmstate-dev:latest"
 RAWHIDE_IMAGE_DEV="quay.io/nmstate/fed-nmstate-dev:rawhide"
 CENTOS_8_STREAM_IMAGE_DEV="quay.io/nmstate/c8s-nmstate-dev"
 CENTOS_9_STREAM_IMAGE_DEV="quay.io/nmstate/c9s-nmstate-dev"
+CENTOS_10_STREAM_IMAGE_DEV="quay.io/nmstate/c10s-nmstate-dev"
 
 
 PYTEST_OPTIONS="--verbose --verbose \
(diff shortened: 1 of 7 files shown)


cathay4t:
1. We still need nispor CLI `npc` in test.
2. Please expand `run-tests.sh` with `--el10` argument. And also `run-tests-in-nmci.sh`. This would show whether the container works or not.
3. You may tag your local build as `quay.io/nmstate/c10s-nmstate-build:latest`, so runtest.sh will use local built result. `podman build -f Dockerfile.c10s-nmstate-build -t quay.io/nmstate/c10s-nmstate-build:latest`

automation/run-tests.sh
```diff
@@ -20,6 +20,7 @@ FEDORA_IMAGE_DEV="quay.io/nmstate/fed-nmstate-dev:latest"
 RAWHIDE_IMAGE_DEV="quay.io/nmstate/fed-nmstate-dev:rawhide"
 CENTOS_8_STREAM_IMAGE_DEV="quay.io/nmstate/c8s-nmstate-dev"
 CENTOS_9_STREAM_IMAGE_DEV="quay.io/nmstate/c9s-nmstate-dev"
+CENTOS_10_STREAM_IMAGE_DEV="quay.io/nmstate/c10s-nmstate-build"
```
cathay4t: Should be `quay.io/nmstate/c10s-nmstate-dev`

packaging/Dockerfile.c10s-nmstate-dev
```diff
@@ -0,0 +1,35 @@
...
+    dnf clean all
+
+RUN go env -w GOSUMDB="sum.golang.org" GOPROXY="https://proxy.golang.org,direct"
+
+COPY network_manager_enable_trace.conf \
+     /etc/NetworkManager/conf.d/97-trace-logging.conf
+COPY network_manager_keyfile.conf \
+     /etc/NetworkManager/conf.d/96-keyfile.conf
+
+# Fedora container has no /etc/systemd/journald.conf file in systemd rpm
```
cathay4t: RHEL/CentOS stream 10

automation/run-tests.sh
vbenes: done

packaging/Dockerfile.c10s-nmstate-dev
vbenes: done as well

cathay4t (approved):
The OVS tests are failing, but good enough for first approach of c10s tests.

PR #2726, opened 2024-07-03
tests: Add the description for loopback integration tests

liangwen12year:


liangwen12year:
```
description: Set the IPv4 address to 192.0.2.251/24 and the IPv6 address to 2001:db8:1::1/64 for loopback device
interfaces:
- name: lo
  type: loopback
  state: up
  ipv4:
    address:
    - ip: 192.0.2.251
      prefix-length: 24
  ipv6:
    address:
    - ip: 2001:db8:1::1
      prefix-length: 64
description: Delete the loopback interface
interfaces:
- name: lo
  type: loopback
  state: absent
description: Set the mtu of the lookback to 12800
interfaces:
- name: lo
  type: loopback
  state: up
  mtu: 12800
description: Delete the loopback interface
interfaces:
- name: lo
  type: loopback
  state: absent
description: Remove the lookback device
interfaces:
- name: lo
  type: loopback
  state: absent
```

tests/integration/loopback_test.yaml
```diff
@@ -0,0 +1,34 @@
+description: Delete the loopback interface
+interfaces:
+- name: lo
+  type: loopback
+  state: absent
+description: Set the address to 192.0.2.251/24 and 2001:db8:1::1/64 for loopback device
```
cathay4t: `Add the address 192.0.2.251/24 and 2001:db8:1::1/64 to loopback device.`

Because we always preserve 127.0.0.1 and ::1 for loopback interface. So set is wrong here.

tests/integration/loopback_test.yaml
```diff
@@ -0,0 +1,34 @@
...
+  state: up
+  ipv4:
+    address:
+    - ip: 192.0.2.251
+      prefix-length: 24
+  ipv6:
+    address:
+    - ip: 2001:db8:1::1
+      prefix-length: 64
+description: Delete the loopback interface
```
cathay4t: Revert the loopback interface to default.

cathay4t (changes requested):

tests/integration/loopback_test.yaml
cathay4t: We never allows user to remove loopback interface, it will break the whole system.

tests/integration/loopback_test.yaml
liangwen12year: I updated it, thanks.

tests/integration/loopback_test.yaml
liangwen12year: I updated it, thanks.

tests/integration/loopback_test.yaml
```diff
@@ -0,0 +1,34 @@
+description: Revert the loopback interface to default
```
tyll, line 4: There should be a comment at the beginning that explains the purpose of this file.

tests/integration/loopback_test.py
```diff
@@ -75,9 +77,12 @@ def test_change_loopback_mtu_and_restore_back(self, loopback_cleanup):
...
             ]
         }
-        libnmstate.apply(desired_state)
+        apply_with_description(
+            "Set the mtu of the lookback to 12800", desired_state
+        )
         assertlib.assert_state_match(desired_state)
-        libnmstate.apply(
+        apply_with_description(
+            "Remove the lookback device",
```
tyll: Gris' previous comment applies here as well, also there is a typo

tests/integration/loopback_test.yaml
```diff
@@ -0,0 +1,34 @@
...
+interfaces:
+- name: lo
+  type: loopback
+  state: absent
+description: Add the address 192.0.2.251/24 and 2001:db8:1::1/64 to loopback device
```
tyll, line 6: There is a separator missing here

tests/integration/loopback_test.yaml
```diff
@@ -0,0 +1,34 @@
+description: Revert the loopback interface to default
+interfaces:
+- name: lo
+  type: loopback
+  state: absent
```
tyll, line 8: There is a separator missing here

tyll: There should be a separator that makes it easy to read this data from a script.

tests/integration/loopback_test.py
liangwen12year: Thanks for pointing it out. I updated it.

tests/integration/loopback_test.yaml
liangwen12year: I added the comment at the beginning that explains the purpose of this file.

tests/integration/loopback_test.yaml
liangwen12year: I added the separator, thanks!

cathay4t (approved):

PR #2734, opened 2024-07-09
net_state: raise the verify timeout

lkundrak:
Nmstate expects the IP addresses to come up when the connection comes up, but fails to set ipv[46].may-fail appropriately.

Moreover, ipv[46].may-fail is controlled separately by the interface.wait-ip key, so this can't really be fixed.

Our only option seems to be to poll significantly longer. We have no way of knowing how long (at least without NetworkManager being extended to tell us know the state of the L2 configurations separately) -- 45 seconds seems to be a good start, because that's the default DHCP timeout.

Resolves: https://issues.redhat.com/browse/RHEL-38607


ihuguet:
According to Gris, nmstate must never wait for DHCP completion, and I think it's reasonable. Also, increasing the timeout to 45s feels like too much for me. If there is an environment with such a slow DHCP server, better to use `nmstatectl apply --timeout 45` manually.

Also, why do you say that nmstate doesn't set `may-fail` well? AFAIK for `Any` it sets `ipv4.may-fail=yes` and `ipv6.may-fail=yes`, which basically means that the connection will succeed if any of ipv4 or ipv6 succeeds.

It must be noted that nmstate doesn't wait until the connection is activated. Once NetworkManager returns an OK from the connection activation method, nmstate starts trying to verify (even if the NM connection is "activating").

So, the conclusion is that the problem might be related to slow DHCP, but it doesn't matter when the connection is finally set to "activated".

ihuguet:
In Jira I see this message in the logs, comming from NetworkManager:
```
[2024-05-24T12:41:53Z INFO  nmstate::nm::query_apply::profile] Got activation failure Bug: Manager(UnknownDevice): No suitable device found for this connection (device eno1 not available because device has no carrier).
``

smyle10:
Apply successfully but the link DOWN detected https://issues.redhat.com/browse/RHEL-38607?focusedId=25278703&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-25278703

cathay4t (changes requested):
The reason we cannot wait DHCP is because linux bridge has STP, bond has LACP, those will only link up after 30+ seconds. There is no reason for nmstate to wait DHCP to pass the verification. Hence increasing verify timeout to wait DHCP is incorrect.

For this issue, I think we should ignore the MPTCP flags set on auto ip and only verify NM has the config correctly stored. User should be warned that nmstate does not verify MPTCP flags on auto-ips.

cathay4t:
Closing as this issue was fixed by #2774

PR #2744, opened 2024-07-30
vlan: make 'base-iface' optional for the desired state

ihuguet:
When creating a VLAN interface we need to know the base interface to
which is attached. However, if the VLAN interface already exists we can
asume that, if it's not specified, base-iface remains the same.

Make 'base-iface' optional and take it from the current state if it's
missing. Emit an error if it's not already defined in the current state.

Also, there was a bug in gen_diff. If base-iface doesn't change,
creating the VlanInterface struct failed because base-iface was
mandatory and it was missing in the diff. As it is no longer
mandatory, this commit fixes the bug. The message of the bug was:
>  "missing field `base-iface`"

Added test cases.

Resolves: https://issues.redhat.com/browse/RHEL-38623


cathay4t (approved):
LGTM

PR #2747, opened 2024-08-01
nm: send lldp with correct D-Bus type INT32

### LINKED ISSUE #2743: lldp nmconnection incorrectly written as lldp=false when it should be lldp=0
## Describe the bug
The nmconnection output writes `lldp=false` instead of using an `int32` value - this causes the NetworkManager to throw a warning and won't modify the LLDP parameter for an interface. 

## Nmstate version
- 2.2.33

## NetworkManager version
- 1.46.0

## Platform
- RHEL 9 / CoreOS 4.16

## How reproducible
- Set lldp enable to false in nmstate config

## Steps to Reproduce
Test yaml to disable `lldp`:
```
    interfaces:
    - name: ens2f1
      type: ethernet 
      state: up
      lldp:
        enabled: false
```
Then run `nmstatectl gc` to see the lldp=false output, i.e.:

```
> nmstatectl gc test-lldp.yaml 
[2024-07-29T16:18:56Z INFO  nmstatectl] Nmstate version: 2.2.33
NetworkManager:
- - ens2f1.nmconnection
  - |
    [connection]
    autoconnect=true
    autoconnect-slaves=-1
    id=ens2f1
    interface-name=ens2f1
    lldp=false <<< this is invalid
    type=802-3-ethernet
 <output truncated>
```

## Nmstate output
- Expected output:
The nmconnection configuration should be an integer, e.g.:
lldp=0 

- Current output:
lldp=false

### Additional context
From the docs [1]:
image

\[1\]: https://networkmanager.pages.freedesktop.org/NetworkManager/NetworkManager/nm-settings-nmcli.html

ihuguet:
Fixes https://github.com/nmstate/nmstate/issues/2743

### DIFF
--- a/rust/src/lib/nm/nm_dbus/connection/conn.rs
+++ b/rust/src/lib/nm/nm_dbus/connection/conn.rs
@@ -390,7 +390,7 @@ impl ToDbusValue for NmSettingConnection {
             ret.insert("slave-type", zvariant::Value::new(v.as_str()));
         }
         if let Some(v) = &self.lldp {
-            ret.insert("lldp", zvariant::Value::new(v));
+            ret.insert("lldp", zvariant::Value::new(*v as i32));
         }
         if let Some(v) = &self.mptcp_flags {
             ret.insert("mptcp-flags", zvariant::Value::new(v));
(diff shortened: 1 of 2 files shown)


cathay4t (changes requested):
Please add test case using `gen_conf_apply` to simulate the original issue.

ihuguet:
I think it cannot be reproduced like that. According to the responses in #2743, the problem only happens when restarting NM.

cathay4t:
I made some cosmetic changes:
 1. Rebased to latest.
 2. Add small explanation of the issue.
 3. Squashed the two patches into single one.

cathay4t (approved):
Reviewed and tested.

PR #2758, opened 2024-08-15
tests: stop NM service before removing its rpms

vbenes:
In EL10 we see respawn issues when removing rpms on running NM service. Stopping the service beforehand should prevent this issue.


cathay4t:
Let's postpone this till we reproduce it or find root cause of it.

cathay4t (approved):

cathay4t:
I have amend the patch by adding `Signed-off-by:` line to the git commit message.

PR #2770, opened 2024-08-30
route: Add support for route source

liangwen12year:
In a scenario where you have a machine with multiple public IP
addresses, typically due to a multi-WAN setup, the src parameter in the
context of routes allows you to specify which source IP address should
be used when sending packets via a specific route. This is crucial when
you want to ensure that outbound traffic uses a specific IP address tied
to a particular network interface, especially when dealing with multiple
WAN connections.

Adding support for the src parameter in routes results in a more
powerful and flexible network configuration capability, especially
important in environments with multiple network interfaces or multiple
IP addresses, it provides better control over traffic routing.

The following is the example for specifying the route src in Nmstate:

```
---
interfaces:
  - name: eth1
    type: ethernet
    state: up
    ipv4:
      address:
        - ip: 192.0.2.251
          prefix-length: 24
        - ip: 192.0.2.252
          prefix-length: 24
      dhcp: false
      enabled: true
routes:
  config:
      source: 192.0.2.251
      next-hop-address: 192.0.2.1
      next-hop-interface: eth1
      table-id: 254
      metric: 150
```

Resolves: https://issues.redhat.com/browse/RHEL-56258

### DIFF
--- a/rust/src/lib/route.rs
+++ b/rust/src/lib/route.rs
@@ -183,6 +183,10 @@ pub struct RouteEntry {
     /// Congestion window clamp
     #[serde(skip_serializing_if = "Option::is_none")]
     pub cwnd: Option<u32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    /// Route source defines which IP address should be used as the source
+    /// for packets routed via a specific route
+    pub source: Option<String>,
 }
 
 #[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq)]
@@ -271,13 +275,16 @@ impl RouteEntry {
         if self.cwnd.is_some() && self.cwnd != other.cwnd {
             return false;
         }
+        if self.source.as_ref().is_some() && self.source != other.source {
+            return false;
+        }
         true
     }
 
     // Return tuple of (no_absent, is_ipv4, table_id, next_hop_iface,
-    // destination, next_hop_addr, weight, cwnd)
+    // destination, next_hop_addr, source, weight, cwnd)
     // Metric is ignored
-    fn sort_key(&self) -> (bool, bool, u32, &str, &str, &str, u16, u32) {
+    fn sort_key(&self) -> (bool, bool, u32, &str, &str, &str, &str, u16, u32) {
         (
             !matches!(self.state, Some(RouteState::Absent)),
             !self
@@ -291,6 +298,7 @@ impl RouteEntry {
                 .unwrap_or(LOOPBACK_IFACE_NAME),
             self.destination.as_deref().unwrap_or(""),
             self.next_hop_addr.as_deref().unwrap_or(""),
+            self.source.as_deref().unwrap_or(""),
             self.weight.unwrap_or_default(),
             self.cwnd.unwrap_or_default(),
         )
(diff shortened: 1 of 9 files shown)


liangwen12year:
@cathay4t , the test failure `test_iface_down_with_routes_in_current` is transient, I ran the test `test_iface_down_with_routes_in_current` 50 times locally, but I can not reproduce the issue.

cathay4t (changes requested):
If we are using `source: Option<String>` instead of `Option<IpAddr>` for unknown future, we need to do sanitize for it in case user ask for `source: 2001:0db8:85a3:0000:0000:8a2e:0370:7331`.

liangwen12year:
> If we are using `source: Option<String>` instead of `Option<IpAddr>` for unknown future, we need to do sanitize for it in case user ask for `source: 2001:0db8:85a3:0000:0000:8a2e:0370:7331`.

I think that I already added the sanitize code, as you can see, 
```
        if let Some(src) = self.source.as_ref() {
            let new_src = format!("{}", src.parse::<std::net::IpAddr>()?);
            if src != &new_src {
                log::warn!(
                    "Route source address {} sanitized to {}",
                    src,
                    new_src
                );
                self.source = Some(new_src);
            }
        }
```
And I tested the sanitize function, as you can see, the src is indeed sanitized
```
[2024-09-03T00:24:28Z WARN  nmstate::route] Route source address 2001:0db8:85a3:0000:0000:8a2e:0370:7001 sanitized to 2001:db8:85a3::8a2e:370:7001
route-rules: {}
routes:
  config:
  - destination: 2003:db8:85a3::8a2e:370:7020/128
    next-hop-interface: eth1
    next-hop-address: 2001:db8:85a3::8a2e:370:1
    metric: 150
    table-id: 254
    source: 2001:db8:85a3::8a2e:370:7001
interfaces: []
ovn: {}
```

cathay4t:
Please create  a unit test case for this sanitize instead of `I tested`

rust/src/lib/route.rs
```diff
@@ -323,6 +331,17 @@ impl RouteEntry {
                 self.next_hop_addr = Some(new_via);
             }
         }
+        if let Some(src) = self.source.as_ref() {
+            let new_src = format!("{}", src.parse::<std::net::IpAddr>()?);
```
cathay4t: Instead of trust auto error converter, I think it is better to `map_err` providing better error message.

rust/src/lib/route.rs
```diff
@@ -323,6 +331,17 @@ impl RouteEntry {
...
+            if src != &new_src {
+                log::warn!(
+                    "Route source address {} sanitized to {}",
```
cathay4t, line 346: It is not a warning, the uncompressed IPv6 address is still legal. Use `log::info`.

tests/integration/route_test.py
```diff
@@ -317,6 +317,26 @@ def test_apply_route_with_route_type_multiple_times(eth1_up):
     assert routes_out_v6.count("unreachable") == 1
 
 
+@pytest.mark.tier1
+def test_add_static_route_with_route_src(eth1_up):
```
cathay4t, line 333: Need test case for IPv6 route src also.

cathay4t (changes requested):

liangwen12year:
> Please create a unit test case for this sanitize instead of `I tested`

I created the unit test case.

liangwen12year:
@cathay4t , all your comments are resolved, thanks !

cathay4t (approved):

PR #2775, opened 2024-09-11
ipsec: support require-id-on-certificate

ihuguet:
Test case included.

Resolves: https://issues.redhat.com/browse/RHEL-56549

### DIFF
--- a/rust/src/lib/ifaces/ipsec.rs
+++ b/rust/src/lib/ifaces/ipsec.rs
@@ -148,6 +148,11 @@ pub struct LibreswanConfig {
     pub hostaddrfamily: Option<LibreswanAddressFamily>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub clientaddrfamily: Option<LibreswanAddressFamily>,
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        rename = "require-id-on-certificate"
+    )]
+    pub require_id_on_certificate: Option<bool>,
 }
 
 impl LibreswanConfig {
@@ -187,6 +192,7 @@ impl std::fmt::Debug for LibreswanConfig {
             .field("kind", &self.kind)
             .field("hostaddrfamily", &self.hostaddrfamily)
             .field("clientaddrfamily", &self.clientaddrfamily)
+            .field("require_id_on_certificate", &self.require_id_on_certificate)
             .finish()
     }
 }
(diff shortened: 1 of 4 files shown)


ihuguet:
The test case uses `assert not retry_till_true_or_timeout` to verify that the IPSEC connection has failed, but I'm not happy because it's slow and it doesn't test that it fails for the right reason. Need to think a bit more about it, any suggestion is welcome.

ihuguet:
NetworkManager doesn't fail immediately if Libreswan fails to connect due to `require-id-on-certificate`. Instead, it stays in `connecting` state for a while. This means that there is no way to avoid a small delay in the test to check for the connection failure.

About ensuring that it fails for the right reason: the whole test case already ensures that it **passes** for the right reason, which is probably enough.

rust/src/lib/ifaces/ipsec.rs
```diff
@@ -148,6 +148,11 @@ pub struct LibreswanConfig {
     pub hostaddrfamily: Option<LibreswanAddressFamily>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub clientaddrfamily: Option<LibreswanAddressFamily>,
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        rename = "require-id-on-certificate"
+    )]
+    pub require_id_on_certificate: Option<bool>,
```
liangwen12year, line 155: Can you explain the use case of the setting `require_id_on_certificate` in the commit message and comments using `///` here (These comments are processed by Rust's documentation tool (rustdoc) to generate HTML documentation)  ?

liangwen12year:
Since require-id-on-certificate controls whether an ID must be present on a certificate for authentication, maybe it would be meaningless or invalid without specifying the certificates for both ends of the IPsec connection (leftcert and rightcert). we probably need the validation for this ?

rust/src/lib/ifaces/ipsec.rs
ihuguet: I didn't put it because the rest of the Libreswan's fields don't have this explanation. I guess it is because each field is a 1-1 match to the raw Libreswan's configs.

ihuguet:
> Since require-id-on-certificate controls whether an ID must be present on a certificate for authentication, maybe it would be meaningless or invalid without specifying the certificates for both ends of the IPsec connection (leftcert and rightcert). we probably need the validation for this ?

Probably, but if something is not OK we can leave NM and/or Libreswan to reject it. Anyway, it is not a problem to specify require-id-on-certificate and then not using certificates: the value will be just ignored, there is no problem with that.

cathay4t (approved):

cathay4t:
Please wait a sec. I am fixing CI.

cathay4t:
Please wait https://github.com/nmstate/nmstate/pull/2780 for CI fixing.

PR #2777, opened 2024-09-17
Policy support unordered references

qinqon:
This change allow the capture commands to be ordered independently of how they reference between them.

It's done by ordering the cmds vector according to how capture entries are referenced between each other, also a cycle reference mechanism is implemeneted.

### DIFF
--- a/rust/src/lib/policy/capture.rs
+++ b/rust/src/lib/policy/capture.rs
@@ -16,6 +16,8 @@ use super::{
     token::{parse_str_to_capture_tokens, NetworkCaptureToken},
 };
 
+pub const CAPTURE_LOOP_LIMIT: u32 = 25;
+
 pub(crate) const PROPERTY_SPLITTER: &str = ".";
 
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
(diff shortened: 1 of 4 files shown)


rust/src/lib/policy/capture.rs
```diff
@@ -32,21 +32,72 @@ impl<'de> Deserialize<'de> for NetworkCaptureRules {
...
+                );
             } else {
                 return Err(serde::de::Error::custom(format!(
                     "Expecting a string, but got {v}"
                 )));
             }
         }
+
+        // Order cmds vector to honor how they get referenced, also
+        // detect looks and fail.
```
cathay4t: locks

rust/src/lib/policy/capture.rs
```diff
@@ -32,21 +32,72 @@ impl<'de> Deserialize<'de> for NetworkCaptureRules {
...
+        let mut visited_caps: HashSet<&String> = HashSet::new();
+        let mut cmds: Vec<(String, NetworkCaptureCommand)> = Vec::new();
+        for (cap_name, cap_cmd) in caps_map.iter() {
+            let mut cmd = cap_cmd;
+            // Implemented this with a stack so we don't have to do recursion
+            let mut stack: Vec<(&String, &NetworkCaptureCommand)> = Vec::new();
+            let mut stacked_caps: HashSet<&String> = HashSet::new();
+            while let Some(key_capture) = cmd.key_capture.as_ref() {
```
cathay4t: I am scared of this might lead to infinity loop for some YAML in the future due to poor patch.

Can you change the way to use method like `set_ifaces_up_priority()` by given a maximum recurrence count?

cathay4t (changes requested):

rust/src/lib/policy/capture.rs
qinqon: > I am scared of this might lead to infinity loop for some YAML in the future due to poor patch.
> 

The PR include some testing, but we can increase the testing with more cases to ensure that we are good enough.

> Can you change the way to use method like `set_ifaces_up_priority()` by given a maximum recurrence count?

This means re-implement the solution using a counter ?

cathay4t: Yes. Counter based. Like ordering controller/port, each round we set a `capture-priority` to its dependent's priority +1. And break the loop if all set. By having a limitation on loop count, we are preventing the dead loop at design instead of trusting test or developer covered all cases.

qinqon: This is different to controlle/ ports, a port has just one owner, but a capture entry can be referenced by mutiple of them.

Also counter implementation can brake by implementation, proper test is always needed

qinqon: Refactored to use a for loop with a limit and simplify it.

qinqon:
@cathay4t this feels right now ?

PR #2778, opened 2024-09-17
tests: add centos mirror list to avoid checksum errors

vbenes:
Let's add mirror lists for CentOS AppStream/BaseOS repos to prevent 
Errors during downloading metadata for repository 'baseos' issues. 
A repo resync likely causes this and should not happen in all 
other repositories. So with this approach, we will find a working 
repo rather than failing in the first checksum failure.

### DIFF
--- a/automation/run-tests.sh
+++ b/automation/run-tests.sh
@@ -183,6 +183,22 @@ function modprobe_ovs {
     #TODO: Install ovs on k8s cluster
 }
 
+function add_baseos_mirrorlist {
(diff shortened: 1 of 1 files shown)


cathay4t (changes requested):
I think the correct fix should be set `fastestmirror=False` in /etc/dnf/dnf.conf, then dnf will try more mirrors instead of the fastest one. And also `best=False` option. Please investigate whether other dnf options fits or not.

vbenes:
That appears to be an infra issue, hopefully solved now.

PR #2785, opened 2024-09-25
packaging: add baseurl repos with low prio as alternative

vbenes:
Add alternative repos to mirrorlink-based with lower priority and enable skip_if_unavailabl=True generally to dnf.conf to avoid broken repos. This should do nothing if repositories are correctly working.


cathay4t (approved):

cathay4t:
@vbenes The CI is complaining
```
STEP 3/29: RUN cat <<EOF > /etc/yum.repos.d/alter-baseos.repo
/bin/sh: line 1: warning: here-document at line 1 delimited by end-of-file (wanted `EOF')
```

cathay4t:
And also please use `git commit --sign-off` or add `Signed-off-by: Foo <foo@example.org>` to the bottom of commit message.

PR #2788, opened 2024-09-26
ipvlan: add IPVLAN interface support

ffmancera:
Example:

```
---
interfaces:
  - name: ipvlan0
    type: ipvlan
    state: up
    ipvlan:
      base-iface: eth1
      mode: l2
```

Integration tests added.

Resolves: https://issues.redhat.com/browse/RHEL-56571


rust/src/lib/ifaces/ipvlan.rs
```diff
@@ -0,0 +1,114 @@
...
+        Ok(())
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Default)]
+#[serde(rename_all = "kebab-case", deny_unknown_fields)]
+#[non_exhaustive]
+pub struct IpVlanConfig {
+    pub base_iface: String,
+    pub mode: IpVlanMode,
```
cathay4t: Both base-iface and mode can be optional when modifying existing interface. It also helps us on `gen_diff()`.

cathay4t: Please also mark `mode` with `derive[serde(default)]`, user does not need to define it for default value.

cathay4t (changes requested):

rust/src/lib/ifaces/ipvlan.rs
ffmancera: Sure! Let me adjust it, thanks!

ffmancera: Done, thank you!

cathay4t (dismissed):
Looks good except we do not validate whether `base-iface` exists.

cathay4t (approved):

PR #2797, opened 2024-10-02
gh: Add option to upload hidden files

qinqon:
As of latest upload github action by default is not uploading hidden files [1]. This change configures the flag to do so, so the .states are uploades.

[1] https://github.com/actions/upload-artifact/releases/tag/v4.4.0

We are going to need to regenerate assets from versions > 2.2.33


ffmancera (approved):
LGTM!

PR #2809, opened 2024-10-30
Avoid some unnecessary deactivations due to routes mismatches

ihuguet:
When comparing the routes with those already added to NM, sometimes
nmstate considered different routes that were actually equal.
- route1 with table=None vs route2 with table=Some(254)
- route1 with next-hop=None vs route2 with next-hop=Some(0.0.0.0)
- route1 is ipv6 with metric=0, route2 is ipv6 with metric=1024 (see explanation below).

The consequence was that unnecessary deactivate/reactivate were done,
instead of reapply. This was because is_route_removed was returning true
with some of the situations described above. This caused that
gen_nm_conn_need_to_deactivate_first also returned true and the
connection was deactivated.

For next-hop, fix it by normalizing to None at creation time, since 0.0.0.0 and ::
are not used by NM nor by the kernel, they just leave it undefined.

For table, fix it by normalizing to Some(254) because we need to always set
the table for a different reason: if the user doesn't define the table in the
desired state we are supposed to save it to the 254 main table. If we left
table empty, it might happen that a different table is used if auto-route-table-id
is also used.

For IPv6 routes metric is also included. When setting metric=0 on a kernel's IPv6 route,
kernel coherces it to 1024. This make that, when reading back the route from kernel
it doesn't match with the expected value of 0. The fix normalizes 0->1024 for IPv6
routes so the comparison doesn't fail.

Additionally, when comparing NmIpRoutes, this PR add code to handle it correctly
for every possible NMConnection. If the route's table is undefined, it falls back to
ipv4/6.route-table, as this is what NM does. Note that for routes defined by
nmstate this won't happen because, since this PR, we always define the table; but
it will handle better the case of already existing NMConnections. The same for the
metric with route-metric.

Note that this only affects NM specific code, not the RouteEntry
entities from the top level nmstate.

Resolves: https://issues.redhat.com/browse/RHEL-65004

### DIFF
--- a/rust/src/lib/nm/query_apply/route.rs
+++ b/rust/src/lib/nm/query_apply/route.rs
@@ -1,50 +1,69 @@
 // SPDX-License-Identifier: Apache-2.0
 
-use super::super::nm_dbus::{NmConnection, NmIpRoute};
+use super::super::nm_dbus::{NmConnection, NmIpRoute, NmSettingIp};
 
-const DEFAULT_TABLE_ID: u32 = 254; // main route table ID
+const NM_IP_SETTING_ROUTE_TABLE_DEFAULT: u32 = 0;
+const NM_IP_SETTING_ROUTE_METRIC_DEFAULT: i64 = -1;
+const IPV6_METRIC_COHERCED_DEFAULT: u32 = 1024; // Coherced by kernel 0->1024
 
 pub(crate) fn is_route_removed(
     new_nm_conn: &NmConnection,
     cur_nm_conn: &NmConnection,
 ) -> bool {
-    is_nm_ip_route_removed(
-        new_nm_conn
-            .ipv4
-            .as_ref()
-            .map(|ip| ip.routes.as_slice())
-            .unwrap_or(&[]),
-        cur_nm_conn
-            .ipv4
-            .as_ref()
-            .map(|ip| ip.routes.as_slice())
-            .unwrap_or(&[]),
-    ) || is_nm_ip_route_removed(
-        new_nm_conn
-            .ipv6
-            .as_ref()
-            .map(|ip| ip.routes.as_slice())
-            .unwrap_or(&[]),
-        cur_nm_conn
-            .ipv6
-            .as_ref()
-            .map(|ip| ip.routes.as_slice())
-            .unwrap_or(&[]),
+    nm_setting_is_route_removed(
+        new_nm_conn.ipv4.as_ref(),
+        cur_nm_conn.ipv4.as_ref(),
+        false,
+    ) || nm_setting_is_route_removed(
+        new_nm_conn.ipv6.as_ref(),
+        cur_nm_conn.ipv6.as_ref(),
+        true,
     )
 }
 
-fn is_nm_ip_route_removed(
-    new_routes: &[NmIpRoute],
-    cur_routes: &[NmIpRoute],
+fn nm_setting_is_route_removed(
+    new_nm_sett: Option<&NmSettingIp>,
+    cur_nm_sett: Option<&NmSettingIp>,
+    is_ipv6: bool,
 ) -> bool {
-    for cur_route in cur_routes {
-        let mut cur_route_norm = cur_route.clone();
-        if cur_route_norm.table.is_none() {
-            cur_route_norm.table = Some(DEFAULT_TABLE_ID);
-        }
-        if !new_routes.contains(&cur_route_norm) {
-            return true;
-        }
+    let new_routes = clone_normalized_routes(new_nm_sett, is_ipv6);
+    let cur_routes = clone_normalized_routes(cur_nm_sett, is_ipv6);
+    cur_routes
+        .iter()
+        .any(|cur_route| !new_routes.contains(cur_route))
+}
+
+fn clone_normalized_routes(
+    ip_sett: Option<&NmSettingIp>,
+    is_ipv6: bool,
+) -> Vec<NmIpRoute> {
+    // Routes defined by nmstate will always has table and metric set, so there
+    // is no problem comparing them.
+    // On routes defined in NM directly, they may depend on the route-metric and
+    // route-table properties of the ipv4 and ipv6 settings. Use them to get the
+    // actual values.
+    // They may even fall back to a globally default value. In that case we can
+    // not know what value is. Use None to fail the comparison so we can
+    // properly install the new desired route, with table and metric defined.
+    let default_table = ip_sett
+        .and_then(|ip| ip.route_table)
+        .filter(|tbl| *tbl != NM_IP_SETTING_ROUTE_TABLE_DEFAULT);
+    let mut default_metric = ip_sett
+        .and_then(|ip| ip.route_metric)
+        .filter(|mtr| *mtr != NM_IP_SETTING_ROUTE_METRIC_DEFAULT)
+        .map(|mtr| mtr as u32);
+    if is_ipv6 && default_metric == Some(0) {
+        default_metric = Some(IPV6_METRIC_COHERCED_DEFAULT);
     }
-    false
+
+    let routes = ip_sett.map(|ip| ip.routes.as_slice()).unwrap_or(&[]);
+    routes
+        .iter()
+        .map(|rt| {
+            let mut new_rt = rt.clone();
+            new_rt.table = rt.table.or(default_table);
+            new_rt.metric = rt.metric.or(default_metric);
+            new_rt
+        })
+        .collect()
 }
--- a/tests/integration/route_test.py
+++ b/tests/integration/route_test.py
@@ -19,8 +19,10 @@ from .testlib import assertlib
 from .testlib import cmdlib
 from .testlib import iprule
 from .testlib.bridgelib import linux_bridge
+from .testlib.dummy import dummy_interface
 from .testlib.env import nm_minor_version
 from .testlib.genconf import gen_conf_apply
+from .testlib.iproutelib import ip_monitor_assert_stable_link_up
 from .testlib.route import assert_routes
 from .testlib.route import assert_routes_missing
 from .testlib.servicelib import disable_service
@@ -110,6 +112,12 @@ def clean_up_route_rule():
             DNS.KEY: {DNS.CONFIG: {}},
         },
     )
+
+
+@pytest.fixture(scope="function")
+def dummy0_up(test_env_setup):
+    with dummy_interface("dummy0") as ifstate:
+        yield ifstate
 
 
 @pytest.mark.tier1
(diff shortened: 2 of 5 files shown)


rust/src/lib/nm/nm_dbus/connection/route.rs
```diff
@@ -26,18 +30,32 @@ pub struct NmIpRoute {
 impl TryFrom<DbusDictionary> for NmIpRoute {
     type Error = NmError;
     fn try_from(mut v: DbusDictionary) -> Result<Self, Self::Error> {
+        let dest = _from_map!(v, "dest", String::try_from)?;
+        let default_next_hop = dest.as_deref().and_then(|dest| {
+            match dest.parse::<std::net::IpAddr>() {
+                Ok(std::net::IpAddr::V4(_)) => Some(IPV4_EMPTY_NEXT_HOP_ADDR),
```
cathay4t:
```rust
Ok(std::net::IpAddr::V4(i)) => Some(i.to_string())
```

rust/src/lib/nm/nm_dbus/connection/route.rs
```diff
@@ -26,18 +30,32 @@ pub struct NmIpRoute {
 impl TryFrom<DbusDictionary> for NmIpRoute {
     type Error = NmError;
     fn try_from(mut v: DbusDictionary) -> Result<Self, Self::Error> {
+        let dest = _from_map!(v, "dest", String::try_from)?;
+        let default_next_hop = dest.as_deref().and_then(|dest| {
+            match dest.parse::<std::net::IpAddr>() {
```
cathay4t: Base on your code it seems you are assuming `dest` in NM is always 0.0.0.0 or ::. Is this correct?

tests/integration/route_test.py
```diff
@@ -143,6 +156,75 @@ def test_add_static_route_without_next_hop_address(eth1_up):
...
 
+@ip_monitor_assert_stable_link_up("eth1")
+def test_route_next_hop_address_default(eth1_up):
+    libnmstate.apply({Interface.KEY: [ETH1_INTERFACE_STATE]})
+
+    # NM's routes without next-hop, nmstate with it's representation of an
+    # empty next-hop: '0.0.0.0' or '::'.
+    # If nmstate wouldn't consider them equal it would do deactivate/reactivate.
+    # Ensure that we only reapply with ip_monitor_assert_stable_link_up.
+    _nmcli_con_modify_routes_reapply("eth1", IPV4_TEST_NET1, IPV6_TEST_NET1)
```
cathay4t: Can you explain why use `nmcli modify` instead apply another nmstate YAML?

I think the original issue report does not use nmcli.

cathay4t (changes requested):

rust/src/lib/nm/nm_dbus/connection/route.rs
cathay4t: And `nm_dbus` should not have this modification code, it should be simple dbus wrapper. This polish/modification should be done at `nm` level.

tests/integration/route_test.py
ihuguet: Because if I use nmstate YAML it is dependant to the internal implementation of nmstate how to create the NM connection. I have seen that in some cases it is adding table=254, for example, while I want to test the case that the NM profile exists without `table`. So what I do here is to force certain combinations of initial NM's state of NM and nmstate desired states.

rust/src/lib/nm/nm_dbus/connection/route.rs
ihuguet: No. Internally, nmstate represents "empty next-hop" as `Some("0.0.0.0")` for IPv4 and `Some("::")` for IPv6. So what I do here is detecting whether it's IPv4 or 6 depending on `dest`, so I can know what the value for "empty next-hop" should be.

Nmstate is doing this because `None` means "ignore next-hop value for matching", not "empty next-hop".

Where can I do the modification at `nm` level? This is where the `NmIpRoute`'s are created, so what I am doing is to create them always with an unified criteria for "default" or "empty" values.

rust/src/lib/nm/nm_dbus/connection/route.rs
ihuguet: If I do that, every time an `NmIpRoute` is created we will have a String allocation, but in most cases it won't be used. As I have done, here I return an `Option<&str>` and only call `to_string()` if it's needed, which is not often because almost all routes have next-hop values.

rust/src/lib/nm/nm_dbus/connection/route.rs
ihuguet: I don't see how to do this at `nm` level. What it's failing is the comparison between `NmIpRoutes`. This is only defined at `nm_dbus` level and it's almost not used outside of `nm_dbus`. Only a few minor usages from `nm/query_apply` and `nm/settings`.

rust/src/lib/nm/nm_dbus/connection/route.rs
ihuguet: Ah, no, I didn't understood your comment. With your change I would use the value of `dest` as "empty next-hop" value, but I have to use IPV6_EMPTY_NEXT_HOP_ADDR / IPV6_EMPTY_NEXT_HOP_ADDR instead.

tests/integration/route_test.py
cathay4t: Please consider use nmstate to reproduce the issue as the reporter did.
Otherwise, move the test case to `tests/integration/nm` folder as nm specific tests which will be dropped when we move away from NM1.

So please consider reproduce this issue use nmstate only.

ihuguet: The bug is caused by NM specific code, it cannot be reliably reproduced with nmstate only. Further modifications to nmstate could make that the integration test always pass, even if the bug were reintroduced.
Moved to /nm subfolder.

ihuguet:
I have changed to a different approach: instead of normalizing to "Some(default_value)", normalize to "None" if the value is the default. This was what we were doing for "weight" and it makes sense to do it also for "table" and "next-hop".

ffmancera (approved):
To me the patch looks good, I did a small test with the reproducer I got and it works.

I prefer this new solution, it makes sense to normalize the defaults to None as weigh was doing.

rust/src/lib/nm/query_apply/route.rs
```diff
@@ -37,14 +35,7 @@ fn is_nm_ip_route_removed(
     new_routes: &[NmIpRoute],
     cur_routes: &[NmIpRoute],
 ) -> bool {
-    for cur_route in cur_routes {
-        let mut cur_route_norm = cur_route.clone();
-        if cur_route_norm.table.is_none() {
-            cur_route_norm.table = Some(DEFAULT_TABLE_ID);
```
liangwen12year, line 43: I added this workaround to prevent the full activation on the existing connection, https://github.com/nmstate/nmstate/pull/2639, but I think that your approach is better, because you convert the predefined default values to None at creation time, which will guarantee that the route table, next-hop will have the consistent default value for routes, therefore, the comparisons are always successful.

liangwen12year (approved):

rust/src/lib/nm/nm_dbus/connection/route.rs
```diff
@@ -26,20 +33,32 @@ pub struct NmIpRoute {
...
-            if weight_num == 0 {
-                weight = None;
-            }
-        }
+        let dest = _from_map!(v, "dest", String::try_from)?;
+
+        let empty_next_hop_val =
+            match dest.as_deref().unwrap_or("").parse::<std::net::IpAddr>() {
+                Ok(std::net::IpAddr::V4(_)) => IPV4_EMPTY_NEXT_HOP_ADDR,
+                Ok(std::net::IpAddr::V6(_)) => IPV6_EMPTY_NEXT_HOP_ADDR,
```
cathay4t: The `nm_dbus` crate should only provide DBUS abstraction without any smart modifications, these smart modification should be done at `nm` level.

```patch
diff --git a/rust/src/lib/nm/query_apply/route.rs b/rust/src/lib/nm/query_apply/route.rs
index d1d6be15..41962016 100644
--- a/rust/src/lib/nm/query_apply/route.rs
+++ b/rust/src/lib/nm/query_apply/route.rs
@@ -3,6 +3,8 @@
 use super::super::nm_dbus::{NmConnection, NmIpRoute};
 
+const EMPTY_NEXT_HOP_ADDRV4: &str = "0.0.0.0";
+const EMPTY_NEXT_HOP_ADDRV6: &str = "::";
 
 pub(crate) fn is_route_removed(
     new_nm_conn: &NmConnection,
@@ -33,16 +35,39 @@ pub(crate) fn is_route_removed(
     )
 }
 
+fn sanitize_nm_ip_route(route: &mut NmIpRoute) {
+    if route.table.is_none() {
+        route.table = Some(DEFAULT_TABLE_ID);
+    }
+    if route.next_hop.as_deref() == Some(EMPTY_NEXT_HOP_ADDRV4)
+        || route.next_hop.as_deref() == Some(EMPTY_NEXT_HOP_ADDRV6)
+    {
+        route.next_hop = None;
+    }
+}
+
 fn is_nm_ip_route_removed(
     new_routes: &[NmIpRoute],
     cur_routes: &[NmIpRoute],
 ) -> bool {
+    let new_routes: Vec<NmIpRoute> = new_routes
+        .iter()
+        .map(|r| {
+            let mut route = r.clone();
+            sanitize_nm_ip_route(&mut route);
+            route
+        })
+        .collect();
+    let cur_routes: Vec<NmIpRoute> = cur_routes
+        .iter()
+        .map(|r| {
+            let mut route = r.clone();
+            sanitize_nm_ip_route(&mut route);
+            route
+        })
+        .collect();
     for cur_route in cur_routes {
-        let mut cur_route_norm = cur_route.clone();
-        if cur_route_norm.table.is_none() {
-            cur_route_norm.table = Some(DEFAULT_TABLE_ID);
-        }
-        if !new_routes.contains(&cur_route_norm) {
+        if !new_routes.contains(&cur_route) {
             return true;
         }
     }
```

cathay4t (changes requested):
The CI test case is still NM specific but the original issue did not use any nmcli command but pure nmstate YAML.
The key reproducer is a route next hop address pointing to a IP not in the CIDR of interface's IP, which NM will additional route with dst 0.0.0.0.

Hence I think the integration test case should be a minimized nmstate only two YAML apply, instead of this nmcli hacks.

rust/src/lib/nm/nm_dbus/connection/route.rs
ihuguet: Fixed all at `nm` level. I've left a small normalization in `nm_dbus`, but it's a condition that should never happen and I've added a warning message so we can notice it it starts happening at some point, so it will be easy to debug.

tests/integration/route_test.py
ihuguet: Changed the test about the next-hop to use nmstate apply only. The test about the table was better using nmcli, I think, and also with the newest changes it cannot be reproduced with nmstate only any more, so it has to remain as NM specific and use nmcli.

tests/integration/conftest.py
```diff
@@ -162,6 +163,12 @@ def eth1_up(test_env_setup):
         yield ifstate
 
 
+@pytest.fixture(scope="function")
+def dummy0_up(test_env_setup):
```
cathay4t: Please store this fixture to `testlib/dummy.py` or `route_test.py`. This `conftest.py` is for fixture used by all test cases.

cathay4t (changes requested):

cathay4t (approved):

Issue #2803, opened 2024-10-18
update zbus crate dependency from v1 to v4 or v5


decathorpe:
The dependency on zbus v1 is very outdated. Fedora currently ships zbus v4 and v1, and upstream released v5 this week.

Please update to v4 or v5 - zbus v1 pulls in tons of obsolete and outdated packages at this point, and I plan to drop zbus v1 from Fedora 42+.

travier:
Looks dependabot is setup on this repo but not looking at Rust crates: https://github.com/nmstate/nmstate/blob/base/.github/dependabot.yml

Fixing this should make regular updates easier.

dustymabe:
@cathay4t - WDYT?

context: https://fedoraproject.org/wiki/Changes/RetireZbusV1

cathay4t:
Thanks. I will create PR to use zbus v4+.

cathay4t:
@decathorpe Hi, what's the progress of zbus V5 in Fedora?
The V4 does not have Clone trait implemented for `zvariant::Value` which is inconvenient to us.

decathorpe:
I have not yet started working on zbus v5 since nothing depends on it yet. But if you would like to make the jump from 1 to 5 directly, I can prioritize that.

cathay4t:
I would like to be the first consumer. It would save me some time on migration from V4 -> V5 in the future, so better bump to latest now.

decathorpe:
Ok, I'll prioritize packaging zbus v5 then.

decathorpe:
zbus v5 is now available in Fedora: https://bodhi.fedoraproject.org/updates/?search=rust-zbus-5.2.0

tyll:
> Looks dependabot is setup on this repo but not looking at Rust crates: [`base`/.github/dependabot.yml](https://github.com/nmstate/nmstate/blob/base/.github/dependabot.yml)
> 
> Fixing this should make regular updates easier.

Could you please report this as a separate issue to make this more visible?

decathorpe:
A "fix" was already proposed here: https://github.com/nmstate/nmstate/pull/2816

PR #2816, opened 2024-11-20
github/dependabot: Add github-actions & Rust crates updates

travier:
github/dependabot: Add github-actions & Rust crates updates

- Enable dependabot update checks for GitHub Actions for GitHub
  Workflows in this repository
- Enable dependabot update checks for Rust dependencies for code in this
  repository

This will make it easier for the progect to keep itself on top of
dependency and GitHub Actions updates.

See: https://github.com/nmstate/nmstate/issues/2803
Signed-off-by: Timothée Ravier <tim@siosm.fr>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,17 @@
 version: 2
 updates:
-- package-ecosystem: pip
-  directory: "/"
-  schedule:
-    interval: daily
-  open-pull-requests-limit: 10
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 3
+  - package-ecosystem: pip
+    directory: "/"
+    schedule:
+      interval: daily
+    open-pull-requests-limit: 10
+  - package-ecosystem: cargo
+    directory: "/rust"
+    schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,7 +1,16 @@
...
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 3
+  - package-ecosystem: pip
+    directory: "/"
+    schedule:
+      interval: daily
+    open-pull-requests-limit: 10
+  - package-ecosystem: cargo
+    schedule:
```
liangwen12year, line 15: According to https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#required-keys, the `directory` field is required for all package-ecosystem entries, including `cargo`.

travier: Thanks, updated.

liangwen12year (commented):
This might be nit picking, but can you elaborate more or summarize the reason for the changes in the commit message, please only consider using links as the reference.

travier:
Updated to hopefully make things clearer.

liangwen12year (approved):
@travier  Thanks for the contribution!

PR #2818, opened 2024-12-02
query_apply: nm: don't do connection down to remove routes

ihuguet:
We were doing connection down & up when a route had to be removed
from an NM profile, instead of doing just a reapply. This was due to
some bugs in NM regarding removing routes:
- https://bugzilla.redhat.com/1837254
- https://issues.redhat.com/browse/RHEL-66262
- https://issues.redhat.com/browse/RHEL-67324

Those bugs are fixed since NM-1.51.5. However, instead of relying in the
version number to know if it's fixed, use the D-Bus VersionInfo
property, as NM is exposing there as a capability whether this bug is
fixed or not. This will facilitate to do the right thing on distros'
downstream builds if they fix the bug in NM via downstream patch.

So, nmstate will do the right thing depending on whether NM is fixed or not.
When setting an interface to down, or setting routes to absent:
- Deactivate and reactivate if NM is not fixed
- Only reapply if NM is fixed

Resolves: https://issues.redhat.com/browse/RHEL-70139

### DIFF
--- a/rust/src/lib/nm/nm_dbus/nm_api.rs
+++ b/rust/src/lib/nm/nm_dbus/nm_api.rs
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::convert::TryFrom;
+use std::str::FromStr;
 use std::time::{Duration, Instant};
 
 use log::debug;
@@ -29,6 +30,18 @@ pub struct NmApi<'a> {
     auto_cp_refresh: bool,
 }
 
+pub struct NmVersionInfo {
+    pub version_encoded: u32,
+    capabilities: Vec<u32>,
+}
+
+#[derive(Eq, PartialEq)]
+pub struct NmVersion {
+    pub major: u8,
+    pub minor: u8,
+    pub micro: u8,
+}
+
 impl NmApi<'_> {
     pub fn new() -> Result<Self, NmError> {
         Ok(Self {
@@ -50,8 +63,14 @@ impl NmApi<'_> {
         self.auto_cp_refresh = value;
     }
 
-    pub fn version(&self) -> Result<String, NmError> {
-        self.dbus.version()
+    pub fn version(&self) -> Result<NmVersion, NmError> {
+        self.dbus.version().and_then(|ver| ver.parse())
+    }
+
+    pub fn version_info(&self) -> Result<NmVersionInfo, NmError> {
+        self.dbus
+            .version_info()
+            .map(|version_info_arr| NmVersionInfo::from(&version_info_arr))
     }
 
     pub fn checkpoint_create(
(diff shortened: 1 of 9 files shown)


ffmancera (commented):
In general, the approach and code looks good to me. But, I have one concern. When would be a good moment to remove this? Could it be possible to add a unit test to do not forget to remove this when the NM minimum required version is greater than the one that contains the fix?

Otherwise, we probably will forget to drop this check.. it is not a big deal.. but it would be nice to have.

liangwen12year:
Please check that reapplying the connection instead of deactivating and activating the connection causes the CI failures on the test `test_remove_specific_route`.

ihuguet:
> In general, the approach and code looks good to me. But, I have one concern. When would be a good moment to remove this? Could it be possible to add a unit test to do not forget to remove this when the NM minimum required version is greater than the one that contains the fix?
> 
> Otherwise, we probably will forget to drop this check.. it is not a big deal.. but it would be nice to have.

AFAIKS the only place where it's checked is in this `check_nm_version`, so if we raise the minimum supported version we will need to change this function. It makes sense that, when you do that, you take a look of the rest of the function. Maybe I can add a comment saying that `route_remove_needs_deactivate` is fixed for all NM versions >= 1.52, so we remember to remove this at that moment. Would that be enough? I don't see what kind of unit test we could add for this.

rust/src/lib/nm/nm_dbus/nm_api.rs
```diff
@@ -29,6 +29,11 @@ pub struct NmApi<'a> {
     auto_cp_refresh: bool,
 }
 
+pub struct NmVersionInfo {
+    version_encoded: u32,
+    capabilities: Vec<u32>,
```
cathay4t, line 35: Can you use `bitflags!` for this? Example https://github.com/rust-netlink/netlink-packet-route/blob/main/src/link/link_flag.rs#L27

The `rust-netlink` has vetted it and since nmstate already depend on rust-netlink, we are not bringing new dependency.

This could make code easier to understand and maintain.

rust/src/lib/nm/query_apply/apply.rs
```diff
@@ -459,22 +466,38 @@ fn gen_nm_conn_need_to_deactivate_first(
...
+            if ver_arr.len() == 2 {
+                nm_ver_encoded = ver_arr[0] << 16 | ver_arr[1] << 8;
             }
         }
+        *route_remove_needs_deactivate = true;
+    }
+
+    if nm_ver_encoded == 0 {
+        log::warn!("Unknown NetworkManager version, expecting >= 1.40");
+    } else if nm_ver_encoded < (1 << 16 | 40 << 8) {
```
cathay4t: Ssome developer are using (major * 1000000 + minor * 1000 + micro), so this 16 and 8 are magic numbers to them.

If you want to improve old `*major < 1 || *minor < 40`. Please do something like:

```
struct NmVersion {
   major: u8
   minor: u8,
   micor: u8,
}

impl Ord for NmVersion {
}

impl TryFrom<&str> for NmVersion {
}
```

cathay4t (changes requested):

rust/src/lib/nm/nm_dbus/nm_api.rs
ihuguet: The problem here is that the capabilities are exposed as a Vec<u32>, each u32 holding some flags. Then, to get the flag number 70, you need to check capabilities[2], bit 7. Bitflags would be great if we only had a single u32, but how to do it with a Vec<u32> of unknown size?

rust/src/lib/nm/query_apply/apply.rs
ihuguet: Done

ihuguet: I'm thinking that, instead of having structs NmVersion and NmVersion info, we could have only NmVersion instead. This NmVersion would hold the capabilities too. I would `NmApi::version_info()` and have only `NmApi::version()`. Then, `NmApi::version()` will try to fetch `VersionInfo` and, if fails, `Version`.

What do you think?

cathay4t:
CI failure is caused by https://issues.redhat.com/browse/RHEL-71686

rust/src/lib/nm/nm_dbus/nm_api.rs
cathay4t: In that case bitflag will not help here.

I am using this https://github.com/rust-netlink/wl-nl80211/blob/main/src/bytes.rs#L30 in rust-netlink.

But you do not need to use it in here, just a reference.

rust/src/lib/nm/nm_dbus/nm_api.rs
```diff
@@ -435,6 +454,93 @@ impl NmApi<'_> {
...
+impl NmVersion {
+    pub fn new(major: u8, minor: u8, micro: u8) -> NmVersion {
+        NmVersion {
+            major,
+            minor,
+            micro,
+        }
+    }
+
+    pub fn encoded(&self) -> u32 {
```
cathay4t, line 494: Optional:

`impl From<u32> for NmVersion`.

rust/src/lib/nm/nm_dbus/nm_api.rs
```diff
@@ -435,6 +454,93 @@ impl NmApi<'_> {
...
+        ((self.major as u32) << 16)
+            | ((self.minor as u32) << 8)
+            | (self.micro as u32)
+    }
+}
+
+impl Ord for NmVersion {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.encoded().cmp(&other.encoded())
```
cathay4t, line 503: Optional to change:

If we have `From<NmVersion> for u32`, we can `u32::from(self).cmp(u32::from(other)`  which is easier to understand.

cathay4t (approved):
Lints can be polished by other PR. Good to me now.

rust/src/lib/nm/nm_dbus/nm_api.rs
ihuguet: I thought about implementing this automatic u32 <-> NmVersion conversions, but I thought that better not to do it. Not any u32 makes sense, it has a specific encoding that is not necessary obvious, so better to be explicit.

PR #2823, opened 2024-12-30
vxlan: Skip serialize destination-port if it is None

liangwen12year:
This commit addresses an issue where the `destination-port` field in VxLAN configurations defaulted to `null` when not explicitly specified, causing serialization and `apply_with_description()` failures. For instance, in the `test_add_vxlan_and_modify_vxlan_id` test, the presence of `null` for `destination-port` resulted in errors such as "Invalid YAML string: interfaces: invalid digit found in string at line 1 column 16." To resolve this, the serialization logic has been updated to completely skip the `destination-port` field if its value is `None`. This ensures that the field is omitted from the serialized output unless explicitly set, preventing invalid configurations and maintaining backward compatibility. The change improves robustness and aligns the serialization behavior with expected standards.

### DIFF
--- a/rust/src/lib/ifaces/vxlan.rs
+++ b/rust/src/lib/ifaces/vxlan.rs
@@ -81,7 +81,8 @@ pub struct VxlanConfig {
     #[serde(
         rename = "destination-port",
         default,
-        deserialize_with = "crate::deserializer::option_u16_or_string"
+        deserialize_with = "crate::deserializer::option_u16_or_string",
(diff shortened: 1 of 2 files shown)


cathay4t (changes requested):
Unit test case required.

liangwen12year:
> Unit test case required.

The unit test has been added. Thanks!

cathay4t (approved):

PR #2828, opened 2025-01-13
test: update the vrf test description

liangwen12year:


rajeshrah22:
The following is the generated state with description from running the test.
```
description: Create the vrf interface test-vrf0 without ports and the vrf route table ID 100
interfaces:
- name: test-vrf0
  type: vrf
  state: up
  vrf:
    port: []
    route-table-id: 100
```

liangwen12year:
LGTM.

PR #2833, opened 2025-01-20
mptcp: fix comparison of flags in is_mptcp_flags_changed()

ffmancera:
When showing and applying a state containing MPTCP flags definition, it was always bringing down and up the interface. This was caused by a wrong comparison in `is_mptcp_flags_changed()` function. This function was checking if the MPTCP flags changed and is used to decide if we can reapply or not.

Fixes: https://issues.redhat.com/browse/RHEL-73700


ffmancera:
I tried to add a test checking the link is stable but MPTCP can only be tested on veth/ethernet.. so we cannot do it. I am looking into adding support for dummy in kernel but that will take longer :)

cathay4t (approved):

PR #2840, opened 2025-02-18
policy: Add --validate option

qinqon:
This change add a --validate option that for now just run the NetworkPolicy serialization and ends with success is this is ok otherwise dump the error.


cathay4t (changes requested):
The validate should be subcommand instead of argument, so we can add more options in the future.

PR #2859, opened 2025-02-28
validate: Check state and policy

qinqon:
Add a `nmstatectl validate` command that will validate state or policy as files or stdin.


cathay4t:
The cli already has [state.rs](http://state.rs/) with helpers of reading state from file. Can you try to use it?

qinqon:
@cathay4t PR ready and integrated with state.rs

Do we need to do deeper validation of policies, or do we do that at follow up ?

cathay4t (approved):

PR #2871, opened 2025-03-24
go: add generating state from policy support to golang bindings

liangwen12year:
Introduce GenerateStateFromPolicy in the Golang bindings to validate YAML files containing nmpolicy. This function behaves similarly to GenerateConfiguration for nmstate YAML, propagating the C binding errors back to the user.

Resolves: https://issues.redhat.com/browse/RHEL-69925

### DIFF
--- a/rust/src/go/nmstate/nmstate.go
+++ b/rust/src/go/nmstate/nmstate.go
@@ -236,3 +236,46 @@ func (n *Nmstate) GenerateConfiguration(state string) (string, error) {
 	}
 	return C.GoString(config), nil
 }
+
+// GenerateStateFromPolicy generates the state from the provided policy and current state.
+// This function returns the state based on the policy and current state provided.
+func (n *Nmstate) GenerateStateFromPolicy(policy, currentState string) (string, error) {
+	var (
+		c_policy       *C.char
+		c_currentState *C.char
+		state          *C.char
+		log            *C.char
+		err_kind       *C.char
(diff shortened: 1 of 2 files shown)


ffmancera:
@qinqon could you take a look to this as it is golang? Thanks!

rust/src/go/nmstate/nmstate.go
```diff
@@ -236,3 +236,46 @@ func (n *Nmstate) GenerateConfiguration(state string) (string, error) {
 	}
 	return C.GoString(config), nil
 }
+
+// GenerateConfigurationFromPolicy generates the configuration from the provided policy and current state.
+// This function returns the configuration files based on the policy and current state provided.
+func (n *Nmstate) GenerateConfigurationFromPolicy(policy, currentState string) (string, error) {
```
qinqon: the "GenerateConfig" comes from the "gc" command at nmstatectl that generate NetworkManager configuration, but this is generate a NMstate witch is different, 

```suggestion
func (n *Nmstate) GenerateStateFromPolicy(policy, currentState string) (string, error) {
```

qinqon (changes requested):

rust/src/go/nmstate/nmstate.go
liangwen12year: I updated it, thanks!

qinqon (approved):
/lgtm

PR #2875, opened 2025-04-01
Python: There is no runtime requirement for setuptools

hroncok:
setuptools|pkg_resources|distutils is only used on build time.

### DIFF
--- a/rust/src/python/requirements.txt
+++ b/rust/src/python/requirements.txt
@@ -6,4 +6,3 @@
 #packageD!=0.13.0,<0.14,>=0.12.0
 
 PyYAML
-setuptools


cathay4t (approved):

cathay4t:
@hroncok Could you add `Signed-off-by: Foo  <foo@example.org>` with your name and email at the bottom of git commit message? It is required for our CI on DCO to pass.

hroncok:
I've edited the file via GitHub web editor. There is no way for me to amend the commit message in the same way. Feel free to do so on my behalf.

hroncok:
I amended the message.

PR #2879, opened 2025-04-06
ip: Support configuring per-device IPv4 sysctl forwarding

liangwen12year:
This feature introduces the ability to configure
net.ipv4.conf.<interface>.forward through nmstate, enabling fine-grained IPv4 forwarding control on individual interfaces. This is especially useful in scenarios such as MetalLB load balancing, where forwarding is needed on specific interfaces without enabling global ip_forward=1.

While sysctl settings can be managed via /etc/sysctl.conf, configuring them for dynamically created or software-defined interfaces (e.g., bridges) is challenging. With this feature, nmstate can declaratively manage per-interface sysctl forwarding, avoiding reliance on fragile or complex dispatch scripts and improving usability in dynamic environments.

Per-interface IPv6 sysctl forwarding is not supported. Enabling it requires turning on the global IPv6 forwarding flag, which can lead to unexpected system behavior (e.g., acting as a router or changing RA handling). To prevent such side effects, nmstate currently supports only IPv4 per-interface forwarding configuration.

Example YAML:

```yml
---
interfaces:
  - name: eth1
    type: ethernet
    state: up
    ipv4:
      address:
        - ip: 192.0.2.251
          prefix-length: 24
      dhcp: false
      enabled: true
      forwarding: true
```

Unit and integration test cases included.

Resolves: https://issues.redhat.com/browse/RHEL-36429

### DIFF
--- a/rust/src/lib/ip.rs
+++ b/rust/src/lib/ip.rs
@@ -103,6 +103,8 @@ struct InterfaceIp {
         rename = "dhcp-custom-hostname"
     )]
     pub dhcp_custom_hostname: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub forwarding: Option<bool>,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Default, Serialize)]
@@ -147,6 +149,12 @@ pub struct InterfaceIpv4 {
     /// which result the IP addresses after first one holding the `secondary`
     /// flag.
     pub addresses: Option<Vec<InterfaceIpAddr>>,
+    /// Whether per‐interface IPv4 sysctl forwarding is enabled.
+    /// When applying with `Some(true)`, nmstate will enable forwarding.
+    /// When applying with `Some(false)`, nmstate will disable forwarding.
+    /// When applying with `None`, the existing sysctl forwarding setting
+    /// is left untouched.
+    pub forwarding: Option<bool>,
     /// Whether to apply DNS resolver information retrieved from DHCP server.
     /// Serialize and deserialize to/from `auto-dns`.
     /// If you want to append static DNS name server before DHCP provided
@@ -299,6 +307,15 @@ impl InterfaceIpv4 {
             }
         }
 
+        if let Some(forwarding) = self.forwarding {
+            if !self.enabled && forwarding && is_desired {
+                log::warn!(
+                    "Ignoring `forwarding: {forwarding}` as IPv4 is disabled",
+                );
+                self.forwarding = None;
+            }
+        }
+
         if let Some(addrs) = self.addresses.as_mut() {
             if is_desired {
                 for addr in addrs.as_slice().iter().filter(|a| a.is_auto()) {
@@ -427,6 +444,7 @@ impl From<InterfaceIp> for InterfaceIpv4 {
             auto_route_metric: ip.auto_route_metric,
             dhcp_send_hostname: ip.dhcp_send_hostname,
             dhcp_custom_hostname: ip.dhcp_custom_hostname,
+            forwarding: ip.forwarding,
             ..Default::default()
         }
     }
@@ -452,6 +470,7 @@ impl From<InterfaceIpv4> for InterfaceIp {
             auto_route_metric: ip.auto_route_metric,
             dhcp_send_hostname: ip.dhcp_send_hostname,
             dhcp_custom_hostname: ip.dhcp_custom_hostname,
+            forwarding: ip.forwarding,
             ..Default::default()
         }
     }
@@ -781,6 +800,12 @@ impl<'de> Deserialize<'de> for InterfaceIpv6 {
                 return Err(serde::de::Error::custom(
                     "dhcp-client-id is not allowed for IPv6",
                 ));
+            }
+            if v_map.contains_key("forwarding") {
+                return Err(serde::de::Error::custom(NmstateError::new(
+                    ErrorKind::InvalidArgument,
+                    "forwarding is not supported for IPv6".to_string(),
+                )));
             }
         }
         let ip: InterfaceIp = match serde_json::from_value(v) {
--- a/rust/src/lib/ip.rs
+++ b/rust/src/lib/ip.rs
@@ -149,11 +149,7 @@ pub struct InterfaceIpv4 {
     /// which result the IP addresses after first one holding the `secondary`
     /// flag.
     pub addresses: Option<Vec<InterfaceIpAddr>>,
-    /// Whether per‐interface IPv4 sysctl forwarding is enabled.
-    /// When applying with `Some(true)`, nmstate will enable forwarding.
-    /// When applying with `Some(false)`, nmstate will disable forwarding.
-    /// When applying with `None`, the existing sysctl forwarding setting
-    /// is left untouched.
+    /// Whether per‐interface IPv4 packets forwarding is enabled.
     pub forwarding: Option<bool>,
     /// Whether to apply DNS resolver information retrieved from DHCP server.
     /// Serialize and deserialize to/from `auto-dns`.
@@ -308,10 +304,13 @@ impl InterfaceIpv4 {
         }
 
         if let Some(forwarding) = self.forwarding {
-            if !self.enabled && forwarding && is_desired {
-                log::warn!(
-                    "Ignoring `forwarding: {forwarding}` as IPv4 is disabled",
-                );
+            if !self.enabled && forwarding {
+                if is_desired {
+                    log::warn!(
+                        "Ignoring `forwarding: {forwarding}` as IPv4 is \
+                         disabled",
+                    );
+                }
                 self.forwarding = None;
             }
         }
@@ -803,7 +802,7 @@ impl<'de> Deserialize<'de> for InterfaceIpv6 {
             }
             if v_map.contains_key("forwarding") {
                 return Err(serde::de::Error::custom(NmstateError::new(
-                    ErrorKind::InvalidArgument,
+                    ErrorKind::NotSupportedError,
                     "forwarding is not supported for IPv6".to_string(),
                 )));
             }
--- a/rust/src/lib/nm/connection.rs
+++ b/rust/src/lib/nm/connection.rs
@@ -23,6 +23,7 @@ pub(crate) fn perpare_nm_conns(
     conn_matcher: &NmConnectionMatcher,
     gen_conf_mode: bool,
     is_retry: bool,
+    forwarding_supported: bool,
 ) -> Result<PerparedNmConnections, NmstateError> {
     let mut nm_conns_to_update: Vec<NmConnection> = Vec::new();
     let mut nm_conns_to_activate: Vec<NmConnection> = Vec::new();
@@ -67,6 +68,32 @@ pub(crate) fn perpare_nm_conns(
             conn_matcher,
             gen_conf_mode,
         )? {
+            // Clear forwarding when not supported by NetworkManager to prevent
+            // failures. TODO: Remove this code once the minimum
+            // supported NetworkManager version is >= 1.54
+            if !forwarding_supported {
+                if let Some(ipv4) = nm_conn.ipv4.as_mut() {
+                    ipv4.forwarding = None;
+                }
+
+                if let Some(desired_iface) = merged_iface.desired.as_ref() {
+                    let forwarding_desired = desired_iface
+                        .base_iface()
+                        .ipv4
+                        .as_ref()
+                        .and_then(|ip4| ip4.forwarding)
+                        .is_some();
+
+                    if forwarding_desired {
+                        log::warn!(
+                            "Clearing unsupported ipv4.forwarding for \
+                             interface '{}'",
+                            desired_iface.name()
+                        );
+                    }
+                }
+            }
+
             if iface.is_up()
                 && (is_retry
                     || !can_skip_activation(
--- a/rust/src/lib/nm/connection.rs
+++ b/rust/src/lib/nm/connection.rs
@@ -86,8 +86,8 @@ pub(crate) fn perpare_nm_conns(
 
                     if forwarding_desired {
                         log::warn!(
-                            "Clearing unsupported ipv4.forwarding for \
-                             interface '{}'",
+                            "Clearing ipv4.forwarding for interface '{}': \
+                             NetworkManager is too old to support this feature",
                             desired_iface.name()
                         );
                     }
--- a/rust/src/lib/nm/nm_dbus/connection/ip.rs
+++ b/rust/src/lib/nm/nm_dbus/connection/ip.rs
@@ -114,6 +114,7 @@ pub struct NmSettingIp {
     pub dhcp_send_hostname: Option<bool>,
     pub dhcp_fqdn: Option<String>,
     pub dhcp_hostname: Option<String>,
+    pub forwarding: Option<i32>,
     _other: HashMap<String, zvariant::OwnedValue>,
 }
 
@@ -156,6 +157,7 @@ impl TryFrom<DbusDictionary> for NmSettingIp {
             )?,
             dhcp_fqdn: _from_map!(v, "dhcp-fqdn", String::try_from)?,
             dhcp_hostname: _from_map!(v, "dhcp-hostname", String::try_from)?,
+            forwarding: _from_map!(v, "forwarding", i32::try_from)?,
             ..Default::default()
         };
 
@@ -236,6 +238,9 @@ impl ToDbusValue for NmSettingIp {
         }
         if let Some(dns_priority) = self.dns_priority {
             ret.insert("dns-priority", zvariant::Value::new(dns_priority));
+        }
+        if let Some(forwarding) = self.forwarding {
+            ret.insert("forwarding", zvariant::Value::new(forwarding));
         }
         if let Some(v) = self.ignore_auto_dns {
             ret.insert("ignore-auto-dns", zvariant::Value::new(v));
--- a/rust/src/lib/nm/query_apply/apply.rs
+++ b/rust/src/lib/nm/query_apply/apply.rs
@@ -21,9 +21,9 @@ use super::super::{
             cur_dns_ifaces_still_valid_for_dns, is_iface_dns_desired,
             purge_global_dns_config, store_dns_config_via_global_api,
         },
-        is_forwarding_supported, is_ipvlan_changed, is_mptcp_flags_changed,
-        is_route_removed, is_veth_peer_changed, is_vlan_changed,
-        is_vrf_table_id_changed, is_vxlan_changed, save_nm_connections,
+        is_ipvlan_changed, is_mptcp_flags_changed, is_route_removed,
+        is_veth_peer_changed, is_vlan_changed, is_vrf_table_id_changed,
+        is_vxlan_changed, save_nm_connections,
     },
     route::store_route_config,
     route_rule::store_route_rule_config,
@@ -45,8 +45,14 @@ pub(crate) async fn nm_apply(
 ) -> Result<(), NmstateError> {
     let mut nm_api = NmApi::new().await.map_err(nm_error_to_nmstate)?;
     let mut nm_route_remove_needs_deactivate = true;
+    let mut forwarding_supported = false;
 
-    check_nm_version(&nm_api, &mut nm_route_remove_needs_deactivate).await;
+    check_nm_version(
+        &nm_api,
+        &mut nm_route_remove_needs_deactivate,
+        &mut forwarding_supported,
+    )
+    .await;
 
     nm_api.set_checkpoint(checkpoint, timeout);
     nm_api.set_checkpoint_auto_refresh(true);
@@ -156,8 +162,6 @@ pub(crate) async fn nm_apply(
             store_dns_config_to_desired_iface(&mut merged_state);
         }
     }
-
-    let forwarding_supported = is_forwarding_supported(&nm_api).await;
 
     let PerparedNmConnections {
         to_store: nm_conns_to_store,
@@ -451,15 +455,19 @@ fn gen_nm_conn_need_to_deactivate_first(
 async fn check_nm_version(
     nm_api: &NmApi<'_>,
     route_remove_needs_deactivate: &mut bool,
+    forwarding_supported: &mut bool,
 ) {
     let version = if let Ok(ver_info) = nm_api.version_info().await {
         *route_remove_needs_deactivate = !ver_info
             .has_capability(NmVersionInfo::CAPABILITY_SYNC_ROUTE_WITH_TABLE);
+        *forwarding_supported =
+            ver_info.has_capability(NmVersionInfo::CAPABILITY_IP4_FORWARDING);
         Ok(ver_info.version())
     } else {
         // VersionInfo was added to NM 1.42. For older version we fallback to
         // parsing from Version string if VersionInfo is not available.
         *route_remove_needs_deactivate = true;
+        *forwarding_supported = false;
         nm_api.version().await
     };
 
--- a/rust/src/lib/nm/query_apply/apply.rs
+++ b/rust/src/lib/nm/query_apply/apply.rs
@@ -21,9 +21,9 @@ use super::super::{
             cur_dns_ifaces_still_valid_for_dns, is_iface_dns_desired,
             purge_global_dns_config, store_dns_config_via_global_api,
         },
-        is_ipvlan_changed, is_mptcp_flags_changed, is_route_removed,
-        is_veth_peer_changed, is_vlan_changed, is_vrf_table_id_changed,
-        is_vxlan_changed, save_nm_connections,
+        is_forwarding_supported, is_ipvlan_changed, is_mptcp_flags_changed,
+        is_route_removed, is_veth_peer_changed, is_vlan_changed,
+        is_vrf_table_id_changed, is_vxlan_changed, save_nm_connections,
     },
     route::store_route_config,
     route_rule::store_route_rule_config,
@@ -156,11 +156,20 @@ pub(crate) async fn nm_apply(
             store_dns_config_to_desired_iface(&mut merged_state);
         }
     }
+
+    let forwarding_supported = is_forwarding_supported(&nm_api).await;
+
     let PerparedNmConnections {
         to_store: nm_conns_to_store,
         to_activate: nm_conns_to_activate,
         to_deactivate: nm_conns_to_deactivate,
-    } = perpare_nm_conns(&merged_state, &conn_matcher, false, is_retry)?;
+    } = perpare_nm_conns(
+        &merged_state,
+        &conn_matcher,
+        false,
+        is_retry,
+        forwarding_supported,
+    )?;
 
     let nm_conns_to_deactivate_first = gen_nm_conn_need_to_deactivate_first(
         &merged_state.interfaces,
--- a/rust/src/lib/nm/query_apply/ip.rs
+++ b/rust/src/lib/nm/query_apply/ip.rs
@@ -4,7 +4,7 @@ use std::collections::HashSet;
 use std::ops::BitXor;
 
 use super::super::nm_dbus::{
-    NmApi, NmIpRouteRuleAction, NmSettingIp, NmSettingIpMethod,
+    NmIpRouteRuleAction, NmSettingIp, NmSettingIpMethod,
 };
 
 use super::dns::nm_dns_to_nmstate;
@@ -311,32 +311,4 @@ fn nm_rules_to_nmstate(
         ret.insert(rule);
     }
     Some(ret)
-}
-
-#[cfg(feature = "query_apply")]
-pub(crate) async fn is_forwarding_supported(nm_api: &NmApi<'_>) -> bool {
-    let version_str = nm_api
-        .version()
-        .await
-        .map(|v| v.to_string())
-        .unwrap_or_default();
-
-    let versions: Vec<&str> = version_str.split('.').collect();
-
-    if versions.len() < 2 {
-        return false;
-    }
-
-    if let (Ok(major), Ok(minor)) =
-        (versions[0].parse::<i32>(), versions[1].parse::<i32>())
-    {
-        major > 1 || (major == 1 && minor >= 54)
-    } else {
-        false
-    }
-}
-
-#[cfg(not(feature = "query_apply"))]
-pub(crate) fn is_forwarding_supported(_nm_api: &NmApi<'_>) -> bool {
-    true
 }
--- a/rust/src/lib/nm/query_apply/ip.rs
+++ b/rust/src/lib/nm/query_apply/ip.rs
@@ -4,7 +4,7 @@ use std::collections::HashSet;
 use std::ops::BitXor;
 
 use super::super::nm_dbus::{
-    NmIpRouteRuleAction, NmSettingIp, NmSettingIpMethod,
+    NmApi, NmIpRouteRuleAction, NmSettingIp, NmSettingIpMethod,
 };
 
 use super::dns::nm_dns_to_nmstate;
@@ -311,4 +311,32 @@ fn nm_rules_to_nmstate(
         ret.insert(rule);
     }
     Some(ret)
+}
+
+#[cfg(feature = "query_apply")]
+pub(crate) async fn is_forwarding_supported(nm_api: &NmApi<'_>) -> bool {
+    let version_str = nm_api
+        .version()
+        .await
+        .map(|v| v.to_string())
+        .unwrap_or_default();
+
+    let versions: Vec<&str> = version_str.split('.').collect();
+
+    if versions.len() < 2 {
+        return false;
+    }
+
+    if let (Ok(major), Ok(minor)) =
+        (versions[0].parse::<i32>(), versions[1].parse::<i32>())
+    {
+        major > 1 || (major == 1 && minor >= 54)
+    } else {
+        false
+    }
+}
+
+#[cfg(not(feature = "query_apply"))]
+pub(crate) fn is_forwarding_supported(_nm_api: &NmApi<'_>) -> bool {
+    true
 }
--- a/rust/src/lib/nm/settings/ip.rs
+++ b/rust/src/lib/nm/settings/ip.rs
@@ -64,6 +64,7 @@ fn gen_nm_ipv4_setting(
     }
     nm_setting.method = Some(method);
     nm_setting.addresses = addresses;
+    nm_setting.forwarding = iface_ip.forwarding.map(i32::from);
     if iface_ip.is_auto() {
         nm_setting.dhcp_timeout = Some(i32::MAX);
         nm_setting.route_metric = iface_ip.auto_route_metric.map(|i| i.into());
--- a/rust/src/lib/unit_tests/ip.rs
+++ b/rust/src/lib/unit_tests/ip.rs
@@ -626,5 +626,5 @@ ipv6:
     let err = result.unwrap_err();
     let err_str = err.to_string();
 
-    assert!(err_str.starts_with("InvalidArgument:"));
+    assert!(err_str.starts_with("NotSupportedError:"));
 }
--- a/rust/src/lib/unit_tests/ip.rs
+++ b/rust/src/lib/unit_tests/ip.rs
@@ -598,3 +598,33 @@ fn test_ip_serlize_allow_extra_address() {
 
     assert_eq!(desired, new);
 }
+
+#[test]
+fn test_forwarding_ipv6_deserialize_error() {
+    let desired = r#"---
+name: eth1
+type: ethernet
+state: up
+ipv6:
+  enabled: true
+  dhcp: false
+  autoconf: false
+  forwarding: true
+  address:
+  - ip: 2001:db8:1::1
+    prefix-length: 64
+"#;
+
+    let value = serde_yaml::from_str::<serde_yaml::Value>(desired).unwrap();
+
+    let result = serde_json::from_value::<Interface>(
+        serde_json::to_value(value).unwrap(),
+    );
+
+    assert!(result.is_err());
+
+    let err = result.unwrap_err();
+    let err_str = err.to_string();
+
+    assert!(err_str.starts_with("InvalidArgument:"));
+}
--- a/tests/integration/static_ip_address_test.py
+++ b/tests/integration/static_ip_address_test.py
@@ -923,8 +923,7 @@ def test_merge_ip_enabled_property_from_current(setup_eth1_static_ip):
 
 @pytest.mark.skipif(
     nm_minor_version() < 54,
-    reason="NetworkManager 1.54- does not support configuring per-device "
-    "IPv4 sysctl forwarding",
+    reason="IPv4 forwarding is not supported on NM 1.54-",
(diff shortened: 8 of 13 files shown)


cgoncalves:
```
     sysctl:
        forwarding: yes
```

I'm wondering whether placing this new field under sysctl is ideal, as it exposes an implementation detail about how IPv4 forwarding is enabled behind the scenes. In NetworkManager, the field is named ipv4.forwarding, with no reference to "sysctl" in the name itself—only in the documentation.

cathay4t:
I agree with @cgoncalves on saving another layer of indent.

rust/src/lib/ip.rs
```diff
@@ -851,6 +872,42 @@ impl From<InterfaceIpv6> for InterfaceIp {
...
+    Default = -1,
+    /// Disable forwarding on this interface.
+    #[serde(rename = "no", alias = "false")]
+    No = 0,
+    /// Enable forwarding on this interface.
+    #[serde(rename = "yes", alias = "true")]
+    Yes = 1,
+    /// Enable forwarding if any shared connection is active.
+    #[serde(rename = "auto")]
+    Auto = 2,
```
cathay4t: Nmstate has no design of shared connection, so this `Forwarding` should be boolean.
Please not use NM design here.

rust/src/lib/nm/query_apply/apply.rs
```diff
@@ -72,6 +72,8 @@ pub(crate) async fn nm_apply(
         }
     }
 
+    let forwarding_supported = is_forwarding_supported(&nm_api).await;
```
cathay4t: You forcing every user to check whether forwarding is supported or not even they are not using this feature.

Instead of check before apply, you should wrap failure error when `connection_add()` for forwarding not supported, so that other user will never been impacted.

cathay4t (changes requested):

tests/integration/static_ip_address_test.py
```diff
@@ -918,6 +922,39 @@ def test_merge_ip_enabled_property_from_current(setup_eth1_static_ip):
     assertlib.assert_state_match(desired_state)
 
 
+@pytest.mark.xfail(
```
cathay4t: not xfail, should be skip.

Please add another test make  sure we fail with proper error on old NM.

cathay4t (changes requested):

rust/src/lib/ip.rs
liangwen12year: I changed the `forwarding` setting to boolean, thanks!

tests/integration/static_ip_address_test.py
liangwen12year: I changed it to `skip` and I also added another test to make sure that the proper error is raised on old NM.

rust/src/lib/nm/query_apply/apply.rs
liangwen12year: I dropped the function `is_forwarding_supported()` and changed it as you suggested, thanks!

tests/integration/static_ip_address_test.py
```diff
@@ -918,6 +923,66 @@ def test_merge_ip_enabled_property_from_current(setup_eth1_static_ip):
...
+            {
+                Interface.NAME: "eth1",
+                Interface.IPV4: {
+                    InterfaceIPv4.FORWARDING: True,
+                },
+            }
+        ]
+    }
+    apply_with_description(
+        "Configure the ethernet device eth1 with IPv4 sysctl forwarding "
```
tyll: This is more verbose than what I expect a human to use since the idea of AI would be to make things simpler which means to type less. Is it better for the AI training to be more verbose, though?

rust/Cargo.toml
```diff
@@ -15,7 +15,7 @@ serde_yaml = "0.9"
 serde = {version = "1.0.137", default-features = false, features = ["derive"]}
 serde_json = { version = "1.0.75", default-features = false }
 nmstate = { path = "src/lib", version = "2.2", default-features = false }
-nispor = "1.2.24"
+nispor = { git = "https://github.com/nispor/nispor.git", rev = "071a2c8c0e406b2248a3082961df7efa057d8aad" }
```
ihuguet: Remember to change this to a version number.

rust/src/lib/ip.rs
```diff
@@ -297,6 +307,17 @@ impl InterfaceIpv4 {
...
+                if !self.enabled {
+                    if is_desired {
+                        log::warn!(
+                            "Ignoring `forwarding: {}` as IPv4 is disabled",
+                            forwarding
+                        );
+                    }
+                    self.forwarding = None;
+                }
+            }
```
ihuguet: This is inside the `if  self.is_auto() {}` so it will be ignored for manual IP config. Additionally, we don't do this check for other properties, so I'd say that it's better not to do it for this one.

rust/src/lib/ip.rs
```diff
@@ -787,6 +810,11 @@ impl<'de> Deserialize<'de> for InterfaceIpv6 {
                     "dhcp-client-id is not allowed for IPv6",
                 ));
             }
+            if v_map.contains_key("forwarding") {
+                return Err(serde::de::Error::custom(
+                    "forwarding is not allowed for IPv6",
```
ihuguet: Suggestion: ~allowed~ supported

rust/src/lib/nm/error.rs
```diff
@@ -37,6 +37,22 @@ pub(crate) fn nm_error_to_nmstate(nm_error: NmError) -> NmstateError {
...
         }
+        NmErrorKind::Connection(NmConnectionError::InvalidProperty) => {
+            if nm_error.msg.starts_with("ipv4.forwarding") {
+                NmstateError::new(
+                    ErrorKind::NotSupportedError,
+                    format!(
+                        "Please upgrade NetworkManager to 1.54 or above to \
+                        support IPv4 sysctl forwarding as indicated by: {nm_error}",
+                    ),
+                )
```
ihuguet: This is not very reliable. It is good to explain that the problem is the NM version. However, this message will be printed for all errors related to ipv4.forwarding, so if it's a different cause it will confuse the user, instead of helping him.
There is no good way to properly detect this, other than [this](https://github.com/nmstate/nmstate/blob/2d8e556f931baf10f20db1aee9fd48c5fb5724c3/rust/src/lib/nm/query_apply/apply.rs#L491), but it's a bit overkill in this case. I would just show the error message returned by NM, as it is done by default, and that's all.

rust/src/lib/nm/error.rs
```diff
@@ -37,6 +37,22 @@ pub(crate) fn nm_error_to_nmstate(nm_error: NmError) -> NmstateError {
...
+            } else {
+                NmstateError::new(
+                    ErrorKind::Bug,
+                    format!("{}: {}", nm_error.kind, nm_error.msg),
+                )
+            }
```
ihuguet: This is changing the error type and message that are used now, which are `            NmstateError::new(ErrorKind::InvalidArgument, nm_error.to_string())`. Either allow this to fallback to the existing lines below, or generate the same error.

rust/src/lib/nm/profile.rs
```diff
@@ -74,6 +91,43 @@ pub(crate) fn perpare_nm_conns(
...
+                .and_then(|ip4| {
+                    ip4.forwarding.and_then(|val| match val {
+                        0 => Some(false),
+                        1 => Some(true),
+                        _ => None,
+                    })
+                });
+
+            // Remove ipv4.forwarding if interface did NOT explicitly request it,
+            // or if the desired setting matches the current (i.e., no change needed)
```
ihuguet: Why do we need to do this? It is not common in nmstate at all, IMO. In an nmstate managed interface we can just force a `forwarding` value to the NM connection. If it's specified in the desired state, use it; if not, take it from the current state.

Or I am missing something? @cathay4t what do you think?

rust/src/lib/nm/query_apply/ip.rs
```diff
@@ -183,6 +184,17 @@ fn nm_dhcp_duid_to_nmstate(nm_setting: &NmSettingIp) -> Option<Dhcpv6Duid> {
     }
 }
 
+pub(crate) fn nm_forwarding_to_nmstate(nm: &NmSettingIp) -> Option<bool> {
+    nm.forwarding.and_then(|val| match val {
+        0 => Some(false),
+        1 => Some(true),
```
ihuguet: Please add some constant variables to the top of the file like FORWARDING_YES and use them here.

rust/src/lib/nm/query_apply/ip.rs
```diff
@@ -183,6 +184,17 @@ fn nm_dhcp_duid_to_nmstate(nm_setting: &NmSettingIp) -> Option<Dhcpv6Duid> {
...
+        other => {
+            log::debug!("ignoring unsupported NM forwarding mode {}", other);
+            None
+        }
```
ihuguet: Note that the most common value configured in NM connections will be `FORWARDING_DEFAULT (-1)`. Less common, but also likely to be configured `FORWARDING_AUTO (2)`. With this, a debug message will be shown in those cases.
What I would do is to silently convert those to None, and emit a warning if we receive unknown values, like it is done in other places of the file.

rust/src/lib/nm/show.rs
```diff
@@ -151,9 +155,18 @@ pub(crate) async fn nm_retrieve(
                 } else {
                     None
                 };
-                if let Some(iface) =
+                if let Some(mut iface) =
                     iface_get(nm_dev, nm_conn, nm_saved_conn, lldp_neighbors)
                 {
+                    // Clear forwarding when unsupported by NetworkManager to prevent failures
+                    // if the user tries to apply the returned state as-is.
```
ihuguet: Please add a `TODO` comment requesting to remove this code when the minimum NM supported version is >= 1.54

ihuguet (changes requested):

rust/src/lib/ip.rs
```diff
@@ -146,6 +148,14 @@ pub struct InterfaceIpv4 {
     /// which result the IP addresses after first one holding the `secondary`
     /// flag.
     pub addresses: Option<Vec<InterfaceIpAddr>>,
+    /// Whether per‐interface IPv4 sysctl forwarding is enabled.
+    /// When applying with `Some(true)`, nmstate will set
+    /// `net.ipv4.conf.<interface>.forward = 1` (enable forwarding).
+    /// When applying with `Some(false)`, nmstate will set
+    /// `net.ipv4.conf.<interface>.forward = 0` (disable forwarding).
```
cathay4t: No need to mention internal implementation in API document. Just state `enable IPv4 forwarding` is enough.

rust/src/lib/ip.rs
cathay4t: We should not raise warning when `forwarding: false` with `enabled: false`.

rust/src/lib/ip.rs
cathay4t: +1

rust/src/lib/nm/error.rs
cathay4t: I agree on showing NM error message unmodified.

rust/src/lib/nm/profile.rs
cathay4t:
```rust
nm_setting.forwarding = iface_ip.forwarding.map(i32::from)
```

Single line is enough, nmstate already has code handling the merging stuff.

rust/src/lib/nm/query_apply/ip.rs
```diff
@@ -74,6 +74,7 @@ pub(crate) fn nm_ip_setting_to_nmstate4(
             } else {
                 None
             },
+            forwarding: nm_forwarding_to_nmstate(nm_ip_setting),
```
cathay4t: NM should not provide this value, nispor should. If you are querying this from NM, then the verification is doing nothing.

rust/src/lib/nm/query_apply/ip.rs
cathay4t: This function should not exist. We should depend on nispor for querying kernel, not NM.

rust/src/lib/nm/settings/ip.rs
```diff
@@ -64,6 +64,7 @@ fn gen_nm_ipv4_setting(
     }
     nm_setting.method = Some(method);
     nm_setting.addresses = addresses;
+    nm_setting.forwarding = iface_ip.forwarding.map(|f| f as i32);
```
cathay4t: trivial:

IMHO `as` is dangerous as it has integer wrapping. Yes, `bool as i32` is safe, but better using `i32::from` which is safer.

rust/src/lib/unit_tests/ip.rs
```diff
@@ -566,3 +566,26 @@ fn test_ip_serlize_allow_extra_address() {
...
+  forwarding: true
+  address:
+  - ip: 2001:db8:1::1
+    prefix-length: 64
+"#,
+    );
+
+    assert!(result.is_err());
+    let err_msg = format!("{}", result.unwrap_err());
+    assert!(err_msg.contains("forwarding is not allowed for IPv6"),);
```
cathay4t: error message is not API. Check `error.kind` instead.

cathay4t (changes requested):
NM should not provide query support as it override nispor data which is incorrect for verifying this kernel property. NM should only override userspace settings in query.

rust/src/lib/nm/query_apply/ip.rs
ihuguet: But nm_ip_setting_to_nmstate4 exists. What is it used for?

rust/Cargo.toml
liangwen12year: I will change it once Nispor has the new release.

tests/integration/static_ip_address_test.py
liangwen12year: You’re right that, on the user side, brevity and clarity are usually best; humans don’t want to type or read more than necessary. But “verbosity” in AI training has a different trade-off:

More context can help the model learn nuance.  Having examples that are richly detailed lets the model pick up on subtleties of style, tone, and structure.

But unnecessary fluff isn’t ideal. AI training data is expensive to process, and overly wordy passages with no new information just waste compute and can teach the model to ramble.

The sweet spot is “concise completeness.” Give the model enough information to learn the pattern we care about, but don’t pad with off-topic sentences.

So in short: it’s not that “more verbose is always better” for training, but rather “rich, clear, context-rich” data is best,  it shouldn’t be longer than it needs to be to convey its full meaning.

tyll: Was this written by AI?

liangwen12year: Not really, I composed it in my own words.

rust/src/lib/nm/error.rs
liangwen12year: I changed it to fall back to the existing lines below. Honestly, I think that it is fine to keep it as `ErrorKind::Bug` when there are other `NmConnectionError::InvalidProperty` triggered but not related to `ipv4.forwarding`,  which means that there are some setting conflicts that are captured by NM but not properly validated in Nmstate yet.

rust/src/lib/ip.rs
liangwen12year: I changed it to raise a warning when setting the `forwarding: true` with `enabled: false`.

rust/src/lib/nm/profile.rs
liangwen12year: Since nmstate already handles merging stuff, it is possible that nmstate will apply the forwarding that is actually retrieved by nispor. The intention of having the code here is to drop the forwarding setting when it is not explicitly desired by the user, or the forwarding for apply is equal to the current forwarding in the connection, so that we are not risking applying the forwarding in a system of which the NM does not support forwarding yet. The idea is similar to the commit f38da12c4 (mptcp: Fix mptcp error on older NetworkManager).  And if we drop the code here, the integration tests will run into error `ipv4.forwarding: unknown property` in a lot of places.

rust/src/lib/nm/query_apply/ip.rs
liangwen12year: Thanks for pointing it out, I dropped the code.

rust/src/lib/nm/show.rs
liangwen12year: I added the TODO comment.

rust/src/lib/nm/query_apply/ip.rs
liangwen12year: I removed this function, thanks!

rust/src/lib/unit_tests/ip.rs
liangwen12year: I changed it to check `error.kind` instead. 

Notice that `serde_yaml::Error` does not have a `kind` field as `NmstateError`, thus I have to wrap the NmstateError inside the `serde_yaml::Error` as follows: 

```
impl<'de> Deserialize<'de> for InterfaceIpv6 {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: Deserializer<'de>,
    {
        let v = serde_json::Value::deserialize(deserializer)?;

        if let Some(v_map) = v.as_object() {
            if v_map.contains_key("dhcp_client_id") {
                return Err(serde::de::Error::custom(
                    "dhcp-client-id is not allowed for IPv6",
                ));
            }
            if v_map.contains_key("forwarding") {
                return Err(serde::de::Error::custom(NmstateError::new(
                    ErrorKind::InvalidArgument,
                    "forwarding is not supported for IPv6".to_string(),
                )));
            }
        }
        let ip: InterfaceIp = match serde_json::from_value(v) {
            Ok(i) => i,
            Err(e) => {
                return Err(serde::de::Error::custom(format!("{e}")));
            }
        };
        let ret = Self::from(ip);
        Ok(ret)
    }
}
```

tests/integration/static_ip_address_test.py
tyll: So what is the conclusion that you draw?

liangwen12year: For AI training, the training data should be as detailed as needed to fully convey the intended meaning, but no longer than necessary.

liangwen12year: `Enable IPv4 forwarding on eth1` is better as the description.

rust/src/lib/nm/nm_dbus/connection/ip.rs
```diff
@@ -167,6 +168,9 @@ impl TryFrom<DbusDictionary> for NmSettingIp {
             setting.dns = _from_map!(v, "dns", parse_nm_dns)?;
         }
 
+        if v.contains_key("forwarding") {
+            setting.forwarding = _from_map!(v, "forwarding", i32::try_from)?;
```
cathay4t: The `_from_map!` already has stuff like `contains_key()`, please move this `_from_map!()` to creation of `NmSettingIp` instead of modify after creation.

rust/src/lib/query_apply/net_state.rs
```diff
@@ -119,9 +119,30 @@ impl NetworkState {
...
+                self.interfaces
+                    .user_ifaces
+                    .values_mut()
+                    .chain(self.interfaces.kernel_ifaces.values_mut())
+                    .for_each(|iface| {
+                        if let Some(ipv4) = iface.base_iface_mut().ipv4.as_mut()
+                        {
+                            ipv4.forwarding = None;
+                        }
+                    });
```
cathay4t: Please do not put merging code at this level, otherwise code are spreading everywhere in the project. Nmstate has 100+ properties, why none of them need merge code at `rust/src/lib/query_apply/net_state.rs`?

Can you spend on sometime understand how nmstate merging nispor query output with nm_retrieve query output?

After that you will understand how this forwarding should be merged.

Meanwhile, there are a lot merged patch adding properties to nmstate schema, they could be very good example on how to do it.

rust/src/lib/unit_tests/ip.rs
```diff
@@ -566,3 +566,33 @@ fn test_ip_serlize_allow_extra_address() {
...
+    let result = serde_json::from_value::<Interface>(
+        serde_json::to_value(value).unwrap(),
+    );
+
+    assert!(result.is_err());
+
+    let err = result.unwrap_err();
+    let err_str = err.to_string();
+
+    assert!(err_str.starts_with("InvalidArgument:"),);
```
cathay4t: Again, error message is not API, only check `NmstateError.kind`

tests/integration/dynamic_ip_test.py
```diff
@@ -1501,6 +1501,7 @@ def clean_state():
         if iface_state.get(Interface.IPV4, {}).get(InterfaceIPv4.ENABLED):
             iface_state[Interface.IPV4][InterfaceIPv4.AUTO_DNS] = False
             iface_state[Interface.IPV4][InterfaceIPv4.AUTO_ROUTES] = False
+            iface_state[Interface.IPV4].pop(InterfaceIPv4.FORWARDING, None)
```
cathay4t: It is very important for nmstate to support applying state produced by `nmstatectl show`.

There are two options:
A: Only show `forwarding` in query when NM support so.
B: NM discard `forwarding` settings when desired but not support, let verification stage found the error.

I prefer option B, with that, you don't need to change existing test codes.

cathay4t (changes requested):

rust/src/lib/nm/error.rs
ihuguet: If we show the unmodified message from NM, there is no need to check if the error is for ipv4.forwarding or not. The only difference is that you are using `ErrorKind::NotSupportedError` but, again, we don't know if that's the actual cause or not. We only know that the error is related to forwarding, nothing else.
Better just remove this change.

tests/integration/static_ip_address_test.py
tyll: What's the best approach to ensure that the AI accepts a variety of prompts for the state and is able to generate descriptions at the desired verbosity? The verbosity is probably something that depends on the user's experience. This discussion should probably move to a different place, please suggest one and reference it here.

rust/src/lib/unit_tests/ip.rs
liangwen12year: Do you suggest validating the `ipv6.forwarding` in the `sanitize()` function?

rust/src/lib/nm/nm_dbus/connection/ip.rs
liangwen12year: Thanks, I changed the code as you suggested.

rust/src/lib/unit_tests/ip.rs
ihuguet: I think that the original `assert!(err_msg.contains("forwarding is not allowed for IPv6"),);` was fine. As the returned error is `serde_yaml::Error` we cannot check the `ErrorKind`. Moreover, the error kind doesn't tell us whether it is because of the forwarding property or because other reason. The test needs to ensure that the error is the one that we expected.

Asserting the string `"forwarding is not allowed for IPv6"` is not ideal either, but I cannot think of a better way of doing it. It is not that bad because at least it checks that the helpful message that we've added is printed, and if we change the message in the future it will only affect to this single test and we can easily adjust it at the same time.

tests/integration/static_ip_address_test.py
liangwen12year: We can create a system using a Random Forest classifier to train on a labeled description of low/medium/high verbosity examples with features like token count, sentence count, type–token ratio, readability scores (e.g. Flesch–Kincaid), n‑gram TF‑IDF vectors, and syntactic depth. The Random Forest classifier will take a description and predict its verbosity tier.

After creating the Random Forest classifier, we can use LLM to generate the description based on the verbosity level. The LLM’s output is then re‑classified by the same Random Forest validator: if it matches the intended verbosity, the description is returned; if not, the pipeline loops back to the LLM (optionally with corrective feedback) until the output passes the random forest classifier check.

The workflow is:

```
       [ State + Verbosity ]
                |
                v
        +------------------+
        |  LLM Generator   | <-----------\
        +------------------+              |
                |                        |
                v                         |
   +--------------------------+          |
   |  Generated Description   |          |
   +--------------------------+          |
                |                        |
                v                        |
        +------------------+             |
        |   RF Validator   |             |
        +------------------+             |
          |            |                |
       Yes|            |No               |
          v            |                 |
  [Return to User]    \-----------------/
  ``` 
  
  
  I will include more details in a discussion page.

Comment (tyll) [reply tests/integration/static_ip_address_test.py]: We can only influence AIs with training data not with changing the pipeline.

**Comment by liangwen12year** [comment]
Hi @cathay4t,  I discussed with @ihuguet  about the best way to check the supportability of forwarding. We agree that the best way to do it is via checking a flag added to NMVersionInfoCapability, similar to the `check_nm_version`  implementation in this patch, https://github.com/nmstate/nmstate/commit/d6589384b159e85c54428af744a552f5340ae668. What do you think about this approach? We do need to add the flag on the NM side first, but that should be easy.

File: rust/src/lib/nm/query_apply/apply.rs [hunk 22]
```diff
@@ -136,6 +136,22 @@ pub(crate) async fn nm_apply(
...
+            .and_then(|d| d.base_iface().ipv4.as_ref())
+            .and_then(|ip4| ip4.forwarding)
+            .is_some()
+    });
+
+    let forwarding_supported = if forwarding_desired {
+        is_forwarding_supported(&nm_api).await
+    } else {
+        true
+    };
```
ihuguet: Instead of this, modify `check_nm_version` and check there if forwarding is supported or not. Also, we should do it by checking the capabilities, and not by version number. Otherwise we cannot backport this to RHEL.

ihuguet:
IMO the changes in lib/nm/show.rs and lib/query_apply/* are not needed, nor desired. Whether NetworkManager supports forwarding or not is something to handle in /lib/nm files only. The only thing that we cannot do is to apply states with `forwarding` with NetworkManager.

liangwen12year:
> IMO the changes in lib/nm/show.rs and lib/query_apply/* are not needed, nor desired. Whether NetworkManager supports forwarding or not is something to handle in /lib/nm files only. The only thing that we cannot do is to apply states with `forwarding` with NetworkManager.

I think that one important functionality for nmstate is to support applying the state produced by `nmstatectl show`, otherwise, if we drop the changes in lib/nm/show.rs and lib/query_apply/*, there will be a lot of CI integration test failures triggered, due to the fact that the desired state contains the forwarding setting, but  the forwarding setting is dropped when applying states with `forwarding` in NetworkManager.

rust/src/lib/nm/query_apply/apply.rs
liangwen12year: @cathay4t , what do you think about this approach? I think that we do need to add the flag on the NM side first, but that should be easy.

ihuguet:
> I think that one important functionality for nmstate is to support applying the state produced by `nmstatectl show`, otherwise, if we drop the changes in lib/nm/show.rs and lib/query_apply/*, there will be a lot of CI integration test failures triggered, due to the fact that the desired state contains the forwarding setting, but the forwarding setting is dropped when applying states with `forwarding` in NetworkManager.

No:
1. You get `forwarding: true` in `nmstatectl show`.
2. When applying, you discard `forwarding` at `nm` level, because NM doesn't support it.
3. NM doesn't change the forwarding, so it stays at `true`, so it passes the verification (unless I'm missing something).

However, if you try to change the forwarding in the desired state, you will get a warning saying that it is discarded, and then a verification error because the final state doesn't match with the desired state. This is fine, IMHO, because you tried to apply a config that we couldn't apply.

liangwen12year:
> > I think that one important functionality for nmstate is to support applying the state produced by `nmstatectl show`, otherwise, if we drop the changes in lib/nm/show.rs and lib/query_apply/*, there will be a lot of CI integration test failures triggered, due to the fact that the desired state contains the forwarding setting, but the forwarding setting is dropped when applying states with `forwarding` in NetworkManager.
> 
> No:
> 
> 1. You get `forwarding: true` in `nmstatectl show`.
> 2. When applying, you discard `forwarding` at `nm` level, because NM doesn't support it.
> 3. NM doesn't change the forwarding, so it stays at `true`, so it passes the verification (unless I'm missing something).
> 
> However, if you try to change the forwarding in the desired state, you will get a warning saying that it is discarded, and then a verification error because the final state doesn't match with the desired state. This is fine, IMHO, because you tried to apply a config that we couldn't apply.

Thanks, I adapted the code as you suggested.

rust/src/lib/nm/query_apply/apply.rs
cathay4t: I think we should do the capacity check instead of version check.

rust/Cargo.toml
cathay4t: I have release nispor 1.2.25. You may use it now.

rust/src/lib/ip.rs
```diff
@@ -782,6 +801,12 @@ impl<'de> Deserialize<'de> for InterfaceIpv6 {
                     "dhcp-client-id is not allowed for IPv6",
                 ));
             }
+            if v_map.contains_key("forwarding") {
+                return Err(serde::de::Error::custom(NmstateError::new(
+                    ErrorKind::InvalidArgument,
```
cathay4t: `ErrorKind::NotSupportedError`

rust/src/lib/nm/connection.rs
```diff
@@ -67,6 +68,32 @@ pub(crate) fn perpare_nm_conns(
...
+                    let forwarding_desired = desired_iface
+                        .base_iface()
+                        .ipv4
+                        .as_ref()
+                        .and_then(|ip4| ip4.forwarding)
+                        .is_some();
+
+                    if forwarding_desired {
+                        log::warn!(
+                            "Clearing unsupported ipv4.forwarding for \
```
cathay4t: This sounds like nmstate does not support ipv4.forwarding, could you rephrase it indicate NetworkManager too old?

tests/integration/static_ip_address_test.py
```diff
@@ -918,6 +922,78 @@ def test_merge_ip_enabled_property_from_current(setup_eth1_static_ip):
...
+        "Disable IPv4 forwarding on eth1",
+        desired_state,
+    )
+    assertlib.assert_state_match(desired_state)
+
+
+@pytest.mark.skipif(
+    nm_minor_version() >= 53,
+    reason="On NM < 1.53, attempting to set per-interface IPv4 forwarding "
+    "should raise an error",
```
cathay4t: Reason="IPv4 forwarding  not supported on NM 1.52-"

tests/integration/static_ip_address_test.py
```diff
@@ -918,6 +922,78 @@ def test_merge_ip_enabled_property_from_current(setup_eth1_static_ip):
     assertlib.assert_state_match(desired_state)
 
 
+@pytest.mark.skipif(
+    nm_minor_version() < 54,
+    reason="NetworkManager 1.54- does not support configuring per-device "
```
cathay4t: Reason="IPv4 forwarding not supported on NM 1.52-"

cathay4t (changes requested):
Just need some trivial changes

tests/integration/static_ip_address_test.py
liangwen12year: I think that the IPv4 forwarding is not supported on NM 1.54-, as you can tell from https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/ea6af6d8066673e39e287c19aa3cbde47c707f47.

rust/src/lib/nm/connection.rs
liangwen12year: I rephrased it, thanks !

rust/src/lib/ip.rs
liangwen12year: updated, thanks!

tests/integration/static_ip_address_test.py
cathay4t: My bad. My suggestion  should be `Reason=IPv4 forwarding is not supported on NM 1.54-` which is simpler string than yours.

Just a lint.

cathay4t (dismissed):

tests/integration/static_ip_address_test.py
```diff
@@ -918,6 +922,76 @@ def test_merge_ip_enabled_property_from_current(setup_eth1_static_ip):
...
+        desired_state,
+    )
+    assertlib.assert_state_match(desired_state)
+
+
+@pytest.mark.skipif(
+    nm_minor_version() >= 54,
+    reason="IPv4 forwarding is not supported on NM 1.54-",
+)
+def test_ipv4_sysctl_forwarding_unsupported_error(setup_eth1_static_ip):
```
cathay4t, line 965: The condition says `>=54`, but the reason says `not supported.


Please also add test case for applying current forwarding value to old NM and nmstate still pass.

cathay4t (changes requested):
Got verification failure when applying this YAML:

```yml
---
interfaces:
  - name: eth1
    type: ethernet
    state: up
    ipv4:
      enabled: false
      forwarding: true
```

rust/src/lib/ip.rs
```diff
@@ -147,6 +149,12 @@ pub struct InterfaceIpv4 {
     /// which result the IP addresses after first one holding the `secondary`
     /// flag.
     pub addresses: Option<Vec<InterfaceIpAddr>>,
+    /// Whether per‐interface IPv4 sysctl forwarding is enabled.
+    /// When applying with `Some(true)`, nmstate will enable forwarding.
+    /// When applying with `Some(false)`, nmstate will disable forwarding.
+    /// When applying with `None`, the existing sysctl forwarding setting
+    /// is left untouched.
```
bengal: `None` leaves the value in the connection set to "default", which corresponds (in absence of a global default) to "auto". "auto" touches the sysctl in some cases, as it enables forwarding in "shared" mode.

liangwen12year:
> Got verification failure when applying this YAML:
> 
> ```yaml
> ---
> interfaces:
>   - name: eth1
>     type: ethernet
>     state: up
>     ipv4:
>       enabled: false
>       forwarding: true
> ```

Do you suggest that we should raise an InvalidArgument error instead ?

rust/src/lib/ip.rs
liangwen12year: how about changing it to: 

```
    /// When applying with `None`, forwarding uses NetworkManager's current
    /// or default setting ("auto"), which may result in NetworkManager
    /// automatically enabling or disabling sysctl forwarding based on the
    /// connections configuration (e.g., enabling it for "shared" mode).
```

liangwen12year:
> Got verification failure when applying this YAML:
> 
> ```yaml
> ---
> interfaces:
>   - name: eth1
>     type: ethernet
>     state: up
>     ipv4:
>       enabled: false
>       forwarding: true
> ```

I fixed the verification failure, thanks!

tests/integration/static_ip_address_test.py
liangwen12year: I reworded the reasons, and I also added the test case for applying the current forwarding value to the old NM.

rust/src/lib/ip.rs
```diff
@@ -147,6 +149,14 @@ pub struct InterfaceIpv4 {
     /// which result the IP addresses after first one holding the `secondary`
     /// flag.
     pub addresses: Option<Vec<InterfaceIpAddr>>,
+    /// Whether per‐interface IPv4 sysctl forwarding is enabled.
+    /// When applying with `Some(true)`, nmstate will enable forwarding.
+    /// When applying with `Some(false)`, nmstate will disable forwarding.
+    /// When applying with `None`, forwarding uses NetworkManager's current
+    /// or default setting ("auto"), which may result in NetworkManager
+    /// automatically enabling or disabling sysctl forwarding based on the
+    /// connections configuration (e.g., enabling it for "shared" mode).
```
ihuguet: I would remove all these lines, except the first one. True and false are obvious, and None just have the same meaning than anywhere else: don't touch current value.

rust/src/lib/ip.rs
```diff
@@ -299,6 +309,18 @@ impl InterfaceIpv4 {
             }
         }
 
+        if let Some(forwarding) = self.forwarding {
+            if !self.enabled && forwarding {
```
ihuguet, line 307: Maybe just `if !self.enabled`? It neither makes much sense to define `false`, doesn't it?

tests/integration/static_ip_address_test.py
```diff
@@ -918,6 +922,108 @@ def test_merge_ip_enabled_property_from_current(setup_eth1_static_ip):
...
+                },
+            }
+        ]
+    }
+    with pytest.raises(libnmstate.error.NmstateVerificationError):
+        apply_with_description(
+            "Configure the ethernet device eth1 with IPv4 sysctl forwarding "
+            "enabled",
+            desired_state,
+        )
```
ihuguet: I think we have to check that the verification error is due to the right reason. For that, maybe we can only check the content of the error message, which is a bit ugly but I think it's OK in this case. If we change the error message we'll just need to adapt a bit a single test.

rust/src/lib/ip.rs
liangwen12year: Since the corresponding IP family is disabled, it’s appropriate to define forwarding: false for the device. This makes the configuration explicit, though omitting the field is also valid since forwarding cannot be enabled while the family is disabled.

rust/src/lib/ip.rs
liangwen12year: @bengal , what do you think ?

cathay4t: This is top level API, it should contains implementation detail of NetworkManager.

No need to explains the detail,  meaning of `Some(true)`,  `Some(false)`  and `None` are obvious.

tests/integration/static_ip_address_test.py
cathay4t: We may check whether the error message contains `interfaces.eth1.ipv4.forwarding`.

liangwen12year: I changed the test to check the error message instead, thanks!

rust/src/lib/ip.rs
liangwen12year: @cathay4t , is there a typo,  it should or shoud not contain implementation details of NetworkManager?

rust/src/lib/ip.rs
```diff
@@ -308,10 +304,13 @@ impl InterfaceIpv4 {
...
-                    "Ignoring `forwarding: {forwarding}` as IPv4 is disabled",
-                );
+            if !self.enabled && forwarding {
+                if is_desired {
+                    log::warn!(
+                        "Ignoring `forwarding: {forwarding}` as IPv4 is \
+                         disabled",
+                    );
+                }
                 self.forwarding = None;
```
ihuguet: Why do we change self.forwarding if !is_desired?

liangwen12year: If we did not change it, applying the following will fail the verification. The verification process compares the sanitized desired state with the sanitized current state. If sanitization rules aren't applied consistently, false verification failure will occur.

```
---
interfaces:
  - name: eth1
    type: ethernet
    state: up
    ipv4:
      enabled: false
      forwarding: true
```

ihuguet (approved):

cathay4t (approved):

PR #2883, opened 2025-04-11
policy: Support regex by adding '=~' operator

qinqon:


qinqon:
This should be `=~` since this is compare with regex at other places like "bash".

cathay4t:
I don't want to introduce big regex as dependency to nmstate unless import use case require so.

For use case you mentioned in test, I prefer syntax like `foo: interface.name == "igb" | "tg4"`

qinqon:
> I don't want to introduce big regex as dependency to nmstate unless import use case require so.
> 
> For use case you mentioned in test, I prefer syntax like `foo: interface.name == "igb" | "tg4"`

Closing, since it introduce a quite big dependency for a small use case

PR #2900, opened 2025-05-22
route rule: ensure idempotency for USE_DEFAULT_PRIORITY

ihuguet:
When the desired state has route rules without a priority defined,
nmstate creates them with increasing values of priority. However,
this was causing that applying the same desired state twice created
the rule twice with different priorities. As a result, the operation
was not idempotent.

Fix it by searching for a matching rule in the current state and, if
there is any, use its priority value to fill the priority value of the
desired rule if it's empty. As a result, new rules are not created if
one that matches already exists.

Fix also a helper function of the integration tests to make it asserting
that the rule is not duplicated.

Signed-off-by: Íñigo Huguet <ihuguet@redhat.com>

### DIFF
--- a/tests/integration/testlib/iprule.py
+++ b/tests/integration/testlib/iprule.py
@@ -37,24 +37,26 @@ def ip_rule_exist_in_os(rule):
     logging.debug(f"Current ip rules in OS: {result[1]}")
     assert result[0] == 0
     current_rules = json.loads(result[1])
-    found = True
+    found = False
     for rule in current_rules:
-        if rule.get("src") == "all" or rule.get("dst") == "all":
-            continue
+        logging.debug(f"Checking ip rule is OS: {rule}")
+        found = True
 
         if rule.get("table") == "main":
             rule["table"] = f"{iplib.KERNEL_MAIN_ROUTE_TABLE_ID}"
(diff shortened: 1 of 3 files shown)


tests/integration/testlib/iprule.py
```diff
@@ -37,24 +37,26 @@ def ip_rule_exist_in_os(rule):
...
-        if ip_from and ip_from != iplib.to_ip_address_full(
-            rule["src"], rule.get("srclen")
-        ):
+        if ip_from is not None and rule["src"] != ip_from:
             found = False
             continue
-        if ip_to and ip_to != iplib.to_ip_address_full(
-            rule["dst"], rule.get("dstlen")
-        ):
+        if ip_to is not None and rule.get("dst") != ip_to:
```
liangwen12year, line 60: Nit, this patch's commit message is a bit hard to understand, maybe rephrase to: 

```
Prior to this change, any kernel rule with src=all or dst=all was
skipped, so asserting rules that actually had “all” never failed and
regressions slipped through. Now we normalize and check every rule
before matching, so tests for src=all/dst=all will correctly pass or fail.
```

ihuguet: Done

cathay4t (approved):

PR #2909, opened 2025-05-29
nm: route rules: don't duplicate rules in NM profile

ihuguet:
Each time that a routing rule was applied it was appended to the routing rules of the NM profile. If it already existed in the profile, it was added again, resulting in duplicated rules. As the kernel does not allow duplicated rules, they were effectively applied only once, though.

Nonetheless, we shouldn't duplicate again and again the same rule in the NM profile. Fix that by converting `InterfaceIPv4/6::rules` from `Vec` to `HashSet`, so we can only add rules that are not present yet.

Resolves: https://issues.redhat.com/browse/NMT-1649

### DIFF
--- a/rust/src/lib/ip.rs
+++ b/rust/src/lib/ip.rs
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
+use std::collections::HashSet;
 use std::fmt::Write;
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 use std::str::FromStr;
@@ -191,7 +192,7 @@ pub struct InterfaceIpv4 {
     /// Deserialize from `dhcp-custom-hostname`
     pub dhcp_custom_hostname: Option<String>,
     pub(crate) dns: Option<DnsClientState>,
-    pub(crate) rules: Option<Vec<RouteRuleEntry>>,
+    pub(crate) rules: Option<HashSet<RouteRuleEntry>>,
 }
 
 impl InterfaceIpv4 {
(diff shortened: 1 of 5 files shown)


cathay4t (changes requested):
Nmstate use HashSet for routes, but use Vec<> for route rules. I think this duplication of route rule should be handled at nmstate top level.

ihuguet:
After generating a `MergedNetworkState`, where the rules are correctly handled without duplication, it happens this:

1. Current routing rules configuration per interface is retrieved from NM [here](https://github.com/nmstate/nmstate/blob/f75ef0474b01f8f5e2a261183547d082a87609f4/rust/src/lib/query_apply/net_state.rs#L122).
2. `nm_apply` is called, which in turn calls to `store_route_rule_config` [here](https://github.com/nmstate/nmstate/blob/f75ef0474b01f8f5e2a261183547d082a87609f4/rust/src/lib/nm/query_apply/apply.rs#L89) to assign the routing rules from the merged state to different interfaces.
3. We append the new rules to the already existing ones [here](https://github.com/nmstate/nmstate/blob/f75ef0474b01f8f5e2a261183547d082a87609f4/rust/src/lib/nm/route_rule.rs#L194).
4. By calling to `prepare_nm_conns`, we end calling to `gen_nm_ip_setting` [here](https://github.com/nmstate/nmstate/blob/f75ef0474b01f8f5e2a261183547d082a87609f4/rust/src/lib/nm/settings/connection.rs#L139), which then calls to `gen_nm_ip_rules`. It uses the rules from the `InterfaceIPv4` or `InterfaceIPv6` to generate the NMSetting.

The rules from `InterfaceIPv4/6` are, as I said in point 3, the combination of the previous rules from the existing NMConnection and the new ones added from the merged state. The rules from the merged state are the top level ones that you say, and there are no duplicates there. The duplication is caused when appending them to those from the NMConnection.

cathay4t (approved):

PR #2912, opened 2025-06-09
go: add policy validation support to golang bindings

liangwen12year:
Introduce ValidatePolicy in the Golang bindings to validate YAML
    files containing nmpolicy. This function behaves similarly to
    GenerateConfiguration for nmstate YAML, propagating the C binding errors
    back to the user.
    
    Resolves: https://issues.redhat.com/browse/RHEL-69925


cathay4t (changes requested):
1. Should be single call `Validate` accepting both state and policy.
2. Take current state as string, optionally empty. In case anyone want precise validation.
3. C library or go library should have *zero* logical code. So you should introduce rust API first as `nmstate::validate` wrapping `NetworkState::validate()` and `NetworkPolicy::validate()`.
4. For now, we only do syntax check and will be expand to more validation tasks. Please state so in the document.

cathay4t:
I will take over this work and create PR for this because Wen is moving away from nmstate works and I find some code should be refactored.

cathay4t:
PR https://github.com/nmstate/nmstate/pull/2972 created for continue work of this one.

PR #2925, opened 2025-07-01
route: add support for advmss option

### LINKED ISSUE #2881: Feature Request: bond option advmss
Hi Team,
I have two openshift clusters across two DCs, openshift clusters are deployed with jumbo frames (9000mtu) however the path inbetween doesnot support jumbo and support 1500mtu only.
I am using separate vlan to for the communication between these two clusters for which i am specifying routes via nmstate nncp, however i do not see any option avaiable to define the mtu in nmstate nncp.
Can team please add these options so i can manage all routes through nncp only instead of defining it via nmcli or by creating machineconfig which will result in mcp rollout whenever i needed to add any new route.

jvaclav-rh:
Example:
```
routes:
  config:
    - destination: 2001:db8:c::/64
      next-hop-interface: eth1
      advmss: 1500
```

Closes #2881

### DIFF
--- a/tests/integration/route_test.py
+++ b/tests/integration/route_test.py
@@ -2447,3 +2447,30 @@ def test_add_route_with_quickack(eth1_up):
     )
     cur_state = libnmstate.show()
     assert_routes(routes, cur_state)
+
+
+# https://github.com/nmstate/nmstate/issues/2881
+@pytest.mark.tier1
+def test_add_route_with_advmss(eth1_up):
+    routes = [
+        {
+            Route.NEXT_HOP_INTERFACE: "eth1",
+            Route.DESTINATION: IPV4_TEST_NET1,
+            Route.NEXT_HOP_ADDRESS: IPV4_ADDRESS1,
+            Route.ADVMSS: 1500,
+        },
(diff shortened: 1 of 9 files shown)


tests/integration/route_test.py
```diff
@@ -2447,3 +2447,29 @@ def test_add_route_with_quickack(eth1_up):
     )
     cur_state = libnmstate.show()
     assert_routes(routes, cur_state)
+
+
+@pytest.mark.tier1
```
cathay4t, line 2453: Please include github issue link to explain why tier1 here.

cathay4t (changes requested):
Also need integration test case for `gen_conf` use case.

jvaclav-rh:
> Also need integration test case for `gen_conf` use case.

@cathay4t For the gen_conf use case, I modified the existing test `test_gen_conf_route_initcwnd_mtu_quickack` to also include advmss. If it makes more sense, I could make a brand new one.

cathay4t:
It's ok to reuse that test case.

cathay4t (approved):

PR #2932, opened 2025-07-05
bond: use as_ref() instead of clone

### LINKED ISSUE #2709: Use `as_ref()` instead of `clone()` in `ifaces/bond.rs`
The code like `let config = self.bond.clone().unwrap_or_default();` should be replaced by `as_ref()`. 

And also these code should be polished:

```rust
let bond_config = self.bond.clone().unwrap_or_default();
if bond_config.port.is_some() && bond_config.ports_config.is_some() {
    let mut port_list = bond_config.port.unwrap_or_default();
    let mut port_config_list: Vec<String> = bond_config
        .ports_config
        .unwrap_or_default()
        .into_iter()
        .map(|p| p.name)
        .collect();


```

The clone or collect `Vec<String>` means the developer don't want to understand memory ownership but blindly duplicate the memory to pass the check of rust compiler.

rajeshrah22:
fixes https://github.com/nmstate/nmstate/issues/2709

Refactored functions to not use `clone()`. User `as_ref()` instead.
Cleaner function, less allocations and copying.

Added a test case for port and ports-config conflict.

### DIFF
--- a/rust/src/lib/ifaces/bond.rs
+++ b/rust/src/lib/ifaces/bond.rs
@@ -109,16 +109,16 @@ impl BondInterface {
 
     fn sort_ports(&mut self) {
         if let Some(ref mut bond_conf) = self.bond {
-            if let Some(ref mut port_conf) = &mut bond_conf.port {
-                port_conf.sort_unstable_by_key(|p| p.clone())
+            if let Some(port_list) = bond_conf.port.as_mut() {
+                port_list.sort_unstable_by_key(|p| p.clone())
             }
         }
     }
 
     fn sort_ports_config(&mut self) {
         if let Some(ref mut bond_conf) = self.bond {
-            if let Some(ref mut port_conf) = &mut bond_conf.ports_config {
-                port_conf.sort_unstable_by_key(|p| p.name.clone())
+            if let Some(ports_conf_list) = bond_conf.ports_config.as_mut() {
+                ports_conf_list.sort_unstable_by_key(|p| p.name.clone())
             }
         }
     }
@@ -167,24 +167,18 @@ impl BondInterface {
         Ok(())
     }
 
-    // Return None when desire state does not mention ports
     pub(crate) fn ports(&self) -> Option<Vec<&str>> {
-        let config = self.bond.clone().unwrap_or_default();
-        if config.port.is_some() {
-            self.bond
+        self.bond.as_ref().and_then(|bond_conf| {
+            bond_conf
+                .port
                 .as_ref()
-                .and_then(|bond_conf| bond_conf.port.as_ref())
                 .map(|ports| {
                     ports.as_slice().iter().map(|p| p.as_str()).collect()
                 })
-        } else {
-            self.bond
-                .as_ref()
-                .and_then(|bond_conf| bond_conf.ports_config.as_ref())
-                .map(|ports| {
+                .or(bond_conf.ports_config.as_ref().map(|ports| {
                     ports.as_slice().iter().map(|p| p.name.as_str()).collect()
-                })
-        }
+                }))
+        })
     }
 
     pub(crate) fn get_port_conf(
@@ -289,39 +283,34 @@ impl BondInterface {
     fn validate_conflict_in_port_and_port_configs(
         &self,
     ) -> Result<(), NmstateError> {
-        let bond_config = self.bond.clone().unwrap_or_default();
-        if bond_config.port.is_some() && bond_config.ports_config.is_some() {
-            let mut port_list = bond_config.port.unwrap_or_default();
-            let mut port_config_list: Vec<String> = bond_config
-                .ports_config
-                .unwrap_or_default()
-                .into_iter()
-                .map(|p| p.name)
-                .collect();
-            port_list.sort_unstable();
-            port_config_list.sort_unstable();
-            let matching = port_list
-                .iter()
-                .zip(port_config_list.iter())
-                .filter(|&(port_name, port_config_name)| {
-                    port_name == port_config_name
-                })
-                .count();
-            if matching != port_list.len() || matching != port_config_list.len()
+        if let Some(bond_conf) = self.bond.as_ref() {
+            if let (Some(ports), Some(ports_config)) =
+                (&bond_conf.port, &bond_conf.ports_config)
             {
-                let e = NmstateError::new(
-                    ErrorKind::InvalidArgument,
-                    format!(
-                        "The port names specified in `port` conflict with the \
-                         port names specified in `ports-config` for bond \
-                         interface: {}",
-                        &self.base.name
-                    ),
-                );
-                log::error!("{e}");
-                return Err(e);
+                let mut ports: Vec<&str> =
+                    ports.iter().map(String::as_str).collect();
+                let mut port_confs: Vec<&str> =
+                    ports_config.iter().map(|p| p.name.as_str()).collect();
+
+                ports.sort_unstable();
+                port_confs.sort_unstable();
+
+                if ports != port_confs {
+                    let e = NmstateError::new(
+                        ErrorKind::InvalidArgument,
+                        format!(
+                            "The port names specified in `port` conflict with \
+                             the port names specified in `ports-config` for \
+                             bond interface: {}",
+                            &self.base.name
+                        ),
+                    );
+                    log::error!("{e}");
+                    return Err(e);
+                }
             }
         }
+
         Ok(())
     }
 
--- a/rust/src/lib/ifaces/bond.rs
+++ b/rust/src/lib/ifaces/bond.rs
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::collections::HashMap;
+use std::collections::HashSet;
 use std::net::Ipv6Addr;
 
 use serde::{
@@ -108,18 +109,16 @@ impl BondInterface {
     }
 
     fn sort_ports(&mut self) {
-        if let Some(ref mut bond_conf) = self.bond {
-            if let Some(port_list) = bond_conf.port.as_mut() {
-                port_list.sort_unstable_by_key(|p| p.clone())
-            }
+        if let Some(ports) = self.bond.as_mut().and_then(|b| b.port.as_mut()) {
+            ports.sort_unstable_by_key(|p| p.clone());
         }
     }
 
     fn sort_ports_config(&mut self) {
-        if let Some(ref mut bond_conf) = self.bond {
-            if let Some(ports_conf_list) = bond_conf.ports_config.as_mut() {
-                ports_conf_list.sort_unstable_by_key(|p| p.name.clone())
-            }
+        if let Some(port_confs) =
+            self.bond.as_mut().and_then(|b| b.ports_config.as_mut())
+        {
+            port_confs.sort_unstable_by_key(|p| p.name.clone());
         }
     }
 
@@ -175,9 +174,15 @@ impl BondInterface {
                 .map(|ports| {
                     ports.as_slice().iter().map(|p| p.as_str()).collect()
                 })
-                .or(bond_conf.ports_config.as_ref().map(|ports| {
-                    ports.as_slice().iter().map(|p| p.name.as_str()).collect()
-                }))
+                .or_else(|| {
+                    bond_conf.ports_config.as_ref().map(|ports| {
+                        ports
+                            .as_slice()
+                            .iter()
+                            .map(|p| p.name.as_str())
+                            .collect()
+                    })
+                })
         })
     }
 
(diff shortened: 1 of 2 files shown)


rajeshrah22:
rev1: used cloning
rev2: reduced cloning I think
rev3: gemini generated code, will need to understand.

This was a good issue to learn about as_ref. Still don't fully understand it.

Gemini converts `&String` to `&str`:
```
ports.iter().map(String::as_str).collect();
```

Don't think it makes a huge difference.

String is 1 word pointer to a struct, which contains a pointer to the string data in heap. 2 levels of indirection.
str is 2 words: (pointer, length). pointer points directly to string in memory. 1 level of indirection.

Let me know if  I'm wrong.

rust/src/lib/unit_tests/bond.rs
```diff
@@ -781,3 +781,32 @@ fn test_bond_opt_ns_ip6_target() {
         assert_eq!(e.kind(), ErrorKind::InvalidArgument);
     }
 }
+
+#[test]
+fn test_bond_ports_ports_config_conflict() {
```
cathay4t: Why we have test case for a clean up patch?

The test case should belong to bug fix patch or standalone commit.

rust/src/lib/ifaces/bond.rs
```diff
@@ -109,16 +109,16 @@ impl BondInterface {
 
     fn sort_ports(&mut self) {
         if let Some(ref mut bond_conf) = self.bond {
-            if let Some(ref mut port_conf) = &mut bond_conf.port {
-                port_conf.sort_unstable_by_key(|p| p.clone())
+            if let Some(port_list) = bond_conf.port.as_mut() {
+                port_list.sort_unstable_by_key(|p| p.clone())
```
cathay4t:
```rust
        self.bond
            .as_mut()
            .and_then(|b| b.port.as_mut())
            .and_then(|ports| ports.sort_unstable_by_key(|p| p.clone()))
```

rust/src/lib/ifaces/bond.rs
```diff
@@ -109,16 +109,16 @@ impl BondInterface {
...
             }
         }
     }
 
     fn sort_ports_config(&mut self) {
         if let Some(ref mut bond_conf) = self.bond {
-            if let Some(ref mut port_conf) = &mut bond_conf.ports_config {
-                port_conf.sort_unstable_by_key(|p| p.name.clone())
+            if let Some(ports_conf_list) = bond_conf.ports_config.as_mut() {
+                ports_conf_list.sort_unstable_by_key(|p| p.name.clone())
```
cathay4t: Use similar code as above comment.

rust/src/lib/ifaces/bond.rs
```diff
@@ -167,24 +167,18 @@ impl BondInterface {
...
-                .and_then(|bond_conf| bond_conf.port.as_ref())
                 .map(|ports| {
                     ports.as_slice().iter().map(|p| p.as_str()).collect()
                 })
-        } else {
-            self.bond
-                .as_ref()
-                .and_then(|bond_conf| bond_conf.ports_config.as_ref())
-                .map(|ports| {
+                .or(bond_conf.ports_config.as_ref().map(|ports| {
```
cathay4t: The use `or` here is incorrect. Please study the difference between `or` and `or_else`.

rust/src/lib/ifaces/bond.rs
```diff
@@ -289,39 +283,34 @@ impl BondInterface {
...
-                         interface: {}",
-                        &self.base.name
-                    ),
-                );
-                log::error!("{e}");
-                return Err(e);
+                let mut ports: Vec<&str> =
+                    ports.iter().map(String::as_str).collect();
+                let mut port_confs: Vec<&str> =
+                    ports_config.iter().map(|p| p.name.as_str()).collect();
```
cathay4t, line 298: I think `HashSet` is better fit here, it could even provide you the difference for better error message.

cathay4t (changes requested):

rust/src/lib/ifaces/bond.rs
rajeshrah22:
```rs
.and_then(|ports| Some(ports.sort_unstable_by_key(|p| p.clone())));
```
Doing this instead. I'm trusting the compiler. `and_then(F)` F should return option type, which makes sense that I need to wrap it around `Some`.

The resulting `sort_ports` function has a `()` return type.

rust/src/lib/ifaces/bond.rs
rajeshrah22: https://doc.rust-lang.org/std/option/#boolean-operators

So, `or_else` lazily evaluates the function passed in if `self` is none, unlike `or`.

rust/src/lib/ifaces/bond.rs
rajeshrah22: With clippy's help, I came to this at the end:
```rs
if let Some(ports) = self.bond.as_mut().and_then(|b| b.port.as_mut()) {
            ports.sort_unstable_by_key(|p| p.clone());
}
```

rajeshrah22:
New error message.

```
$ echo '
interfaces:
- name: bond99
  type: bond
  state: up
  link-aggregation:
    mode: active-backup
    port:
    - eth1
    - eth3
    ports-config:
    - name: eth1
      queue-id: 0
    - name: eth2
      queue-id: 0' | sudo ./rust/target/debug/nmstatectl apply -
[2025-07-08T19:45:15Z INFO  nmstatectl] Nmstate version: 2.2.47
[2025-07-08T19:45:15Z INFO  nmstate::ifaces::inter_ifaces] Ignoring interface wlp0s20f3 type wifi
[2025-07-08T19:45:15Z INFO  nmstate::nm::show] Got unsupported interface type 802-11-wireless: wlp0s20f3, ignoring
[2025-07-08T19:45:15Z INFO  nmstate::nm::show] Got unsupported interface type wifi-p2p: p2p-dev-wlp0s20f3, ignoring
[2025-07-08T19:45:16Z INFO  nmstate::ifaces::inter_ifaces] Ignoring interface wlp0s20f3 type wifi
[2025-07-08T19:45:16Z INFO  nmstate::nm::show] Got unsupported interface type 802-11-wireless: wlp0s20f3, ignoring
[2025-07-08T19:45:16Z INFO  nmstate::nm::show] Got unsupported interface type wifi-p2p: p2p-dev-wlp0s20f3, ignoring
[2025-07-08T19:45:16Z INFO  nmstate::ifaces::inter_ifaces] Ignoring interface wlp0s20f3 type unknown
[2025-07-08T19:45:16Z INFO  nmstate::iface] Include interface eth3 to edit as its controller required so
[2025-07-08T19:45:16Z INFO  nmstate::iface] Include interface eth1 to edit as its controller required so
[2025-07-08T19:45:16Z ERROR nmstate::ifaces::bond] InvalidArgument: The port names specified in `port` conflict with the port names specified in `ports-config` for bond interface: bond99. Ports in `port`` but not in `ports_config`: eth3
NmstateError: InvalidArgument: The port names specified in `port` conflict with the port names specified in `ports-config` for bond interface: bond99. Ports in `port`` but not in `ports_config`: eth3
```

cathay4t (approved):

rust/src/lib/ifaces/bond.rs
ihuguet: Yes, that looks good.
However, I wonder why do we need the `p.clone()`... ports is just a `Vec<String>`, it should be possible to order it with `ports.sort_unstable()`.

rust/src/lib/ifaces/bond.rs
ihuguet: Yes, use similar code to the one that you has proposed above. Note that you can remove one nesting level with that.

rust/src/lib/unit_tests/bond.rs
```diff
@@ -781,3 +781,32 @@ fn test_bond_opt_ns_ip6_target() {
...
+        ",
+    )
+    .unwrap();
+
+    let mut merged_iface = MergedInterface::new(Some(iface), None).unwrap();
+    let result = merged_iface.post_inter_ifaces_process_bond();
+    assert!(result.is_err());
+    if let Err(e) = result {
+        assert_eq!(e.kind(), ErrorKind::InvalidArgument);
+    }
```
ihuguet: Please assert that the InvalidArgument is because of the right reason (port names conflict), and not a different reason. For that, you need to check that the error message is the one that you expected. For example, assert that the error message contains "conflict with the port names".

ihuguet:
> Gemini converts `&String` to `&str`:
> ```
> ports.iter().map(String::as_str).collect();
> ```
> Don't think it makes a huge difference.
> 
> String is 1 word pointer to a struct, which contains a pointer to the string data in heap. 2 levels of indirection. str is 2 words: (pointer, length). pointer points directly to string in memory. 1 level of indirection.

It doesn't matter much, but Gemini is right here. Generally is a good practice to use `&str` instead of `&String` unless you specifically need a String to manipulate it. Not because of performance or data size, but for compatibility with function and trait arguments. If you have a `&String` it's trivial to convert it to `&str` if you need, but if you have `Vec<&String>` it's not trivial to get a `Vec<&str>` or `&[&str]` if you need one to pass to a function or to do a comparison with another Vec. So, the common practice is to use `&str` instead of `&String` as much as possible.

ihuguet:
Oh, I didn't see it was merged already. Well, feel free to ignore my comments, or do a small follow-up if you want. Good work, BTW.

rust/src/lib/unit_tests/bond.rs
rajeshrah22: I was thinking of doing something like this but wasn't sure if it was too much logic for a test case.

rajeshrah22:
> Oh, I didn't see it was merged already. Well, feel free to ignore my comments, or do a small follow-up if you want. Good work, BTW.

thanks! Yeah the test case was not included, maybe it should be added?

rajeshrah22:
> > Gemini converts `&String` to `&str`:
> > ```
> > ports.iter().map(String::as_str).collect();
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Don't think it makes a huge difference.
> > String is 1 word pointer to a struct, which contains a pointer to the string data in heap. 2 levels of indirection. str is 2 words: (pointer, length). pointer points directly to string in memory. 1 level of indirection.
> 
> It doesn't matter much, but Gemini is right here. Generally is a good practice to use `&str` instead of `&String` unless you specifically need a String to manipulate it. Not because of performance or data size, but for compatibility with function and trait arguments. If you have a `&String` it's trivial to convert it to `&str` if you need, but if you have `Vec<&String>` it's not trivial to get a `Vec<&str>` or `&[&str]` if you need one to pass to a function or to do a comparison with another Vec. So, the common practice is to use `&str` instead of `&String` as much as possible.

I learned that `String` owns the memory that it refers to and is a managed structure like `Vec`. Whereas `&str` is a slice type, which is a read-only reference to an un owned string.

What you said makes sense with this understanding of the ownership.

rust/src/lib/ifaces/bond.rs
rajeshrah22: I think we try to follow clippy's reccomendations. it doesnt like me using this format:
```rs
        self.bond
            .as_mut()
            .and_then(|b| b.port.as_mut())
            .and_then(|ports| ports.sort_unstable())
```

ihuguet: Really? What does it says?

rajeshrah22:
```
    Checking nmstate v2.2.48 (/home/rrajesh/projects/nmstate/rust/src/lib)
error: using `Option.and_then(|x| Some(y))`, which is more succinctly expressed as `map(|x| y)`
   --> src/lib/ifaces/bond.rs:112:9
    |
112 | /         self.bond
113 | |             .as_mut()
114 | |             .and_then(|b| b.port.as_mut())
115 | |             .and_then(|ports| {
116 | |                 ports.sort_unstable();
117 | |                 Some(())
118 | |             });
    | |______________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#bind_instead_of_map
    = note: `-D clippy::bind-instead-of-map` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::bind_instead_of_map)]`
help: use `map` instead
    |
115 ~             .map(|ports| {
116 |                 ports.sort_unstable();
117 ~                 ()
    |

error: could not compile `nmstate` (lib) due to 1 previous error
```

ihuguet: Ah, yes. Sorry, I meant something like this:
```
if let Some(ports) = self.bond.as_mut().and_then(|b| b.port.as_mut()) {
    ports.sort_unstable();
}
```

ihuguet: I think this would also work, but looks less intuitive:
```
        self.bond
            .as_mut()
            .and_then(|b| b.port.as_mut())
            .map(|ports| ports.sort_unstable())
```

PR #2935, opened 2025-07-07
route: detect nexthop conflicts and raise error

jvaclav-rh:
Currently, when a user tries to create a route which already exists, but has a different nexthop, nmstate returns a cryptic error:

Got netlink unknown error: code -22, msg: Invalid argument (os error 22)

When this happens, it is unclear whether the user
wants to create a new route, or alter existing
routes. Raise an error explaining what needs to
be changed by the user.

Resolves: https://issues.redhat.com/browse/NMT-1537

### DIFF
--- a/rust/src/lib/route.rs
+++ b/rust/src/lib/route.rs
@@ -766,6 +766,8 @@ impl MergedRoutes {
         merged_routes.sort_unstable();
         merged_routes.dedup();
 
+        validate_conflicting_routes(&merged_routes, &desired_routes)?;
+
         let mut merged: HashMap<String, Vec<RouteEntry>> = HashMap::new();
 
         for rt in merged_routes {
@@ -873,5 +875,64 @@ fn validate_route_dst(route: &RouteEntry) -> Result<(), NmstateError> {
             return Ok(());
         }
     }
+    Ok(())
+}
+
+fn validate_conflicting_routes(
+    merged: &[RouteEntry],
+    desired: &[RouteEntry],
+) -> Result<(), NmstateError> {
+    let mut conflict_map = HashMap::new();
+    let desired: HashSet<&RouteEntry> = HashSet::from_iter(desired.iter());
+
+    for route in merged.iter() {
+        let destination =
+            if let Some(destination) = route.destination.as_deref() {
+                destination
+            } else {
+                return Err(NmstateError::new(
+                    ErrorKind::Bug,
+                    format!(
+                        "Route to {} is present, but has no destination",
+                        route.next_hop_addr.as_deref().unwrap_or("")
+                    ),
+                ));
+            };
+
+        let nexthop = route.next_hop_addr.as_deref().unwrap_or("");
+        let existing_nexthop = conflict_map.insert(
+            (
+                destination,
+                route.metric.unwrap_or_default(),
+                route.table_id.unwrap_or(DEFAULT_TABLE_ID),
+                route
+                    .route_type
+                    .as_ref()
+                    .map(|t| u8::from(*t))
+                    .unwrap_or_default(),
+            ),
+            nexthop,
+        );
+
+        if existing_nexthop.is_some()
+            && existing_nexthop != Some(nexthop)
+            && route.weight.is_none()
+            && desired.contains(route)
+        {
+            let nexthop_a = existing_nexthop.unwrap_or("");
+            let nexthop_b = route.next_hop_addr.as_deref().unwrap_or("");
+
+            return Err(NmstateError::new(
+                ErrorKind::InvalidArgument,
+                format!(
+                    "Route to {destination} has conflicting nexthop \
+                     ({nexthop_a}; {nexthop_b}). Either set the existing \
+                     route to 'absent', or set their 'weight' (for ECMP) or \
+                     'metric' value"
+                ),
+            ));
+        }
+    }
+
     Ok(())
 }
--- a/rust/src/lib/unit_tests/route.rs
+++ b/rust/src/lib/unit_tests/route.rs
@@ -864,3 +864,157 @@ fn test_route_advmss_deserialize_from_string() {
 
     assert_eq!(route.advmss, Some(1500));
 }
+
+#[test]
+fn test_implicit_route_nexthop_change_forbidden() {
+    let cur_routes: Routes = serde_yaml::from_str(
+        r"
+        config:
+        - destination: 0.0.0.0/0
+          next-hop-interface: eth1
+          next-hop-address: 192.0.2.1
+          table-id: 200
+        - destination: ::/0
+          next-hop-interface: eth1
+          next-hop-address: 2001:db8:1::3
+          table-id: 200
+        ",
+    )
+    .unwrap();
+
+    let des_routes: Routes = serde_yaml::from_str(
+        r"
+        config:
+        - destination: 0.0.0.0/0
+          next-hop-address: 192.0.2.2
+          next-hop-interface: eth1
+          table-id: 200
+        - destination: ::/0
+          next-hop-address: 2001:db8:1::2
+          next-hop-interface: eth1
+          table-id: 200
+        ",
+    )
+    .unwrap();
+
+    let (merged_ifaces, _current_ifaces) = gen_merged_ifaces_for_route_test();
+    let merged_err =
+        MergedRoutes::new(des_routes, cur_routes, &merged_ifaces).unwrap_err();
+
+    assert_eq!(merged_err.kind(), ErrorKind::InvalidArgument);
+    assert!(merged_err.msg().contains("conflicting nexthop"));
+}
+
+#[test]
+fn test_explicit_route_nexthop_change() {
+    let cur_routes: Routes = serde_yaml::from_str(
+        r"
+        config:
+        - destination: 0.0.0.0/0
+          next-hop-interface: eth1
+          next-hop-address: 192.0.2.1
+          table-id: 200
+        - destination: ::/0
+          next-hop-interface: eth2
+          next-hop-address: 2001:db8:1::3
+          table-id: 200
+        ",
+    )
+    .unwrap();
+
+    let des_routes: Routes = serde_yaml::from_str(
+        r"
+        config:
+        - destination: 0.0.0.0/0
+          next-hop-interface: eth1
+          next-hop-address: 192.0.2.1
+          table-id: 200
+          state: absent
+        - destination: 0.0.0.0/0
+          next-hop-address: 192.0.2.2
+          next-hop-interface: eth1
+          table-id: 200
+        - destination: ::/0
+          next-hop-interface: eth2
+          next-hop-address: 2001:db8:1::3
+          table-id: 200
+          state: absent
+        - destination: ::/0
+          next-hop-address: 2001:db8:1::2
+          next-hop-interface: eth2
+          table-id: 200
+        ",
+    )
+    .unwrap();
+
+    let (merged_ifaces, _current_ifaces) = gen_merged_ifaces_for_route_test();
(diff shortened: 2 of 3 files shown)


cathay4t:
My concerns:

1. This issue is for kernel mode. But you change the behavior for NM mode also.
2. Can you check whether you can do the same YAML in NM(NM automatically setup metric or multipath route)? If it used to works, then we are introducing a regression.

rust/src/lib/route.rs
```diff
@@ -386,6 +387,21 @@ impl RouteEntry {
         true
     }
 
+    fn conflict_key(&self) -> (&str, i64, Vec<u32>) {
+        (
+            self.destination.as_deref().unwrap_or(""),
```
ihuguet: Routes without `destination` are not valid except if they have `state: absent`. And if they are (or going to be) absent, they don't cause conflicts. So, better skip routes with `state: absent` for the conflict detection, and for the rest just assume that will have a destination. If they don't have a destination, log an error of kind `ErrorKind::Bug` and skip.

rust/src/lib/route.rs
```diff
@@ -386,6 +387,21 @@ impl RouteEntry {
         true
     }
 
+    fn conflict_key(&self) -> (&str, i64, Vec<u32>) {
```
ihuguet: Don't create a specific function for this, just inline the code in validate_conflicting_routes.

Also, why do you create the key of type `(&str, i64, Vec<u32>)` and not `(&str, i64, u32, u32)`?

rust/src/lib/route.rs
```diff
@@ -875,3 +893,39 @@ fn validate_route_dst(route: &RouteEntry) -> Result<(), NmstateError> {
...
+fn validate_conflicting_routes(
+    merged: &[RouteEntry],
+    desired: &[RouteEntry],
+) -> Result<(), NmstateError> {
+    let mut conflict_map = HashMap::new();
+    let desired: HashSet<&RouteEntry> = HashSet::from_iter(desired.iter());
+
+    for route in merged.iter() {
+        let existing_nexthop = conflict_map
+            .insert(route.conflict_key(), route.next_hop_addr.as_ref());
```
ihuguet: Instead of saving the next_hop as `Option<&String>`, why don't you save directly an `&str`?
```
        let existing_nexthop = conflict_map
            .insert(route.conflict_key(), route.next_hop_addr.map(String::as_str).unwrap_or(""));
```

rust/src/lib/route.rs
```diff
@@ -875,3 +893,39 @@ fn validate_route_dst(route: &RouteEntry) -> Result<(), NmstateError> {
...
+
+        if existing_nexthop.is_some()
+            && existing_nexthop != Some(route.next_hop_addr.as_ref())
+            && route.weight.is_none()
+            && desired.contains(route)
+        {
+            let nexthop_a =
+                existing_nexthop.flatten().map(|v| v.deref()).unwrap_or("");
+            let nexthop_b = route.next_hop_addr.as_deref().unwrap_or("");
```
ihuguet, line 923: `as_deref` is not commonly used, at least I'm not used to it and it's harder to understand. It's easier to understand `route.next_hop_addr.map(String::as_str).unwrap_or("")`

rust/src/lib/unit_tests/route.rs
```diff
@@ -864,3 +864,107 @@ fn test_route_advmss_deserialize_from_string() {
...
+          table-id: 200
+        ",
+    )
+    .unwrap();
+
+    let (merged_ifaces, _current_ifaces) = gen_merged_ifaces_for_route_test();
+    let merged_err = MergedRoutes::new(des_routes, cur_routes, &merged_ifaces);
+
+    assert!(merged_err.is_err());
+    assert_eq!(merged_err.unwrap_err().kind(), ErrorKind::InvalidArgument);
```
ihuguet: Please check that the InvalidArgument is because of the right reason. I.e. check that the error message contains the substring "conflicting nexthop".

rust/src/lib/unit_tests/route.rs
```diff
@@ -864,3 +864,107 @@ fn test_route_advmss_deserialize_from_string() {
 
     assert_eq!(route.advmss, Some(1500));
 }
+
+#[test]
+fn test_implicit_route_nexthop_change_forbidden() {
```
ihuguet, line 869: Add another test to check that various routes with no `destination` and with `state: absent` don't cause a conflict (as I've mentioned in other comment).

ihuguet (changes requested):

rust/src/lib/route.rs
jvaclav-rh: If you do this, clippy will complain:
```
warning: called `.as_ref().map(String::as_str)` on an `Option` value
   --> src/lib/route.rs:903:12
    |
903 |         ), route.next_hop_addr.as_ref().map(String::as_str).unwrap_or(""));
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: consider using as_deref: `route.next_hop_addr.as_deref()`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#option_as_ref_deref
```

(`route.next_hop_addr` is a `Option<String>`, and `String::as_str` takes a reference, so as_ref() is necessary here).

rust/src/lib/route.rs
jvaclav-rh: > Also, why do you create the key of type `(&str, i64, Vec<u32>)` and not `(&str, i64, u32, u32)`?

This was mostly inspired by the sort_key() function in this class, which returns a `(Vec<bool>, Vec<&str>, Vec<u32>)`. Although, it does handle a lot more values, so `&str, i64, u32, u32` would be better here.

rust/src/lib/route.rs
ihuguet: Ok then. It seems that the problem was mine, anyway, the code base has lot of `as_deref`.

rust/src/lib/route.rs
```diff
@@ -875,3 +877,64 @@ fn validate_route_dst(route: &RouteEntry) -> Result<(), NmstateError> {
...
+                let err = NmstateError::new(
+                    ErrorKind::Bug,
+                    format!(
+                        "Route to {} is present, but has no destination",
+                        route.next_hop_addr.as_deref().unwrap_or("")
+                    ),
+                );
+                log::warn!("{err}");
+                continue;
+            };
```
ihuguet: If I'm not wrong, here is it normal that `merged` contains routes with `absent`. We should just skip them (`continue`) without raising an error.

What I described about raising an error is for the code following this one. All the code executed after this one can assume that `destination != None`. The only way to have `destination == None` is that we introduce a bug in the future. In that case, there are 3 options to read the `destination` field:
1. `unwrap`: in case of this future bug, nmstate would crash. Bad.
2. `unwrap_or("")`: better because it doesn't crash, but it's not correct because it could detect false duplicates with `destination == ""`, leading to logic errors.
3. Skip the route and raise an error. This is what I'd choose.

But, as I said, that's only for code executed after this one. Given your current code, there is no need for that, because you don't access to `route.destination` anywhere else, only to the local variable `destination`. So much better!

TLDR: don't raise an error here, just skip the route.

rust/src/lib/unit_tests/route.rs
```diff
@@ -864,3 +864,155 @@ fn test_route_advmss_deserialize_from_string() {
...
+        Some("::/0".to_string())
+    );
+    assert_eq!(
+        merged_routes.merged.get("eth2").unwrap()[0].next_hop_addr,
+        Some("2001:db8:1::2".to_string())
+    );
+}
+
+#[test]
+fn test_nexthop_change_absent_routes() {
```
ihuguet, line 976: Good test! Please a concise comment explaining what is this test guarding against. It's obvious for us now, but we'll forget in the future.

ihuguet (changes requested):

rust/src/lib/route.rs
jvaclav-rh: > If I'm not wrong, here is it normal that `merged` contains routes with `absent`. We should just skip them (`continue`) without raising an error.

If I understand the code in `MergedRoutes::new()` correctly, `merged_routes` should always end up with present routes only:

https://github.com/nmstate/nmstate/blob/29a1ab8f9cc83563669f71d9f8a3967d98686ade/rust/src/lib/route.rs#L722-L765

So, all routes should have a `destination` unless something went very wrong.
But I could check for that here also, so that it does not rely on the behavior of MergedRoutes::new.

ihuguet: But this block will make that all absent routes from the desired state are added, doesn't it? https://github.com/nmstate/nmstate/blob/29a1ab8f9cc83563669f71d9f8a3967d98686ade/rust/src/lib/route.rs#L751

jvaclav-rh: If a route in `desired_routes` is absent, it would be filtered by:
https://github.com/nmstate/nmstate/blob/29a1ab8f9cc83563669f71d9f8a3967d98686ade/rust/src/lib/route.rs#L755

And so the for loop body would never be executed for it.

ihuguet: :facepalm: yeah, thanks, I missed that. Then, logging the error message makes sense. Don't change it.

jvaclav-rh:
> My concerns:
> 
> 1. This issue is for kernel mode. But you change the behavior for NM mode also.
> 2. Can you check whether you can do the same YAML in NM(NM automatically setup metric or multipath route)? If it used to works, then we are introducing a regression.

I had some discussions with @ihuguet, but forgot to respond here. The configuration of having two+ routes with different nexthops works in NM, because NM appends the routes instead of creaing them. The resulting routes though, are not ECMP, or load balanced. It serves no practical purpose, and it seems to me that it's just confusing for the user. So, I think it would be okay to break this behavior and raise an error message, so the user does not have nonsensical routes configured.

rust/src/lib/route.rs
```diff
@@ -875,3 +877,64 @@ fn validate_route_dst(route: &RouteEntry) -> Result<(), NmstateError> {
...
+        {
+            let nexthop_a = existing_nexthop.unwrap_or("");
+            let nexthop_b = route.next_hop_addr.as_deref().unwrap_or("");
+
+            return Err(NmstateError::new(
+                ErrorKind::InvalidArgument,
+                format!(
+                    "Route to {destination} has conflicting nexthop \
+                     ({nexthop_a}; {nexthop_b}). Either remove one of the \
+                     routes, or set their 'weight' (for ECMP) or 'metric' \
```
ihuguet: On second though, maybe instead of "either remove one of the routes" use "either set the existing route to 'absent'". It is more explicit about how to do the removal, more helpful for the user.

ihuguet (approved):

rust/src/lib/route.rs
```diff
@@ -875,3 +877,64 @@ fn validate_route_dst(route: &RouteEntry) -> Result<(), NmstateError> {
...
+                destination
+            } else {
+                let err = NmstateError::new(
+                    ErrorKind::Bug,
+                    format!(
+                        "Route to {} is present, but has no destination",
+                        route.next_hop_addr.as_deref().unwrap_or("")
+                    ),
+                );
+                log::warn!("{err}");
```
cathay4t: If you just log a warning message, you should not use `NmstateError` here.

cathay4t (changes requested):
Just a lint.

rust/src/lib/route.rs
ihuguet: Then better log an error. Because it is an error (of type "Bug").

cathay4t (approved):

PR #2946, opened 2025-07-11
rpm: don't call to setup.py directly

ihuguet:
Building by directly calling to setup.py has been deprecated for 5 years now by upstream Setuptools. Because of that, Fedora is deprecating it and it will stop working in Fedora 45, having to change to a pyproject.toml based macros. Macros calling to setup.py are those starting by %py3_*, and the new ones are %pyproject_*. See https://fedoraproject.org/wiki/Changes/DeprecateSetuppyMacros.

Replace the %py3_* macros by their %pyproject_* counterparts.

Also, adapt the pyproject.toml [build-system] section per Setuptool documentation:
- Don't explicitly add the "wheel" dependency. It's not directly used in setup.py, so not needed, see [1].
- Add "build-backend" value.

[1] Historically this documentation has unnecessarily listed wheel in the requires list, and many projects still do that. This is not recommended, as the backend no longer requires the wheel package, and listing it explicitly causes it to be unnecessarily required for source distribution builds. You should only include wheel in requires if you need to explicitly access it during build time (e.g. if your project needs a setup.py script that imports wheel).

### DIFF
--- a/packaging/nmstate.spec.in
+++ b/packaging/nmstate.spec.in
@@ -29,7 +29,12 @@ Source3:        %{url}/releases/download/v%{version}/%{srcname}-vendor-%{version
 %endif
 %endif
 BuildRequires:  python3-devel
+%if 0%{?fedora} >= 43
+%define use_pyproject 1
+%else
+%define use_pyproject 0
 BuildRequires:  python3-setuptools
+%endif
 BuildRequires:  gnupg2
 BuildRequires:  systemd-rpm-macros
 %if 0%{?rhel}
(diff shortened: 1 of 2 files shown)


cathay4t (approved):

ihuguet:
Marking as draft as the build with the pyproject macros doesn't work well, probably due to a bug in them: https://src.fedoraproject.org/rpms/python-rpm-macros/pull-request/187

jvaclav-rh (approved):
Looks good to me.

PR #2955, opened 2025-07-14
bond: followup for PR 2832

rajeshrah22:
Use `sort_unstable` instead of `sort_unstable_by_key`. Fix extra ` in error message.

### DIFF
--- a/rust/src/lib/unit_tests/bond.rs
+++ b/rust/src/lib/unit_tests/bond.rs
@@ -773,3 +773,60 @@ fn test_bond_opt_ns_ip6_target() {
         assert_eq!(e.kind(), ErrorKind::InvalidArgument);
     }
 }
+
+#[test]
+fn test_bond_ports_ports_config_conflict() {
+    let iface: Interface = serde_yaml::from_str(
+        r"---
+        name: bond99
+        type: bond
+        state: up
+        link-aggregation:
+          mode: active-backup
+          ports:
+          - eth1
+          - eth2
+          ports-config:
+          - name: eth1
+            queue-id: 0
+          - name: eth2
+            queue-id: 0
+        ",
+    )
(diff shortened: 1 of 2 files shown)


rajeshrah22:
@ihuguet should I add test case to this PR?

ihuguet (approved):

rust/src/lib/unit_tests/bond.rs
```diff
@@ -781,3 +781,60 @@ fn test_bond_opt_ns_ip6_target() {
...
+
+    assert!(result.is_err());
+
+    let error = result.as_ref().err().unwrap();
+
+    assert_eq!(error.kind(), ErrorKind::InvalidArgument);
+    assert!(
+        error.msg().contains("conflict with ")
+            && error.msg().contains("The port names specified in")
+    );
```
cathay4t, line 831: The error message is not API, so we do not assert error message content.

But you may use `assert!(error.msg().contains("eth3") && error.msg().contains("eth2")`.

cathay4t (changes requested):

rust/src/lib/unit_tests/bond.rs
ihuguet: It's not API but we need to test that the error is raised for the right reason. Testing the error message here doesn't mean that we convert it into an API promise, we will just have to adapt the test if we modify the error message, that's all.

Checking "eth3" and "eth2" is not better. We are still checking the error message, but now in a less specific way. It's easy to think that other error related to the bond can contain the name of the ports.

cathay4t: @rajeshrah22 I don't have strong opinion on this, so please change it back to original assertion.

cathay4t (approved):

PR #2958, opened 2025-07-18
dns: don't add DNS search and/or options to down NM connection

### LINKED ISSUE #2798: Interface state erroneously reported as up, when set to up but no carrier
## Describe the bug
nmstate will show an interface as `up`, as long as the interface flags (as reported by nispor) contain _either_ up _or_ running. I believe this is incorrect, as for example an interface can be set "up" using `ip l set ens0 up`, but have no cable plugged in, so be actually shown as DOWN in the ip-link state output

## Nmstate version
- Version 2.2.37

## NetworkManager version
- n/a

## Platform
- Platform: Rocky 9

## How reproducible
- 100%

## Steps to Reproduce
1. Have an (e.g Ethernet) interface with no cable plugged in
2. Set the state flag to up (e.g `ip l set ens0 up`)
3. Observe ip link-state shows as down (`ip l show ens0`) -> `..state DOWN...`
4.

## Nmstate output
- Expected output:
`nmstatectl show -k ens0` -> `state: down`

- Current output:
- Expected output:
`nmstatectl show -k ens0` -> `state: up`

### Additional context

I believe this is caused because of [this line](https://github.com/nmstate/nmstate/blob/base/rust/src/lib/nispor/base_iface.rs#L41), the condition is currently:
```rust
        if *state == nispor::IfaceState::Up
            || flags.contains(&nispor::IfaceFlag::Up)
            || flags.contains(&nispor::IfaceFlag::Running)
        {
```

But it should be (maybe?):

```rust
        if *state == nispor::IfaceState::Up
            || ( flags.contains(&nispor::IfaceFlag::Up)
                 && flags.contains(&nispor::IfaceFlag::Running))
        {
```

ihuguet:
When configuring only search and/or options in dns-resolver we add them to a NM connection and not to NM's global config, but nmstate may choose a down interface, which results on errors at apply time or even on the configurations being accepted but stay unused by NetworkManager. Fix that by always checking that the interface is valid for DNS, and by adding the requirement of the interface being up.

Additionally, in certain uncommon circumstances we consider an interface as "up" incorrectly despite it being actually down. As a consequence, if the interface doesn't appear in the desired state we rely only on the detected current state, and we might consider that the interface is up and put some DNS configurations on it. Fix the detection of down interfaces.

Fixes https://github.com/nmstate/nmstate/issues/2798

Resolves: https://issues.redhat.com/browse/RHEL-102333

### DIFF
--- a/rust/src/lib/dns.rs
+++ b/rust/src/lib/dns.rs
@@ -382,6 +382,9 @@ pub(crate) fn parse_dns_ipv6_link_local_srv(
 impl MergedInterface {
     // IP stack is merged with current at this point.
     pub(crate) fn is_iface_valid_for_dns(&self, is_ipv6: bool) -> bool {
+        if !self.merged.is_up() {
+            return false;
+        }
         if is_ipv6 {
             self.merged.base_iface().ipv6.as_ref().map(|ip_conf| {
                 ip_conf.enabled && (ip_conf.is_static() || (ip_conf.is_auto()))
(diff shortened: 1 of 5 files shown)


ihuguet:
Got acks by chat from @bengal @jvaclav-rh and @smyle10, so merging.

PR #2960, opened 2025-07-22
hsr: copy mac address from port1 to port2 and hsr interfaces

jvaclav-rh:
Currently, when a PRP interface is configured, the MAC address is not shared among ports, and the interface loses failover ability.

Make it so that the MAC address is shared among the PRP interface and its ports. The MAC address is chosen in order of preference:
- mac-address / copy-mac-from set on hsr interface
- mac-address / copy-mac-from set on port1 interface
- permanent-mac-address of port1 interface

This issue is already handled similarly in kernel: https://lore.kernel.org/netdev/20250409101911.3120-1-ffmancera@riseup.net/

This fix is a (re)implementation of this logic to maintain support for older distributions.

Resolves: https://issues.redhat.com/browse/RHEL-100758

### DIFF
--- a/rust/src/lib/ifaces/hsr.rs
+++ b/rust/src/lib/ifaces/hsr.rs
@@ -2,7 +2,10 @@
 
 use serde::{Deserialize, Serialize};
 
-use crate::{BaseInterface, InterfaceType, NmstateError};
+use crate::{
+    BaseInterface, ErrorKind, Interface, InterfaceType, MergedInterfaces,
+    NmstateError,
+};
 
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
 #[serde(rename_all = "kebab-case")]
@@ -21,6 +24,13 @@ use crate::{BaseInterface, InterfaceType, NmstateError};
 ///       multicast-spec: 40
 ///       protocol: prp
 /// ```
+///
+/// nmstate will configure the MAC addresses of HSR ports
+/// to be matching, to retain failover ability. It will
+/// use the MAC address of port1 by default, and apply it
+/// to port2 and the HSR interface. The MAC address can
+/// be overridden by setting the `mac-address` property
+/// on the HSR interface itself.
 pub struct HsrInterface {
     #[serde(flatten)]
     pub base: BaseInterface,
@@ -61,6 +71,156 @@ impl HsrInterface {
                 }
             }
         }
+        Ok(())
+    }
+}
+
+impl MergedInterfaces {
+    pub(crate) fn get_hsr_mac(&self, iface: &HsrInterface) -> Option<String> {
+        if iface.base.mac_address.is_some() {
+            return iface.base.mac_address.clone();
+        }
+
+        let hsr_conf = iface.hsr.as_ref()?;
+
+        for use_permanent in [false, true] {
+            for port in [&hsr_conf.port1, &hsr_conf.port2] {
+                let mac = self
+                    .get_iface(port, InterfaceType::Unknown)
+                    .and_then(|iface| {
+                        let base = iface.merged.base_iface();
+                        if use_permanent {
+                            base.permanent_mac_address.clone()
+                        } else {
+                            base.mac_address.clone()
+                        }
+                    });
+                if mac.is_some() {
+                    return mac;
+                }
+            }
+        }
+
+        None
+    }
+
+    pub(crate) fn validate_hsr_mac(&self) -> Result<(), NmstateError> {
+        for hsr_iface in self.kernel_ifaces.iter().filter_map(|(_, iface)| {
+            if let Interface::Hsr(hsr_iface) = iface.desired.as_ref()? {
+                let hsr_conf = hsr_iface.hsr.as_ref()?;
+
+                if hsr_conf.protocol == HsrProtocol::Prp {
+                    return Some(hsr_iface);
+                }
+            }
+
+            None
+        }) {
+            let macs = [
+                hsr_iface.base.mac_address.as_deref(),
+                hsr_iface
+                    .hsr
+                    .as_ref()
+                    .and_then(|hsr_conf| {
+                        self.get_iface(&hsr_conf.port1, InterfaceType::Unknown)
+                    })
+                    .and_then(|iface| iface.desired.as_ref())
+                    .and_then(|iface| {
+                        iface.base_iface().mac_address.as_deref()
+                    }),
+                hsr_iface
+                    .hsr
+                    .as_ref()
+                    .and_then(|hsr_conf| {
+                        self.get_iface(&hsr_conf.port2, InterfaceType::Unknown)
+                    })
+                    .and_then(|iface| iface.desired.as_ref())
+                    .and_then(|iface| {
+                        iface.base_iface().mac_address.as_deref()
+                    }),
+            ];
+
+            let mut acc = None;
+
+            for &mac in macs.iter().flatten() {
+                if acc.is_some() && Some(mac) != acc {
+                    return Err(NmstateError::new(
+                        ErrorKind::InvalidArgument,
+                        format!(
+                            "HSR ports on interface {} cannot have different \
+                             MAC addresses",
+                            hsr_iface.base.name
+                        ),
+                    ));
+                }
+
+                acc = Some(mac);
+            }
+        }
+
+        Ok(())
+    }
+
+    pub(crate) fn copy_hsr_mac(&mut self) -> Result<(), NmstateError> {
+        let mut pending_changes = Vec::new();
+
+        for (hsr_iface, hsr_conf, mac) in
+            self.kernel_ifaces.iter().filter_map(|(_, iface)| {
+                if !iface.is_desired() {
+                    return None;
+                }
+
+                if let Interface::Hsr(hsr_iface) = &iface.merged {
+                    let mac = self.get_hsr_mac(hsr_iface)?;
+                    let hsr_conf = hsr_iface.hsr.as_ref()?;
+
+                    if hsr_conf.protocol != HsrProtocol::Prp {
+                        return None;
+                    }
+
+                    if iface.current.is_some() {
+                        if Some(mac.as_str())
+                            != self.kernel_ifaces.get(&hsr_conf.port2).and_then(
+                                |iface| {
+                                    iface
+                                        .merged
+                                        .base_iface()
+                                        .mac_address
+                                        .as_deref()
+                                },
+                            )
+                        {
+                            log::warn!(
+                                "Existing HSR PRP interface {} has \
+                                 mismatching MAC addresses on port1 and \
+                                 port2, which may break functionality",
+                                iface.merged.name()
+                            );
+                        }
+
+                        return None;
+                    }
+
+                    return Some((hsr_iface, hsr_conf, mac));
+                }
+
+                None
+            })
+        {
+            for ifname in
+                [&hsr_iface.base.name, &hsr_conf.port1, &hsr_conf.port2]
+            {
+                pending_changes.push((ifname.clone(), mac.clone()));
+            }
+        }
+
+        for (ifname, mac) in pending_changes {
+            if let Some(iface) = self.kernel_ifaces.get_mut(&ifname) {
+                iface.mark_as_changed();
+                iface.set_copy_from_mac(mac.clone());
+            }
+        }
+
         Ok(())
     }
 }
--- a/rust/src/lib/ifaces/inter_ifaces.rs
+++ b/rust/src/lib/ifaces/inter_ifaces.rs
@@ -904,6 +904,8 @@ impl MergedInterfaces {
     fn process(&mut self) -> Result<(), NmstateError> {
         self.process_allow_extra_ovs_patch_ports_for_apply();
         self.apply_copy_mac_from()?;
+        self.validate_hsr_mac()?;
+        self.copy_hsr_mac()?;
         self.validate_controller_and_port_list_confliction()?;
         self.resolve_port_name_ref()?;
         self.resolve_parent_name_ref()?;
--- a/rust/src/lib/unit_tests/ifaces.rs
+++ b/rust/src/lib/unit_tests/ifaces.rs
@@ -2,10 +2,10 @@
 
 use crate::{
     unit_tests::testlib::{
-        new_eth_iface, new_ovs_br_iface, new_ovs_iface, new_unknown_iface,
-        new_vlan_iface,
+        get_mac, new_eth_iface, new_ovs_br_iface, new_ovs_iface,
+        new_unknown_iface, new_vlan_iface,
     },
-    BondMode, Interface, InterfaceState, InterfaceType, Interfaces,
+    BondMode, ErrorKind, Interface, InterfaceState, InterfaceType, Interfaces,
     MergedInterfaces,
 };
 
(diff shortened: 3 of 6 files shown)


ihuguet:
At first glance, given that this is fixed in kernel I don't think that we should add an nmstate's workaround. If I'm not wrong RHEL is going to pick the kernel fix and other distros should too.

rust/src/lib/ifaces/inter_ifaces.rs
```diff
@@ -1017,6 +1018,64 @@ impl MergedInterfaces {
...
+                            .insert(hsr_conf.port2.clone(), mac.to_string());
+                    }
+                }
+            }
+        }
+
+        for (iface_name, mac) in pending_changes.drain() {
+            if let Some(iface) = self.kernel_ifaces.get_mut(&iface_name) {
+                if iface.merged.base_iface().mac_address.as_deref()
+                    != Some(&mac)
```
cathay4t: No need to do the check again. If it is already in pending_changes, it should be applied without questioning again.

rust/src/lib/ifaces/inter_ifaces.rs
```diff
@@ -1017,6 +1018,64 @@ impl MergedInterfaces {
         Ok(())
     }
 
+    fn copy_hsr_mac(&mut self) -> Result<(), NmstateError> {
+        let mut pending_changes = HashMap::new();
+
+        for (iface_name, merged_iface) in self.kernel_ifaces.iter() {
```
cathay4t: Use `iter().filter()` or `iter().filter_map()` could save you from these nested `if let`.

rust/src/lib/ifaces/inter_ifaces.rs
```diff
@@ -1017,6 +1018,64 @@ impl MergedInterfaces {
...
+            if !merged_iface.is_desired() {
+                continue;
+            }
+
+            if let Interface::Hsr(hsr_iface) = &merged_iface.merged {
```
cathay4t: We should only apply this workaround for newly created HSR interface instead of touching existing ones.

rust/src/lib/ifaces/inter_ifaces.rs
```diff
@@ -1017,6 +1018,64 @@ impl MergedInterfaces {
...
+
+                    if let Some(mac) = mac {
+                        pending_changes
+                            .insert(iface_name.clone(), mac.to_string());
+
+                        pending_changes
+                            .insert(hsr_conf.port1.clone(), mac.to_string());
+
+                        pending_changes
+                            .insert(hsr_conf.port2.clone(), mac.to_string());
```
cathay4t: If user is desiring different MAC address on HSR ports, we should raise error message. Example YAML:

```yml
---
interfaces:
  - name: eth1
    type: ethernet
    state: up
    mac-address: 00:23:45:67:89:1a
    ipv4:
      enabled: false
    ipv6:
      enabled: false
  - name: eth2
    type: ethernet
    state: up
    mac-address: 00:23:45:67:89:1b
    ipv4:
      enabled: false
    ipv6:
      enabled: false
  - name: hsr0
    type: hsr
    state: up
    hsr:
      port1: eth1
      port2: eth2
      multicast-spec: 40
      protocol: prp
```

cathay4t (changes requested):
Besides requested changes, unit test should also cover all combination of:
1. One or both HSR port has MAC address defined in desired state or not.
2. HSR controller has MAC address defined or not.

Please also state in rust document of `HsrInterface` under which conditions nmstate will set the MAC address of HSR port.

rust/src/lib/ifaces/inter_ifaces.rs
```diff
@@ -1017,6 +1018,64 @@ impl MergedInterfaces {
...
+                    {
+                        if let Some(set_mac) =
+                            port1.merged.base_iface().mac_address.as_ref()
+                        {
+                            Some(set_mac.as_ref())
+                        } else {
+                            port1
+                                .merged
+                                .base_iface()
+                                .permanent_mac_address
```
rajeshrah22: Just a general question from me as I am not aware: what is the difference between permanent_mac_address and mac_address?

jvaclav-rh: `permanent_mac_address` is the MAC address assigned to the interface by the manufacturer, and is permanently etched into some ROM (for physical devices). `mac_address` can be the (temporary) address that is desired by the user

rust/src/lib/ifaces/inter_ifaces.rs
```diff
@@ -1017,6 +1019,106 @@ impl MergedInterfaces {
...
+            if !iface.is_desired() {
+                return None;
+            }
+
+            if let Interface::Hsr(hsr_iface) = iface.desired.as_ref()? {
+                return Some(hsr_iface);
+            }
+
+            None
+        }) {
```
ihuguet: Small stylistic request, for better readability:
```
let hsr_ifaces = self.kernel_ifaces.iter().filter_map(...);

for hsr_iface in hsr_ifaces {...}
```

rust/src/lib/ifaces/inter_ifaces.rs
```diff
@@ -1017,6 +1019,106 @@ impl MergedInterfaces {
...
+                    } else {
+                        port1
+                            .merged
+                            .base_iface()
+                            .permanent_mac_address
+                            .as_deref()
+                    }
+                } else {
+                    None
+                };
```
ihuguet: What happens if the user defines port2's mac-address only? We don't raise a validation error, but we will ignore it and use the port1's permanent address instead... I think that's wrong. If not HSR iface nor port1 defines mac-address but port2 does, we should use port2's mac-address. Fallback to port1's permanent address if none of the other 3 are defined.

rust/src/lib/ifaces/inter_ifaces.rs
```diff
@@ -1017,6 +1019,106 @@ impl MergedInterfaces {
...
+                if let Some(mac) = mac {
+                    pending_changes
+                        .insert(hsr_iface.base.name.clone(), mac.to_string());
+
+                    pending_changes
+                        .insert(hsr_conf.port1.clone(), mac.to_string());
+
+                    pending_changes
+                        .insert(hsr_conf.port2.clone(), mac.to_string());
+                }
```
ihuguet: Can't we just do the changes here instead of filling this `pending_changes` variable? You can extract the part of getting the MAC to a separate function to avoid the problem with the borrow of the hsr_iface.
```
if let Some(mac) = get_hsr_mac(&hsr_iface) {
    let port1 = self.kernel_ifaces.get(&hsr_conf.port1);
    let port2 = self.kernel_ifaces.get(&hsr_conf.port2);
    for iface in &[&mut hsr_iface, &mut port1, &mut port2] {
        iface.mark_as_changed();
        iface.set_copy_from_mac(mac);
     }
}
```

rust/src/lib/ifaces/inter_ifaces.rs
```diff
@@ -1017,6 +1019,106 @@ impl MergedInterfaces {
...
+            if !i.is_desired() {
+                return None;
+            }
+
+            if let Interface::Hsr(hsr_iface) = i.desired.as_ref()? {
+                return Some(hsr_iface);
+            }
+
+            None
+        }) {
```
ihuguet: Same stylistic request.

rust/src/lib/ifaces/inter_ifaces.rs
```diff
@@ -1017,6 +1019,106 @@ impl MergedInterfaces {
...
+                    pending_changes
+                        .insert(hsr_conf.port1.clone(), mac.to_string());
+
+                    pending_changes
+                        .insert(hsr_conf.port2.clone(), mac.to_string());
+                }
+            }
+        }
+
+        for (iface_name, mac) in pending_changes.drain() {
```
ihuguet: No need to `drain()`, with `for ... in pending_changes` will consume pending_changes.

rust/src/lib/unit_tests/ifaces.rs
```diff
@@ -426,3 +426,303 @@ fn test_unknown_iface_type() {
...
+    let mut iface_with_mac = new_eth_iface("eth1");
+    iface_with_mac.base_iface_mut().mac_address =
+        Some("AA:BB:CC:DD:EE:FF".to_string());
+    cur_ifaces.push(iface_with_mac);
+    cur_ifaces.push(new_eth_iface("eth2"));
+
+    let mut des_ifaces = Interfaces::new();
+    des_ifaces.push(new_hsr_iface("hsr0", "eth1", "eth2"));
+    des_ifaces.push(new_eth_iface("eth1"));
+    des_ifaces.push(new_eth_iface("eth2"));
```
ihuguet: Was it not possible to use yaml syntax here? It's more readable, IMHO.

rust/src/lib/unit_tests/ifaces.rs
```diff
@@ -426,3 +426,303 @@ fn test_unknown_iface_type() {
...
+
+    let hsr_iface =
+        merged_ifaces.get_iface("hsr0", InterfaceType::Hsr).unwrap();
+    assert_eq!(
+        hsr_iface
+            .for_apply
+            .as_ref()
+            .unwrap()
+            .base_iface()
+            .mac_address,
```
ihuguet: Create a `_get_mac` helper function so you can do more readable asserts: `assert_eq!(_get_mac(hsr_iface), "AA:BB:CC:DD:EE:FF")`, here and in the rest of the tests.

rust/src/lib/unit_tests/ifaces.rs
```diff
@@ -426,3 +426,303 @@ fn test_unknown_iface_type() {
         "foo_type\n"
     );
 }
+
+#[test]
+fn test_hsr_mac_sync_no_mac_addresses_defined() {
```
ihuguet, line 431: Remember to add a test for only port2.mac-address being defined.

ihuguet (changes requested):

rust/src/lib/ifaces/inter_ifaces.rs
jvaclav-rh: The problem is that we are borrowing `self.kernel_ifaces`. Here, on `let port1/2` lines, we would have to use `get_mut` instead, since we are modifying it, which would trigger the borrow checker due to the earlier `filter_map()` that we run over it

ihuguet: Ah, I forgot the first borrow in the filter_map. OK, although there should be a better way of doing it, it is not obvious to me which one, ATM, so let's keep this.

rust/src/lib/unit_tests/ifaces.rs
jvaclav-rh: Sorry, a lot of the old tests were using these functions, so I figured to just follow the convention. I wrote the remaining tests using YAML syntax.

rust/src/lib/ifaces/inter_ifaces.rs
```diff
@@ -1017,6 +1019,117 @@ impl MergedInterfaces {
         Ok(())
     }
 
+    fn get_hsr_mac(&self, iface: &HsrInterface) -> Option<String> {
```
cathay4t: This is a function dedicate for HSR, not a function shared by multiple interface types, so please move this function to `ifaces/hsr.rs` like:

```rust
impl HsrInterface {

fn get_hsr_mac(&self, merged_ifaces: MergedInterfaces) {
}
}
```

rust/src/lib/ifaces/inter_ifaces.rs
```diff
@@ -1017,6 +1019,117 @@ impl MergedInterfaces {
...
+                if mac.is_some() {
+                    return mac;
+                }
+            }
+        }
+
+        None
+    }
+
+    fn get_desired_mac(&self, name: &str) -> Option<String> {
```
cathay4t: You only searched `kernel_ifaces`, but function name does not reflect this, future developer might use this function in a wrong way.

Personally I don't think we should wrap this simple stuff into function.

rust/src/lib/ifaces/inter_ifaces.rs
```diff
@@ -1017,6 +1019,117 @@ impl MergedInterfaces {
...
+        self.kernel_ifaces
+            .get(name)
+            .and_then(|merged| merged.desired.as_ref())
+            .and_then(|iface| iface.base_iface().mac_address.clone())
+    }
+
+    fn validate_hsr_mac(&mut self) -> Result<(), NmstateError> {
```
cathay4t: Again, this should belong to `HsrInterface`, not `MergedInterfaces`.

tests/integration/hsr_test.py
```diff
@@ -41,3 +42,39 @@ def test_add_hsr_and_remove(eth1_up, eth2_up):
...
+
+    try:
+        libnmstate.apply(desired_state)
+        hsr_mac = get_mac_address(HSR0)
+        eth2_mac = get_mac_address(ETH2)
+
+        assert hsr_mac == eth1_mac
+        assert hsr_mac == eth2_mac
+        assertlib.assert_state_match(desired_state)
+    finally:
```
cathay4t: Use pytest fixture instead of `try-finally`

cathay4t (changes requested):

rust/src/lib/ifaces/hsr.rs
```diff
@@ -63,6 +72,81 @@ impl HsrInterface {
         }
         Ok(())
     }
+
+    pub(crate) fn get_hsr_mac(
+        &self,
+        merged_ifaces: &MergedInterfaces,
```
ihuguet: As I was saying, move this functions that are aware of MergedInterfaces to an `impl MergedInterfaces` block inside this file.
That will allow you to move the rest of HSR related functions from `ifaces/inter_ifaces.rs` to this new block.

rust/src/lib/ifaces/inter_ifaces.rs
ihuguet: Better to move it to `ifaces/hsr.rs`, as you say, but inside an `impl MergedInterfaces` block, not inside the `impl HsrInterface`. An HsrInterface (or any Interface) should not know anything about MergedInterface, because an Interface alone doesn't know if itself is a current, merged or desired interface.

Look how process_allow_extra_ovs_patch_ports_for_apply and process_veth_peer_changes are implemented as an example.

rust/src/lib/ifaces/hsr.rs
```diff
@@ -63,6 +72,81 @@ impl HsrInterface {
...
+                .and_then(|iface| iface.desired.as_ref())
+                .and_then(|iface| iface.base_iface().mac_address.as_deref()),
+            self.hsr
+                .as_ref()
+                .and_then(|hsr_conf| {
+                    merged_ifaces
+                        .get_iface(&hsr_conf.port2, InterfaceType::Unknown)
+                })
+                .and_then(|iface| iface.desired.as_ref())
+                .and_then(|iface| iface.base_iface().mac_address.as_deref()),
```
ihuguet: Shouldn't we check here from iface.merged and not from iface.desired?
For example:
```
# Current state:
- name: port1
  mac-address: xx:xx:xx:xx:xx:xx
- name: port2

# This desired state should be invalid
- name: port1
- name: port2
  mac-address: yy:yy:yy:yy:yy:yy
```

rust/src/lib/unit_tests/ifaces.rs
ihuguet: Please rewrite all to yaml syntax. They're easier to read, and there are already multiple tests with that style in the surroundings so it's fine.

ihuguet (changes requested):

rust/src/lib/ifaces/hsr.rs
jvaclav-rh: The problem is that this would always fail, because (at least on my machine) all interfaces have the `mac_address` property set by default, even if it's just the same as `permanent_mac_address`. For example, if we have:

current state:
```yaml
---
  - name: eth1
    type: ethernet
    state: up
    mac-address: aa:bb:cc:dd:ee:ff
  - name: eth2
    type: ethernet
    state: up
    mac-address: bb:cc:dd:ee:ff:00
```

and desired state:
```yaml
---
  - name: eth1
    type: ethernet
    state: up
  - name: eth2
    type: ethernet
    state: up
  - name: hsr0
    type: hsr
    state: up
    mac-address: bb:cc:ee:ff:00:11
    hsr:
      port1: eth1
      port2: eth2
      multicast-spec: 40
      protocol: prp
```

the error would be raised if we are checking against `merged`.

ihuguet: You are totally right

rust/src/lib/ifaces/inter_ifaces.rs
```diff
@@ -1017,6 +1019,41 @@ impl MergedInterfaces {
...
+        let mut pending_changes = Vec::new();
+
+        for (hsr_iface, hsr_conf, mac) in
+            self.kernel_ifaces.iter().filter_map(|(_, iface)| {
+                if !iface.is_desired() {
+                    return None;
+                }
+
+                if let Interface::Hsr(hsr_iface) = iface.desired.as_ref()? {
+                    let mac = self.get_hsr_mac(hsr_iface)?;
```
ihuguet: Here we pass a desired interface (of type Interface) to get_hsr_mac. Indeed, get_hsr_mac assumes it is a desired interface, but nothing prevents us to use it wrong. Better pass the MergedInterface and let get_hsr_mac to properly use `.desired`.

rust/src/lib/ifaces/hsr.rs
```diff
@@ -65,6 +75,89 @@ impl HsrInterface {
     }
 }
 
+impl MergedInterfaces {
+    pub(crate) fn get_hsr_mac(&self, iface: &HsrInterface) -> Option<String> {
+        if iface.base.mac_address.is_some() {
+            return iface.base.mac_address.clone();
+        }
```
ihuguet, line 82: Here, iface is a desired iface. It may not contain any mac-address value, but one might exist from the current iface. If mac-address was already defined, we should use it, right?

But then, there is an added complexity. If there existed a mac-address, was it defined by the user or copied from a port?

For example:
```
# Case 1: explicitly define a mac-address, we want to preserve it
- name: hsr0
  mac-address: xx:xx:xx:xx:xx:xx
  hsr:
    port1: eth1
    port2: eth2

# Later, apply (what to do with this address? The user explicitly configured other address in hsr0):
- name: eth1
  mac-address: yy:yy:yy:yy:yy:yy:yy

# Case 2: define the mac-address in the port, we want the HSR to always get it from the port
- name: hsr0
  hsr:
    port1: eth1
    port2: eth2
- name: eth1
  mac-address: xx:xx:xx:xx:xx:xx

# Later, apply (we want the new mac-address to be used by hsr0):
- name: eth1
  mac-address: yy:yy:yy:yy:yy:yy:yy
```

We have finally clearly defined how it should behave on the first apply, but not on next applies.

rust/src/lib/ifaces/hsr.rs
```diff
@@ -65,6 +75,89 @@ impl HsrInterface {
...
+
+        let hsr_conf = iface.hsr.as_ref()?;
```
ihuguet, line 84: If it's a desired iface, it may not content any 'hsr' section. Use the merged iface here.

rust/src/lib/ifaces/inter_ifaces.rs
```diff
@@ -1017,6 +1019,41 @@ impl MergedInterfaces {
...
+                    let hsr_conf = hsr_iface.hsr.as_ref()?;
```
ihuguet: Same, if it's a desired iface, it may not content any 'hsr' section. Use the merged one.

rust/src/lib/ifaces/inter_ifaces.rs
```diff
@@ -1017,6 +1019,41 @@ impl MergedInterfaces {
         Ok(())
     }
 
+    fn copy_hsr_mac(&mut self) -> Result<(), NmstateError> {
```
ihuguet: Move this to the MergedInterfaces block of hsr.rs too.

ihuguet (changes requested):

rust/src/lib/ifaces/hsr.rs
ihuguet: The only reasonable solution that I can see is:
- If the HSR device doesn't have a mac-address, copy from ports
- If the HSR device does have a mac-address, don't copy from ports

This means that only the first time that you apply the HSR will take the address from the ports.

But then, on second applies, users used to define the address in the ports won't see any effect...

rust/src/lib/ifaces/hsr.rs
```diff
@@ -65,6 +75,89 @@ impl HsrInterface {
...
+                            base.permanent_mac_address.clone()
+                        } else {
+                            base.mac_address.clone()
+                        }
+                    });
+                if mac.is_some() {
+                    return mac;
+                }
+            }
+        }
```
ihuguet, line 102: I don't really understand why this work... you said that mac-address was always defined in the merged interface, so it would always pick the address from port1, never from port2. Clearly I'm missing something because you have a test for that...

cathay4t (changes requested):
1. The kernel patch indicate only PRP protocol should have MAC address cloned.
2. Nmstate should only clone MAC address when creating new PRP interface, for existing PRP interface, just log a warning. This could save us from complex corner cases.

cathay4t (changes requested):
The production code looks good to me. Please add unit test cases for preferring permanent mac address.

cathay4t (approved):

PR #3017, opened 2025-10-29
Add support for the lsc-interrupt option for OVS DPDK interfaces

bengal:
Add support for configuring the Link State Change (LSC) detection mode for OVS DPDK interfaces.

Resolves: https://issues.redhat.com/browse/RHEL-53956

### DIFF
--- a/rust/src/lib/ifaces/ovs.rs
+++ b/rust/src/lib/ifaces/ovs.rs
@@ -659,6 +659,16 @@ pub struct OvsDpdkConfig {
     /// Must be power of 2 in the range of 1 to 4096.
     /// Setting to 0 means remove this setting from OVS database.
     pub n_txq_desc: Option<u32>,
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        rename = "lsc-interrupt",
+        alias = "dpdk-lsc-interrupt"
+    )]
+    /// Configures the Link State Change (LSC) detection mode for dpdk ports.
+    /// Setting to true enables interrupt mode; false disables interrupt mode
+    /// and configures the interface in poll mode. You may also use the OVS
+    /// terminology `dpdk-lsc-interrupt` for this property.
+    pub lsc_interrupt: Option<bool>,
 }
 
 const POWER_2_BETWEEN_1_4096: [u32; 13] =
--- a/rust/src/lib/nm/nm_dbus/connection/ovs.rs
+++ b/rust/src/lib/nm/nm_dbus/connection/ovs.rs
@@ -289,6 +289,7 @@ pub struct NmSettingOvsDpdk {
     pub n_rxq: Option<u32>,
     pub n_rxq_desc: Option<u32>,
     pub n_txq_desc: Option<u32>,
+    pub lsc_interrupt: Option<bool>,
     _other: HashMap<String, zvariant::OwnedValue>,
 }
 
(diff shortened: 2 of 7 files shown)


rust/src/lib/nm/nm_dbus/connection/ovs.rs
```diff
@@ -300,6 +301,8 @@ impl TryFrom<DbusDictionary> for NmSettingOvsDpdk {
             n_rxq: _from_map!(v, "n-rxq", u32::try_from)?,
             n_rxq_desc: _from_map!(v, "n-rxq-desc", u32::try_from)?,
             n_txq_desc: _from_map!(v, "n-txq-desc", u32::try_from)?,
+            lsc_interrupt: _from_map!(v, "lsc-interrupt", i32::try_from)?
+                .map(|i| i == 1),
```
ihuguet: Maybe better `i != 0`?

rust/src/lib/ifaces/ovs.rs
```diff
@@ -659,6 +659,11 @@ pub struct OvsDpdkConfig {
     /// Must be power of 2 in the range of 1 to 4096.
     /// Setting to 0 means remove this setting from OVS database.
     pub n_txq_desc: Option<u32>,
+    #[serde(skip_serializing_if = "Option::is_none", rename = "lsc-interrupt")]
+    /// Configures the Link State Change (LSC) detection mode for dpdk ports.
+    /// Setting to true enables interrupt; false disables interrupt and configures
+    /// the interface in poll mode.
+    pub lsc_interrupt: Option<bool>,
```
ihuguet: I think that `rename=` is not needed, `_` is replaced for `-` automatically, if I'm not wrong.

ihuguet (approved):
LGTM

rust/src/lib/ifaces/ovs.rs
bengal: If I understand it correctly, it's done automatically when the struct has attribute `#[serde(rename_all = "kebab-case")]`. This struct doesn't have it because some fields like `n_rxq_desc`, `n_txq_desc` use snake case (I don't know why), and so it needs to be added to the new field.

rust/src/lib/nm/nm_dbus/connection/ovs.rs
bengal: Changed.

rust/src/lib/ifaces/ovs.rs
ihuguet: Yes, it seems so, you are right.

cathay4t: We have to align with OVS manpage says so people could easily search the meaning of them.

The manpage of `ovs-vswitchd.conf.db` uses `dpdk-lsc-interrupt`, so we should also.

cathay4t (changes requested):
Please use `dpdk-lsc-interrupt`

rust/src/lib/ifaces/ovs.rs
cathay4t: Or at lease create an alias for `dpdk-lsc-interrupt`.

bengal: Okay, I have added an alias

ihuguet: If we add it as alias, how will it be displayed in `nmstatectl show`? If showing the same names than ovs is desired, maybe an alias is not good enough...

bengal: It will be displayed as the main name `lsc-interrupt`.

>  If showing the same names than ovs is desired, maybe an alias is not good enough...

I don't know, @cathay4t, what do you prefer?

cathay4t (approved):

PR #3025, opened 2025-11-20
nm: deactivate hsr interface before reactivating when changed

jvaclav-rh:
NM has special handling for "virtual" interfaces, and does not remove them from kernel when connections are changed. This is a problem because now-modified properties are not properly applied to such interfaces.

HSR is one of such virtual interfaces, and since it does not have a `changelink` handler in kernel, implementing reapply is out of the question. Thus, we have no other option but to deactivate the interface, to force NM into recreating it from scratch.

Resolves: https://issues.redhat.com/browse/RHEL-100773

### DIFF
--- a/tests/integration/hsr_test.py
+++ b/tests/integration/hsr_test.py
@@ -64,3 +64,17 @@ def test_hsr_mac_address_sync(hsr0_with_eths):
     assert hsr_mac is not None
     assert hsr_mac == eth1_mac
     assert hsr_mac == eth2_mac
+
+
+# https://issues.redhat.com/browse/RHEL-100773
+@pytest.mark.tier1
+def test_hsr_update_protocol(hsr0_with_eths):
+    # Break if the default protocol in fixture is changed,
+    # since it may render this test otherwise useless.
+    assert (
+        "prp"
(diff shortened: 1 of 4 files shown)


tests/integration/hsr_test.py
```diff
@@ -64,3 +64,37 @@ def test_hsr_mac_address_sync(hsr0_with_eths):
...
+    try:
+        libnmstate.apply(desired_state)
+        assertlib.assert_state_match(desired_state)
+
+        desired_state[Interface.KEY][0][Hsr.CONFIG_SUBTREE][
+            Hsr.PROTOCOL
+        ] = "hsr"
+        libnmstate.apply(desired_state)
+        assertlib.assert_state_match(desired_state)
+    finally:
```
cathay4t: Please place these setup and clean up code into a pytest fixture.

cathay4t (changes requested):

tests/integration/hsr_test.py
jvaclav-rh: Fixed in amended commit

cathay4t (approved):

PR #3032, opened 2025-11-26
deps: use time instead of chrono

mkowalski:
This PR changes the `get_utc_time_in_rfc3339_format` function to use `time` instead of `chrono`. This is in order to reduce dependencies of the project to get rid of unwanted Android libraries (namely `android-tzdata` and `android_system_properties`i).

PR includes unit tests for the modified function.

### DIFF
--- a/rust/src/cli/policy.rs
+++ b/rust/src/cli/policy.rs
@@ -149,5 +149,161 @@ fn store_capture_states_from_file(
 }
 
 fn get_utc_time_in_rfc3339_format() -> String {
-    chrono::Utc::now().to_rfc3339_opts(chrono::SecondsFormat::Secs, true)
+    time::OffsetDateTime::now_utc()
+        .format(&time::format_description::well_known::Rfc3339)
+        .expect("Failed to format time as RFC3339")
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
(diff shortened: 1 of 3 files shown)


rust/src/cli/policy.rs
```diff
@@ -149,5 +149,161 @@ fn store_capture_states_from_file(
...
+            minute
+        );
+
+        // Verify second is valid (00-59)
+        let second: u32 = time_part[6..8]
+            .parse()
+            .expect("Second should be a valid number");
+        assert!(
+            second <= 59,
+            "Second should be between 0 and 59: {}",
```
tyll: FYI: At least RFC3339 also allows leap seconds:
https://datatracker.ietf.org/doc/html/rfc3339#appendix-D

mkowalski: OMG, you caught me off guard with that one. Just fixed

mkowalski:
/retest

Looks like test flakiness

```
>           raise map_error(err_kind, err_msg)
E           libnmstate.error.NmstateVerificationError: Verification failure: eth1.101.interface.vlan.protocol desire '"802.1q"', current '"802.1ad"'
```

cathay4t (approved):

cathay4t:
That CI failure is known issue which require too much effort for such a corner case( Changing VLAN protocol of existing VLAN).

Anyway, the PR looks good. Merged.

PR #2953, opened 2025-07-12
nispor: delete ip addr not in desired state

rajeshrah22:
Previously, new IP addr on same interface was being appended to addresses list instead of being updated.

This solution asks nispor to delete ip addresses in the current state not contained in the desired state.

Resolves: https://issues.redhat.com/browse/RHEL-54544
Resolves: #2893

PR #3033, opened 2025-11-26
nispor: consider re-ordering when assigning static ip

rajeshrah22:
Fixes Pull Request #2953

The original pull request fixed the problem of new ip address appending instead of updating. This was done by removing ip address in current state that are *not* in the desired state and adding new ip addresses.

However, if the ip addresses are re-ordered in the new state, the previous logic will not update the ordering. The new logic will ~instead purge all previous ip addresses and add the new ip addresses in their correct ordering.~ also consider the ordering of the ip addresses. It will delete ip address in current state if it does not exist in desired state or it is in a different index.

Example case:

Old state
```
        name: lo
        type: loopback
        ipv4:
          enabled: true
          address:
          - ip: 127.0.0.1
            prefix-length: 8
          - ip: 127.0.0.2
            prefix-length: 32
        ipv6:
          enabled: true
          address:
          - ip: ::1
            prefix-length: 128
          - ip: ::2
            prefix-length: 128
```
New state
```
        name: lo
        type: loopback
        ipv4:
          enabled: true
          address:
          - ip: 127.0.0.2
            prefix-length: 32
          - ip: 127.0.0.1
            prefix-length: 8
        ipv6:
          enabled: true
          address:
          - ip: ::2
            prefix-length: 128
          - ip: ::1
            prefix-length: 128
```

### DIFF
--- a/rust/src/lib/nispor/ip.rs
+++ b/rust/src/lib/nispor/ip.rs
@@ -155,45 +155,50 @@ pub(crate) fn nmstate_ipv4_to_np(
 ) -> nispor::IpConf {
     let mut np_ip_conf = nispor::IpConf::default();
 
-    // delete ip addresses not in desired state
-    if let Some(nms_cur_iface) = nms_merged_iface.current.as_ref() {
-        if let (Some(nms_cur_ipv4), Some(nms_des_ipv4)) = (
-            &nms_cur_iface.base_iface().ipv4,
-            &nms_merged_iface.merged.base_iface().ipv4,
-        ) {
-            let des_ips: Vec<_> = nms_des_ipv4
-                .addresses
-                .as_deref()
-                .unwrap_or_default()
-                .iter()
-                .collect();
+    let des_ips = nms_merged_iface
+        .merged
+        .base_iface()
+        .ipv4
+        .as_ref()
+        .and_then(|ipv4| ipv4.addresses.as_deref())
+        .unwrap_or_default();
 
-            for nms_addr in
-                nms_cur_ipv4.addresses.as_deref().unwrap_or_default()
-            {
-                if !des_ips.contains(&nms_addr) {
-                    np_ip_conf.addresses.push({
-                        let mut ip_conf = nispor::IpAddrConf::default();
-                        ip_conf.address = nms_addr.ip.to_string();
-                        ip_conf.prefix_len = nms_addr.prefix_length;
-                        ip_conf.remove = true;
-                        ip_conf
-                    });
-                }
-            }
-        }
-    }
+    let cur_ips = nms_merged_iface
+        .current
+        .as_ref()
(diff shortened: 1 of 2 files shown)


rajeshrah22:
I am slightly confused by the fact that my changes in ordering Nispor conf did not affect ipv6? Maybe ipv6 ordering works differently in the kernel?

cathay4t:
If we purge IP then readd, we will break on-going TCP/UDP communication.

I think the correct way is do minimum changes by only delete changed/reordered IP address.

rajeshrah22:
> If we purge IP then readd, we will break on-going TCP/UDP communication.
> 
> I think the correct way is do minimum changes by only delete changed/reordered IP address.

That makes sense, so we need to handle re-ordering as well.

So instead of deleting everything and re-adding, we only delete and re-add ip addresses that are changed or their index had changed.  

By break, do you mean a TCP/UDP connection will completely stop, or is it that some re-tries will have to happen? Just curious.

cathay4t:
I think TCP socket will be terminated once IP been purged. I have not test it yet, just my guess.

A good blog material  anyway, I will allocate time to investigate this.

rajeshrah22:
> I am slightly confused by the fact that my changes in ordering Nispor conf did not affect ipv6? Maybe ipv6 ordering works differently in the kernel?

Yeah, I replicated same messages using `ip` commands, and the ordering of ipv6 addresses is different than ipv4. Not sure if `ipv6 is just plain un-ordered or there is a different ordering scheme. From the little testing I did, it looks like ipv4 addresses stay in the order they were added.

cathay4t:
Please also consider the use case describe in https://github.com/cathay4t/NetworkManager/commit/a9d81d9909b255b7d4ba95a72c572617aada2deb

rajeshrah22:
> Please also consider the use case describe in [cathay4t/NetworkManager@a9d81d9](https://github.com/cathay4t/NetworkManager/commit/a9d81d9909b255b7d4ba95a72c572617aada2deb)

In what cases can you use the `replace` flag? If the prefix of an IP address itself is changed in the desired state, then it is a completely different IP address correct, and would need to be removed and added.

cathay4t:
Can you search the answer by yourself?

cathay4t:
For fixing the issue. I think nmstate should:
1. For append IP address or removal IP address, we do incremental changes.
2. Otherwise, we do purge and set again. So we got expected IP address ordering.

rajeshrah22:
> For fixing the issue. I think nmstate should:
> 
>     1. For append IP address or removal IP address, we do incremental changes.
> 
>     2. Otherwise, we do purge and set again. So we got expected IP address ordering.

See this https://github.com/nmstate/nmstate/issues/3083 issue for my investigation and discussion on this.

PR #3035, opened 2025-12-01
hsr: add support for HSRv1/2012 protocol

jvaclav-rh:
Introduces a new protocol enum value called Hsr2012.

Protocol version is included as part of protocol property, because they are directly tied together, and configuring them separately would be error prone.

Test case included.

Resolves: https://issues.redhat.com/browse/RHEL-100763

### DIFF
--- a/tests/integration/hsr_test.py
+++ b/tests/integration/hsr_test.py
@@ -9,6 +9,8 @@ from libnmstate.schema import InterfaceType
 from libnmstate.schema import Hsr
 
 from .testlib import assertlib
+from .testlib.env import kernel_newer_than
+from .testlib.env import nm_minor_version
 from .testlib.hsrlib import hsr_interface
 from .testlib.ifacelib import get_mac_address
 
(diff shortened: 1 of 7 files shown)


tests/integration/hsr_test.py
```diff
@@ -78,3 +80,18 @@ def test_hsr_update_protocol(hsr0_with_eths):
...
+# https://issues.redhat.com/browse/RHEL-100763
+@pytest.mark.tier1
+@pytest.mark.skipif(
+    not kernel_newer_than(6, 18) or nm_minor_version() < 56,
+    reason=(
+        "HSR protocol version is only supported by NetworkManager 1.56+, "
+        "and kernel exposes this attribute only since 6.19+"
+    ),
+)
+def test_add_hsr_with_protocol_version(eth1_up, eth2_up):
```
cathay4t: Please use `@pytest.mark.parametrize` to test all protocols:
 * `hsr`
 * `hsr-2010`
 * `hsr-2012`

cathay4t (changes requested):

tests/integration/hsr_test.py
jvaclav-rh: Done.

cathay4t (approved):

PR #3046, opened 2025-12-15
hsr: add support for configuring "interlink" property

jvaclav-rh:
Integration test case included.

Resolves: https://issues.redhat.com/browse/RHEL-100766

### DIFF
--- a/rust/src/lib/ifaces/hsr.rs
+++ b/rust/src/lib/ifaces/hsr.rs
@@ -76,6 +76,22 @@ impl HsrInterface {
                  state."
             );
         }
+
+        if is_desired
+            && let Some(conf) = &mut self.hsr
+            && conf.interlink.is_some()
+            && conf.protocol == HsrProtocol::Prp
+        {
+            return Err(NmstateError::new(
+                ErrorKind::InvalidArgument,
+                format!(
+                    "The interlink property is not supported on PRP interface \
+                     {}",
+                    self.base.name
+                ),
+            ));
+        }
+
         Ok(())
     }
 }
@@ -259,6 +275,9 @@ pub struct HsrConfig {
     pub port1: String,
     /// The port2 interface name.
     pub port2: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    /// The interlink interface name.
+    pub interlink: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     /// The MAC address used for the supervision frames. This property is
     /// read-only.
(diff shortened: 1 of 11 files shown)


tests/integration/bond_test.py
```diff
@@ -228,7 +228,7 @@ def test_conflict_port_name_between_port_and_ports_config(eth1_up, eth2_up):
     state[Interface.KEY][0][Bond.CONFIG_SUBTREE][Bond.PORT] = ["eth1"]
     with pytest.raises(NmstateValueError):
         libnmstate.apply(state)
-    state[Interface.KEY][0][Bond.CONFIG_SUBTREE][Bond.PORT] = ["eth1", "eth3"]
+    state[Interface.KEY][0][Bond.CONFIG_SUBTREE][Bond.PORT] = ["eth1", "eth4"]
```
cathay4t: You don't need to change this. The test still valid even eth3 exists.

cathay4t (changes requested):

tests/integration/hsr_test.py
```diff
@@ -101,3 +102,20 @@ def test_add_hsr_with_protocol_version(eth1_up, eth2_up, protocol):
         )
         assertlib.assert_state(state)
     assertlib.assert_absent(HSR0)
+
+
+# https://issues.redhat.com/browse/RHEL-100766
+@pytest.mark.tier1
+@pytest.mark.skipif(
+    not kernel_newer_than(6, 19) or nm_minor_version() < 56,
```
cathay4t: `nm_minor_version() < 55`

cathay4t (changes requested):
Please also include sanitize check on desire state has `protocol: prp` along with `interlink` property.

tests/integration/hsr_test.py
jvaclav-rh: Done.

rust/src/lib/ifaces/hsr.rs
```diff
@@ -76,6 +76,20 @@ impl HsrInterface {
                  state."
             );
         }
+
+        if is_desired
+            && let Some(conf) = &mut self.hsr
+            && conf.interlink.is_some()
+            && conf.protocol == HsrProtocol::Prp
+        {
+            log::warn!(
```
cathay4t: This is a error instead of warning. We should stop user from doing this clearly wrong config.

cathay4t (changes requested):

cathay4t (approved):

PR #3067, opened 2026-01-19
ci: increase integration test timeout to 45 minutes

jvaclav-rh:
Some jobs are now running out of time before they are able to complete all tests:

https://github.com/nmstate/nmstate/actions/runs/21132725440/job/60772534942

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -123,7 +123,7 @@ jobs:
         run: ./automation/run-lint-tests.sh
 
   integ:
-    timeout-minutes: 30
+    timeout-minutes: 45
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


cathay4t:
For the test time, can you investigate more which test case is slow?  In https://github.com/nmstate/nmstate/actions/runs/20995270003 , the maximum run time is c10s and fedora where compile is required, only takes 20 minutes. So it is hard for me to understand why suddenly your test need 10 more minutes.

cathay4t (changes requested):
I need a good reason why we need to expend the test time.

jvaclav-rh:
Not needed anymore. The cause for the increase was the added construction of `eth3`, worked around it by only configuring it where it's necessary.

PR #3079, opened 2026-02-05
ipv4: Add new parameter prefix-route-metric

### LINKED ISSUE #3080: Bug on route metric  - not possible to set ipv4.route-metric / ipv6.route-metric for anything other than DHCP / autoconf.
Currently, with nmstate, it is not possible to set ipv4.route-metric / ipv6.route-metric for anything other than DHCP / autoconf.

However, it is important to set this value to influence the metric of directly connected routes even for static addressing.

https://networkmanager.dev/docs/api/latest/nm-settings-nmcli.html
```
ipv4.route-metric
	
The default metric for routes that don't explicitly specify a metric. The default value -1 means that the metric is chosen automatically based on the device type.
The metric applies to dynamic routes, manual (static) routes that don't have an explicit metric setting, address prefix routes, and the default route.
Note that for IPv6, the kernel accepts zero (0) but coerces it to 1024 (user default). Hence, setting this property to zero effectively mean setting it to 1024.
For IPv4, zero is a regular value for the metric.

Format: integer

Valid values: -1 - 4294967295
```

This is needed, for example, to install BGP multipath routes with FRR, as newer versions of FRR respect and compare the nexthop routes' metrics.

Right now, nmstate picks up auto-route-metric from nmcli ipv4.route-metric / ipv6.route-metric unconditionally, but doesn't write it back when its changed (with static address configuration)

In the code, it mentions right now that auto_route_metric should be for DHCP only:
```
    /// Metric for routes retrieved from DHCP server.
    /// Only available for DHCPv4 enabled interface.
    /// Deserialize from `auto-route-metric`
```

I think this stems from a misunderstanding about route-metric and that route-metric is actually also useful in the context of static IP addressing.

In the code:
```
    if iface_ip.is_auto() {
(...)
        nm_setting.route_metric = iface_ip.auto_route_metric.map(|i| i.into());
(...)
```
but I think that the route_metric should be set regardless of the iface mode.

Whereas, while reading the value, it is parsed unconditionally:
rust/src/lib/nm/query_apply/ip.rs
```
            auto_route_metric: nm_ip_setting.route_metric.map(|i| i as u32),
```

```
#  nmcli conn show enp2s0 | grep metric
ipv4.route-metric:                      149
ipv6.route-metric:                      -1
```

```
# nmstatectl show | grep -B20 auto-route-metric
[2026-02-05T11:57:08Z INFO  nmstatectl] Nmstate version: 2.2.57
  - name: enx5254002a62ef
  ethernet:
    auto-negotiation: false
- name: enp2s0
  type: ethernet
  driver: virtio_net
  state: up
  identifier: name
  mac-address: 52:54:00:3F:5D:16
  permanent-mac-address: 52:54:00:3F:5D:16
  mtu: 1500
  min-mtu: 68
  max-mtu: 65535
  wait-ip: any
  ipv4:
    enabled: true
    dhcp: false
    address:
    - ip: 192.168.123.10
      prefix-length: 24
    auto-route-metric: 149
(...)
```

```
# cat nmstate.yml 
interfaces:
- name: enp2s0
  type: ethernet
  state: up
  ipv4:
    enabled: true
    dhcp: false
    address:
    - ip: 192.168.123.10
      prefix-length: 24
    auto-route-metric: 151
    forwarding: false
  ipv6:
    enabled: false
    dhcp: false
    autoconf: false
- name: enp3s0
  type: ethernet
  state: up
  ipv4:
    enabled: true
    dhcp: false
    address:
    - ip: 192.168.124.10
      prefix-length: 24
    auto-route-metric: 151
    forwarding: false
  ipv6:
    enabled: false
    dhcp: false
    autoconf: false
```

```
#  nmcli apply nmstate.yaml
(...)
#  nmcli conn show enp2s0 | grep metric
ipv4.route-metric:                      149
ipv6.route-metric:                      -1
# nmstatectl show | grep auto-route-metric
[2026-02-05T11:59:01Z INFO  nmstatectl] Nmstate version: 2.2.57
    auto-route-metric: 149
    auto-route-metric: 149
```

andreaskaris:
Introducing `ipv4.prefix-route-metric` for setting route metric for
prefix routes. Example YAML for set prefix route (192.0.2.0/24) metric
to 100:

```yml
---
interfaces:
- name: eth1
  state: up
  ipv4:
    enabled: true
    dhcp: false
    prefix-route-metric: 100
    address:
    - ip: 192.0.2.252
      prefix-length: 24
```

Due to limitation of NetworkManager, we do not support
`prefix-route-metric` different from `auto-routes-metric`. An error will
be raise if otherwise.

Valid value range is [RouteEntry::USE_DEFAULT_METRIC, u32::MAX].

NetworkManager does not create IPv6 prefix route, so this property does
not exist in ipv6.

Unit test cases and integration test cases included.

Resolves: https://github.com/nmstate/nmstate/issues/3080
Resolves: https://redhat.atlassian.net/browse/RHEL-147209

### DIFF
--- a/rust/src/lib/ip.rs
+++ b/rust/src/lib/ip.rs
@@ -11,7 +11,7 @@ use serde::{Deserialize, Deserializer, Serialize};
 
 use crate::{
     BaseInterface, DnsClientState, ErrorKind, MergedInterface,
-    MptcpAddressFlag, NmstateError, RouteRuleEntry,
+    MptcpAddressFlag, NmstateError, RouteEntry, RouteRuleEntry,
 };
 
 const AF_INET: u8 = 2;
@@ -83,9 +83,9 @@ struct InterfaceIp {
         skip_serializing_if = "Option::is_none",
         rename = "auto-route-metric",
         default,
-        deserialize_with = "crate::deserializer::option_u32_or_string"
+        deserialize_with = "crate::deserializer::option_i64_or_string"
     )]
-    pub auto_route_metric: Option<u32>,
+    pub auto_route_metric: Option<i64>,
     #[serde(skip_serializing_if = "Option::is_none", rename = "addr-gen-mode")]
     pub addr_gen_mode: Option<Ipv6AddrGenMode>,
     #[serde(
@@ -105,12 +105,15 @@ struct InterfaceIp {
         rename = "dhcp-custom-hostname"
     )]
     pub dhcp_custom_hostname: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub forwarding: Option<bool>,
     #[serde(
         skip_serializing_if = "Option::is_none",
+        rename = "prefix-route-metric",
         default,
-        deserialize_with = "crate::deserializer::option_bool_or_string"
+        deserialize_with = "crate::deserializer::option_i64_or_string"
     )]
-    pub forwarding: Option<bool>,
+    pub prefix_route_metric: Option<i64>,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Default, Serialize)]
@@ -184,7 +187,7 @@ pub struct InterfaceIpv4 {
     /// Metric for routes retrieved from DHCP server.
     /// Only available for DHCPv4 enabled interface.
     /// Deserialize from `auto-route-metric`
-    pub auto_route_metric: Option<u32>,
+    pub auto_route_metric: Option<i64>,
     /// Whether to include hostname in DHCP request.
     /// If the hostname is FQDN, the `Fully Qualified Domain Name (FQDN)`
     /// option(81) defined in RFC 4702 will be used.
@@ -203,6 +206,9 @@ pub struct InterfaceIpv4 {
     pub dhcp_custom_hostname: Option<String>,
     pub(crate) dns: Option<DnsClientState>,
     pub(crate) rules: Option<HashSet<RouteRuleEntry>>,
+    /// Metric for routes of prefixes directly connected to the interface.
+    /// Deserialize from `prefix-route-metric`
+    pub prefix_route_metric: Option<i64>,
 }
 
 impl InterfaceIpv4 {
@@ -276,6 +282,69 @@ impl InterfaceIpv4 {
         self.sanitize(false).ok();
     }
 
+    pub(crate) fn sanitize_auto_metric_and_prefix_metric(
+        &mut self,
+    ) -> Result<(), NmstateError> {
+        if let Some(prefix_metric) = self.prefix_route_metric
+            && prefix_metric != RouteEntry::USE_DEFAULT_METRIC
+            && let Some(auto_metric) = self.auto_route_metric
+            && auto_metric != RouteEntry::USE_DEFAULT_METRIC
+            && prefix_metric != auto_metric
+        {
+            return Err(NmstateError::new(
+                ErrorKind::NotSupportedError,
+                format!(
+                    "Different value of ipv4.auto-route-metric ({}) and \
+                     ipv4.prefix-route-metric ({}) are not supported yet by \
+                     NetworkManager",
+                    auto_metric, prefix_metric,
+                ),
+            ));
+        }
+
+        if let Some(metric) = self.prefix_route_metric
+            && (metric < RouteEntry::USE_DEFAULT_METRIC
+                || metric > u32::MAX as i64)
+        {
+            return Err(NmstateError::new(
+                ErrorKind::InvalidArgument,
+                format!(
+                    "Invalid value for ipv4.prefix-route-metric: {metric}, \
+                     should be in range of [{};{}]",
+                    RouteEntry::USE_DEFAULT_METRIC,
+                    u32::MAX,
+                ),
+            ));
+        }
+        if let Some(metric) = self.auto_route_metric
+            && (metric < RouteEntry::USE_DEFAULT_METRIC
+                || metric > u32::MAX as i64)
+        {
+            return Err(NmstateError::new(
+                ErrorKind::InvalidArgument,
+                format!(
+                    "Invalid value for ipv4.auto-route-metric: {metric}, \
+                     should be in range of [{};{}]",
+                    RouteEntry::USE_DEFAULT_METRIC,
+                    u32::MAX,
+                ),
+            ));
+        }
+
+        if self.prefix_route_metric == Some(RouteEntry::USE_DEFAULT_METRIC)
+            && self.auto_route_metric.is_some()
+        {
+            self.prefix_route_metric = self.auto_route_metric;
+        }
+        if self.auto_route_metric == Some(RouteEntry::USE_DEFAULT_METRIC)
+            && self.prefix_route_metric.is_some()
+        {
+            self.auto_route_metric = self.prefix_route_metric;
+        }
+
+        Ok(())
+    }
+
     // * Remove link-local address
     // * Set auto_dns, auto_gateway and auto_routes to true if DHCP enabled and
     //   those options is None
@@ -283,6 +352,7 @@ impl InterfaceIpv4 {
     // * Remove auto IP address.
     // * Set DHCP options to None if DHCP is false
     // * Remove mptcp_flags is they are for query only
+    // * Validate auto metric and prefix metric
     pub(crate) fn sanitize(
         &mut self,
         is_desired: bool,
@@ -359,6 +429,7 @@ impl InterfaceIpv4 {
         if !self.enabled {
             self.dhcp = None;
             self.addresses = None;
+            self.prefix_route_metric = None;
         }
 
         if self.dhcp != Some(true) {
@@ -393,6 +464,9 @@ impl InterfaceIpv4 {
             for addr in addrs.iter_mut() {
                 addr.mptcp_flags = None;
             }
+        }
+        if is_desired && self.enabled {
+            self.sanitize_auto_metric_and_prefix_metric()?;
         }
         Ok(())
     }
@@ -446,6 +520,7 @@ impl From<InterfaceIp> for InterfaceIpv4 {
             dhcp_send_hostname: ip.dhcp_send_hostname,
             dhcp_custom_hostname: ip.dhcp_custom_hostname,
             forwarding: ip.forwarding,
+            prefix_route_metric: ip.prefix_route_metric,
             ..Default::default()
         }
     }
@@ -472,6 +547,7 @@ impl From<InterfaceIpv4> for InterfaceIp {
             dhcp_send_hostname: ip.dhcp_send_hostname,
             dhcp_custom_hostname: ip.dhcp_custom_hostname,
             forwarding: ip.forwarding,
+            prefix_route_metric: ip.prefix_route_metric,
             ..Default::default()
         }
     }
@@ -552,7 +628,7 @@ pub struct InterfaceIpv6 {
     /// Metric for routes retrieved from DHCP server.
     /// Only available for autoconf enabled interface.
     /// Deserialize from `auto-route-metric`.
-    pub auto_route_metric: Option<u32>,
+    pub auto_route_metric: Option<i64>,
     /// IETF draft(expired) Tokenised IPv6 Identifiers. Should be only
     /// containing the tailing 64 bites for IPv6 address.
     pub token: Option<String>,
@@ -684,9 +760,11 @@ impl InterfaceIpv6 {
             self.auto_gateway = None;
             self.auto_routes = None;
             self.auto_table_id = None;
-            self.auto_route_metric = None;
             self.dhcp_send_hostname = None;
             self.dhcp_custom_hostname = None;
+        }
+        if self.autoconf == Some(false) {
+            self.auto_route_metric = None;
         }
         if let Some(addrs) = self.addresses.as_mut() {
             for addr in addrs.iter_mut() {
@@ -794,6 +872,11 @@ impl<'de> Deserialize<'de> for InterfaceIpv6 {
         let v = serde_json::Value::deserialize(deserializer)?;
 
         if let Some(v_map) = v.as_object() {
+            if v_map.contains_key("prefix-route-metric") {
+                return Err(serde::de::Error::custom(
+                    "prefix-route-metric is not allowed for IPv6",
+                ));
+            }
             if v_map.contains_key("dhcp_client_id") {
                 return Err(serde::de::Error::custom(
                     "dhcp-client-id is not allowed for IPv6",
--- a/rust/src/lib/nm/query_apply/ip.rs
+++ b/rust/src/lib/nm/query_apply/ip.rs
@@ -8,7 +8,7 @@ use super::{
 };
 use crate::{
     AddressFamily, Dhcpv4ClientId, Dhcpv6Duid, InterfaceIpv4, InterfaceIpv6,
-    Ipv6AddrGenMode, RouteRuleAction, RouteRuleEntry, WaitIp,
+    Ipv6AddrGenMode, RouteEntry, RouteRuleAction, RouteRuleEntry, WaitIp,
 };
 
 const ADDR_GEN_MODE_EUI64: i32 = 0;
@@ -54,7 +54,8 @@ pub(crate) fn nm_ip_setting_to_nmstate4(
             } else {
                 None
             },
-            auto_route_metric: nm_ip_setting.route_metric.map(|i| i as u32),
+            // -1 is a special value meaning use NetworkManager's default metric
+            auto_route_metric: nm_ip_setting.route_metric.or(Some(-1)),
             dhcp_send_hostname: if enabled && dhcp == Some(true) {
                 Some(dhcp_send_hostname)
             } else {
@@ -72,6 +73,9 @@ pub(crate) fn nm_ip_setting_to_nmstate4(
             } else {
                 None
             },
+            prefix_route_metric: nm_ip_setting
+                .route_metric
+                .or(Some(RouteEntry::USE_DEFAULT_METRIC)),
             ..Default::default()
         }
     } else {
@@ -123,7 +127,8 @@ pub(crate) fn nm_ip_setting_to_nmstate6(
                     None
                 }
             },
-            auto_route_metric: nm_ip_setting.route_metric.map(|i| i as u32),
+            // -1 is a special value meaning use NetworkManager's default metric
+            auto_route_metric: nm_ip_setting.route_metric.or(Some(-1)),
             dhcp_send_hostname: if enabled && dhcp == Some(true) {
                 Some(nm_ip_setting.dhcp_send_hostname.unwrap_or(true))
             } else {
--- a/rust/src/lib/nm/settings/ip.rs
+++ b/rust/src/lib/nm/settings/ip.rs
@@ -65,9 +65,16 @@ fn gen_nm_ipv4_setting(
     nm_setting.method = Some(method);
     nm_setting.addresses = addresses;
     nm_setting.forwarding = iface_ip.forwarding.map(i32::from);
+
+    if iface_ip.prefix_route_metric.is_some() {
+        nm_setting.route_metric = iface_ip.prefix_route_metric;
+    }
+
     if iface_ip.is_auto() {
         nm_setting.dhcp_timeout = Some(i32::MAX);
-        nm_setting.route_metric = iface_ip.auto_route_metric.map(|i| i.into());
+        if iface_ip.auto_route_metric.is_some() {
+            nm_setting.route_metric = iface_ip.auto_route_metric;
+        }
         nm_setting.dhcp_client_id = Some(nmstate_dhcp_client_id_to_nm(
             iface_ip
                 .dhcp_client_id
@@ -189,6 +196,7 @@ fn gen_nm_ipv6_setting(
     nm_setting.addresses = addresses;
     nm_setting.addr_gen_mode =
         Some(nmstate_addr_gen_mode_to_nm(iface_ip.addr_gen_mode.as_ref()));
+
     if iface_ip.is_auto() {
         nm_setting.dhcp_timeout = Some(i32::MAX);
         nm_setting.ra_timeout = Some(i32::MAX);
@@ -207,7 +215,9 @@ fn gen_nm_ipv6_setting(
                 nm_setting.token = Some(token.to_string());
             }
         }
-        nm_setting.route_metric = iface_ip.auto_route_metric.map(|i| i.into());
+        if iface_ip.auto_route_metric.is_some() {
+            nm_setting.route_metric = iface_ip.auto_route_metric;
+        }
         apply_dhcp_opts(
             &mut nm_setting,
             iface_ip.auto_dns,
--- a/tests/integration/dynamic_ip_test.py
+++ b/tests/integration/dynamic_ip_test.py
@@ -2330,7 +2330,7 @@ def test_append_static_dns_before_auto_dns(dhcpcli_up_with_dns_cleanup):
 def test_static_ipv6_route_not_covert_auto_ip(dhcpcli_up_with_dynamic_ip):
     iface_state = dhcpcli_up_with_dynamic_ip[Interface.KEY][0]
 
-    (pre_dhcpv4_addrs, pre_dhcpv6_addrs) = get_dhcp_addr(iface_state)
+    pre_dhcpv4_addrs, pre_dhcpv6_addrs = get_dhcp_addr(iface_state)
 
     state_yml = """---
     routes:
@@ -2349,7 +2349,7 @@ def test_static_ipv6_route_not_covert_auto_ip(dhcpcli_up_with_dynamic_ip):
     assert _poll(_has_dhcpv6_addr)
 
     current_state = statelib.show_only((DHCP_CLI_NIC,))
-    (new_dhcpv4_addrs, new_dhcpv6_addrs) = get_dhcp_addr(
+    new_dhcpv4_addrs, new_dhcpv6_addrs = get_dhcp_addr(
         current_state[Interface.KEY][0]
     )
     assert pre_dhcpv4_addrs == new_dhcpv4_addrs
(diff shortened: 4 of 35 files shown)


ihuguet:
LGTM, but let's wait to the discussion in https://github.com/nmstate/nmstate/issues/3080

cathay4t:
Nmstate is not wrapper of NM API, if we expand the definition of `ipv4.auto-route-metric` to what ever NetworkManager1 does right now, it means we are forcing future NetworkManager 2 or other backend to do the same.

Currently, this `ipv4.route-metrics` in NM1 does three things:
1. Metric for static route which does not have metric. Nmstate already have `routes` section, user should set it there instead this property.
2. Metric for address prefix route. I intend to isolate this out to dedicate property `address-prefix-route-metric`.
3. Metric for routes retrieved from DHCPv4 and IPv6-RA. We should preserve whatever current nmstate API definition on `ipv4.auto-route-metric`.

Have the single property for all 3 things above is preventing use case `different metric for address prefix route and auto retrieved routes` when future backend need to support so.

ihuguet:
After a discussion about this with @cathay4t , I agree with him. Better to introduce a new property for direct prefix routes. When using NM backend, this new `prefix_route_metric` and `auto_route_metric` needs to have the same value or an error will be raised, because with NM backend it is impossible to support different values on them.

cathay4t:
@andreaskaris Would you amend this PR to design mentioned above or let us do it?

andreaskaris:
@cathay4t I'll amend it today

tests/integration/dynamic_ip_test.py
```diff
@@ -1738,12 +1738,14 @@ def test_auto_route_metric(dhcpcli_up_with_dynamic_ip):
                         InterfaceIPv4.ENABLED: True,
                         InterfaceIPv4.DHCP: True,
                         InterfaceIPv4.AUTO_ROUTE_METRIC: 901,
+                        InterfaceIPv4.PREFIX_ROUTE_METRIC: 901,
```
andreaskaris: q: wouldn't this requirement brake backwards compatibility?

rust/src/lib/nm/settings/ip.rs
```diff
@@ -65,9 +66,30 @@ fn gen_nm_ipv4_setting(
     nm_setting.method = Some(method);
     nm_setting.addresses = addresses;
     nm_setting.forwarding = iface_ip.forwarding.map(i32::from);
+
+    if iface_ip.dhcp == Some(true)
+        && iface_ip.prefix_route_metric != iface_ip.auto_route_metric
```
ihuguet: The error must be raised if both are set (they're not `None`) and they're different. They can be different if one is `Some(...)` and other `None`. Also, it doesn't matter if dhcp is enabled or not: if both are defined and mismatch, raise error.

rust/src/lib/nm/settings/ip.rs
```diff
@@ -65,9 +66,30 @@ fn gen_nm_ipv4_setting(
...
+                on interface {} when DHCP is enabled",
+                auto_str, prefix_str, iface_name,
+            ),
+        ));
+    }
+    nm_setting.route_metric = iface_ip.prefix_route_metric.map(|i| i.into());
+
     if iface_ip.is_auto() {
         nm_setting.dhcp_timeout = Some(i32::MAX);
-        nm_setting.route_metric = iface_ip.auto_route_metric.map(|i| i.into());
```
ihuguet, line 70: Don't remove this line. If prefix-route-metric is not defined, then we need to take the value from auto-route-metric (but only if `is_auto() == true`).

ihuguet (changes requested):

rust/src/lib/nm/settings/ip.rs
andreaskaris: O.k., so I misunderstood and will address this, what we want is:
- if none is set, reset metric to default
- if only one is Some(), set to that value
- if both are Some() and have the same value, set to that value
- if both are Some() and have different values, throw an error

I'll change it with the next iteration

--------------------------

My assumption was that they always had to match, regardless, but because of the sanitize logic:
https://github.com/nmstate/nmstate/blob/base/rust/src/lib/ip.rs#L365
https://github.com/nmstate/nmstate/blob/base/rust/src/lib/ip.rs#L683
I had to add the dhcp/auto if condition

rust/src/lib/nm/settings/ip.rs
andreaskaris: O.k. - yes, I misunderstood the earlier discussion (see my comment above), taking into account what you said above, that line's needed, I'll add it back

andreaskaris: we still need to do it conditionally, right? if prefix-route is defined, and is_auto() == true, and auto-route is empty, then we still want to go with the prefix route?

rust/src/lib/nm/settings/ip.rs
ihuguet: Almost there, but not exactly:

> if none is set, reset metric to default

If none is set: don't touch `nm_setting.route_metric`, don't reset it to default. This way, we will keep the previous value that was stored in NM, if any.

> if only one is Some(), set to that value

Yes, but for `auto_route_metric`, only if `is_auto()`. I know that, because of the sanitization, this will always happen. But better to be explicit here as well, so the code is more robust to changes.

> if both are Some() and have the same value, set to that value

Yes, as a consequence of the previous point.

> if both are Some() and have different values, throw an error

Yes.

rust/src/lib/nm/settings/ip.rs
ihuguet: Yes, if prefix-route is defined, we always have to set the value.

If prefix-route is not defined, but auto-route is, we only set the value if `is_auto()`.

rust/src/lib/nm/settings/ip.rs
andreaskaris: So if we're not resetting to default is the field isn't there, then how can admins explicitly reset back to the default (-1):
```
# /home/akaris/development/nmstate/rust/target/debug/nmstatectl apply nmstate.2b.yml 
[2026-02-13T14:19:55Z INFO  nmstatectl] Nmstate version: 2.2.59
Provide file is not valid NetworkState or NetworkPolicy: interfaces: invalid type: integer `-1`, expected unsigned integer or string at line 2 column 1
[root@metallb-k8s ~]# cat nmstate.2b.yml 
interfaces:
- name: enp2s0
  type: ethernet
  state: up
  ipv4:
    enabled: true
    dhcp: true
    address:
    - ip: 192.168.123.10
      prefix-length: 24
    forwarding: false
    prefix-route-metric: -1
  ipv6:
    enabled: true
    dhcp: false
    autoconf: false
- name: enp3s0
  type: ethernet
  state: up
  ipv4:
    enabled: true
    dhcp: true
    address:
    - ip: 192.168.124.10
      prefix-length: 24
    forwarding: false
    auto-route-metric: -1
  ipv6:
    enabled: true
    dhcp: false
    autoconf: false
```

I.e.: can set to 0 but cannot unset:
```
# nmcli conn show enp3s0  | grep metric
ipv4.route-metric:                      0
ipv6.route-metric:                      -1
```

andreaskaris: I had a look at the deserializers.rs stuff and at the current logic and with Some(uint32) there's really no way how to interpret this it seems ..
```
    #[serde(
        skip_serializing_if = "Option::is_none",
        rename = "prefix-route-metric",
        default,
        deserialize_with = "crate::deserializer::option_u32_or_string"
    )]
    pub prefix_route_metric: Option<u32>,
```

Would allow:
- interpret None as don't touch
---> no reset to default possible
- interpet None as reset to default
---> cannot leave the value as is

Unless I use the **option_i32_or_string** for both auto_route and prefix_route?

btw, with the old code, reset to default happens when is_auto() remains true (prerequisite) and the value for auto-route is removed so prior behavior afaict is a bit inconsistent

andreaskaris: @ihuguet @cathay4t TL;DR: if parameters prefix_route_metric and auto_route_metric are absent, should they reset nmcli metric to the default (-1). If they should not, how can an admin reset to the default of -1 given that this is an `option<u32>`? This is the same issue with auto_route_metric with current versions of nmstate which reset to -1 if is_auto() is still on but when an admin disables is_auto, it will keep the metric:

Current production versions of nmstate auto-route-metric reset to default when DHCP is on:

```
# nmstatectl version
[2026-02-16T11:15:13Z INFO  nmstatectl] Nmstate version: 2.2.57
nmstatectl 2.2.57
# cat nmstate.auto-prefix.yml 
interfaces:
- name: enp2s0
  type: ethernet
  state: up
  ipv4:
    enabled: true
    dhcp: true
    address:
    - ip: 192.168.123.10
      prefix-length: 24
    forwarding: false
    auto-route-metric: 149
  ipv6:
    enabled: true
    dhcp: false
    autoconf: false
# nmstatectl apply nmstate.auto-prefix.yml 
[2026-02-16T11:08:13Z INFO  nmstatectl] Nmstate version: 2.2.57
[2026-02-16T11:08:13Z INFO  nmstate::ip] Static addresses 192.168.123.10/24 defined when dynamic IP is enabled
[2026-02-16T11:08:13Z INFO  nmstate::query_apply::net_state] Created checkpoint /org/freedesktop/NetworkManager/Checkpoint/3
[2026-02-16T11:08:13Z INFO  nmstate::nm::query_apply::connection] Modifying connection b9551a15-9c8e-3f37-804e-93630040a1de: enp2s0/802-3-ethernet
[2026-02-16T11:08:13Z INFO  nmstate::nm::query_apply::connection] Reapplying connection b9551a15-9c8e-3f37-804e-93630040a1de: enp2s0/802-3-ethernet
[2026-02-16T11:08:13Z INFO  nmstate::query_apply::net_state] Destroyed checkpoint /org/freedesktop/NetworkManager/Checkpoint/3
[2026-02-16T11:08:13Z INFO  nmstate::ip] Static addresses 192.168.123.10/24 defined when dynamic IP is enabled
route-rules: {}
routes: {}
interfaces:
- name: enp2s0
  type: ethernet
  state: up
  ipv4:
    dhcp: true
    auto-dns: true
    auto-gateway: true
    auto-routes: true
    auto-route-metric: 149
ovn: {}

# nmcli conn show enp2s0 | grep metric
ipv4.route-metric:                      149
ipv6.route-metric:                      -1
# vim nmstate.auto-prefix.yml
# cat nmstate.auto-prefix.yml 
interfaces:
- name: enp2s0
  type: ethernet
  state: up
  ipv4:
    enabled: true
    dhcp: true
    address:
    - ip: 192.168.123.10
      prefix-length: 24
    forwarding: false
    # auto-route-metric: 149
  ipv6:
    enabled: true
    dhcp: false
    autoconf: false
# nmstatectl apply nmstate.auto-prefix.yml 
[2026-02-16T11:08:39Z INFO  nmstatectl] Nmstate version: 2.2.57
[2026-02-16T11:08:39Z INFO  nmstate::ip] Static addresses 192.168.123.10/24 defined when dynamic IP is enabled
[2026-02-16T11:08:39Z INFO  nmstate::query_apply::net_state] Created checkpoint /org/freedesktop/NetworkManager/Checkpoint/4
[2026-02-16T11:08:39Z INFO  nmstate::nm::query_apply::connection] Modifying connection b9551a15-9c8e-3f37-804e-93630040a1de: enp2s0/802-3-ethernet
[2026-02-16T11:08:39Z INFO  nmstate::nm::query_apply::connection] Reapplying connection b9551a15-9c8e-3f37-804e-93630040a1de: enp2s0/802-3-ethernet
[2026-02-16T11:08:39Z INFO  nmstate::query_apply::net_state] Destroyed checkpoint /org/freedesktop/NetworkManager/Checkpoint/4
[2026-02-16T11:08:39Z INFO  nmstate::ip] Static addresses 192.168.123.10/24 defined when dynamic IP is enabled
route-rules: {}
routes: {}
interfaces: []
ovn: {}

[root@metallb-k8s ~]# nmcli conn show enp2s0 | grep metric
ipv4.route-metric:                      -1
ipv6.route-metric:                      -1
```

But not when I also disable dhcp:
```
# nmstatectl apply nmstate.auto-prefix.yml 
[2026-02-16T11:10:37Z INFO  nmstatectl] Nmstate version: 2.2.57
[2026-02-16T11:10:37Z INFO  nmstate::ip] Static addresses 192.168.123.10/24 defined when dynamic IP is enabled
[2026-02-16T11:10:37Z INFO  nmstate::query_apply::net_state] Created checkpoint /org/freedesktop/NetworkManager/Checkpoint/5
[2026-02-16T11:10:37Z INFO  nmstate::nm::query_apply::connection] Modifying connection b9551a15-9c8e-3f37-804e-93630040a1de: enp2s0/802-3-ethernet
[2026-02-16T11:10:37Z INFO  nmstate::nm::query_apply::connection] Reapplying connection b9551a15-9c8e-3f37-804e-93630040a1de: enp2s0/802-3-ethernet
[2026-02-16T11:10:37Z INFO  nmstate::query_apply::net_state] Destroyed checkpoint /org/freedesktop/NetworkManager/Checkpoint/5
[2026-02-16T11:10:37Z INFO  nmstate::ip] Static addresses 192.168.123.10/24 defined when dynamic IP is enabled
route-rules: {}
routes: {}
interfaces:
- name: enp2s0
  type: ethernet
  state: up
  ipv4:
    auto-route-metric: 149
ovn: {}

# nmcli conn show enp2s0 | grep metric
ipv4.route-metric:                      149
ipv6.route-metric:                      -1
# vim nmstate.auto-prefix.yml
# cat nmstate.auto-prefix.yml 
interfaces:
- name: enp2s0
  type: ethernet
  state: up
  ipv4:
    enabled: true
    dhcp: false
    address:
    - ip: 192.168.123.10
      prefix-length: 24
    forwarding: false
    # auto-route-metric: 149
  ipv6:
    enabled: true
    dhcp: false
    autoconf: false
# nmstatectl apply nmstate.auto-prefix.yml 
[2026-02-16T11:10:59Z INFO  nmstatectl] Nmstate version: 2.2.57
[2026-02-16T11:10:59Z INFO  nmstate::query_apply::net_state] Created checkpoint /org/freedesktop/NetworkManager/Checkpoint/6
[2026-02-16T11:10:59Z INFO  nmstate::nm::query_apply::connection] Modifying connection b9551a15-9c8e-3f37-804e-93630040a1de: enp2s0/802-3-ethernet
[2026-02-16T11:10:59Z INFO  nmstate::nm::query_apply::connection] Reapplying connection b9551a15-9c8e-3f37-804e-93630040a1de: enp2s0/802-3-ethernet
[2026-02-16T11:10:59Z INFO  nmstate::query_apply::net_state] Destroyed checkpoint /org/freedesktop/NetworkManager/Checkpoint/6
route-rules: {}
routes: {}
interfaces:
- name: enp2s0
  type: ethernet
  state: up
  ipv4:
    dhcp: false
ovn: {}

# nmcli conn show enp2s0 | grep metric
ipv4.route-metric:                      149
ipv6.route-metric:                      -1
```

andreaskaris: I decided to switch it to i64 and accept -1 to explicitly set to the default value (the exact same as nmcli). That would also work for other backends, I suppose, given that valid metrics are likely [0 -> u32::MAX] everywhere
I'm open to other suggestions (e.g., accepting a "default" string, etc.), though the internal representation, I suppose, would be either i64 or an enum that accepts something like None, Some(), Default ?

Keeping it as 2 commits for review and to make changes easier, otherwise I'd squash them into a single commit once everything is o.k.

rust/src/lib/nm/settings/ip.rs
```diff
@@ -370,3 +412,57 @@ pub(crate) fn fix_ip_dhcp_timeout(nm_conns: &mut [NmConnection]) {
...
+    is_auto: bool,
+) -> Result<(), NmstateError> {
+    if prefix_route_metric.is_some()
+        && auto_route_metric.is_some()
+        && prefix_route_metric != auto_route_metric
+    {
+        let auto_str =
+            auto_route_metric.map_or("unset".to_string(), |v| v.to_string());
+        let prefix_str =
+            prefix_route_metric.map_or("unset".to_string(), |v| v.to_string());
```
cathay4t: Strange AI does not comment on this.

```rust
if let Some(prefix_metric) = prefix_route_metric &&
    let Some(auto_metric) = auto_route_metric &&
    prefix_metric != auto_metric {
// raise error.

}
```

rust/src/lib/nm/settings/ip.rs
```diff
@@ -370,3 +412,57 @@ pub(crate) fn fix_ip_dhcp_timeout(nm_conns: &mut [NmConnection]) {
...
+                "When using NetworkManager backend, {} auto-route-metric \
+                ({}) and prefix-route-metric ({}) must have the same value \
+                on interface {}",
+                address_family, auto_str, prefix_str, iface_name,
+            ),
+        ));
+    }
+    // -1 is a special value meaning "use NetworkManager's default metric"
+    if let Some(metric) = prefix_route_metric
+        && !(-1..=4294967295).contains(&metric)
```
cathay4t: u32::MAX as i64

rust/src/lib/nm/settings/ip.rs
```diff
@@ -370,3 +412,57 @@ pub(crate) fn fix_ip_dhcp_timeout(nm_conns: &mut [NmConnection]) {
...
+    {
+        return Err(NmstateError::new(
+            ErrorKind::NotSupportedError,
+            format!(
+                "When using NetworkManager backend, {} prefix-route-metric \
+                ({}) must be in interval [-1;4294967295] on interface {}",
```
cathay4t: u32::MAX

rust/src/lib/nm/settings/ip.rs
```diff
@@ -370,3 +412,57 @@ pub(crate) fn fix_ip_dhcp_timeout(nm_conns: &mut [NmConnection]) {
...
+    }
+    if let Some(metric) = auto_route_metric
+        && !(-1..=4294967295).contains(&metric)
+        && is_auto
+    {
+        return Err(NmstateError::new(
+            ErrorKind::NotSupportedError,
+            format!(
+                "When using NetworkManager backend, {} auto-route-metric \
+                ({}) must be in interval [-1;4294967295] on interface {}",
```
cathay4t: u32::MAX

rust/src/lib/ip.rs
```diff
@@ -199,6 +206,9 @@ pub struct InterfaceIpv4 {
     pub dhcp_custom_hostname: Option<String>,
     pub(crate) dns: Option<DnsClientState>,
     pub(crate) rules: Option<HashSet<RouteRuleEntry>>,
+    /// Metric for routes of prefixes directly connected to the interface.
+    /// Deserialize from `prefix-route-metric`
```
cathay4t, line 210: Please add document state `Some(RouteEntry::USE_DEFAULT_METRIC)` means use default value of backend.

tests/integration/static_ip_address_test.py
```diff
@@ -1299,3 +1299,319 @@ def test_kernel_mode_change_static_ip(cleanup_veth1_kernel_mode):
...
+                ],
+            },
+        },
+    }
+    for description, test_case in test_cases.items():
+        apply_with_description(
+            description,
+            test_case["configuration"],
+        )
+        assertlib.assert_state(test_case["desired"])
```
cathay4t: Use `assertlib.assert_state_match` instead, otherwise we will get error when adding more properties to ip section.

tests/integration/static_ip_address_test.py
```diff
@@ -1299,3 +1299,319 @@ def test_kernel_mode_change_static_ip(cleanup_veth1_kernel_mode):
...
+                                    InterfaceIPv6.ADDRESS_IP: IPV6_ADDRESS1,
+                                    InterfaceIPv6.ADDRESS_PREFIX_LENGTH: 64,
+                                }
+                            ],
+                        },
+                    }
+                ],
+            },
+        },
+    }
```
cathay4t: This test case is too big and very hard to understand.

Please split it into two test case with proper name.

tests/integration/static_ip_address_test.py
```diff
@@ -1299,3 +1299,319 @@ def test_kernel_mode_change_static_ip(cleanup_veth1_kernel_mode):
...
+            },
+            "desired": {
+                Interface.KEY: [
+                    {
+                        Interface.NAME: "eth1",
+                        Interface.TYPE: InterfaceType.ETHERNET,
+                        Interface.STATE: InterfaceState.UP,
+                        Interface.IPV4: {
+                            InterfaceIPv4.ENABLED: True,
+                            InterfaceIPv4.AUTO_ROUTE_METRIC: 901,
```
cathay4t: There is no DHCP on eth1, I see no point of having `InterfaceIPv4.AUTO_ROUTE_METRIC`

rust/src/lib/nm/query_apply/ip.rs
```diff
@@ -54,7 +54,8 @@ pub(crate) fn nm_ip_setting_to_nmstate4(
             } else {
                 None
             },
-            auto_route_metric: nm_ip_setting.route_metric.map(|i| i as u32),
+            // -1 is a special value meaning use NetworkManager's default metric
+            auto_route_metric: nm_ip_setting.route_metric.or(Some(-1)),
```
cathay4t, line 58: This property should be `None` when dhcpv4 or autoconf is false.

tests/integration/static_ip_address_test.py
```diff
@@ -1299,3 +1299,319 @@ def test_kernel_mode_change_static_ip(cleanup_veth1_kernel_mode):
...
+                            InterfaceIPv6.ADDRESS_IP: IPV6_ADDRESS1,
+                            InterfaceIPv6.ADDRESS_PREFIX_LENGTH: 64,
+                        }
+                    ],
+                    InterfaceIPv6.PREFIX_ROUTE_METRIC: 902,
+                    InterfaceIPv6.AUTO_ROUTE_METRIC: 902,
+                },
+            }
+        ]
+    }
```
cathay4t: These two dicts are duplicate code.

cathay4t (dismissed):
Besides change requests, unit test is also required for cases:

1. Raise error when ipv4.dhcp=true and two auto metric different.
2. Raise error when ipv6.autoconf=true and two auto metric different.
3. `for_apply` should have `auto_route_metric` set to None when ipv4.dhcp=false or ipv6.autoconf=false. No error even two auto metics different.

Regardless how NM behaves, the unit test case should make sure we behave exactly what we describe in API document.

For the problem of nmstate reset NM `ipv4.route-metric` when not desired, that should in another issue or PR.

cathay4t:
@andreaskaris Do you still have time to work on this? I can take over and finish the remaining works.

andreaskaris:
@cathay4t  hi! if you have time, feel free to do it. I'm a bit busy with other things atm; otherwise, I might be able to get back to it by the end of the week or early next week .. but if you have time to do it this week, I'd prefer if you could do it :)

cathay4t:
Let me work on this. Thank you!

cathay4t:
Changes I made(almost rewrite):

1. `ipv6.prefix-route-metric` is not supported by NetworkManager because NM does not create prefix-route for IPv6, hence removed.
2. The change on `test_auto_route_metric` has been removed. Using `prefix_route_metric` for `auto-routes` is hack, we will never claim support on that use case.
3. Add many use tests.
4. Moved prefix_route_metric test into `route_test.py` and rewrite from scratch.
5. Moved the sanitation code from `nm/ip.rs` to `ip.rs`.
6. Do not raise error of `prefix-route-metric` and `auto-route-metric` different but one of them is `RouteEntry::USE_DEFAULT_METRIC`. And also copy the non-default value to override default value.

cathay4t:
@andreaskaris Could you review my patch?

andreaskaris:
I will do so during the day, thanks!

rust/src/lib/nm/query_apply/ip.rs
ihuguet, line 58: Maybe create a `NM_DEFAULT_ROUTE_METRIC` constant?

rust/src/lib/nm/query_apply/ip.rs
```diff
@@ -72,6 +73,9 @@ pub(crate) fn nm_ip_setting_to_nmstate4(
             } else {
                 None
             },
+            prefix_route_metric: nm_ip_setting
+                .route_metric
+                .or(Some(RouteEntry::USE_DEFAULT_METRIC)),
```
ihuguet, line 78: USE_DEFAULT_METRIC is a value from nmstate. Here is more correct to use the new constant that I've suggested to create, or just `-1`.

rust/src/lib/nm/query_apply/ip.rs
```diff
@@ -123,7 +127,8 @@ pub(crate) fn nm_ip_setting_to_nmstate6(
                     None
                 }
             },
-            auto_route_metric: nm_ip_setting.route_metric.map(|i| i as u32),
+            // -1 is a special value meaning use NetworkManager's default metric
+            auto_route_metric: nm_ip_setting.route_metric.or(Some(-1)),
```
ihuguet, line 131: As we are converting NM-to-nmstate here, use `USE_DEFAULT_METRIC` instead of `-1`.

ihuguet (approved):

rust/src/lib/nm/query_apply/ip.rs
cathay4t: Fixed by using `nm_route_metric_to_nmstate()` and `nmstate_route_metric_to_nm()`.

rust/src/lib/ip.rs
```diff
@@ -684,10 +760,12 @@ impl InterfaceIpv6 {
             self.auto_gateway = None;
             self.auto_routes = None;
             self.auto_table_id = None;
-            self.auto_route_metric = None;
```
rajeshrah22, line 687: @coderabbitai why do you think this line was changed? It seems like when no explicit `autoconf` value is set, the default is None so the if condition below `if self.autoconf == Some(false)` will evaluate to false so `auto_route_metric` will stay and will fail verification later.

PR #3084, opened 2026-02-10
route: add support for lock-mtu option

cgoncalves:
When lock-mtu is set to true, the route MTU is locked, preventing Path MTU Discovery (PMTUD) from automatically adjusting it.

Example:
```
routes:
  config:
    - destination: 2001:db8:c::/64
      next-hop-interface: eth1
      mtu: 1280
      lock-mtu: true
```

### DIFF
--- a/rust/src/lib/route.rs
+++ b/rust/src/lib/route.rs
@@ -303,6 +303,13 @@ pub struct RouteEntry {
         deserialize_with = "crate::deserializer::option_u32_or_string"
     )]
     pub advmss: Option<u32>,
+    /// Lock route MTU, preventing Path MTU Discovery from adjusting it
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        default,
+        deserialize_with = "crate::deserializer::option_bool_or_string"
+    )]
+    pub lock_mtu: Option<bool>,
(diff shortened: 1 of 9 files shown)


rust/src/lib/route.rs
```diff
@@ -303,6 +303,14 @@ pub struct RouteEntry {
         deserialize_with = "crate::deserializer::option_u32_or_string"
     )]
     pub advmss: Option<u32>,
+    /// Lock route MTU, preventing Path MTU Discovery from adjusting it
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        default,
+        deserialize_with = "crate::deserializer::option_bool_or_string",
+        rename = "lock-mtu"
```
cathay4t: This rename is not required. We already have this line for `RouteEntry`

```rust
#[serde(rename_all = "kebab-case")]
```

cathay4t (changes requested):
The code looks great especially for first contributor.

Please just fix what AI mentioned and the small lint on `serde rename`.

cathay4t (approved):

cathay4t:
Merged. Thanks for the contribution!

PR #3098, opened 2026-03-09
ip: ipv4.forwarding deserialize from string also

rajeshrah22:
Be able to use "true" in place of true in ipv4.forwarding:
```
    interfaces:
      - name: eth1
        type: ethernet
        state: up
        ipv4:
          address:
          - ip: 192.0.2.1
            prefix-length: 24
          dhcp: false
          enabled: true
          forwarding: "true"
```

Resolves: https://issues.redhat.com/browse/RHEL-153618

### DIFF
--- a/rust/src/lib/deserializer.rs
+++ b/rust/src/lib/deserializer.rs
@@ -15,7 +15,7 @@ where
         if let Some(i) = i {
             Ok(i)
         } else {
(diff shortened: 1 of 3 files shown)


cathay4t (approved):
LGTM

cathay4t:
The CI failure is unrelated to this patch and a NM bug in nm-main.

PR #3105, opened 2026-03-24
bridge: allow extra ports in verification stage

rajeshrah22:
Ignores extra ports found in current state bridge during query_apply verification stage.

This avoids failiure when third party tools configures extra ports in a bridge at the same time as when nmstate is creating it. For example, CNI in OpenStack/OCP may attach extra ports to enable communication between pods/VMs.

Resolves: https://redhat.atlassian.net/browse/RHEL-153297

### DIFF
--- a/rust/src/lib/ifaces/inter_ifaces.rs
+++ b/rust/src/lib/ifaces/inter_ifaces.rs
@@ -903,6 +903,7 @@ impl MergedInterfaces {
     // Contains all the smart modifications, validations among interfaces
     fn process(&mut self) -> Result<(), NmstateError> {
         self.process_allow_extra_ovs_patch_ports_for_apply();
+        self.process_allow_extra_bridge_ports_for_apply();
         self.apply_copy_mac_from()?;
         self.validate_hsr_mac()?;
         self.copy_hsr_mac()?;
@@ -946,6 +947,15 @@ impl MergedInterfaces {
              interfaces in desire state to place controller before its ports"
                 .to_string(),
         ))
+    }
+
+    fn process_allow_extra_bridge_ports_for_apply(&mut self) {
+        for merged_iface in self
+            .iter_mut()
+            .filter(|mi| mi.desired.as_ref().map(|i| i.is_up()) == Some(true))
+        {
+            merged_iface.process_allow_extra_bridge_ports();
+        }
     }
 
     fn apply_copy_mac_from(&mut self) -> Result<(), NmstateError> {
--- a/rust/src/lib/ifaces/linux_bridge.rs
+++ b/rust/src/lib/ifaces/linux_bridge.rs
@@ -345,6 +345,12 @@ impl LinuxBridgeInterface {
 /// Linux bridge specific configuration.
 pub struct LinuxBridgeConfig {
     #[serde(skip_serializing_if = "Option::is_none")]
+    /// Only validate for applying, when set to `true`, Ignore extra ports
+    /// attached to bridge. Default is false. This property will not be
+    /// persisted, every time you modify ports of specified bridge, you
+    /// need to explicitly define this property if not using default value.
+    pub allow_extra_ports: Option<bool>,
+    #[serde(skip_serializing_if = "Option::is_none")]
     /// Linux bridge options. When applying, existing options will merged into
     /// desired.
     pub options: Option<LinuxBridgeOptions>,
--- a/rust/src/lib/query_apply/inter_ifaces.rs
+++ b/rust/src/lib/query_apply/inter_ifaces.rs
@@ -196,6 +196,7 @@ impl MergedInterfaces {
         current.remove_ignored_ifaces(self.ignored_ifaces.as_slice());
         current.remove_unknown_type_port();
         merged.process_allow_extra_ovs_patch_ports_for_verify(&mut current);
+        merged.process_allow_extra_bridge_ports_for_verify(&mut current);
 
         for iface in current
             .kernel_ifaces
@@ -254,5 +255,25 @@ impl MergedInterfaces {
             }
         }
         Ok(())
+    }
+
+    pub(crate) fn process_allow_extra_bridge_ports_for_verify(
+        &mut self,
+        post_apply_current: &mut Interfaces,
+    ) {
+        for merged_iface in self.iter().filter(|merged_iface| {
+            if let Some(i) = merged_iface.desired.as_ref() {
+                i.allow_extra_ports() == Some(true) && i.is_up()
+            } else {
+                false
+            }
+        }) {
+            if let Some(des_iface) = merged_iface.desired.as_ref()
+                && let Some(cur_iface) = post_apply_current
+                    .get_iface_mut(des_iface.name(), des_iface.iface_type())
+            {
+                des_iface.remove_extra_ports(cur_iface);
+            }
+        }
     }
 }
--- a/rust/src/lib/unit_tests/bridge.rs
+++ b/rust/src/lib/unit_tests/bridge.rs
@@ -769,3 +769,146 @@ fn test_linux_bridge_is_default_pvid_changed() {
 
     assert!(!merged_iface.is_default_pvid_changed())
 }
+
+#[test]
+fn test_linux_bridge_ignore_extra_ports_for_verify() {
+    let des_ifaces: Interfaces = serde_yaml::from_str(
+        r"---
+        - name: eth1
+          type: ethernet
+          state: up
+        - name: eth2
+          type: ethernet
+          state: up
+        - name: br0
+          type: linux-bridge
+          state: up
+          bridge:
+            allow-extra-ports: true
+            port:
+            - name: eth1
+            - name: eth2
+        ",
+    )
+    .unwrap();
+    let pre_apply_cur_ifaces: Interfaces = serde_yaml::from_str(
+        r"---
+        - name: eth1
+          type: ethernet
+          state: up
+        - name: eth2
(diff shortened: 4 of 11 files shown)


rust/src/lib/ifaces/linux_bridge.rs
```diff
@@ -339,11 +339,49 @@ impl LinuxBridgeInterface {
     }
 }
 
+impl MergedInterface {
+    // If `bridge.allow_extra_ports`, prevent the extra ports from being
+    // deleted when applying by removing the extra ports from current interface
```
rajeshrah22: This may not make sense and may not be needed for bridge ports actually. I took this idea from `allow_extra_patch_ports` patch.

rust/src/lib/query_apply/inter_ifaces.rs
```diff
@@ -255,4 +256,76 @@ impl MergedInterfaces {
...
+                    }
+                    (
+                        b.base.name.clone(),
+                        InterfaceType::LinuxBridge,
+                        b.ports()
+                            .unwrap_or_default()
+                            .into_iter()
+                            .map(str::to_string)
+                            .collect(),
+                    )
```
cathay4t: This long block of assignment is not for human been. Please polish it.

rust/src/lib/query_apply/inter_ifaces.rs
```diff
@@ -255,4 +256,76 @@ impl MergedInterfaces {
...
+                _ => continue,
+            };
+            if let Some(cur_iface) = current.get_iface_mut(&name, iface_type) {
+                remove_extra_ports(cur_iface, &desired_ports);
+            }
+        }
+    }
+}
+
+fn remove_extra_ports(cur_iface: &mut Interface, desired_ports: &[String]) {
```
cathay4t: This small function is used only once. Inline is better.

rust/src/lib/unit_tests/bridge.rs
```diff
@@ -769,3 +769,140 @@ fn test_linux_bridge_is_default_pvid_changed() {
...
+    .unwrap();
+
+    let for_apply_bridge = merged_ifaces
+        .get_iface("br0", InterfaceType::LinuxBridge)
+        .unwrap()
+        .for_apply
+        .as_ref()
+        .unwrap();
+
+    assert_eq!(for_apply_bridge.ports(), Some(vec!["eth1", "eth2"]));
```
cathay4t, line 912: Also need to check `MergedInterface` of `eth3` is holding `for_apply: None`.

rust/src/lib/unit_tests/ovs.rs
```diff
@@ -1022,3 +1022,140 @@ bridge:
...
+    .unwrap();
+
+    let for_apply_bridge = merged_ifaces
+        .get_iface("br0", InterfaceType::OvsBridge)
+        .unwrap()
+        .for_apply
+        .as_ref()
+        .unwrap();
+
+    assert_eq!(for_apply_bridge.ports(), Some(vec!["eth1", "eth2"]));
```
cathay4t, line 1165: Please check `eth3` holds `for_apply: None`.

cathay4t (changes requested):
Integration test case like `test_allow_extra_ovs_patch_ports` required. You may use `dummy` interface as extra besides desired eth1 and eth2.

rust/src/lib/query_apply/inter_ifaces.rs
```diff
@@ -255,4 +256,70 @@ impl MergedInterfaces {
         }
         Ok(())
     }
+
+    pub(crate) fn process_allow_extra_bridge_ports_for_verify(
```
cathay4t, line 260: The `process_allow_extra_bridge_ports_for_apply()` and `process_allow_extra_bridge_ports_for_verify()`  are almost identical. Please merge them.

rust/src/lib/ifaces/inter_ifaces.rs
```diff
@@ -1129,6 +1130,58 @@ impl MergedInterfaces {
...
+                    if b.bridge.as_ref().and_then(|c| c.allow_extra_ports)
+                        == Some(true) =>
+                {
+                    if let Some(Interface::LinuxBridge(cur_iface)) =
+                        mi.current.as_mut()
+                    {
+                        let des_ports = b.ports().unwrap_or_default();
+                        let Some(cur_ports) = cur_iface.ports() else {
+                            return;
+                        };
```
cathay4t: To simplify this function, you may add `Interface::allow_extra_port() -> Option<bool>`, so you can remove duplicate code here on linux bridge and ovs bridge because we can use `Interface::ports()` and `Interface::remove_port()`.

Meanwhile, use chained `if let`, something like:

```rust
    let pending_changes: HashMap<String, String> = HashMap::new();
    if merged_iface
        .desired
        .as_ref()
        .and_then(|i| i.allow_extra_port())
        == Some(true)
        && let Some(des_ports) =
            merged_iface.desired.as_ref().and_then(|i| i.ports())
        && let Some(cur_ports) =
            merged_iface.current.as_ref().and_then(|i| i.ports())
    {
        for cur_port in cur_ports.iter().filter(|p| !des_ports.contains(p)) {
            pending_changes.insert(
                merged_iface.merged.name().to_string(),
                cur_port.to_string(),
            );
        }
    }
    // Remove ports from cur_iface using pending changes.
```

cathay4t (changes requested):

rust/src/lib/query_apply/inter_ifaces.rs
rajeshrah22: I have found it difficult to merge into one function. One way is the create a function with signature `process_allow_extra_bridge_ports(&self, current: Option<&mut Interfaces>, for_verify: bool)` to differentiate between a pre-apply stage call and a post-apply stage verification call.

In post-apply verification stage call, we must find and mutate iface in `current: &mut Interfaces` vec, but this is not available in the pre-apply stage `MergedInterfaces::process()` call.

I have gone with an alternative way:
```rust
// called from MergedInterfaces::process()
fn process_allow_extra_bridge_ports_for_apply(&mut self) {
     for merged_iface in self.iter_mut() {
        if let Some(mut cur_iface) = merged_iface.current.take() {
            merged_iface
                .process_allow_extra_bridge_ports(&mut cur_iface, false);
            merged_iface.current = Some(cur_iface);
        }
    }
}
```

```rust
// called from MergedInterfaces::verify()
pub(crate) fn process_allow_extra_bridge_ports_for_verify(
    &mut self,
    current: &mut Interfaces,
) {
    for merged_iface in self.iter_mut() {
        if let Some(des_iface) = merged_iface.desired.as_ref()
            && let Some(cur_iface) = current
                .get_iface_mut(des_iface.name(), des_iface.iface_type())
        {
            merged_iface.process_allow_extra_bridge_ports(cur_iface, true);
        }
    }
}
```

```rust
    pub(crate) fn process_allow_extra_bridge_ports(
        &mut self,
        cur_iface: &mut Interface,
        for_verify: bool,
    ) {
        if self
            .desired
            .as_ref()
            .and_then(|des_iface| des_iface.allow_extra_ports())
            != Some(true)
        {
            return;
        }
        if for_verify && let Some(i) = self.for_verify.as_mut() {
            i.remove_allow_extra_ports();
        }
        let Some(des_ports) = self.desired.as_ref().and_then(|i| i.ports()) else {
            return;
        };
        let Some(cur_ports) = cur_iface.ports() else {
            return;
        };
        let to_delete: Vec<String> = cur_ports
            .iter()
            .filter(|p| !des_ports.contains(p))
            .map(|p| p.to_string())
            .collect();
        if to_delete.is_empty() {
            return;
        }
        for port in to_delete {
            cur_iface.remove_port(&port);
        }
    }
```

cathay4t: The `for_verify` does not need `remove_extra_ports()`. Only `pre_apply_current` and `post_apply_current`

rajeshrah22: Maybe we need to rename that function but `remove_extra_ports()` just sets `allow_extra_ports` to `None`.
We get verification failure if we don't do that in `for_verify`:
```
thread 'unit_tests::bridge::test_linux_bridge_ignore_extra_ports_for_verify' (1373951) panicked at src/lib/unit_tests/bridge.rs:846:39:
called `Result::unwrap()` on an `Err` value: NmstateError { kind: VerificationError, msg: "Verification failure: br0.interface.bridge.allow-extra-ports desire 'true', current 'null'", line: "", position: 0 }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- unit_tests::ovs::test_ovs_bridge_ignore_extra_ports_for_verify stdout ----

thread 'unit_tests::ovs::test_ovs_bridge_ignore_extra_ports_for_verify' (1374147) panicked at src/lib/unit_tests/ovs.rs:1099:39:
called `Result::unwrap()` on an `Err` value: NmstateError { kind: VerificationError, msg: "Verification failure: br0.interface.bridge.allow-extra-ports desire 'true', current 'null'", line: "", position: 0 }
```

cathay4t (changes requested):
Please add a integration test for creating linux and ovs bridge with out

rajeshrah22:
@cathay4t I changed code to filter on `allow_extra_ports == Some(true)` and `iface.is_up()` to avoid unit tests failures.

cathay4t:
Please fix the CI failure.

cathay4t (approved):

PR #3109, opened 2026-03-27
governance: Add project governance document

mkowalski:
Based on the OVN-K8s governance with some changes.


mkowalski:
Part of https://github.com/cncf/sandbox/issues/476

cathay4t (approved):

ffmancera:
LGTM, thanks!

(I believe we drafted something like this in the past but we never finished it for some reason..)

PR #3119, opened 2026-04-08
policy: support array property capture

jvaclav-rh:
Extend the policy capture JSON matching to handle matching against any element in an array.

For example, this allows capturing interfaces by alt-name:
```
---
capture:
  base-iface: interfaces.alt-names.name == "primary"
desiredState:
  interfaces:
    - name: br1
      type: linux-bridge
      state: up
      mac-address: "{{ capture.base-iface.interfaces.0.mac-address }}"
      ipv4: "{{ capture.base-iface.interfaces.0.ipv4 }}"
      bridge:
        port:
          - name: "{{ capture.base-iface.interfaces.0.name }}"
```

Unit tests and example included.

Resolves: https://redhat.atlassian.net/browse/RHEL-145076

### DIFF
--- a/examples/policy/bridge-on-alt-name/current.yml
+++ b/examples/policy/bridge-on-alt-name/current.yml
@@ -1,0 +1,14 @@
+---
+interfaces:
+  - name: eth0
+    type: ethernet
+    state: up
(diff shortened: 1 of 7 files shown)


cathay4t (approved):
Nice work!

PR #3120, opened 2026-04-13
tests: remove unnecessary eth1_up fixture from mac identifier test

vbenes:
The `eth1_up` fixture creates a name-based `eth1` connection that competes with the test's MAC-based `port1` connection during verify, causing ~30% failure rate in CI.

The test only needs the kernel interface for its MAC address, not an NM-managed connection.

See analysis: https://github.com/nmstate/nmstate/pull/3111#issuecomment-4236905670


ihuguet:
Closing as per this comment: https://github.com/nmstate/nmstate/pull/3111#issuecomment-4236968153

PR #3122, opened 2026-04-15
test: add integration tests for policy capture by alt-name

fpokryvk:
Add tier1 integration tests for NMT-2202 verifying that nmpolicy capture rules can match interfaces by alt-name.

Assisted-by: Claude Opus 4.6

### DIFF
--- a/tests/integration/alt_name_test.py
+++ b/tests/integration/alt_name_test.py
@@ -433,3 +433,85 @@ class TestAltNames:
             "eth1",
             [],
         )
+
+    # https://issues.redhat.com/browse/NMT-2202
+    @pytest.mark.tier1
+    def test_policy_capture_by_alt_name(self, eth1_with_alt_names):
+        libnmstate.apply(
+            load_yaml(
+                """---
+                interfaces:
+                - name: eth1
+                  type: ethernet
+                  state: up
+                  ipv4:
+                    address:
+                    - ip: 192.0.2.10
+                      prefix-length: 24
+                    dhcp: false
+                    enabled: true
+                """
+            )
+        )
+        eth1_state = show_only(("eth1",))[Interface.KEY][0]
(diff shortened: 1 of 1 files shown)


tests/integration/alt_name_test.py
```diff
@@ -433,3 +434,128 @@ def test_both_alt_name_iface_and_kernel_iface_mark_as_absent(
...
+            br_state = show_only((TEST_BRIDGE_NIC,))[Interface.KEY][0]
+            assert state_match(
+                [{"name": "eth1"}],
+                br_state[LinuxBridge.CONFIG_SUBTREE][
+                    LinuxBridge.PORT_SUBTREE
+                ],
+            )
+
+    # https://issues.redhat.com/browse/NMT-2202
+    def test_policy_capture_by_alt_name_not_found(
```
jvaclav-rh: This is already tested by:
https://github.com/nmstate/nmstate/blob/18c56d45ef9f4c195c2e46f63036f1da12507a52/rust/src/lib/unit_tests/policy/error.rs#L497-L511

I don't think it's necessary to test it here also.

tests/integration/alt_name_test.py
```diff
@@ -433,3 +434,128 @@ def test_both_alt_name_iface_and_kernel_iface_mark_as_absent(
...
         )
+
+    # https://issues.redhat.com/browse/NMT-2202
+    @pytest.mark.tier1
+    def test_policy_capture_by_alt_name(self, eth1_with_alt_names):
+        """
+        Capture an interface by alt-name in nmpolicy and use its properties
+        to configure another interface. This is the core OCP use case where
+        nmpolicy copies MAC/IP from a physical interface to a bridge.
+        """
```
jvaclav-rh: these docstrings are probably not needed (also below)

ihuguet (approved):

PR #3124, opened 2026-04-15
fix: fail when multiple devices match mac-identifier

### LINKED ISSUE #3123: resolve_mac_identifier_in_desired() is non-deterministic with duplicate MACs
## Problem

When two interfaces have the same MAC address, `resolve_mac_identifier_in_desired()` in `inter_ifaces.rs` resolves to a random interface. This is because it iterates `current.kernel_ifaces.values()` which is a Rust `HashMap` — iteration order is non-deterministic.

The function takes the first matching MAC and breaks, so the result depends on HashMap internal ordering which changes between runs.

https://github.com/nmstate/nmstate/blob/base/rust/src/lib/ifaces/inter_ifaces.rs#L411

## Reproducer

```bash
# Setup: two veth pairs with same MAC
ip link del eth1 2>/dev/null; ip link del eth2 2>/dev/null
nmcli con del eth1 eth2 port1 2>/dev/null
ip link add eth1 type veth peer name eth1.ep
ip link add eth2 type veth peer name eth2.ep
ip link set eth1 up
ip link set eth2 up
ip link set eth1 address 02:00:00:00:00:01
ip link set eth2 address 02:00:00:00:00:01
nmcli device set eth1 managed yes
nmcli device set eth2 managed yes

# Reproduce: apply state with identifier: mac-address
# Expected: always resolves to the same interface
# Actual: randomly resolves to eth1 or eth2
for i in $(seq 1 10); do
  nmcli con del port1 2>/dev/null
  python3 -c "
import libnmstate, yaml
state = yaml.safe_load('''
interfaces:
- name: port1
  type: ethernet
  identifier: mac-address
  mac-address: 02:00:00:00:00:01
  description: port1
  state: down
''')
libnmstate.apply(state)
current = libnmstate.show()
for iface in current['interfaces']:
    if iface.get('description') == 'port1':
        print(f'run \$i: port1 landed on', iface['name'])
"
done
```

## Output

```
run 1: port1 landed on eth2
run 3: port1 landed on eth2
run 4: port1 landed on eth1
run 5: port1 landed on eth1
run 6: port1 landed on eth2
run 7: port1 landed on eth1
run 8: port1 landed on eth2
run 9: port1 landed on eth1
run 10: port1 landed on eth1
```

## Impact

This causes ~30% intermittent failure rate of `test_description_on_down_interface_ref_by_mac` in CI when bridge tests leave eth1 and eth2 with the same MAC address.

In production, duplicate MACs can occur with bonding (bond sets same MAC on all ports and NM may not restore original MAC after bond removal) or manual configuration.

vbenes:
## Summary
- `resolve_mac_identifier_in_desired()` iterates a Rust `HashMap` to find interfaces by MAC address. HashMap iteration is non-deterministic, so when multiple interfaces share the same MAC, the result is random.
- Sort interface names before iterating to always pick the same interface (lowest name alphabetically).
- Add tier2 integration test that creates a duplicate-MAC interface and verifies deterministic resolution.

Fixes: #3123

## Reproducer
See https://github.com/nmstate/nmstate/issues/3123 — 10/10 FAIL without fix, the test resolves to random interface each run.

## Test plan
- [x] New integration test `test_mac_identifier_deterministic_with_duplicate_mac` — 10/10 FAIL without fix
- [ ] CI passes with this change

### DIFF
--- a/rust/src/lib/ifaces/inter_ifaces.rs
+++ b/rust/src/lib/ifaces/inter_ifaces.rs
@@ -428,6 +428,18 @@ impl Interfaces {
                             .map(|m| m.to_ascii_uppercase())
                     };
                     if cur_mac_addr.as_deref() == Some(&mac_address) {
+                        if has_match {
+                            return Err(NmstateError::new(
+                                ErrorKind::InvalidArgument,
+                                format!(
+                                    "Desired interface {} has `identifier: \
+                                     mac-address` with MAC address \
+                                     {mac_address}, but multiple interfaces \
+                                     are holding that MAC address",
+                                    iface.name()
+                                ),
+                            ));
+                        }
                         let mut new_iface = if iface.iface_type()
                             == InterfaceType::Unknown
                         {
@@ -451,7 +463,6 @@ impl Interfaces {
                             cur_iface.name().to_string();
                         changed_ifaces.push(new_iface);
                         has_match = true;
-                        break;
                     }
                 }
                 if has_match {
--- a/tests/integration/mac_identifer_test.py
+++ b/tests/integration/mac_identifer_test.py
@@ -609,3 +609,36 @@ def test_description_on_down_interface_ref_by_mac(eth1_up):
           """
     )
     assert_state_match(expected_state)
+
+
+@pytest.mark.tier2
+def test_mac_identifier_error_on_duplicate_mac(eth1_up, clean_up):
+    """Duplicate MACs with mac-identifier should raise an error."""
+    from .testlib.cmdlib import exec_cmd
+
+    eth1_mac = get_mac_address("eth1")
+
+    exec_cmd(
+        "ip link add ethX type veth peer name ethX.ep".split(), check=True
+    )
+    exec_cmd(["ip", "link", "set", "ethX", "address", eth1_mac], check=True)
+    exec_cmd("ip link set ethX up".split(), check=True)
+    exec_cmd("nmcli device set ethX managed yes".split(), check=True)
+
+    try:
+        with pytest.raises(libnmstate.error.NmstateValueError):
+            libnmstate.apply(
+                load_yaml(
+                    f"""---
+                    interfaces:
+                    - name: port1
+                      type: ethernet
+                      identifier: mac-address
+                      mac-address: {eth1_mac}
+                      state: down
+                    """
+                )
+            )
+    finally:
+        exec_cmd("ip link del ethX".split())
+        exec_cmd("ip link del ethX.ep".split())


vbenes:
@gemini-code-assist review

rust/src/lib/ifaces/inter_ifaces.rs
```diff
@@ -408,7 +408,13 @@ impl Interfaces {
...
             // `mac_address`
             for use_permanent_addr in [true, false] {
-                for cur_iface in current.kernel_ifaces.values() {
+                // Sort interface names for deterministic resolution
+                // when multiple interfaces share the same MAC address
+                let mut sorted_names: Vec<_> =
+                    current.kernel_ifaces.keys().collect();
+                sorted_names.sort();
+                for name in sorted_names {
+                    let cur_iface = &current.kernel_ifaces[name];
```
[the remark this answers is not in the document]
vbenes: done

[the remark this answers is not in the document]
vbenes: done

ihuguet:
I think that the correct solution would be to raise an error if you try to apply a desired state matching by MAC, and multiple devices match.

Picking the lowest alphabetically is slightly better than picking one pseudo-randomly. But from a user's point of view, I don't think they are going to rely on that. If they try to match by MAC, and there are two identical MACs, I'm almost sure that it's a user mistake, and better to report it to the user.

cathay4t:
How about we change the code to raise error when desired state want to match by mac on multiple interface?

Then we can tell which test case caused the random CI failure and how.

ihuguet:
> How about we change the code to raise error when desired state want to match by mac on multiple interface?
> 
> Then we can tell which test case caused the random CI failure and how.

Vlad had already identified the test that want to match by MAC. But the bug was caused by other tests that didn't do a proper cleanup, leaving eth1 and eth2 with the same MAC.

However, this PR makes sense because if a user wants to match by MAC, it should fail if there are 2 devices with the same MAC.

cathay4t (approved):

rust/src/lib/ifaces/inter_ifaces.rs
ihuguet, line 470: Because of this, if one device matches by permanent MAC and other by MAC, we won't raise an error. Probably that's OK, though. WDYT @cathay4t ?

tests/integration/mac_identifer_test.py
```diff
@@ -609,3 +609,36 @@ def test_description_on_down_interface_ref_by_mac(eth1_up):
           """
     )
     assert_state_match(expected_state)
+
+
+@pytest.mark.tier2
+def test_mac_identifier_error_on_duplicate_mac(eth1_up, clean_up):
+    """Duplicate MACs with mac-identifier should raise an error."""
+    from .testlib.cmdlib import exec_cmd
+
```
ihuguet, line 618: Put this at the top of the file, please

tests/integration/mac_identifer_test.py
```diff
@@ -609,3 +609,36 @@ def test_description_on_down_interface_ref_by_mac(eth1_up):
...
+    exec_cmd(
+        "ip link add ethX type veth peer name ethX.ep".split(), check=True
+    )
+    exec_cmd(["ip", "link", "set", "ethX", "address", eth1_mac], check=True)
+    exec_cmd("ip link set ethX up".split(), check=True)
+    exec_cmd("nmcli device set ethX managed yes".split(), check=True)
+
+    try:
+        with pytest.raises(libnmstate.error.NmstateValueError):
```
ihuguet, line 629: Please check a relevant portion of the error message, like "multiple interfaces are holding that MAC", so we're sure that the error is the one that we expected.

tests/integration/mac_identifer_test.py
```diff
@@ -609,3 +609,36 @@ def test_description_on_down_interface_ref_by_mac(eth1_up):
...
+                      type: ethernet
+                      identifier: mac-address
+                      mac-address: {eth1_mac}
+                      state: down
+                    """
+                )
+            )
+    finally:
+        exec_cmd("ip link del ethX".split())
+        exec_cmd("ip link del ethX.ep".split())
```
ihuguet, line 644: This line is not needed, both peers are deleted with the first command.

PR #3129, opened 2026-04-22
alt-name: use OriginalName match rule for non-ethernet interfaces

jvaclav-rh:
When a non-ethernet interface with alt-names already exists in the kernel at apply time, the .link file [Match] section was generated using MACAddress instead of OriginalName. This happened because the interface type check was only applied when current state was absent (first apply), but not when current state was present.

For a VLAN on top of a bond, the MAC is inherited from the bond's active port and can change on failover. Using MACAddress for the match rule is both unreliable and can match the wrong interface since the bond and all its VLANs share the same MAC.

Hoist the non-ethernet type check to the top of
gen_systemd_network_link_match_rules() so that OriginalName is always used for non-ethernet interfaces (as implied by the original comment) regardless of whether current state exists.

Resolves: https://redhat.atlassian.net/browse/RHEL-167955

Assisted-by: Claude Opus 4.6

### DIFF
--- a/rust/src/lib/nispor/alt_name.rs
+++ b/rust/src/lib/nispor/alt_name.rs
@@ -282,7 +282,11 @@ async fn save_systemd_network_link_file(
     }
     let iface_name = des_base_iface.name.as_str();
 
-    let match_rules = match des_base_iface.identifier.as_ref() {
+    let identifier = &des_base_iface
+        .identifier
+        .or_else(|| cur_base_iface.and_then(|i| i.identifier));
+
+    let match_rules = match identifier {
         Some(InterfaceIdentifier::MacAddress) => {
             let mut match_rules = if let Some(mac) =
                 des_base_iface.permanent_mac_address.as_ref()
--- a/rust/src/lib/nispor/alt_name.rs
+++ b/rust/src/lib/nispor/alt_name.rs
@@ -217,6 +217,10 @@ fn gen_systemd_network_link_match_rules(
     des_iface: &BaseInterface,
     cur_iface: Option<&BaseInterface>,
 ) -> String {
+    if des_iface.iface_type != InterfaceType::Ethernet {
+        return format!("OriginalName={}", des_iface.name);
+    }
+
     if let Some(cur_iface) = cur_iface {
         match (
             cur_iface.permanent_mac_address.as_ref(),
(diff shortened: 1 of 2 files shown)


rust/src/lib/nispor/alt_name.rs
```diff
@@ -217,6 +217,10 @@ fn gen_systemd_network_link_match_rules(
     des_iface: &BaseInterface,
     cur_iface: Option<&BaseInterface>,
 ) -> String {
+    if des_iface.iface_type != InterfaceType::Ethernet {
+        return format!("OriginalName={}", des_iface.name);
+    }
```
ihuguet, line 222: I don't like much the idea of making Ethernet so special. We should aim to a more generic solution, if possible.

Idea: function `save_systemd_network_link_file` uses `des_base_iface.identifier` to define a criteria. If we don't have a desired state, but only a current state, we can also use `cur_base_iface.identifier`. It is filled from NM's data in `lib/nm/show.rs:fill_identifier`.

By doing that we will allow respecting the identifier that was initially set for any device type. Otherwise this may happen for non-ethernet devices:

```
# Per save_systemd_network_link_file, this will generate a link file with `MacAddress=...`
interfaces:
- identifier: mac-address
  mac-address: ...
  alt-names:
  - name: my-alt-name

# Now we apply a new state changing alt-names, it will change the link file to `OriginalName=`
interfaces:
- name: vlan0
  alt-names
  - name: my-new-alt-name
```

Even with my suggestion, non-ethernet devices with `identifier: empty/name` will still fallback to use `OriginalName`. That's still not ideal, but we can keep it as out of scope for this PR. I mean, with `identifier: empty/name`, keep ethernet as a special case, as it is now.

cathay4t: The udev might change `OriginalName` when kernel driver decided to expose more information(like switch port number, pci function number, device type, etc), and we see it happening in real customer environment.

The whole point of using `alt-names` is make sure user get a  consistent interface name regardless how kernel/OS upgrades, depending on udev `OriginalName` is not the use case we try to solve.

For use care where non-ethernet interfaces holds alt-names but desired alt-names changes, we need to understand what's their real use case and design it properly.

cathay4t (changes requested):
I also dislike make Ethernet special on alt-names.

jvaclav-rh:
Superseded by #3171

PR #3131, opened 2026-04-22
release: refactor and harden upstream release script

jvaclav-rh:


automation/upstream_release.sh
```diff
@@ -25,6 +25,46 @@ CODE_BASE_DIR=$(readlink -f "$(dirname -- "$0")/..");
...
+    echo "Cargo registry token not found. Please log in using:" >&2
+    echo " $ cargo login" >&2
+    echo "and confirm your email on crates.io: https://crates.io/settings/profile" >&2
+    echo "The token you use to log in should have at least 'publish-update' permissions." >&2
+    exit 1
+fi
+
+CARGO_REGISTRY_TOKEN=$(grep -oP 'token = "\K[^"]+' ~/.cargo/credentials.toml)
+if ! curl -s "https://crates.io/api/v1/crates?following=1" \
+          -H "Authorization: Bearer $CARGO_REGISTRY_TOKEN" | grep -q "perform this action"; then
```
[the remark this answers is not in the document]
jvaclav-rh: This is expected, if the token is invalid a different error is returned ("authentication failed")

automation/upstream_release.sh
```diff
@@ -25,6 +25,46 @@ CODE_BASE_DIR=$(readlink -f "$(dirname -- "$0")/..");
...
+
+if ! [ -f ~/.cargo/credentials.toml ]; then
+    echo "Cargo registry token not found. Please log in using:" >&2
+    echo " $ cargo login" >&2
+    echo "and confirm your email on crates.io: https://crates.io/settings/profile" >&2
+    echo "The token you use to log in should have at least 'publish-update' permissions." >&2
+    exit 1
+fi
+
+CARGO_REGISTRY_TOKEN=$(grep -oP 'token = "\K[^"]+' ~/.cargo/credentials.toml)
```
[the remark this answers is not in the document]
jvaclav-rh: we do not care about macos

automation/upstream_release.sh
```diff
@@ -25,6 +25,46 @@ CODE_BASE_DIR=$(readlink -f "$(dirname -- "$0")/..");
...
+    err=0
+    if ! command -v $cmd > /dev/null 2>&1; then
+        echo "$cmd not found. Please install it." >&2
+        err=1
+    fi
+
+    if [ $err = 1 ]; then
+        exit 1
+    fi
+done
```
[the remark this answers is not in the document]
ihuguet: I agree, I would print an additional line with all the required commands.
Also, just `exit 1` from within the loop, no need for `err`.

automation/upstream_release.sh
```diff
@@ -25,6 +25,53 @@ CODE_BASE_DIR=$(readlink -f "$(dirname -- "$0")/..");
...
+if [ $err = 1 ]; then
+    exit 1
+fi
+
+if ! git fetch upstream 2>/dev/null; then
+    echo "Upstream repository not found. If you use ssh keys to push, add it using:" >&2
+    echo " $ git remote add upstream $UPSTREAM_GIT_SSH" >&2
+    echo "Otherwise, if you use HTTPS, add it using:" >&2
+    echo " $ git remote add upstream $UPSTREAM_GIT_HTTPS" >&2
```
ihuguet, line 44: Using it via HTTPS may be a source of confusion, as you have to provide a token as password. I would suggest to set it only as ssh.
IMHO it's OK to require ssh access, this is only for maintainers, not for normal users.

automation/upstream_release.sh
```diff
@@ -25,6 +25,53 @@ CODE_BASE_DIR=$(readlink -f "$(dirname -- "$0")/..");
...
+
+if ! [ -f ~/.cargo/credentials.toml ]; then
+    echo "Cargo registry token not found. Please log in using:" >&2
+    echo " $ cargo login" >&2
+    echo "and confirm your email on crates.io: https://crates.io/settings/profile" >&2
+    echo "The token you use to log in should have at least 'publish-update' permissions." >&2
+    exit 1
+fi
+
+CARGO_REGISTRY_TOKEN=$(grep -oP 'token = "\K[^"]+' ~/.cargo/credentials.toml)
```
ihuguet: I think that you can do `cargo logout` and then the credentials.toml file will exist, but without a token. I think that the error message must be shown also if the grep command fails.

ihuguet:
I've checked with Gris and there was no reason to not updating nmstate-clib. He has confirmed that we should add the command to publish it into upstream_release.sh.

automation/upstream_release.sh
jvaclav-rh: I left it there since some maintainer might use HTTPS, but I can remove it. I assumed Gris uses https:// for git operations, and that's why the script defaulted to it

automation/upstream_release.sh
jvaclav-rh: I merged the file check and the token check into one `if`, should be cleaner now.

automation/upstream_release.sh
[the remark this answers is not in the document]
jvaclav-rh: I left the `if` in the loop here by mistake, I fixed it by moving it out. The goal is to have it print out all the missing packages, like:
```
hub not found. Please install it.
vim not found. Please install it.
```
and only exit after.

automation/upstream_release.sh
ihuguet: @cathay4t any concern about suggesting only ssh acces to the 'upstream' remote?

automation/upstream_release.sh
```diff
@@ -89,6 +129,8 @@ cd $CODE_BASE_DIR/rust/src/lib
 cargo publish --allow-dirty
 cd $CODE_BASE_DIR/rust/src/cli
 cargo publish --allow-dirty
+cd $CODE_BASE_DIR/rust/src/clib
+cargo publish --allow-dirty
```
ihuguet: Maybe we should move this at least below `hub release create`. If something fails doing the release, we don't publish the crates. For example if `make release` fails.

jvaclav-rh: I wonder if this could actually be entirely moved out of this file, maybe into a CI action that triggers when a new release is published. It does not need any signing like the tarball, so it should be doable. It might also be easier to rectify when it fails, since you can just re-trigger the action, and it always runs on the relevant commit.

ihuguet: Maybe it's doable, but it would imply to give permissions to the crates in crates.io to a bot account, or putting one of the maintainer's API key in the CI. So, it would relax the security of the crates, with no actual benefit, because for practical purposes, there is no real problem on doing it from the script.

jvaclav-rh: It would be one less thing you have to be careful about when doing an upstream release, and we could skip this entire setup with cargo, but I don't mind it being it either place.

automation/upstream_release.sh
cathay4t: Since this is for maintainer, better fail early than modify user's git config by adding a remote.

For me I use HTTPS token based authentication in git.

automation/upstream_release.sh
cathay4t: Allowing CI to release is possible, but I find it provides more risk than convenience:
1. nmstate tarbal has gpg sign, you want to generate a private key to CI and announce it as nmstate sponsored signer?
2. I like to polish the changelog manually before release, it is a simple task but give me a good chance to review whether we have breakage or missed pending patches.
3. For me `make upstream_release` is a very simple task.

jvaclav-rh: I am not proposing to move the entire release process into CI, since as you mentioned we must sign the tarball (and I would prefer not to give bots a signing key). Just the `cargo publish` part, since it's not dependent on anything that requires human interaction, and can be executed when the release is created.

ihuguet: IMHO that could make sense, in any case, if the entire release process is done by CI. But splitting it part to manual and part to CI is just adding complexity for nothing. Also, it would imply to store a crates.io API token in the pipeline secrets, so it's an added risk with no benefit.

Let's keep it in the release script, it's just easier.

ihuguet (approved):

ihuguet:
@jvaclav-rh please squash and rebase

jvaclav-rh:
@ihuguet Done.

PR #3133, opened 2026-04-23
ethtool: alias tso and tcp-segmentation-offload to tx-tcp-segmentation

karthiksundaravel:
YAML using `tso` or the ethtool -k name `tcp-segmentation-offload` now normalizes to the kernel feature name used in query and verify, matching `gro`/`gso` behavior and avoiding false verification failures.

### DIFF
--- a/rust/src/lib/unit_tests/ethtool.rs
+++ b/rust/src/lib/unit_tests/ethtool.rs
@@ -1,6 +1,9 @@
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::{EthernetInterface, EthtoolFeatureConfig, EthtoolFecConfig};
+use crate::ifaces::ETHTOOL_FEATURE_CLI_ALIAS;
+use crate::{
+    EthernetInterface, EthtoolConfig, EthtoolFeatureConfig, EthtoolFecConfig,
+};
 
 #[test]
 fn test_ethtool_stringlized_attributes() {
@@ -112,6 +115,23 @@ ethtool:
     assert_eq!(ring.rx_mini_max, Some(205));
     assert_eq!(ring.tx, Some(206));
     assert_eq!(ring.tx_max, Some(207));
+}
+
+#[test]
+fn test_ethtool_apply_feature_alias_normalizes_all_cli_aliases() {
+    for (alias, name) in ETHTOOL_FEATURE_CLI_ALIAS {
+        let mut ec: EthtoolConfig = serde_yaml::from_str(&format!(
+            r"---
+            feature:
+              {alias}: false",
+        ))
+        .unwrap();
+
+        ec.apply_feature_alias();
+        let f = ec.feature.as_ref().unwrap();
+        assert_eq!(f.get(name), Some(&false), "alias {alias} -> {name}");
+        assert_eq!(f.get(alias), None, "alias {alias}");
+    }
 }
 
 #[test]
--- a/tests/integration/ethtool_test.py
+++ b/tests/integration/ethtool_test.py
@@ -121,6 +121,65 @@ def test_ethtool_feature_using_ethtool_cli_alias_rx_checksumming(eth1_up):
     assertlib.assert_state_match({Interface.KEY: [desire_iface_state]})
 
 
+def veth_support_tso():
+    """True if eth1 reports tx-tcp-segmentation in nmstate ethtool query."""
+    try:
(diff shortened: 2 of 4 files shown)


rust/src/lib/unit_tests/ethtool.rs
```diff
@@ -114,6 +114,27 @@ ethtool:
...
+        ec.apply_feature_alias();
+        let f = ec.feature.as_ref().unwrap();
+        assert_eq!(
+            f.get("tx-tcp-segmentation"),
+            Some(&expected),
+            "alias {alias}"
+        );
+        assert_eq!(f.get(alias), None, "alias {alias}");
+    }
+}
```
ihuguet, line 135: I think this would be a better test, more readable and testing all the aliases, not only the new one:

You'll need to add proper `pub(crate)` to ETHTOOL_FEATURE_CLI_ALIAS so it can be imported.

```
#[test]
fn test_ethtool_tso_aliases_normalize_to_tx_tcp_segmentation() {
    for (alias, name) in ETHTOOL_FEATURE_CLI_ALIAS {
        let mut ec: EthtoolConfig = serde_yaml::from_str(&format!(
            r"---
            feature:
              {alias}: false",
        ))
        .unwrap();

        ec.apply_feature_alias();
        let f = ec.feature.as_ref().unwrap();
        assert_eq!(f.get(name), Some(&false));
        assert_eq!(f.get(alias), None);
    }
}
```

karthiksundaravel: Done in the following push

cathay4t (changes requested):
Please include integration test for this.

The eth1 (veth) support tso. You may try to turn it on and off with both alias and kernel long name.

tests/integration/ethtool_test.py
```diff
@@ -121,6 +121,65 @@ def test_ethtool_feature_using_ethtool_cli_alias_rx_checksumming(eth1_up):
...
+    """
+    Apply TSO on eth1 (session veth) using CLI-style aliases and the kernel
+    name, then assert against canonical 'tx-tcp-segmentation' in show/verify.
+    """
+    if not _eth1_changeable_tx_tcp_segmentation():
+        pytest.skip(
+            "eth1 does not list tx-tcp-segmentation in changeable ethtool "
+            "features"
+        )
+
```
cathay4t, line 163: You may move this to function line as decorator like:

```python
@pytest.mark.skipif(
    not veth_support_tso(), 
    reason="Current kernel does not support changing ethtool tso on veth")
```

tests/integration/ethtool_test.py
```diff
@@ -121,6 +121,65 @@ def test_ethtool_feature_using_ethtool_cli_alias_rx_checksumming(eth1_up):
...
+def _eth1_changeable_tx_tcp_segmentation():
+    features = (
+        show_only(["eth1"])[Interface.KEY][0]
+        .get(Ethtool.CONFIG_SUBTREE, {})
+        .get(Ethtool.Feature.CONFIG_SUBTREE, {})
+    )
+    return "tx-tcp-segmentation" in features
+
+
+def _eth1_feature_desire_state(feature_key, value):
```
cathay4t: These three simple functions only been used once, it force developers to jump between lines to see the full big picture. Please change them into inline.

cathay4t (changes requested):

tests/integration/ethtool_test.py
karthiksundaravel: It didn't work. Seems the eth1 is not created while @pytest.mark.skipif() is executed.

tests/integration/ethtool_test.py
karthiksundaravel: Will do

tests/integration/ethtool_test.py
cathay4t: In that case, please add a comment indicate why we cannot use `@pytest.mark.skipif` in  the code, so people don't ask the same question again.

karthiksundaravel: added

cathay4t (changes requested):
Besides the pylint failure, looks good to me.

karthiksundaravel:
> Besides the pylint failure, looks good to me.
Thanks @cathay4t.

cathay4t (approved):

PR #3144, opened 2026-05-12
fix: extend checkpoint timeout for SR-IOV operations

vbenes:
When SR-IOV VFs are being created or modified, the NM activation can take longer than the default 60s checkpoint timeout, causing the checkpoint to expire and NM to auto-rollback.

Use VERIFY_RETRY_COUNT_SRIOV_MAX (300s) as checkpoint timeout when SR-IOV VFs are detected in the desired state, not only for the special PF state case.

### DIFF
--- a/rust/src/lib/query_apply/ethernet.rs
+++ b/rust/src/lib/query_apply/ethernet.rs
@@ -191,6 +191,26 @@ impl NetworkState {
     }
 }
 
+async fn query_sriov_vf_count(iface_name: &str) -> u32 {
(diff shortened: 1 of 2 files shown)


cathay4t:
We already has `get_proper_verify_retry_count()`, is this AI PR again?

vbenes:
Those ixgbe cards are very old (back to 2015). Ice and i40e cards do not show such issues. We are solving the kernel bug with such optimizations, and that is what we shouldn't do. It's just a driver bug.

PR #3146, opened 2026-05-20
ci: upgrade Fedora to F44, fix COPR deps, add c10s jobs

vbenes:
- Upgrade Fedora dev container
  - Add c10s integration test jobs: nm_stable-integ_tier2, nm_main-integ_tier1,
    nm_main-integ_tier2
  - Install the missing dependency tcpreplay from EPEL in c10s
  - Temporary: skip a few IPSec tests that are failing in c10s, likely because a bug in the kernel, NM-libreswan or Libreswan (bug already reported)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -143,6 +143,9 @@ jobs:
           - job_type: "c9s-nm_min-integ_slow"
           - job_type: "fed-nm_main-integ_tier1"
           - job_type: "c10s-nm_stable-integ_tier1"
+          - job_type: "c10s-nm_stable-integ_tier2"
+          - job_type: "c10s-nm_main-integ_tier1"
+          - job_type: "c10s-nm_main-integ_tier2"
     steps:
       - uses: actions/checkout@v6
 
--- a/packaging/Dockerfile.fed-nmstate-dev:latest
+++ b/packaging/Dockerfile.fed-nmstate-dev:latest
@@ -1,6 +1,6 @@
 FROM registry.fedoraproject.org/fedora:latest
 
-RUN echo "2025-03-12" > /build_time
+RUN echo "2026-05-25" > /build_time
 
 RUN dnf update -y && \
     dnf -y install --setopt=install_weak_deps=False \
--- a/tests/integration/ipsec_test.py
+++ b/tests/integration/ipsec_test.py
@@ -11,6 +11,7 @@ from libnmstate.schema import InterfaceIPv6
 
 
 from .testlib import cmdlib
+from .testlib.env import is_el10
 from .testlib.env import is_k8s
 from .testlib.env import nm_libreswan_version_int
 from .testlib.env import version_str_to_int
@@ -684,6 +685,11 @@ def test_ipsec_ipv6_host_to_subnet(ipsec_srv_host_to_site):
     )
 
 
+# https://redhat.atlassian.net/browse/RHEL-176474
+@pytest.mark.skipif(
+    is_el10(),
+    reason="Waive: currently failing on EL10 due to known bug",
+)
 @pytest.mark.xfail(
     nm_libreswan_version_int() < version_str_to_int("1.2.22"),
     reason="Need NetworkManager-libreswan 1.2.22+ to support IPv6",
@@ -855,6 +861,11 @@ def test_ipsec_ipv4_libreswan_change_ipsec_iface(ipsec_psk_with_ipsec_iface):
 
 
 # DHCPv4 off with empty IP address means IP disabled for IPSec interface
+# https://redhat.atlassian.net/browse/RHEL-176474
+@pytest.mark.skipif(
+    is_el10(),
+    reason="Waive: currently failing on EL10 due to known bug",
+)
 def test_ipsec_dhcpv4_off_and_empty_ip_addr(
     ipsec_srv_rsa_gw,
 ):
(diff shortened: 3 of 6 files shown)


packaging/Dockerfile.fed-nmstate-dev:latest
```diff
@@ -1,6 +1,6 @@
-FROM registry.fedoraproject.org/fedora:latest
+FROM registry.fedoraproject.org/fedora:44
```
cathay4t: As long as the Fedora 44 released, we should not change this line.

Doing a auto or manual rebuild in quay.io should be enough with `build_time` changed.

.github/workflows/main.yml
```diff
@@ -143,6 +143,9 @@ jobs:
           - job_type: "c9s-nm_min-integ_slow"
           - job_type: "fed-nm_main-integ_tier1"
           - job_type: "c10s-nm_stable-integ_tier1"
+          - job_type: "c10s-nm_stable-integ_tier2"
```
cathay4t, line 146: I think the main reason we didn't enable this is due to lack of test dependencies, have you test it in c10s VM? Otherwise, constant failure only waste dev's review time.

cathay4t (dismissed):

.github/workflows/main.yml
vbenes: Yes, I plan to get the whole test suite green on C10s.

packaging/Dockerfile.fed-nmstate-dev:latest
vbenes: Here, we had dependency issues with NM and glib2. It's better to stick with Fedora's latest release. We have Fedora 44 out for quite some time now (weeks). No idea why 'fedora: latest' is still F43.

.github/workflows/main.yml
vbenes: This is what I am talking about: we have some IPsec issues on C10s that are not visible in the current testing. I think C10s is about to be more relevant than C9s. It's probably not nmstate fault, but it's there and should be solved.

packaging/Dockerfile.fed-nmstate-dev:latest
ihuguet: Fedora:latest was still F43 because the last build in quay.io was 2 months ago, while F44 was released 1 month ago. I don't care much about using `:44` or `:latest`. The first is more explicit, and we need to update the Dockerfile with the "build_time" to trigger a rebuild, anyway. Do whatever you prefer.

automation/run-tests.sh
```diff
@@ -228,12 +228,19 @@ function upgrade_nm_from_copr {
...
+    # disable all repos and force COPR.  On Fedora the COPR NM is always
+    # newer, and we need the base repos enabled for dependency resolution
+    # (e.g. glib2).
+    if [[ $CONTAINER_IMAGE == *"fed-"* ]]; then
+        exec_cmd_with_retry "dnf install --assumeyes NetworkManager \
+            NetworkManager-ovs --enablerepo '${copr_repo_id}'"
+    else
+        exec_cmd_with_retry "dnf install --assumeyes NetworkManager \
+            NetworkManager-ovs --disablerepo '*' --enablerepo '${copr_repo_id}'"
+    fi
```
ihuguet: Is this really necessary? It hasn't been until now. Probably once the Fedora image is rebuilt in Quay, this won't be needed?

vbenes: The point here was missing glib2, which is in fact in repos (so having repos enabled fixes it), any idea who can regenerate Fedora:latest to F44? And yes, not needed once this is working

packaging/Dockerfile.fed-nmstate-dev:latest
vbenes: Aha, can we drive this rebuild? I don't think so, can we? fedora:latest is better for the future, of course.

vbenes: ok, moving back to F44 as I can do the rebuild myself, yes, latest seems to be better.

.github/workflows/main.yml
vbenes: The only missing dependency was tcpreplay, added as a link to koji.

automation/run-tests.sh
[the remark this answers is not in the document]
vbenes: addressed

vbenes: removed

tests/integration/ipsec_test.py
```diff
@@ -684,6 +685,11 @@ def test_ipsec_ipv6_host_to_subnet(ipsec_srv_host_to_site):
     )
 
 
+# https://redhat.atlassian.net/browse/RHEL-176474
+@pytest.mark.skipif(
+    is_el10(),
+    reason="IPSec 4in6/6in4 not supported on EL10",
```
ihuguet: Please be more specific in the "reason". It's supported, but we have a bug. Better say something like "Waive: currently failing on EL10 due to known bug"

.github/workflows/main.yml
vbenes: ok, better to add epel and install from there

tests/integration/ipsec_test.py
vbenes: addressed

ihuguet (approved):
Approved. I will wait to see if Gris has something else to comment, and merge it if not.

PR #3147, opened 2026-05-26
tests: make test_missing_operation compatible with clap 3 and 4

vbenes:
The error message format changed between clap 3 and clap 4:
- clap 3: "Found argument 'no-such-oper' which wasn't expected"
- clap 4: "unrecognized subcommand 'no-such-oper'"

Relax the assertion to check for 'no-such-oper' in the error output, which works with both versions.

### DIFF
--- a/tests/integration/nmstatectl_test.py
+++ b/tests/integration/nmstatectl_test.py
@@ -130,7 +130,7 @@ def test_missing_operation():
     rc, out, err = ret
 
     assert rc != cmdlib.RC_SUCCESS, cmdlib.format_exec_cmd_result(ret)
-    assert "unrecognized subcommand 'no-such-oper'" in err
+    assert "no-such-oper" in err
 
 
 def test_show_command_with_json():


vbenes:
I've filed the above c9s nm_stable-integ here: https://github.com/nmstate/nmstate/issues/3150 AFAIK, NM profile removal and kernel removal operations are async, so there might be a race.

ihuguet (approved):
I didn't digged into the CI failure, but certainly is not related to this MR. Approved.

tests/integration/nmstatectl_test.py
```diff
@@ -130,7 +130,7 @@ def test_missing_operation():
     rc, out, err = ret
 
     assert rc != cmdlib.RC_SUCCESS, cmdlib.format_exec_cmd_result(ret)
-    assert "unrecognized subcommand 'no-such-oper'" in err
+    assert "no-such-oper" in err
```
tyll, line 133: This assertion seems to be pretty useless for me. Asserting the full error message would help to notice changes in the output to ensure that documentation with command output stays accurate or to ensure that message do not accidentally change. In this case, checking that the operation that does not exist is mentioned in the error message is a very weak test.

tyll (commented):
Since it seems that nmstate depends on the hard coded error messages in clap, how would these be translated if necessary? Also, this seems to be testing a feature in clap and not in nmstate if there is no explicit expecation in the assertion about the error message. It seems to make more sense to just remove the assertion if you do not want to test the message properly.

PR #3155, opened 2026-06-05
ci: bump kubevirtci to k8s-1.35 and fix integ_tier1-k8s

mkowalski:
## Why

`pull-nmstate-integ_tier1-k8s` has been broken since the kubevirtci `k8s-1.27` provider image was EOL'd — the pinned node image references the retired Kubic cri-o 1.27 dnf mirror, which now returns HTTP 404 during `dnf upgrade` in `k8s/up.sh`. Every CI run dies at cluster provisioning before any test executes.

## What

Bump kubevirtci to tag `2606030700-d560ba85` (`k8s-1.35`) and fix three classes of test failures exposed by the new node image:

### 1. Physical-NIC name collision (`testlib/veth.py`)

The k8s-1.35 node boots with `net.ifnames=0`, so its secondary NIC is named `eth1` — the same name the test fixtures use for veth pairs. A new `_free_iface_name()` helper detects and parks any pre-existing physical NIC out of the way before creating the test veth.

### 2. udev 'strictly unmanaged' veths (`k8s/up.sh`)

NetworkManager ships `/usr/lib/udev/rules.d/85-nm-unmanaged.rules` which sets `NM_UNMANAGED=1` for every veth not matching `eth[0-9]*`. This makes the devices strictly unmanaged at the udev level — neither `nmcli device set managed yes` nor an NM conf `[device]` section can override it. The fix installs an override in `/etc/udev/rules.d/` that keeps the VirtualBox/VMWare/Parallels/gadget rules but drops the blanket veth rule.

### 3. Node environment alignment with kubernetes-nmstate (`k8s/up.sh`)

Follow what `kubernetes-nmstate/cluster/up.sh` does:
- **Upgrade NetworkManager** (`dnf upgrade -y NetworkManager --allowerasing`) — the base node image ships an older NM whose 'strictly unmanaged' enforcement and OVS activation paths differ from the container CI
- **Remove the dhclient drop-in** (`002-dhclient.conf`) — keeps the NM internal DHCP client consistent with the container CI
- **Enable persistent journal** — so test logs survive node reboots

### Removed the broken `dnf upgrade openvswitch NetworkManager NetworkManager-ovs`

The original `k8s/up.sh` tried to upgrade all three packages in a single `dnf upgrade` command. The bare `openvswitch` package name doesn't resolve in the node repos — this command only ever succeeded as a side effect of also upgrading NM. Replaced with the targeted `dnf upgrade -y NetworkManager --allowerasing` that kubernetes-nmstate uses.

### DIFF
--- a/k8s/kubevirtci.sh
+++ b/k8s/kubevirtci.sh
@@ -1,5 +1,5 @@
-export KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER:-'k8s-1.27'}
-export KUBEVIRTCI_TAG='2310060757-0afa095'
+export KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER:-'k8s-1.35'}
+export KUBEVIRTCI_TAG='2606030700-d560ba85'
 
 KUBEVIRTCI_REPO='https://github.com/kubevirt/kubevirtci.git'
 KUBEVIRTCI_PATH="${PWD}/_kubevirtci"
--- a/k8s/up.sh
+++ b/k8s/up.sh
@@ -15,21 +15,53 @@ if [[ "$KUBEVIRT_PROVIDER" =~ ^(okd|ocp)-.*$$ ]]; then \
 fi
 
 for node in $(./k8s/kubectl.sh get nodes --no-headers | awk '{print $1}'); do
+    # Upgrade NetworkManager to the latest available version.
+    # The base node image ships an older NM whose "strictly unmanaged"
+    # enforcement and OVS activation paths differ from the container CI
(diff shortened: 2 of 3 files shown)


mkowalski:
\gemini full review

qinqon:
Hey @mkowalski as we commented on matrix now the prow job is gone (https://github.com/kubevirt/project-infra/pull/5128) can we repurpose this PR to remove the k8s part of nmstate/nmstate ? since CI is not testing it we don't know if it really works, let's just remove it.

mkowalski:
Yeah we should definitely cleanup this code one day but today is not this day

PR #3158, opened 2026-06-12
Add Ján Václav as maintainer

ihuguet:
Also remove Wen's gpg key.


mkowalski:
Totally +1.

Basic sanity check — is the GPG private key at least password-protected at rest or (preferably?) stored on a key? I guess company will soon stop being fan of keys just lying around our laptops unprotected

jvaclav-rh:
> Basic sanity check — is the GPG private key at least password-protected at rest

Yes, but I am hoping to get a security key in the next few weeks, so I might have to rotate my current key to one stored on it 😄

PR #3159, opened 2026-06-15
loopback: Allow IPv6 disabled when kernel-disabled

pfeifferj:
On a system booted with `ipv6.disable=1`, `nmstatectl show` reports loopback with `ipv6: enabled: false`, but reapplying that output was rejected by the loopback sanitize check.

Move the check to `MergedInterface::post_inter_ifaces_process_loopback()` where the current state is known, and only reject disabling IPv6 when the IPv6 stack is currently enabled or unknown.

Resolves: https://redhat.atlassian.net/browse/NMT-2213

### DIFF
--- a/rust/src/lib/ifaces/loopback.rs
+++ b/rust/src/lib/ifaces/loopback.rs
@@ -18,10 +18,10 @@ use crate::{
 ///  * The [InterfaceState::Absent] can only restore the loopback configure back
 ///    to default.
 ///  * Cannot disable IPv4.
-///  * Cannot disable IPv6 unless the kernel IPv6 stack is already off (e.g.
-///    booted with `ipv6.disable=1`).
-///  * Even when not desired, `127.0.0.1/8` and `::1` (the latter only when
-///    IPv6 is enabled) are always appended to the static IP address list.
+///  * Cannot disable IPv6 unless it is already reported disabled, so `show`
+///    output round-trips (e.g. on a host booted with `ipv6.disable=1`).
+///  * Even when not desired, `127.0.0.1/8` and `::1` (the latter only when IPv6
+///    is enabled) are always appended to the static IP address list.
 ///  * Require NetworkManager 1.41+ unless in kernel only mode.
 ///
 /// Example yaml outpuf of `[crate::NetworkState]` with loopback interface:
@@ -116,9 +116,8 @@ impl LoopbackInterface {
 }
 
 impl MergedInterface {
-    // Allow disabling IPv6 on loopback only when the kernel IPv6 stack is
-    // already off (e.g. `ipv6.disable=1`), so `nmstatectl show` output can be
-    // reapplied on such a host.
+    // Allow disabling IPv6 on loopback only when it is already reported
+    // disabled, so `nmstatectl show` output can be reapplied.
     pub(crate) fn post_inter_ifaces_process_loopback(
         &self,
     ) -> Result<(), NmstateError> {
--- a/rust/src/lib/ifaces/loopback.rs
+++ b/rust/src/lib/ifaces/loopback.rs
@@ -5,8 +5,9 @@ use std::net::{Ipv4Addr, Ipv6Addr};
 use serde::{Deserialize, Serialize};
 
 use crate::{
-    BaseInterface, ErrorKind, InterfaceIpAddr, InterfaceIpv4, InterfaceIpv6,
-    InterfaceState, InterfaceType, NmstateError,
+    BaseInterface, ErrorKind, Interface, InterfaceIpAddr, InterfaceIpv4,
+    InterfaceIpv6, InterfaceState, InterfaceType, MergedInterface,
+    NmstateError,
 };
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
@@ -16,9 +17,11 @@ use crate::{
 ///  * Cannot enable DHCP or autoconf.
 ///  * The [InterfaceState::Absent] can only restore the loopback configure back
 ///    to default.
-///  * Ignore the request of disable IPv4 or IPv6.
-///  * Even not desired, the `127.0.0.1/8` and `::1` are always appended to
-///    static IP address list.
+///  * Cannot disable IPv4.
+///  * Cannot disable IPv6 unless the kernel IPv6 stack is already off (e.g.
+///    booted with `ipv6.disable=1`).
+///  * Even when not desired, `127.0.0.1/8` and `::1` (the latter only when
+///    IPv6 is enabled) are always appended to the static IP address list.
 ///  * Require NetworkManager 1.41+ unless in kernel only mode.
 ///
 /// Example yaml outpuf of `[crate::NetworkState]` with loopback interface:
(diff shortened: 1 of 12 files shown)


rust/src/lib/nm/query_apply/ip.rs
```diff
@@ -110,7 +118,9 @@ pub(crate) fn nm_ip_setting_to_nmstate6(
             parse_dhcp_opts(nm_ip_setting);
         let mut ret = InterfaceIpv6 {
             enabled,
-            enabled_defined: true,
+            // The kernel owns loopback IPv6 (e.g. `ipv6.disable=1`); let
+            // nispor's state win the merge.
+            enabled_defined: iface_type != &InterfaceType::Loopback,
```
rajeshrah22: Should `ipv6.method: auto` in NM cause nmstate to report `ipv6.enabled: true`?
The kernel IPv6 stack can be disabled even when the NM profile says `auto`, so maybe not.

```
# cat /proc/sys/net/ipv6
cat: /proc/sys/net/ipv6: No such file or directory   # kernel IPv6 disabled

# nmcli connection show "Wired connection 1" | grep ipv6.method
ipv6.method: auto

# nmstatectl show enp4s0 | grep -A 2 ipv6:
  ipv6:
    enabled: true   # wrong
    dhcp: true
```

I think NM should only own `enabled_defined` when `ipv6.method` is `disabled` because that's
the only case where NM is actively deciding to not use IPv6. All other methods
(`auto`, `manual`, `link-local`, etc.) require the kernel stack to be present, so
nispor's view of whether the stack is actually up should win.

```rust
enabled_defined: iface_type != &InterfaceType::Loopback
    && nm_ip_method == &NmSettingIpMethod::Disabled
```

What do you think?

pfeifferj: I think it makes more sense to keep this scoped to loopback.

NM's method is only ambiguous there, since our apply path writes auto for lo whether IPv6 is on or kernel-disabled  while every other type gets `method=disabled` if IPv6 is off, so NM stays authoritative everywhere else and deferring to nispor is just the existing Ignore-arm idiom applied where NM can't disambiguate

generalizing to all non-disabled methods would hand enabled to nispor's `!addresses.is_empty()` check and flip admin-down or mid-DAD interfaces to `enabled: false`

it is worth adding a direct global-disable detection rather than a wider carve-out for the enp4s0 case under global `ipv6.disable=1` IMO, but as a follow-up

what are your thoughts on this?

rajeshrah22: Yeah good points, we should re-visit this another time as nispor's enabled = `!addresses.is_empty()` doesn't seem ideal to me.

pfeifferj: I created a followup issue to track this: https://github.com/nmstate/nmstate/issues/3162

rust/src/lib/nm/settings/ip.rs
```diff
@@ -380,3 +391,47 @@ pub(crate) fn fix_ip_dhcp_timeout(nm_conns: &mut [NmConnection]) {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
```
rajeshrah22, line 396: I haven't seen the tests in the same file as the implementation in nmstate before. But I don't see a problem with it being in the same file. Other than maybe for agentic development where you want to implementing agent to not be able view the test cases.

Maybe jan or Gris will have other thoughts.

rajeshrah22 (approved):

rust/src/lib/nm/settings/ip.rs
pfeifferj: since `gen_nm_ip_setting` is crate-internal, an external test file wouldn't be able to call it

this is the standard, idiomatic rust convention https://doc.rust-lang.org/book/ch11-03-test-organization.html#testing-private-functions

jvaclav-rh: I think it's fine, we have similarly tests in `rust/src/lib/gen_conf.rs`

rust/src/lib/nm/query_apply/ip.rs
```diff
@@ -93,6 +95,12 @@ pub(crate) fn nm_ip_setting_to_nmstate6(
             NmSettingIpMethod::LinkLocal
             | NmSettingIpMethod::Manual
             | NmSettingIpMethod::Shared => (true, Some(false), Some(false)),
+            // NM sets loopback to `auto` but never runs autoconf/DHCPv6 on it.
+            NmSettingIpMethod::Auto
+                if iface_type == &InterfaceType::Loopback =>
```
jvaclav-rh: maybe:
```suggestion
                if *iface_type == InterfaceType::Loopback =>
```

rust/src/lib/nm/query_apply/ip.rs
jvaclav-rh: same as above

rust/src/lib/nm/query_apply/ip.rs
jvaclav-rh: Also maybe it would be clearer to do
```rust
NmSettingIpMethod::Auto => {
  if ... {
    ...
  } else {
    ...
  }
}
```

but I don't have a strong preference on that

jvaclav-rh (approved):
lgtm % nits

rust/src/lib/nm/query_apply/ip.rs
pfeifferj: hmm I checked and it looks like the rust book mentions a guarded variant arm immediately followed by an unguarded fall-through arm as the canonical idiom [1].

so if you don't have a preference, I would keep this part the way it is

[1] https://doc.rust-lang.org/book/ch19-03-pattern-syntax.html#adding-conditionals-with-match-guards

rust/src/lib/ifaces/loopback.rs
```diff
@@ -119,3 +114,32 @@ impl LoopbackInterface {
         Ok(())
     }
 }
+
+impl MergedInterface {
+    // Allow disabling IPv6 on loopback only when the kernel IPv6 stack is
+    // already off (e.g. `ipv6.disable=1`), so `nmstatectl show` output can be
+    // reapplied on such a host.
```
ihuguet: The comment says that it will check the kernel stack on/off, but the code actually checks the current state of the interface. That's not very solid, and moreover, having the IPv6 stack disabled is not the only case where `ipv6.enabled=false`. Try, for example, `ip link set down lo; nmstatectl show lo`.

It would be better if it really checks if the stack is disabled, maybe like #3163?

Also, try `sysctl net.ipv6.conf.lo.disable_ipv6=1`. You can disable the IPv6 stack in different ways, not only by the global kernelarg. So the solution from the other PR looks incomplete?

pfeifferj: fixed the comment that implied a stack probe. the gate just mirrors reported state wherein disabling ipv6 on lo is allowed only when `show` already reports it disabled, so `show | apply` round-trips.

the other causes you mention (`ip link set down lo` etc) make show emit `enabled: false` too, so apply has to accept them or the round-trip breaks. it's not trying to assert the stack is off.

stack detection (global + per-iface `disable_ipv6` sysctl) is the separate concern in #3163.

ihuguet: But then this is a change in behaviour. Previously we didn't allow to have IPv6 disabled. Now, we allow it in certain case: it it was already disabled. It is not a backwards incompatible change, because we're allowing new things, but it is a change. So, the question is, why didn't we allow to have ipv6.enabled=false? Without knowing the reason it is difficult to know whether it is OK to allow it now.

Also, the new behaviour is not very nice because, normally, ipv6.enable=false is not allowed. However, if I disable it manually and apply a ipv6.enable=false, then it works. IMHO, if ipv6.enable=false is an acceptable behaviour, we should allow it always. If it's not, we should only allow it when strictly necessary (when the stack is not available). So, again, the main question is why it wasn't allowed and whether it is OK to allow it now.

ihuguet: The Git history gives no clues. Only commit 82a78a3314 is kind of relevant, and it just says that disabling IPv4 and IPv6 is not allowed, but not why. Therefore, I'd like to ask @cathay4t if is there any reason for that. If we don't get a response, I would be OK to allow lifting that restriction, but: both for IPv4 and 6, and allowing it in every case, not only when it was already disabled.

ihuguet: Makes sense?

pfeifferj: yes, makes sense. waiting to hear if Gris has some thoughts on this :)

jvaclav-rh (approved):

PR #3160, opened 2026-06-16
route: catch metric-less route conflicts in kernel mode

pfeifferj:
Kernel-mode conflict validation keyed on the raw route metric, so a desired route that omits the metric (None) never collided with an existing route the kernel reports at its coerced default. IPv6 default routes carry metric 1024, so a metric-less IPv6 route bypassed pre-apply validation and failed only at post-apply verification.

Key the conflict map on the metric and table the kernel actually assigns (unspecified/0 IPv6 -> 1024, IPv4 -> 0; unset table -> main).

Resolves: https://redhat.atlassian.net/browse/NMT-2067

### DIFF
--- a/rust/src/lib/route.rs
+++ b/rust/src/lib/route.rs
@@ -16,6 +16,9 @@ use crate::{
 
 const DEFAULT_TABLE_ID: u32 = 254; // main route table ID
 const LOOPBACK_IFACE_NAME: &str = "lo";
+// Kernel metric for user/static IPv6 routes with unset/0 metric.
+#[cfg(feature = "query_apply")]
+const IP6_RT_PRIO_USER: u32 = 1024;
 
 #[derive(Debug, Clone, PartialEq, Eq, Default, Serialize, Deserialize)]
 #[non_exhaustive]
@@ -558,6 +561,27 @@ impl RouteEntry {
     pub(crate) fn is_ipv6(&self) -> bool {
         self.destination.as_ref().map(|d| is_ipv6_addr(d.as_str()))
             == Some(true)
+    }
+
+    // Metric the kernel will assign, for conflict detection. Unset/0 IPv6
+    // metric is coerced to 1024, IPv4 to 0.
+    #[cfg(feature = "query_apply")]
+    pub(crate) fn effective_metric(&self) -> u32 {
+        match self.metric.and_then(|m| u32::try_from(m).ok()) {
+            Some(m) if m > 0 => m,
+            _ if self.is_ipv6() => IP6_RT_PRIO_USER,
+            _ => 0,
+        }
+    }
+
+    // Table the kernel will assign, for conflict detection. Unset table is
+    // the main table.
+    #[cfg(feature = "query_apply")]
+    pub(crate) fn effective_table_id(&self) -> u32 {
+        match self.table_id {
+            None | Some(Self::USE_DEFAULT_ROUTE_TABLE) => DEFAULT_TABLE_ID,
+            Some(t) => t,
+        }
     }
 
     pub(crate) fn is_unicast(&self) -> bool {
--- a/tests/integration/testlib/servicelib.py
+++ b/tests/integration/testlib/servicelib.py
@@ -6,29 +6,32 @@ import time
 from . import cmdlib
 
 
+def _wait_until(cmd, want_success, timeout=5):
+    while timeout > 0:
+        ret, _, _ = cmdlib.exec_cmd(cmd, check=False)
+        if (ret == 0) == want_success:
+            return
+        time.sleep(1)
+        timeout -= 1
+
+
 @contextmanager
 def disable_service(service):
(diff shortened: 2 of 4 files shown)


rajeshrah22:
Hi, I validated that your fix works; however, I do not understand the reproducer attached to the issue. I am not sure why the validation error is only triggered after changing the ip address of the eth1 interface. Do you happen to figure out why?

pfeifferj:
> Hi, I validated that your fix works; however, I do not understand the reproducer attached to the issue. I am not sure why the validation error is only triggered after changing the ip address of the eth1 interface. Do you happen to figure out why?

@rajeshrah22 the ip swap is a bit of a red herring. it doesn't trigger the validation

for ipv6 the "metric" is really route priority, and the kernel fills in a default when you don't give one

`IP6_RT_PRIO_USER = 1024` [1]. so a metric-less `::/0` you apply reads back from the kernel at metric 1024. ipv4 doesn't do this, unset just means 0

nmstate compared the metric you typed against the metric the kernel stored without accounting for the kernel filling in 1024

the conflict check before apply is supposed to catch "two routes, same dst, same metric, same table" since the kernel won't keep both

does this clear things up?

[1] https://github.com/torvalds/linux/blob/502d801f0ab03e4f32f9a33d203154ce84887921/include/uapi/linux/ipv6_route.h#L61

rajeshrah22:
> > Hi, I validated that your fix works; however, I do not understand the reproducer attached to the issue. I am not sure why the validation error is only triggered after changing the ip address of the eth1 interface. Do you happen to figure out why?
> 
> @rajeshrah22 the ip swap is a bit of a red herring. it doesn't trigger the validation

yeahh, okay.

> 
> for ipv6 the "metric" is really route priority, and the kernel fills in a default when you don't give one
> 
> `IP6_RT_PRIO_USER = 1024` [1]. so a metric-less `::/0` you apply reads back from the kernel at metric 1024. ipv4 doesn't do this, unset just means 0
> 
> nmstate compared the metric you typed against the metric the kernel stored without accounting for the kernel filling in 1024
> 
> the conflict check before apply is supposed to catch "two routes, same dst, same metric, same table" since the kernel won't keep both
> 
> does this clear things up?

Yeah, it makes sense now. Thanks.
> 
> [1] https://github.com/torvalds/linux/blob/502d801f0ab03e4f32f9a33d203154ce84887921/include/uapi/linux/ipv6_route.h#L61

rajeshrah22 (approved):

rust/src/lib/route.rs
```diff
@@ -16,6 +16,8 @@ use crate::{
 
 const DEFAULT_TABLE_ID: u32 = 254; // main route table ID
 const LOOPBACK_IFACE_NAME: &str = "lo";
+#[cfg(feature = "query_apply")]
+const IPV6_DEFAULT_METRIC: u32 = 1024; // Coerced by kernel 0->1024
```
jvaclav-rh: Maybe rename to IP6_RT_PRIO_USER or add a comment mentioning it? The variable name is a bit misleading, since the default metric for e.g. RA routes is 256

pfeifferj: good point. I changed this as suggested

tests/integration/testlib/servicelib.py
```diff
@@ -23,6 +23,7 @@ def disable_service(service):
     try:
         yield
     finally:
+        cmdlib.exec_cmd(("systemctl", "reset-failed", service), check=False)
```
jvaclav-rh: I think this was left here by accident/doesn't belong in this commit?

rust/src/lib/route.rs
```diff
@@ -16,6 +16,9 @@ use crate::{
 
 const DEFAULT_TABLE_ID: u32 = 254; // main route table ID
 const LOOPBACK_IFACE_NAME: &str = "lo";
+#[cfg(feature = "query_apply")]
+// Kernel metric for user/static IPv6 routes with unset/0 metric.
+const IP6_RT_PRIO_USER: u32 = 1024;
```
jvaclav-rh: Maybe do:

```suggestion
// Kernel metric for user/static IPv6 routes with unset/0 metric.
#[cfg(feature = "query_apply")]
const IP6_RT_PRIO_USER: u32 = 1024;
```
It seems easier to read to me.

pfeifferj: good point, this also also consistent with the others. thank you

jvaclav-rh (approved):

tests/integration/testlib/servicelib.py
pfeifferj: ah this was to make the teardown in the 2 failing ci jobs work properly. I think we need to revert this removal

jvaclav-rh: Could you maybe put it in a separate commit with some explanation in the description why it was introduced?

PR #3163, opened 2026-06-22
ipv6: report disabled when kernel has no IPv6 stack

### LINKED ISSUE #3162: show reports `ipv6.enabled: true` under `ipv6.disable=1`
Followup to this discussion: https://github.com/nmstate/nmstate/pull/3159#discussion_r3431086689

On a host booted with `ipv6.disable=1`, `nmstatectl show` reports `ipv6: enabled: true` for any interface whose NM profile uses a non-disabled
method (e.g. `method=auto`). The kernel has no IPv6 stack, so that state is unrealizable.

I would propose that we detect the global signal directly when `/proc/sys/net/ipv6` is absent, force `enabled: false` on all interfaces. Fires only when no interface can have IPv6, so the normal path is untouched. No such check exists in `rust/src/lib` yet.

pfeifferj:
On a host booted with `ipv6.disable=1` the kernel has no IPv6 stack, but `show` still reported `ipv6: enabled: true` for interfaces whose NM profile uses `method=auto`: the NM backend wins the merge over nispor, which correctly reports the interface as disabled. That state is unrealizable.

Resolves: https://github.com/nmstate/nmstate/issues/3162

### DIFF
--- a/rust/src/lib/ifaces/inter_ifaces.rs
+++ b/rust/src/lib/ifaces/inter_ifaces.rs
@@ -186,11 +186,20 @@ impl Interfaces {
     // Reset every present IPv6 section to a disabled state on interfaces where
     // `is_disabled` returns true.
     #[cfg(feature = "query_apply")]
-    pub(crate) fn disable_ipv6_where(&mut self, is_disabled: fn(&str) -> bool) {
+    pub(crate) fn disable_ipv6_where(
+        &mut self,
+        is_disabled: impl Fn(&str) -> bool,
+    ) {
         for iface in self.iter_mut() {
             let base = iface.base_iface_mut();
-            if is_disabled(base.name.as_str()) {
-                reset_disabled_ipv6(base);
+            if is_disabled(base.name.as_str())
+                && let Some(ipv6_conf) = base.ipv6.as_mut()
+            {
+                *ipv6_conf = crate::InterfaceIpv6 {
+                    enabled: false,
+                    enabled_defined: true,
+                    ..Default::default()
+                };
             }
         }
     }
@@ -727,17 +736,6 @@ impl Interfaces {
                     Some(iface.name().to_string());
             }
         }
-    }
-}
-
-#[cfg(feature = "query_apply")]
-fn reset_disabled_ipv6(base: &mut crate::BaseInterface) {
-    if let Some(ipv6_conf) = base.ipv6.as_mut() {
-        *ipv6_conf = crate::InterfaceIpv6 {
-            enabled: false,
-            enabled_defined: true,
-            ..Default::default()
-        };
     }
 }
 
--- a/rust/src/lib/ifaces/inter_ifaces.rs
+++ b/rust/src/lib/ifaces/inter_ifaces.rs
@@ -183,6 +183,18 @@ impl Interfaces {
             .chain(self.kernel_ifaces.values_mut())
     }
 
+    // Reset every present IPv6 section to a disabled state on interfaces where
+    // `is_disabled` returns true.
+    #[cfg(feature = "query_apply")]
+    pub(crate) fn disable_ipv6_where(&mut self, is_disabled: fn(&str) -> bool) {
+        for iface in self.iter_mut() {
+            let base = iface.base_iface_mut();
+            if is_disabled(base.name.as_str()) {
+                reset_disabled_ipv6(base);
+            }
+        }
+    }
+
     // Not allowing changing veth peer away from ignored peer unless previous
     // peer changed from ignore to managed
     // Not allowing creating veth without peer config
@@ -715,6 +727,17 @@ impl Interfaces {
                     Some(iface.name().to_string());
             }
         }
+    }
+}
+
+#[cfg(feature = "query_apply")]
+fn reset_disabled_ipv6(base: &mut crate::BaseInterface) {
+    if let Some(ipv6_conf) = base.ipv6.as_mut() {
+        *ipv6_conf = crate::InterfaceIpv6 {
+            enabled: false,
+            enabled_defined: true,
+            ..Default::default()
+        };
     }
 }
 
--- a/rust/src/lib/query_apply/net_state.rs
+++ b/rust/src/lib/query_apply/net_state.rs
@@ -138,6 +138,13 @@ impl NetworkState {
         // Mark routes next hop to ignored interface as ignored
         self.routes.apply_ignored_ifaces(&self.interfaces);
 
+        // A kernel booted with `ipv6.disable=1` has no IPv6 stack, but NM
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -47,3 +47,5 @@ nmstate-*.tar.gz.asc
 
 rust/src/clib/nmstate.pc
 rust/src/python/build
+issue-triage.md
+issue-triage-comments.sh
```
rajeshrah22: you may have accidentally committed these.

pfeifferj: thank you! yes, I was running a local experiment

rajeshrah22:
My thoughts:

What if we decided that we also want to detect runtime sysctl `disable_ipv6`, which could also be per interface?

Maybe this code should belong to nispor? Seems simple enough to add in nmstate though, and would avoid the complexity of updating both dependencies.

I did see somewhere else that Gris commented that sysfs related code should move to nispor folder or directly in nispor itself: https://github.com/nmstate/nmstate/pull/3153#pullrequestreview-4414595727

pfeifferj:
> My thoughts:
> 
> What if we decided that we also want to detect runtime sysctl `disable_ipv6`, which could also be per interface?
> 
> Maybe this code should belong to nispor? Seems simple enough to add in nmstate though, and would avoid the complexity of updating both dependencies.
> 
> I did see somewhere else that Gris commented that sysfs related code should move to nispor folder or directly in nispor itself: [#3153 (review)](https://github.com/nmstate/nmstate/pull/3153#pullrequestreview-4414595727)

you're right sysfs reads ideally belong in nispor. but moving it doesn't make sense in this case imo because the override has to live in nmstate no matter where the signal comes from 

nispor already reports this iface [correctly](https://github.com/nmstate/nmstate/blob/9563770892f570aac553494f72e34530afd9bdbc/rust/src/lib/nispor/ip.rs#l86). the `enabled: true` comes from the [nm backend mapping](https://github.com/nmstate/nmstate/blob/9563770892f570aac553494f72e34530afd9bdbc/rust/src/lib/nm/query_apply/ip.rs#l96). so even if nispor had an "ipv6 stack present" flag, nmstate still needs a post-merge comparison to act on it 

and we already do this [in mptcp](https://github.com/nmstate/nmstate/blob/9563770892f570aac553494f72e34530afd9bdbc/rust/src/lib/nm/settings/mptcp.rs#l15) where it reads `/proc/sys/net/mptcp/enabled` directly 

`/proc/sys/net/ipv6` is a stable path, absent only when the kernel has no inet6 stack (booted `ipv6.disable=1` or `config_ipv6=n`). per [kernel docs](https://docs.kernel.org/networking/ipv6.html) the `disable` param loads the module but skips sysctl registration, so the whole tree just isn't there

sidenote on the runtime per-iface case: [kernel ip-sysctl docs](https://docs.kernel.org/networking/ip-sysctl.html) say reading `conf/all/disable_ipv6` "does not have any particular meaning" and can return 1 even with addresses configured. so the absent-tree check is the only reliable "no stack" signal, and runtime per-iface disable is a separate, messier problem

wdyt?

rajeshrah22:
> wdyt?

Yeah makes sense.

I was also wondering if this would break any `nmstatectl show | nmstatectl apply -` or `nmstatectl edit` use cases. I don't think so.

I wonder how the apply path of `enabled: false` looks like...

pfeifferj:
>I was also wondering if this would break any nmstatectl show | nmstatectl apply - or nmstatectl edit use cases. I don't think so.

ohhh good catch. I tested this in my vm and it did indeed break

```
❯ ssh nm-vm 'nmstatectl show | nmstatectl apply -'
[2026-06-26T06:26:24Z INFO  nmstatectl] Nmstate version: 2.2.61
[2026-06-26T06:26:24Z INFO  nmstatectl] Nmstate version: 2.2.61
NmstateError: InvalidArgument: Loopback interface cannot be have IPv6 disabled
```

I'll add a skip condition 

---

edit: actually never mind, #3159 addresses this already. I tested the same scenario again with both patches applied and it behaves as expected

rust/src/lib/query_apply/net_state.rs
```diff
@@ -461,6 +468,13 @@ impl NetworkState {
     }
 }
 
+// `/proc/sys/net/ipv6` is absent only when the kernel has no IPv6 stack at all.
+// The runtime `disable_ipv6` sysctl keeps the directory present and is
+// overridable per interface, so it is deliberately not forced off here.
+fn ipv6_disabled_in_kernel() -> bool {
+    !std::path::Path::new("/proc/sys/net/ipv6").exists()
```
ihuguet: IPv6 can be disabled in kernel per interface too, with `sysctl net.ipv6.conf.lo.disable_ipv6=1`

pfeifferj: heyy, I made changes that also consider the case you pointed out. could you take another look whenever you have time and let me know if you think this PR is good to go now?

rust/src/lib/ifaces/inter_ifaces.rs
```diff
@@ -183,6 +183,18 @@ impl Interfaces {
...
+    // `is_disabled` returns true.
+    #[cfg(feature = "query_apply")]
+    pub(crate) fn disable_ipv6_where(&mut self, is_disabled: fn(&str) -> bool) {
+        for iface in self.iter_mut() {
+            let base = iface.base_iface_mut();
+            if is_disabled(base.name.as_str()) {
+                reset_disabled_ipv6(base);
+            }
+        }
+    }
```
ihuguet, line 205: I think the solution is good, just maybe a bit convoluted code? I think something like this would be easier to read (untested):
```
fn handle_ipv6_disabled_in_kernel(&mut self) {
    for ifaces in &mut self {
        let base = iface.base_iface_mut();
        if ipv6_disabled_in_kernel() || ipv6_disabled_in_kernel_for_iface(&base.name) {
            if let Some(ipv6_conf) = base.ipv6.as_mut() {
               *ipv6_conf = crate::InterfaceIpv6 {
                enabled: false,
                enabled_defined: true,
                ..Default::default()
            };
        }
    }
}
```

Then, reset_disabled_ipv6 is not needed, and you just call `handle_ipv6_disabled_in_kernel(self)` from net_state.rs.

pfeifferj: thanku, I simplified this a bit but kept the predicate split. `query_apply/net_state.rs` owns the `/proc/sys/net/ipv6` checks, while `inter_ifaces.rs` owns the pure "reset present ipv6 sections where predicate holds" loop so the unit tests can use the reset logic with `|_| true` and `|name| name == "eth1"` without touching `/proc/sys`.

would this version work for you?

jvaclav-rh (approved):

PR #3167, opened 2026-06-29
ovs: Support internal interfaces in down state

qinqon:
## What

Support OVS internal interfaces in `state: down`, behaving like `ovs-vsctl add-br`: the internal/local port stays **attached** to the bridge but its kernel link is administratively **DOWN** (no kernel-datapath traffic). Needed for OVS-DPDK.

Jira: NMT-2268

## Before

A down OVS internal interface was treated as absent → NM deactivated it → detached from bridge → `VerificationError` ("still found" / bridge port `null`).

```yaml
interfaces:
- name: br0
  type: ovs-interface
  state: down
- name: br0
  type: ovs-bridge
  state: up
  bridge:
    port: [{name: br0}]
```

## Why nispor (NetworkManager limitation)

Verified in NM source: NM has **no** way to keep a connection active while the link is admin-down.
- Activation always brings the link up — `nm-device.c:11094` `nm_device_bring_up_full()` → `:16061` `nm_platform_link_change_flags(IFF_UP, TRUE)`; OVS internal adds explicit `nm_device_bring_up()` at `nm-device-ovs-interface.c:198,288,449`.
- No "active + admin-down" connection property exists (`SET_ADMIN_STATE` ties admin-down to **unmanaging** the device, `nm-device.c:15073`).
- NM's only "down" is deactivation, which detaches the OVS port from ovsdb — `nm-device-ovs-port.c:247` → `nm-ovsdb.c:1378`.
- ovsdb membership is driven by attach/detach (activation lifecycle), **not** IFF_UP — so clearing IFF_UP out-of-band leaves the port attached and the connection active.

So we set the kernel link state directly via nispor after the NM apply — the same pattern nmstate already uses for `apply_ifaces_alt_names` and `set_running_hostname` right after `nm_apply` (`query_apply/net_state.rs`).

## Change

- `nm/connection.rs`: do not deactivate a down OVS internal interface (deactivation destroys the local port).
- `query_apply/inter_ifaces.rs`: down OVS internal not treated as absent in verification.
- `nispor/ovs.rs`: new `enforce_ovs_internal_link_state()`, invoked from `query_apply/net_state.rs` next to `apply_ifaces_alt_names` — sets the OVS internal link DOWN when desired, UP otherwise.

## Test

Manual (`nmstatectl`) on a NetworkManager + OVS host: down/up/down cycle, LOCAL port + extra internal port, idempotent, parity with `ovs-vsctl add-br`. Added `test_create_ovs_bridge_with_internal_port_down` (attached + DOWN + up-toggle). fmt + clippy clean.

## Open / RFC

- The kernel admin-down is a 5x retry (OVS link may appear late after NM activation) — could be tightened.
- `nmstatectl show` reports the profile up while the link is down.
- Modelling: `state:down` vs an explicit flag — feedback welcome.
- Proper long-term fix likely needs a NetworkManager RFE (admin-link-down property, or decouple OVS bring-up).

### DIFF
--- a/rust/src/lib/nispor/ovs.rs
+++ b/rust/src/lib/nispor/ovs.rs
@@ -1,0 +1,69 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{ErrorKind, InterfaceType, MergedInterfaces, NmstateError};
+
+/// NetworkManager always brings an OVS internal interface link up on
+/// activation and offers no way to keep it administratively down while the
+/// connection stays active (the only "down" it has detaches the port from the
+/// bridge). To support `state: down` on OVS internal interfaces — keeping them
+/// attached to the bridge with the link DOWN, like `ovs-vsctl add-br` — we set
+/// the kernel link state directly via nispor after the NetworkManager apply.
+/// See NMT-2268.
+pub(crate) async fn enforce_ovs_internal_link_state(
+    merged_ifaces: &MergedInterfaces,
+) -> Result<(), NmstateError> {
+    let np_ifaces: Vec<nispor::IfaceConf> = merged_ifaces
+        .kernel_ifaces
+        .values()
+        .filter_map(|i| {
+            let iface = i.for_apply.as_ref()?;
+            if !i.is_desired()
+                || iface.iface_type() != InterfaceType::OvsInterface
+            {
+                return None;
+            }
+            let state = if iface.is_down() {
+                nispor::IfaceState::Down
+            } else if iface.is_up() {
+                nispor::IfaceState::Up
+            } else {
+                return None;
+            };
+            let mut np_iface = nispor::IfaceConf::default();
+            np_iface.name = iface.name().to_string();
+            np_iface.iface_type = Some(nispor::IfaceType::OpenvSwitch);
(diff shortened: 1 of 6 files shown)


rust/src/lib/nispor/ovs.rs
```diff
@@ -0,0 +1,69 @@
...
+
+use crate::{ErrorKind, InterfaceType, MergedInterfaces, NmstateError};
+
+/// NetworkManager always brings an OVS internal interface link up on
+/// activation and offers no way to keep it administratively down while the
+/// connection stays active (the only "down" it has detaches the port from the
+/// bridge). To support `state: down` on OVS internal interfaces — keeping them
+/// attached to the bridge with the link DOWN, like `ovs-vsctl add-br` — we set
+/// the kernel link state directly via nispor after the NetworkManager apply.
+/// See NMT-2268.
```
ihuguet, line 11: From my side, I refuse such kind of workaround. It will only be a cause of problems when NetworkManager will get confused, or will put up the interface again, etc. Please request to add this feature to NM first.

qinqon: I will is how feasible is that.

PR #3170, opened 2026-07-06
sriov: only verify VF netdevs referenced in desired state

rjarry:
After creating SR-IOV VFs and rebinding some of them to vfio-pci (or moving them to another network namespace), verify_sriov still checks that every VF has a valid kernel network interface. Since the rebound VFs no longer have one, this triggers up to 300 retries (one per second) before the error is silently ignored. This wastes a significant amount of time and floods logs with spurious warnings on every subsequent apply.

VF properties such as MAC address, VLAN, trust and spoof check are all configured through the PF netlink/sysfs interface. The NM connection profile for VFs is also part of the PF. The VF netdev is not involved in any of this. The only case where the VF netdev must exist is when it is explicitly referenced elsewhere in the desired state, for example as a port in a bridge, bond or OVS bridge.

Restrict verify_sriov to only wait for VF interfaces that actually appear as desired interfaces in the applied configuration. This covers both literal names (e.g. "ens1f0v2") and resolved sriov: references (e.g. "sriov:ens1f0np0:2") since name resolution runs before verification. VFs that are only configured through the PF and not referenced by name are no longer checked, which eliminates the spurious retries entirely.

### DIFF
--- a/rust/src/lib/query_apply/inter_ifaces.rs
+++ b/rust/src/lib/query_apply/inter_ifaces.rs
@@ -1,5 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
+use std::collections::HashSet;
+
 use crate::{
     ErrorKind, Interface, InterfaceType, Interfaces, MergedInterfaces,
     NmstateError,
@@ -217,6 +219,14 @@ impl MergedInterfaces {
             iface.sanitize_desired_for_verify();
         }
 
+        let desired_iface_names: HashSet<String> = merged
+            .iter()
+            .filter(|i| i.is_desired())
+            .filter_map(|i| i.for_verify.as_ref())
+            .filter(|i| i.is_up())
+            .map(|i| i.name().to_string())
+            .collect();
+
         for des_iface in merged.iter_mut().filter(|i| i.is_desired()) {
             let iface = if let Some(i) = des_iface.for_verify.as_mut() {
                 i
@@ -240,7 +250,8 @@ impl MergedInterfaces {
                     if let Interface::Ethernet(eth_iface) = iface
                         && eth_iface.sriov_is_enabled()
                     {
-                        eth_iface.verify_sriov(&current)?;
+                        eth_iface
+                            .verify_sriov(&current, &desired_iface_names)?;
                     }
                 }
             } else if iface.is_up() {
--- a/rust/src/lib/query_apply/sriov.rs
+++ b/rust/src/lib/query_apply/sriov.rs
@@ -2,6 +2,8 @@
 
 use std::collections::hash_map::Entry;
 
+use std::collections::HashSet;
+
 use crate::{
     BaseInterface, ErrorKind, EthernetConfig, EthernetInterface, Interface,
     InterfaceType, Interfaces, NmstateError, SrIovConfig, SrIovVfConfig,
@@ -47,6 +49,7 @@ impl SrIovConfig {
         &self,
         pf_name: &str,
         cur_ifaces: &Interfaces,
+        desired_iface_names: &HashSet<String>,
     ) -> Result<(), NmstateError> {
         let cur_pf_iface =
             match cur_ifaces.get_iface(pf_name, InterfaceType::Ethernet) {
@@ -86,6 +89,9 @@ impl SrIovConfig {
             return Ok(());
         };
         for vf in vfs {
+            if !desired_iface_names.contains(&vf.iface_name) {
+                continue;
+            }
             if vf.iface_name.is_empty() {
                 return Err(NmstateError::new(
                     ErrorKind::SrIovVfNotFound,
--- a/rust/src/lib/unit_tests/sriov.rs
+++ b/rust/src/lib/unit_tests/sriov.rs
@@ -59,6 +59,55 @@ fn test_sriov_vf_mac_mix_case() {
 }
 
 #[test]
+fn test_verify_sriov_skip_unreferenced_vfs() {
+    let mut pre_apply_cur_ifaces = Interfaces::new();
+    pre_apply_cur_ifaces.push(new_eth_iface("eth1"));
+
+    let mut cur_ifaces = Interfaces::new();
+    let mut cur_iface = new_eth_iface("eth1");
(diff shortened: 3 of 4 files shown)


rust/src/lib/query_apply/inter_ifaces.rs
```diff
@@ -217,6 +219,14 @@ impl MergedInterfaces {
             iface.sanitize_desired_for_verify();
         }
 
+        let desired_iface_names: HashSet<String> = merged
+            .iter()
+            .filter(|i| i.is_desired())
+            .filter_map(|i| i.for_verify.as_ref())
+            .filter(|i| i.is_up())
+            .map(|i| i.name().to_string())
+            .collect();
```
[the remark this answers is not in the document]
CrumblyBread: Copilot is right. Build the set from interfaces that are changed (desired or pulled in by a controller), and use `for_apply` when `for_verify` is absent. Add a unit test mirroring `test_resolve_sriov_port_name_bond` that runs through `MergedInterfaces::verify()` and asserts the referenced VF is still checked.

rust/src/lib/query_apply/sriov.rs
```diff
@@ -2,6 +2,8 @@
 
 use std::collections::hash_map::Entry;
 
+use std::collections::HashSet;
```
CrumblyBread, line 5: Should be:
```
use std::collections::{HashSet, hash_map::Entry};
```

rust/src/lib/query_apply/inter_ifaces.rs
```diff
@@ -217,6 +219,14 @@ impl MergedInterfaces {
             iface.sanitize_desired_for_verify();
         }
 
+        let desired_iface_names: HashSet<String> = merged
```
CrumblyBread, line 222: In enable-and-use path in net_state.rs the first apply uses `get_sriov_pf_conf_state()` which strips VF interfaces. With this change during that phase contains only VF names, so all VF netdev checks will be skipped.  I think this warrants a comment or a test so it's obvious this is intended behavior

rust/src/lib/unit_tests/sriov.rs
```diff
@@ -58,6 +58,55 @@ fn test_sriov_vf_mac_mix_case() {
     merged_ifaces.verify(&cur_ifaces).unwrap();
 }
 
+#[test]
```
CrumblyBread, line 61: Might be a good idea to cover cases like a referenced VF but `iface_name = ""`, referenced VF with name set but netdev missing or 2 VFs but only one is referenced.

rust/src/lib/query_apply/inter_ifaces.rs
rajeshrah22: Ur right that `get_sriov_pf_conf_state` strips VFs so in the first phase where the `pf_state` is applied, there won't be any VF references like `sriov:eth1:1`.

Because VFs can take some time to come up, we actually need to wait for those VFs to have a kernel interface in the first apply if we want them to configured in the next apply with the references resolved. So, stripping them out of the `pf_state` will lead to them not being the `desired_iface_names` set. We need to find a way to include those in the `desired_iface_names` set.

rajeshrah22, line 222: Probably should create set that has desired `(pf, vf_id)` pairs so that in the first apply, we can still wait for VFs that are not yet up in the first phase of enable and use.

rajeshrah22 (changes requested):
I think that the code needs to be changed slightly so that the VFs are still waited on the in the first phase of `apply_with_nm_backend` where the configuration for physical functions are applied. It will take some time for VFs to come up after they are requested. I do not have much experience with SR-IOV so you may know better.

I think it would be nice if you could try this on SR-IOV hardware yourself and send us logs as it is difficult to know for sure how this change will work on real hardware.

rjarry:
Thanks all for the thorough review. I reworked the change based on the feedback.

**What changed since the first version**

- `verify_sriov` now tracks referenced VFs by `(PF name, VF id)` pairs collected from the raw desired state, instead of matching on resolved interface names. This keeps the empty-name `SrIovVfNotFound` branch reachable — a referenced VF whose netdev has not appeared yet is still waited for, which the previous name-only check had turned into dead code (@gemini-code-assist, @Copilot).
- That `(pf, vf_id)` set is threaded through both apply phases. In the enable-and-use path, the first PF-only phase now waits for the referenced VFs to come up before their names are resolved in the second phase (@rajeshrah22).
- The name set used during verification now also includes interfaces pulled in as controller ports (changed but not desired, falling back to `for_apply`), not just explicitly desired ones (@CrumblyBread, @Copilot).
- Fixed the import grouping, and added unit tests: referenced VF with empty name, referenced VF present, two VFs with only one referenced, referenced VF as a bond port present, and referenced bond port with a missing netdev.

**Hardware validation** — Intel E810 (`ice`), CentOS Stream 10, kernel 6.12, comparing upstream base vs this branch. External actors rebind one VF to vfio-pci and move another to a netns.

| Scenario | base | this branch |
|---|---|---|
| Configure PF only; missing VFs unreferenced | 31.3s, 29 spurious retries | 0.16s, 0 |
| Referenced present VF as bridge port, others missing | 30.9s, 29 spurious retries | 1.4s, 1 (legit bridge-port settle) |
| Configure a present VF while others missing | 31.0s, 29 spurious retries | 0.20s, 0 |
| Enable-and-use single txn from 0 VFs | 3.49s, waits VF2 **and** VF5 | 3.11s, waits **only** VF5 |

The enable-and-use log from this branch shows phase 1 waiting only for the referenced VF before resolution (base also waits for the unreferenced VF2):

```
Retrying on: SrIovVfNotFound: Failed to find VF 5 interface name of PF eno12399np0
SR-IOV VF sriov:eno12399np0:5 resolved to interface name eno12399v5
```

I only had the E810 available, so this is not tested on mlx5 (ConnectX-6 Lx).

Unrelated aside: I noticed a pre-existing issue where a VF entry combining `mac` + `vlan-id` + `qos` + `trust` + `spoof-check` produces an empty diff and applies nothing (mac-only works fine). I'll chase that down separately.

PR #3175, opened 2026-07-16
tests: handle RPM tilde pre-release versions in NM version parsing

rajeshrah22:
## Summary
- Fix `nm_minor_version()` and `version_str_to_int()` to handle RPM pre-release version strings containing `~` (e.g. `1.58~rc1`)
- Split on `~` before parsing the numeric version segments to strip suffixes like `~rc1`, `~beta2`, etc.

## Problem
NetworkManager `1.58~rc1` causes 9 test collection errors because `int("58~rc1")` raises `ValueError`.

### DIFF
--- a/tests/integration/testlib/env.py
+++ b/tests/integration/testlib/env.py
@@ -25,7 +25,7 @@ def nm_minor_version():
         "rpm -q NetworkManager --qf %{VERSION}".split(),
         check=True,
     )[1]
-    return int(version_str.split(".")[1])
+    return int(version_str.split("~")[0].split(".")[1])
 
 
 def is_k8s():
@@ -46,7 +46,15 @@ def version_int(major, minor, micro=0):
 
 
 def version_str_to_int(version_str):
-    versions = version_str.split(".")
+    parts = version_str.split("~")
+    versions = parts[0].split(".")
+
+    # if it is a rc version, then use 90 based versioning to ensure
+    # that a release candidate is less than a full release.
+    # Example: "1.58~rc1" would become 1.57.90 but "1.58.0" would stay 1.58.0.
+    if len(parts) == 2:
+        rc_num = int(parts[1].lstrip("rc"))
+        return version_int(int(versions[0]), int(versions[1]) - 1, 89 + rc_num)
     return version_int(int(versions[0]), int(versions[1]), int(versions[2]))
 
 


tyll:
IMHO the commit message should explain what "handle the tilde" actually means by providing examples and the expected behavior in English. Without this context I find it hard to make a judgement whether the change is correct or not.

ihuguet:
You are converting 1.58~rc1, ~rc2 or any ~rcX to 1.58.0. That way, you cannot compare between RC releases, or an RC release with the final 1.58.0 release.

I suggest to convert it to the old 90-based version schema. I.e 1.58~rc1 -> 1.57.90, ~rc2 -> 1.57.91, etc.

rajeshrah22:
@tyll I have added more context in the commit message.

@ihuguet thanks for your feedback, I have revised the function.

tests/integration/testlib/env.py
```diff
@@ -46,7 +46,12 @@ def version_int(major, minor, micro=0):
 
 
 def version_str_to_int(version_str):
-    versions = version_str.split(".")
+    parts = version_str.split("~")
+    versions = parts[0].split(".")
+    if len(parts) == 2:
+        rc_num = int(parts[1].lstrip("rc"))
+        return version_int(int(versions[0]), int(versions[1]) - 1, 89 + rc_num)
```
tyll, line 57: This new logic is now not explained in the commit message or the function.
FYI: AFAIU, using a single int to compare the versions is not necessary. Python can compare tupels with ints just fine and it would be easier to read IMHO. For example a tuple like

(major_version, minor_version, patch_version/stable_indicator, release_candidate) will compare just fine.

patch/stable_indicator can then be -1 for pre-releases and the release_candidate can be the release number, for stable releases, the patch_version can then count up. With the incorrect minor version, tests will also have to use this versioning to be enabled/disabled.

Also, I believe it would be better not to check based on the version but based on the expected features to be available. The test runner should define what features NM/Nmstate needs to have and test them accordingly. This avoids accidentally testing the wrong version/build. This will require a re-design of the tests, though.

rajeshrah22: > This new logic is now not explained in the commit message or the function.

I have addressed this in the new commit message. I think a comment would also help.

> (major_version, minor_version, patch_version/stable_indicator, release_candidate) will compare just fine.
> 
> patch/stable_indicator can then be -1 for pre-releases and the release_candidate can be the release number, for stable releases, the patch_version can then count up. With the incorrect minor version, tests will also have to use this versioning to be enabled/disabled.

IMHO this can be done another time to keep things simple for this PR.

> Also, I believe it would be better not to check based on the version but based on the expected features to be available. The test runner should define what features NM/Nmstate needs to have and test them accordingly. This avoids accidentally testing the wrong version/build. This will require a re-design of the tests, though.

I can create an issue for this so we can remember to come back to this at some point.

jvaclav-rh (approved):

tests/integration/testlib/env.py
```diff
@@ -46,7 +46,16 @@ def version_int(major, minor, micro=0):
...
+    parts = version_str.split("~")
+    versions = parts[0].split(".")
+
+    # if it is a rc version, then use 90 based versioning to ensure
+    # that a release candidate is less than a full release.
+    # Example: "1.58~rc1" would become 1.57.90 but "1.58.0" would stay 1.58.0.
+    if len(parts) == 2:
+        rc_num = int(parts[1].lstrip("rc"))
+        return version_int(int(versions[0]), int(versions[1]) - 1, 89 + rc_num)
+    versions = [v.split("~")[0] for v in versions]
```
jvaclav-rh: Is this necessary? I don't think strings in `versions` can contain `~` at this point, since they are derived from parts[0] which comes from splitting version_str by `~`

rajeshrah22: Done

tyll: Your "Is this necessary?" question made me wonder why there is actually a need to distinguish release candidates from final releases. The functions are used to choose which tests to run for which versions and needing different tests for the release candidate vs the final release seems to be an indication that something is wrong. Tests should be the same for release candidates and final versions, otherwise the tests or the release needs to be adjusted.

With this patch, tests for features that will be released in 1.60 need to declared as 59 or newer. I believe this is unnecessary complexity.

PR #3180, opened 2026-07-23
test: fix 802.1x tests for PrivateTmp and hostapd timing

vbenes:
Copy cert files to ~/.cert/nmstate/ so NetworkManager can access them when PrivateTmp=true is set in its systemd unit. Also add a short sleep after starting hostapd to avoid race condition during EAP authentication.


jvaclav-rh:
@coderabbitai review

vbenes:
review

vbenes:
@coderabbitai review

jvaclav-rh (approved):

PR #3181, opened 2026-07-23
automation: allow base repos when installing NM from COPR

jvaclav-rh:
The COPR NM builds now require glib2 >= 2.88.2, which is not available in the COPR itself. Replace --disablerepo/--enablerepo with --setopt priority=1 for the COPR repo, so dnf resolves dependencies from the base distro repos while still preferring COPR for NM packages.


jvaclav-rh:
Will be merged in #3180

PR #3182, opened 2026-07-23
dns: auto-activate down ifaces for IPv6 link-local DNS and route-rule iif

CrumblyBread:
Restore auto-activation of managed but disconnected interfaces when they are only referenced by IPv6 link-local DNS (fe80::…%iface) or by route-rule iif, so apply no longer requires an explicit interfaces: [{ name, state: up }] entry.

Regression from 282b5b9e (is_iface_valid_for_dns() requiring state: up): link-local DNS validation failed with a misleading “IPv6 disabled” error on down/disconnected ifaces that still had IPv6. Route-rule iif similarly left merged down, so activation/verification could fail.

This change adds MergedInterface::mark_as_up_for_apply() and uses it for those reference paths. Unmanaged/ignored ifaces still need an explicit desired entry; DNS search/options are still not placed on arbitrary down ifaces.

Resolves: https://redhat.atlassian.net/browse/RHEL-113828

### DIFF
--- a/rust/src/lib/dns.rs
+++ b/rust/src/lib/dns.rs
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::{
+    collections::HashSet,
     net::{Ipv4Addr, Ipv6Addr},
     str::FromStr,
 };
@@ -294,22 +295,40 @@ impl MergedDnsState {
         self.servers.is_empty()
             && (!self.searches.is_empty() || !self.options.is_empty())
     }
+
+    /// Servers explicitly set in desired `dns-resolver.config.server`.
+    ///
+    /// Returns `None` when desired DNS omits `server` (including when DNS is
+    /// omitted entirely). Inherited current servers are not included.
+    pub(crate) fn desired_servers(&self) -> Option<&[String]> {
+        self.desired
+            .as_ref()
+            .and_then(|d| d.config.as_ref())
+            .and_then(|c| c.server.as_deref())
+    }
 }
 
 impl MergedNetworkState {
     // * Specified interface is valid for hold IPv6 DNS config.
     // * Cannot have more than one IPv6 link-local DNS interface.
+    // * Disconnected/down interfaces referenced by *desired* link-local DNS are
+    //   auto-activated so validation and apply can succeed without an explicit
+    //   `interfaces: [{name, state: up}]` entry.
+    // * Inherited current link-local DNS must not trigger auto-activation.
     pub(crate) fn validate_ipv6_link_local_address_dns_srv(
-        &self,
+        &mut self,
     ) -> Result<(), NmstateError> {
-        let mut iface_names = Vec::new();
-        for srv in self.dns.servers.as_slice() {
+        let Some(desired_servers) = self.dns.desired_servers() else {
+            return Ok(());
+        };
+
+        // Immutable checks while borrowing desired DNS; collect link-local
+        // targets so we can mutably update interfaces afterwards.
+        let mut pending = Vec::new();
+        for srv in desired_servers {
             if let Some((_, iface_name)) = parse_dns_ipv6_link_local_srv(srv)? {
-                let iface = if let Some(iface) =
-                    self.interfaces.kernel_ifaces.get(iface_name)
-                {
-                    iface
-                } else {
+                let Some(iface) = self.interfaces.kernel_ifaces.get(iface_name)
+                else {
                     return Err(NmstateError::new(
                         ErrorKind::InvalidArgument,
                         format!(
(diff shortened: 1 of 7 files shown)


rust/src/lib/unit_tests/nm/dns.rs
```diff
@@ -94,6 +94,98 @@ fn test_dns_ipv6_link_local_iface_has_ipv6_disabled() {
...
+    assert!(eth1.merged.is_down());
+    assert!(!eth1.is_changed());
+
+    store_dns_config_to_iface(&mut merged_state, &[], &[]).unwrap();
+
+    let eth1 = merged_state.interfaces.kernel_ifaces.get("eth1").unwrap();
+    assert!(eth1.merged.is_down());
+    if let Some(apply_iface) = eth1.for_apply.as_ref() {
+        assert!(apply_iface.is_down());
+    }
```
jvaclav-rh: In unit tests, if we expect a value to be present, we `.unwrap()` it instead of using `if let Some(...)`

rust/src/lib/unit_tests/nm/dns.rs
```diff
@@ -94,6 +94,98 @@ fn test_dns_ipv6_link_local_iface_has_ipv6_disabled() {
...
+    .unwrap();
+
+    let merged_state =
+        MergedNetworkState::new(desired, current, Default::default(), false)
+            .unwrap();
+
+    let iface = merged_state.interfaces.kernel_ifaces.get("eth1").unwrap();
+    assert!(iface.merged.is_up());
+    assert!(iface.is_changed());
+    assert_eq!(iface.for_apply.as_ref().map(|i| i.is_up()), Some(true));
```
jvaclav-rh: Similarly here, we can simply unwrap the Option, and then we can simply use assert!() instead of assert_eq!()

jvaclav-rh (commented):
One of the commits also has a signoff of `CrumblyBread` instead of `Branislav Kadlec`, they should probably be both the full name because otherwise DCO complains

rust/src/lib/iface.rs
```diff
@@ -1110,6 +1110,31 @@ impl MergedInterface {
...
+    /// Used when a disconnected/down interface is referenced by other desired
+    /// config (IPv6 link-local DNS, route-rule `iif`, etc.) and should be
+    /// auto-activated without the user listing it under `interfaces`.
+    ///
+    /// Does nothing when the interface is already in the desired state, so an
+    /// explicit `state: down` is not overridden.
+    pub(crate) fn mark_as_up_for_apply(&mut self) {
+        if self.desired.is_some() {
+            return;
+        }
```
jvaclav-rh: This function will set interfaces which were implicitly marked as `absent` (e.g. in mark_orphan_interface_as_absent() when a VLAN's parent bond is removed) as `up`, because `self.desired` is None for those, yet they were changed.

jvaclav-rh: Maybe the following would work:?
```
        if self.desired.is_some() || self.is_changed() {
            return;
        }
        self.mark_as_changed();
        // ...
```

It would be good to introduce a test for this as well.

rust/src/lib/dns.rs
```diff
@@ -319,8 +337,25 @@ impl MergedNetworkState {
                         ),
                     ));
                 };
+                if iface.merged.is_ignore() {
+                    return Err(NmstateError::new(
+                        ErrorKind::InvalidArgument,
+                        format!(
+                            "Desired IPv6 link local DNS server {srv} is \
+                             pointing to unmanaged/ignored interface \
```
jvaclav-rh: "unmanaged" is NM-specific terminology, we should probably just say ignored.

rust/src/lib/dns.rs
```diff
@@ -294,22 +294,40 @@ impl MergedDnsState {
...
+    //   `interfaces: [{name, state: up}]` entry.
+    // * Inherited current link-local DNS must not trigger auto-activation.
     pub(crate) fn validate_ipv6_link_local_address_dns_srv(
-        &self,
+        &mut self,
     ) -> Result<(), NmstateError> {
+        let Some(desired_servers) = self.dns.desired_servers() else {
+            return Ok(());
+        };
+        let desired_servers = desired_servers.to_vec();
```
jvaclav-rh: I think this is unnecessary, it should be possible to iterate over `desired_servers` directly already.

jvaclav-rh (changes requested):
The two commits should be probably squashed into one, I suspect the first commit is not correct just by itself.

rust/src/lib/dns.rs
```diff
@@ -348,24 +348,33 @@ impl MergedNetworkState {
...
+            }
+        }
+
+        let mut iface_names = Vec::new();
+        for (srv, iface_name) in pending {
+            let iface = self
+                .interfaces
+                .kernel_ifaces
+                .get_mut(iface_name.as_str())
+                .expect("BUG: interface existence checked above");
```
jvaclav-rh: We do not use .expect() in library code, the function should handle this gracefully and return an error with ErrorKind::Bug

jvaclav-rh:
@coderabbitai review

jvaclav-rh (approved):

PR #3185, opened 2026-07-29
nm: deactivate when auto-route-table-id changes

CrumblyBread:
## Summary
- Detect changes to NM `ipv4.route-table` / `ipv6.route-table` (`auto-route-table-id`).
- Treat them like other settings that need deactivate/reactivate instead of in-place reapply.
- Fixes leftover IPv6 `proto kernel` routes in the old table after changing `auto-route-table-id` (e.g. 1000 → 2000).

Resolves: https://redhat.atlassian.net/browse/RHEL-178772

Assisted-By: Claude Sonnet 5

### DIFF
--- a/rust/src/lib/nm/query_apply/apply.rs
+++ b/rust/src/lib/nm/query_apply/apply.rs
@@ -14,8 +14,8 @@ use super::super::{
         delete_orphan_ovs_ports, dispatch::apply_dispatch_script,
         dns::store_dns_config, is_hsr_changed, is_ip_tunnel_changed,
         is_ipvlan_changed, is_mptcp_flags_changed, is_route_removed,
-        is_veth_peer_changed, is_vlan_changed, is_vrf_table_id_changed,
-        is_vxlan_changed, save_nm_connections,
+        is_route_table_changed, is_veth_peer_changed, is_vlan_changed,
+        is_vrf_table_id_changed, is_vxlan_changed, save_nm_connections,
     },
     route::store_route_config,
     route_rule::store_route_rule_config,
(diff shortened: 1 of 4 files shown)


CrumblyBread:
@coderabbitai review

CrumblyBread:
@coderabbitai review

rust/src/lib/nm/query_apply/apply.rs
jvaclav-rh, line 1: In the commit message, the `Resolves` link should be to the RHEL- ticket, I don't think the NMT one is public.

tests/integration/dynamic_ip_test.py
```diff
@@ -450,6 +450,56 @@ def test_ipv6_dhcp_set_table_id_with_autoconf(dhcpcli_up):
...
+
+    assert _poll(_has_dhcpv4_addr)
+    assert _poll(_has_dhcpv6_addr)
+    assert _poll(_has_iface_route_in_table, 4, 2000)
+    assert _poll(_has_kernel_route_in_table, 6, 2000)
+    # _poll_till_not returns False once the condition clears.
+    assert not _poll_till_not(_has_iface_route_in_table, 4, 1000)
+    assert not _poll_till_not(_has_kernel_route_in_table, 6, 1000)
+    assert not _poll_till_not(_has_iface_route_in_table, 6, 1000)
+
```
jvaclav-rh: This seems to me as a very confusing way to express these test expectations, but it's not really the fault of this PR. I feel like we ought to refactor this eventually and instead put a `not` inside _poll_till_not(), considering all of its invocations are `assert not _poll_till_not(...)`

CrumblyBread:
https://github.com/coderabbitai review

rust/src/lib/nm/query_apply/ip.rs
jvaclav-rh, line 1: This change does not really belong in ip.rs, it would be more suitable to put it in route.rs.

rust/src/lib/nm/query_apply/apply.rs
jvaclav-rh, line 357: This comment should most likely be updated.

jvaclav-rh (approved):

PR #3188, opened 2026-08-01
ifaces: stop silently dropping invalid desired state input

pfeifferj:
Several places in desired-state deserialization silently accepted input that had no effect

### DIFF
--- a/rust/src/lib/iface.rs
+++ b/rust/src/lib/iface.rs
@@ -301,26 +301,34 @@ impl<'de> Deserialize<'de> for Interface {
                 .map_err(serde::de::Error::custom)?,
             Some(InterfaceState::Absent)
         ) {
-            let mut new_value = serde_json::map::Map::new();
-            if let Some(n) = v.get("name") {
-                new_value.insert("name".to_string(), n.clone());
+            const ABSENT_KEPT_PROPS: &[&str] = &[
+                "name",
+                "type",
+                "state",
+                "identifier",
+                "mac-address",
+                "pci-address",
+            ];
+            if let Some(obj) = v.as_object_mut() {
+                let mut discarded: Vec<String> = Vec::new();
+                obj.retain(|k, _| {
+                    let keep = ABSENT_KEPT_PROPS.contains(&k.as_str());
+                    if !keep {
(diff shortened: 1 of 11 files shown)


rust/src/lib/iface.rs
```diff
@@ -301,24 +301,35 @@ impl<'de> Deserialize<'de> for Interface {
...
+                            .unwrap_or("unknown")
+                    );
+                }
+                for prop in ABSENT_KEPT_PROPS {
+                    if let Some(val) = obj.get(*prop) {
+                        new_value.insert(prop.to_string(), val.clone());
+                    }
+                }
             }
             v = serde_json::value::Value::Object(new_value);
```
jvaclav-rh: I think this could be simplified a bit by taking `v.as_object_mut()`, and then .retain()ing the kept props in the original object, which should let us avoid `new_value` entirely as well as the reassignment

ihuguet: I agree. In the function passed to `retain()`, apart from returning true/false for the filtering, you can create on the fly a Vec with the discarded keys so you can use it later for the warn message.

rust/src/lib/unit_tests/testlib.rs
```diff
@@ -304,3 +304,58 @@ pub(crate) fn gen_merged_ifaces_for_route_test()
 
     (merged, current)
 }
+
+pub(crate) mod log_capture {
```
ihuguet, line 309: I think I don't like this much, but I cannot tell exactly why, so I won't block it. Won't it be better to use a dedicated crate?

ihuguet (commented):
Mostly LGTM. Great improvements, thanks!

rust/src/lib/unit_tests/testlib.rs
pfeifferj: fair, i went looking for a crate first but the lib crate currently only has serde_yaml as a dev-dependency, and releases ship a vendored tarball, so anything i add gets vendored into every rpm build. which seems like a lot of overhead for a small test helper.

jvaclav-rh (approved):

PR #3189, opened 2026-08-04
test: fix race in IPv6 link-local detection

vbenes:
Some kernels/NM versions are slow to assign the link-local address after IPv6 is enabled, causing a race condition where the address key is missing in the returned state. Use retry_till_true_or_timeout to wait up to 10 seconds for the link-local address to appear.


jvaclav-rh (approved):

vbenes:
[retest]

jvaclav-rh:
@coderabbitai review

PR #3191, opened 2026-08-04
test: kernel mode test group

CrumblyBread:
Introduce a dedicated `@pytest.mark.kernel group` for kernel-mode (nispor / kernel_only) integration tests.
Keep those cases out of auto-tier2 so they run in their own group (tests that are also tier1 still run there).
Add integ_kernel to automation/run-tests.sh and document it.
Enforce the group in CI via mandatory integ_kernel matrix jobs (c9s stable/main/min, fed main, c10s stable/main).

Resolves: https://redhat.atlassian.net/browse/NMT-1381


.github/workflows/main.yml
```diff
@@ -127,19 +127,25 @@ jobs:
...
           - job_type: "fed-nm_main-integ_tier1"
+          - job_type: "fed-nm_main-integ_kernel"
           - job_type: "c10s-nm_stable-integ_tier1"
           - job_type: "c10s-nm_stable-integ_tier2"
+          - job_type: "c10s-nm_stable-integ_kernel"
           - job_type: "c10s-nm_main-integ_tier1"
           - job_type: "c10s-nm_main-integ_tier2"
+          - job_type: "c10s-nm_main-integ_kernel"
     steps:
       - uses: actions/checkout@v7
```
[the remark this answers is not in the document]
jvaclav-rh: I don't really understand what ramifications the addition or lack of this `persist-credentials`, but i definitely think it does not belong into this PR.

automation/run-tests.sh
```diff
@@ -106,8 +107,8 @@ function install_nmstate {
 }
 
 function run_tests {
-    if [ $TEST_TYPE == $TEST_TYPE_ALL ] || \
-       [ $TEST_TYPE == $TEST_TYPE_RUST_GO ];then
+    if [ "$TEST_TYPE" == "$TEST_TYPE_ALL" ] || \
+       [ "$TEST_TYPE" == "$TEST_TYPE_RUST_GO" ];then
```
jvaclav-rh: $TEST_TYPE is always a single token without whitespace, so this is not necessary (same below)

automation/run-tests.sh
```diff
@@ -337,6 +355,22 @@ done
...
+    "$TEST_TYPE_INTEG_SLOW"|\
+    "$TEST_TYPE_INTEG_KERNEL")
+        ;;
+    *)
+        echo "Invalid --test-type value: '$TEST_TYPE'" >&2
+        echo "Expected one of: all, integ, integ_tier1, integ_tier2, integ_slow, integ_kernel, rust_go" >&2
+        exit 1
+        ;;
+esac
+
```
jvaclav-rh: If we are going to add invalid input handling, we should probably just do an else-if in run_tests() and put this in the `else` branch

tests/integration/conftest.py
```diff
@@ -41,6 +41,10 @@ def pytest_configure(config):
     config.addinivalue_line("markers", "slow: mark time consuming test")
     config.addinivalue_line("markers", "tier2")
     config.addinivalue_line("markers", "tier1")
+    config.addinivalue_line(
+        "markers",
+        "kernel: mark kernel-mode (nispor, no NetworkManager) tests",
+    )
```
jvaclav-rh: i think just:
```suggestion
    config.addinivalue_line("markers", "kernel")
```

is fine

jvaclav-rh (approved):

PR #3196, opened 2026-08-10
route: Reject conflicting special routes

CrumblyBread:
Reject desired states that put more than one special route (blackhole / unreachable / prohibit) on the same destination and table. The kernel only keeps one and nmstate used to apply silently with no error.
Handle no-next-hop special routes in MergedRoutes so a route-type change marks the old route absent and shows up in gen_diff.
Unit and integration coverage for the collision error and for reporting a type change.

Resolves: https://redhat.atlassian.net/browse/RHEL-82785

### DIFF
--- a/rust/src/lib/route.rs
+++ b/rust/src/lib/route.rs
@@ -930,7 +930,7 @@ fn validate_route_dst(route: &RouteEntry) -> Result<(), NmstateError> {
                         ));
                     }
                 };
-                if prefix >= 8 && route.is_unicast() {
+                if prefix >= 8 && route.is_unicast() && !route.is_absent() {
                     let e = NmstateError::new(
                         ErrorKind::InvalidArgument,
                         "0.0.0.0/8 and its subnet cannot be used as the route \
@@ -941,7 +941,7 @@ fn validate_route_dst(route: &RouteEntry) -> Result<(), NmstateError> {
                     log::error!("{e}");
                     return Err(e);
                 }
-            } else if route.is_unicast() {
+            } else if route.is_unicast() && !route.is_absent() {
                 let e = NmstateError::new(
                     ErrorKind::InvalidArgument,
                     "0.0.0.0/8 and its subnet cannot be used as the route \
--- a/rust/src/lib/route.rs
+++ b/rust/src/lib/route.rs
@@ -17,7 +17,6 @@ use crate::{
 const DEFAULT_TABLE_ID: u32 = 254; // main route table ID
 const LOOPBACK_IFACE_NAME: &str = "lo";
 // Kernel metric for user/static IPv6 routes with unset/0 metric.
-#[cfg(feature = "query_apply")]
 const IP6_RT_PRIO_USER: u32 = 1024;
 
 #[derive(Debug, Clone, PartialEq, Eq, Default, Serialize, Deserialize)]
@@ -565,7 +564,6 @@ impl RouteEntry {
 
     // Metric the kernel will assign, for conflict detection. Unset/0 IPv6
     // metric is coerced to 1024, IPv4 to 0.
-    #[cfg(feature = "query_apply")]
     pub(crate) fn effective_metric(&self) -> u32 {
         match self.metric.and_then(|m| u32::try_from(m).ok()) {
             Some(m) if m > 0 => m,
@@ -576,7 +574,6 @@ impl RouteEntry {
 
     // Table the kernel will assign, for conflict detection. Unset table is
     // the main table.
-    #[cfg(feature = "query_apply")]
     pub(crate) fn effective_table_id(&self) -> u32 {
         match self.table_id {
             None | Some(Self::USE_DEFAULT_ROUTE_TABLE) => DEFAULT_TABLE_ID,
@@ -587,6 +584,18 @@ impl RouteEntry {
     pub(crate) fn is_unicast(&self) -> bool {
         self.route_type.is_none()
             || u8::from(self.route_type.unwrap()) == RTN_UNICAST
+    }
+
+    // True when both are non-unicast routes to the same destination, table,
+    // and metric. Netlink rejects a second route for the same triple on
+    // normal `ip route add` (IPv4/IPv6 fib); different metrics may coexist.
+    pub(crate) fn is_special_route_conflict_with(&self, other: &Self) -> bool {
+        !self.is_unicast()
+            && !other.is_unicast()
+            && self.destination.is_some()
+            && self.destination == other.destination
+            && self.effective_table_id() == other.effective_table_id()
+            && self.effective_metric() == other.effective_metric()
     }
 }
 
@@ -643,6 +652,9 @@ impl std::fmt::Display for RouteEntry {
         }
         if let Some(v) = self.table_id.as_ref() {
             props.push(format!("table-id: {v}"));
+        }
+        if let Some(v) = self.route_type.as_ref() {
+            props.push(format!("route-type: {v}"));
         }
         if let Some(v) = self.weight {
             props.push(format!("weight: {v}"));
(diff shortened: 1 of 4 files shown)


rust/src/lib/route.rs
```diff
@@ -941,7 +1030,7 @@ fn validate_route_dst(route: &RouteEntry) -> Result<(), NmstateError> {
                     log::error!("{e}");
                     return Err(e);
                 }
-            } else if route.is_unicast() {
+            } else if route.is_unicast() && !route.is_absent() {
                 let e = NmstateError::new(
```
jvaclav-rh, line 1042: It's not very clear to me how this relates to the special routes change, it should be probably a separate commit with an explanation in the description.

rust/src/lib/unit_tests/route.rs
```diff
@@ -937,3 +937,223 @@ fn test_route_lock_mtu_deserialize_from_string() {
...
+        r#"
+        config:
+        - destination: 0.0.0.0/8
+          route-type: blackhole
+          metric: 200
+          table-id: 99
+        - destination: 0.0.0.0/8
+          route-type: prohibit
+          metric: 300
+          table-id: 99
```
jvaclav-rh, line 972: This test is completely wrong, this is a valid config. See:
```
# ip route add blackhole 0.0.0.0/8 proto static metric 200 table 99
# ip route add prohibit 0.0.0.0/8 proto static metric 300 table 99
# ip route list table 99
blackhole 0.0.0.0/8 proto static metric 200 
prohibit 0.0.0.0/8 proto static metric 300 
```

rust/src/lib/unit_tests/route.rs
```diff
@@ -937,3 +937,223 @@ fn test_route_lock_mtu_deserialize_from_string() {
...
+
+    let merged_ifaces = MergedInterfaces::new(
+        Interfaces::default(),
+        cur_ifaces,
+        Default::default(),
+        false,
+    )
+    .unwrap();
+
+    MergedRoutes::new(routes, Routes::new(), &merged_ifaces).unwrap();
```
jvaclav-rh: We should probably assert something here (and other tests with no asserts), like that the routes will truly be applied - something like assert_eq!(...changed_routes().len, 2)

rust/src/lib/route.rs
```diff
@@ -576,18 +576,34 @@ impl RouteEntry {
...
+
+    // True when both are non-unicast routes to the same destination and table.
+    // The kernel only allows one special route type per destination/table.
+    pub(crate) fn is_special_route_conflict_with(&self, other: &Self) -> bool {
+        !self.is_unicast()
+            && !other.is_unicast()
+            && self.destination.is_some()
+            && self.destination == other.destination
+            && self.table_id_or_default() == other.table_id_or_default()
+    }
```
jvaclav-rh, line 599: This is missing a comparison of the (effective) metrics, which then results in two routes with distinct metrics being rejected, e.g.:
```
  - destination: 0.0.0.0/8
    route-type: blackhole
    metric: 100
    table-id: 99
  - destination: 0.0.0.0/8
    route-type: prohibit
    metric: 200
    table-id: 99
```

rust/src/lib/route.rs
```diff
@@ -576,18 +576,34 @@ impl RouteEntry {
...
         }
     }
 
+    // Table the kernel will assign, for conflict detection. Unset table is
+    // the main table.
+    #[cfg(feature = "query_apply")]
+    pub(crate) fn effective_table_id(&self) -> u32 {
+        self.table_id_or_default()
+    }
+
```
jvaclav-rh: I don't think this is necessary, gating is_special_route_conflict_with with cfg(query_apply) should just work?

tests/integration/route_test.py
```diff
@@ -301,6 +301,83 @@ def test_apply_route_with_route_type_multiple_times(eth1_up):
...
+                            Route.DESTINATION: "0.0.0.0/8",
+                            Route.ROUTETYPE: Route.ROUTETYPE_BLACKHOLE,
+                            Route.TABLE_ID: 99,
+                        },
+                    ]
+                }
+            }
+        )
+
+
```
jvaclav-rh, line 365: The try-catch-finally here is redundant, we have clean_up_route_rule(). Irt should be just written as a normal function body, see surrounding tests.

tests/integration/route_test.py
```diff
@@ -301,6 +301,83 @@ def test_apply_route_with_route_type_multiple_times(eth1_up):
...
+        changed = diff.get(Route.KEY, {}).get(Route.CONFIG, [])
+        assert any(
+            rt.get(Route.ROUTETYPE) == Route.ROUTETYPE_PROHIBIT
+            and rt.get(Route.STATE) != Route.STATE_ABSENT
+            for rt in changed
+        ), f"expected prohibit route in gen_diff, got {changed}"
+        libnmstate.apply({Route.KEY: {Route.CONFIG: prohibit}})
+        routes_out = _get_routes_from_iproute(4, 99)
+        assert Route.ROUTETYPE_PROHIBIT in routes_out
+        assert Route.ROUTETYPE_BLACKHOLE not in routes_out
```
jvaclav-rh: We are missing an assert here to check that the blackhole route appears as absent.

rust/src/lib/route.rs
```diff
@@ -907,6 +959,43 @@ impl MergedRoutes {
...
+        .iter()
+        .filter(|rt| !rt.is_absent() && !rt.is_unicast())
+    {
+        let dst = if let Some(dst) = route.destination.as_deref() {
+            dst
+        } else {
+            continue;
+        };
+        let table = route.table_id_or_default();
+        if let Some(existing) = special_routes.insert((dst, table), route) {
```
jvaclav-rh: same deal with the metric here as in is_special_route_conflict_with()

jvaclav-rh (changes requested):

rust/src/lib/route.rs
```diff
@@ -799,6 +819,25 @@ impl MergedRoutes {
...
+                    !des_rt.is_absent()
+                        && des_rt.is_special_route_conflict_with(rt)
+                        && des_rt.route_type != rt.route_type
+                })
+            }
+            #[cfg(not(feature = "query_apply"))]
+            {
+                let _ = rt;
+                false
+            }
```
jvaclav-rh: I think it would be nicer to just un-cfg() is_special_route_conflict_with() and effective_route()

rust/src/lib/route.rs
jvaclav-rh: ^

cathay4t:
• P1:  route.rs:821  infers deletion of the old route, but  revert/route.rs:13-33  does not restore it. Reverting blackhole→prohibit leaves neither route.
• P1:  unit_tests/route.rs:955  fails because absent  0.0.0.0/8  still reaches  validate_route_dst() .
• P2:  route.rs:976  rejects identical duplicate special routes ( blackhole  vs  blackhole ) although merging previously deduplicated them.

Kernel limitation is only partial: normal  ip route add  rejects same destination/table/metric for IPv4 and IPv6 ( fib_trie.c:1232 ,  ip6_fib.c:1120 ). Different metrics coexist. However, IPv4 can store both types using  append ; IPv6 cannot because duplicate-nexthop comparison ignores route type ( ip6_route.h:351 ).  replace  keeps only the new route. The PR should describe this as backend/netlink semantics, not a universal kernel storage limitation.

The commit body is also unrelated SR-IOV/max-vfs text with the wrong Jira reference.

tests/integration/route_test.py
```diff
@@ -301,6 +301,66 @@ def test_apply_route_with_route_type_multiple_times(eth1_up):
...
+            Route.TABLE_ID: 99,
+        },
+    ]
+    with pytest.raises(NmstateValueError):
+        libnmstate.apply({Route.KEY: {Route.CONFIG: routes}})
+
+
+@pytest.mark.tier1
+def test_change_special_route_type_is_reported():
+    blackhole = [
```
jvaclav-rh, line 328: Please add a comment with the RHEL- ticket link to these (see e.g. test_add_route_with_initcwnd_and_initrwnd)

jvaclav-rh (approved):
lgtm with small comment

PR #3198, opened 2026-08-12
test/ip_tunnel: add gretap regression test

jvaclav-rh:
Adds a test for the regression reported in https://redhat.atlassian.net/browse/RHEL-170926

### DIFF
--- a/tests/integration/ip_tunnel_test.py
+++ b/tests/integration/ip_tunnel_test.py
@@ -3,12 +3,39 @@
 import pytest
 
 import libnmstate
+from libnmstate.schema import Interface
+from libnmstate.schema import InterfaceType
 
 from .testlib import assertlib
+from .testlib.cmdlib import exec_cmd
 from .testlib.iptunnellib import ip_tunnel_interface
 
 IP_TUNNEL0 = "ip_tunnel_test0"
+GRETAP0 = "gretap_test0"
 ETH1 = "eth1"
+
+
+@pytest.fixture(scope="function")
+def gretap_up():
+    exec_cmd(
+        [
+            "ip",
+            "link",
+            "add",
+            GRETAP0,
+            "type",
+            "gretap",
+            "local",
+            "192.0.2.1",
+            "remote",
+            "192.0.2.2",
+        ],
+        check=True,
+    )
+    try:
+        yield
+    finally:
+        exec_cmd(["ip", "link", "del", GRETAP0])
 
 
 # https://issues.redhat.com/browse/RHEL-79801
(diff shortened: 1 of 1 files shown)


tests/integration/ip_tunnel_test.py
```diff
@@ -3,14 +3,41 @@
...
+            "192.0.2.1",
+            "remote",
+            "192.0.2.2",
+        ],
+        check=True,
+    )
+    try:
+        yield
+    finally:
+        exec_cmd(["ip", "link", "del", GRETAP0])
```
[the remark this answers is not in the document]
jvaclav-rh: if the interface doesn't exist, raising an exception in finally would mask the original error.

tests/integration/ip_tunnel_test.py
```diff
@@ -59,3 +86,15 @@ def test_ipip6_tunnel_and_remove():
...
+# https://issues.redhat.com/browse/RHEL-170926
+@pytest.mark.tier1
+def test_gretap_show(gretap_up):
+    state = libnmstate.show()
+    iface_names = [i[Interface.NAME] for i in state[Interface.KEY]]
+    assert GRETAP0 in iface_names
+    gretap_iface = next(
+        i for i in state[Interface.KEY] if i[Interface.NAME] == GRETAP0
+    )
+    assert gretap_iface[Interface.TYPE] != InterfaceType.ETHERNET
```
CrumblyBread, line 100: I think this would still pass if the interface were reported as unknown, other, or some other incorrect type. `assert gretap_iface[Interface.TYPE] != InterfaceType.IP_TUNNEL` would be better here.

CrumblyBread, line 100: It might be useful to verify the full state with:
 ```
ip_tunnel = gretap_iface[IpTunnel.CONFIG_SUBTREE]
assert ip_tunnel[IpTunnel.LOCAL] == "192.0.2.1"
assert ip_tunnel[IpTunnel.REMOTE] == "192.0.2.2"
```
to prevent a future regression. Might be premature, idk

File: tests/integration/ip_tunnel_test.py [hunk 3]
```diff
@@ -59,3 +86,15 @@ def test_ipip6_tunnel_and_remove():
         libnmstate.apply(state)
         assertlib.assert_state(state)
     assertlib.assert_absent(IP_TUNNEL0)
+
+
+# https://issues.redhat.com/browse/RHEL-170926
+@pytest.mark.tier1
+def test_gretap_show(gretap_up):
```
CrumblyBread, line 93: Test name too vague? I would choose something like `test_gretap_not_reported_as_ethernet`

jvaclav-rh: Good point

tests/integration/ip_tunnel_test.py
jvaclav-rh: gretap here is an unknown interface, so it will not have any IpTunnel subtree

jvaclav-rh: I'll narrow it to ` == "gretap"`

PR #3199, opened 2026-08-17
Revert "ci: pin crun path to /usr/local/bin"

jvaclav-rh:
This reverts commit 6249dc90fd3c9ae1fac9d8608d22361358fe2a9d.

The underlying problem should now be resolved, so this workaround is no longer needed.


CrumblyBread (approved):
Maybe a short comment about it would be nice like the AI suggests but otherwise LGTM

PR #3203, opened 2026-08-19
nispor: adapt ip6tnl flow-label to nispor 2.0.3

pfeifferj:
netlink-packet-route 0.33 (via nispor 2.0.3) now decodes IFLA_IPTUN_FLOWINFO into host byte order, so the u32::from_be() compensation double-swaps and every ip6tnl flow-label reads as 0, failing verification. Drop the swap and require nispor >= 2.0.3.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -16,7 +16,7 @@ serde_yaml = "0.9"
 serde = {version = "1.0.137", default-features = false, features = ["derive"]}
 serde_json = { version = "1.0.75", default-features = false }
 nmstate = { path = "src/lib", version = "2.2", default-features = false }
-nispor = "2.0.2"
+nispor = "2.0.3"
 uuid = { version = "1.1 ", default-features = false, features = ["v4"] }
 nix = { version = "0.31", default-features = false, features = ["feature", "hostname"] }
 zbus = { version = "5.1.0", default-features = false, features = ["tokio"] }
--- a/rust/src/lib/nispor/ip_tunnel.rs
+++ b/rust/src/lib/nispor/ip_tunnel.rs
@@ -37,9 +37,9 @@ pub(crate) fn np_ip_tunnel_to_nmstate(
                 remote: np_ip_tunnel_info.remote,
                 ttl: np_ip_tunnel_info.ttl,
                 tos: np_ip_tunnel_info.tos,
-                flow_label: np_ip_tunnel_info.flow_info.map(|flow_info| {
-                    u32::from_be(flow_info) & IPV6_FLOWLABEL_MASK
-                }),
+                flow_label: np_ip_tunnel_info
+                    .flow_info
+                    .map(|flow_info| flow_info & IPV6_FLOWLABEL_MASK),
                 ip6tun_flags: np_ip_tunnel_info
                     .ip6tun_flags
                     .as_deref()


rust/src/lib/nispor/ip_tunnel.rs
```diff
@@ -37,9 +37,9 @@ pub(crate) fn np_ip_tunnel_to_nmstate(
                 remote: np_ip_tunnel_info.remote,
                 ttl: np_ip_tunnel_info.ttl,
                 tos: np_ip_tunnel_info.tos,
-                flow_label: np_ip_tunnel_info.flow_info.map(|flow_info| {
-                    u32::from_be(flow_info) & IPV6_FLOWLABEL_MASK
-                }),
+                flow_label: np_ip_tunnel_info
+                    .flow_info
+                    .map(|flow_info| flow_info & IPV6_FLOWLABEL_MASK),
```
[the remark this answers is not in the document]
pfeifferj: flow_info is host order as of netlink-packet-route 0.33 (parse_u32_be), which the nispor >= 2.0.3 floor guarantees at build time; restoring the conversion would reintroduce the double-swap this PR fixes. The contract is documented in the commit message.

rust/src/lib/nispor/ip_tunnel.rs
```diff
@@ -95,3 +95,22 @@ impl From<nispor::IpTunnelMode> for IpTunnelMode {
...
+        np_tunnel.mode = nispor::IpTunnelMode::Ip6ip6;
+        np_tunnel.flow_info = Some(0x02A0_1080);
+        let mut np_iface = nispor::Iface::default();
+        np_iface.ip_tunnel = Some(np_tunnel);
+
+        let iface =
+            np_ip_tunnel_to_nmstate(&np_iface, BaseInterface::default());
+
+        assert_eq!(iface.ip_tunnel.unwrap().flow_label, Some(0x1080));
+    }
```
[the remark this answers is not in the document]
pfeifferj: A host-order literal is the expected representation under the 2.0.3 contract; to_be would encode the opposite.
This test intentionally pins only mask extraction (hence its name); the byte-order contract is enforced by the version floor and covered by netlink-packet-route's decode tests plus the ip_tunnel integration tests.

rust/Cargo.toml
```diff
@@ -16,7 +16,7 @@ serde_yaml = "0.9"
 serde = {version = "1.0.137", default-features = false, features = ["derive"]}
 serde_json = { version = "1.0.75", default-features = false }
 nmstate = { path = "src/lib", version = "2.2", default-features = false }
-nispor = "2.0.2"
+nispor = "2.0.3"
```
jvaclav-rh, line 19: I think bumping the version here by me in 44bf4b51f71768f43aa74c598393e181aaa5b2c5 was a mistake, it should have stayed on 2.0.0 (I did not at the time realize that Rust uses semantic versioning, if there is no other modifier on the version)

rust/src/lib/nispor/ip_tunnel.rs
jvaclav-rh: this is probably not necessary, we already have other tests which catch this

pfeifferj: the integ tests only run in the full ci lanes. I'd rather have this fail earlier/more transparently on our laptops but if you feel strongly about this, we can drop it

jvaclav-rh: i feel like this is kinda a one off thing that should hopefully never change again

pfeifferj: "hopefully never changes again" was also true right up until yesterday but it's fine I dropped it :p

jvaclav-rh (approved):

PR #3216, opened 2026-09-08
ovs: support ovs-db settings on bridge ports

jvaclav-rh:
Allow ovs-db under bridge.port so Linux bonds used as OVS system ports can configure Port table settings such as other_config:stp-path-cost. Apply the settings to the OVS port connection and read them back from OVSDB so they survive show/apply round-trips.

Keep the existing link-aggregation.ovs-db representation for OVS bonds.

Unit and integration tests and example included.

Resolves: https://redhat.atlassian.net/browse/RHEL-124922


examples/ovsbridge_linux_bond_port.yml
```diff
@@ -0,0 +1,35 @@
...
+  state: up
+  bridge:
+    options:
+      stp: true
+    port:
+    - name: br0
+    - name: bond0
+      ovs-db:
+        other_config:
+          stp-path-cost: "2"
```
rajeshrah22, line 21: So, the `other_config` is a key value pair that can have any key? Or is there a list of accepted keys?

If there are restrictions, how will NM handle bad keys?

rajeshrah22 (approved):
I think it looks good. I tried the example in my VM and also ran the integration test.

Just a question about `other_config`.

rust/src/lib/unit_tests/ovs.rs
```diff
@@ -1165,3 +1165,118 @@ fn test_ovs_bridge_ignore_extra_ports_for_apply() {
...
+    conflicting.bond.as_mut().unwrap().ovsdb = Some(Default::default());
+    iface.bridge.as_mut().unwrap().ports = Some(vec![conflicting]);
+    assert_eq!(
+        iface.sanitize(true).unwrap_err().kind(),
+        ErrorKind::InvalidArgument
+    );
+}
+
+#[test]
+fn test_ovs_port_db_verify_clear() {
```
rajeshrah22, line 1251: What is the intention of this test?

examples/ovsbridge_linux_bond_port.yml
jvaclav-rh: It's not anyhow filtered on nmstate side, I think NM does not validate that they are allowed either, that should be responsibility of OVS.

rust/src/lib/unit_tests/ovs.rs
jvaclav-rh: The intention is to check that clearing other_config passes verification even when show omits the empty ovs-db subtree.

PR #3219, opened 2026-09-12
tests: add docs about tests

rajeshrah22:
I think we should add this as a contributor didn't
know about the tiered tests and what the requirements
of a tier1 test are.

General info about the how to the tests are setup
and structured.

Signed-off-by: Rahul Rajesh <rajeshrah22@gmail.com>


jvaclav-rh (approved):
lgtm so far, is there something you would like to add/change before merging?

rajeshrah22:
> lgtm so far, is there something you would like to add/change before merging?

I might change a few things.

PR #3221, opened 2026-09-15
test: add OVS br-ex rollback integration tests

CrumblyBread:
## Summary

- Add `tests/integration/ovs_rollback_test.py` with integration tests for OVS `br-ex` rollback on the three common OCP day1 uplink topologies
- Each test applies a known-good `br-ex` configuration, triggers a failed modification (STP change + `invalid_key`), and verifies nmstate rolls back to the previous state
- Mark all three tests as `tier1` since checkpoint rollback is critical for kubernetes-nmstate and OCP day1 networking

## Test topologies

| Test | Topology |
|------|----------|
| `test_rollback_br_ex_over_vlan_of_ethernet` | `br-ex` → VLAN (`eth101`) → `eth1` |
| `test_rollback_br_ex_over_bond` | `br-ex` → Linux bond (active-backup) → `eth1`/`eth2` |
| `test_rollback_br_ex_over_vlan_of_bond` | `br-ex` → bond → VLAN → `eth1` |

Resolves: https://redhat.atlassian.net/browse/NMT-1148


tests/integration/ovs_rollback_test.py
```diff
@@ -0,0 +1,236 @@
...
+                    Interface.STATE: InterfaceState.ABSENT,
+                },
+                {
+                    Interface.NAME: VLAN_IFNAME,
+                    Interface.STATE: InterfaceState.ABSENT,
+                },
+            ]
+        }
+    )
+
```
jvaclav-rh, line 287: Maybe this could be merged into one shared cleanup and used by all the test cases? E.g. +- some pseudocode

```
apply({ NAME: BR_EX, TYPE: OVS_BRIDGE, STATE: ABSENT },
{ NAME: BOND1, TYPE: BOND, STATE: ABSENT },
{ NAME: VLAN_IFNAME, TYPE: VLAN, STATE: ABSENT })
```

rust/src/lib/nm/query_apply/connection.rs
```diff
@@ -82,15 +84,26 @@ pub(crate) async fn save_nm_connections(
...
+            // protocol change which requires full activate/recreate).
+            let nm_acs = nm_api
+                .active_connections_get()
+                .await
+                .map_err(nm_error_to_nmstate)?;
+            let acs_by_uuid: HashMap<&str, &NmActiveConnection> = nm_acs
+                .iter()
+                .map(|nm_ac| (nm_ac.uuid.as_str(), nm_ac))
+                .collect();
+
```
jvaclav-rh, line 102: How is this change related to what is described in the commit message? If it fixes a regression, it should be in a separate commit with an explanation + regression test which was red before and green after.

rust/src/lib/nm/query_apply/connection.rs
jvaclav-rh, line 1: To test this fix, maybe we could extend `test_change_vlan_protocol` like so:
```diff
diff --git i/tests/integration/vlan_test.py w/tests/integration/vlan_test.py
index 7904ad6a..156c2c27 100644
--- i/tests/integration/vlan_test.py
+++ w/tests/integration/vlan_test.py
@@ -312,7 +312,7 @@ def test_preserve_existing_vlan_conf(eth1_up):
         assertlib.assert_state(desired_state)


-def test_change_vlan_protocol(vlan_on_eth1):
+def test_change_vlan_protocol(vlan_on_eth1, caplog):
     dot1q_state = {
         Interface.KEY: [
             {
@@ -348,11 +348,27 @@ def test_change_vlan_protocol(vlan_on_eth1):
     )
     assertlib.assert_state_match(qinq_state)

+    caplog.clear()
+    caplog.set_level("INFO", logger="libnmstate")
     apply_with_description(
         "Create the vlan interface eth1.101 with ID 102, "
         "using vlan protocol 802.1q",
         dot1q_state,
     )
+    # Changing the VLAN protocol requires deactivation and activation.
+    # Check the logs because failed reapply also falls back to activation.
+    vlan_logs = [
+        message
+        for message in caplog.messages
+        if f": {VLAN_IFNAME}/vlan" in message
+    ]
+    assert any(
+        "Deactivating connection " in msg for msg in vlan_logs
+    ), vlan_logs
+    assert any("Activating connection " in msg for msg in vlan_logs), vlan_logs
+    assert not any(
+        "Reapplying connection " in msg for msg in vlan_logs
+    ), vlan_logs
     assertlib.assert_state_match(dot1q_state)


```

WDYT?

CrumblyBread: Worth adding. Final state alone does not catch that regression. reapply_or_activate() falls back to activate when reapply fails, so the test can stay green while still taking the wrong path. I will get on it.

rust/src/lib/nm/query_apply/apply.rs
```diff
@@ -247,6 +255,21 @@ async fn delete_ifaces(
         }
     }
 
+    for uuid in &ovs_iface_uuids_to_deactivate {
+        log::info!(
+            "Deactivating OVS internal interface connection {uuid} before \
+             deletion"
+        );
+        if let Err(e) = nm_api.connection_deactivate(uuid).await {
```
jvaclav-rh, line 263: I think this `.await` actually only waits for the D-Bus reply from NM rather than for the connection to actually become deactivated?

It looks like deletion could still overlap teardown. Do we have a reproducer showing that this change fixes the stale ports without the separate DOWN step in the test helper?

tests/integration/testlib/genconf.py
```diff
@@ -32,22 +55,28 @@ def gen_conf_apply(desire_state):
...
-        absent_state = {
+        # OVS needs interfaces brought down before absent so ports detach from
+        # the bridge cleanly. Other virtual types (e.g. mac-vtap) fail
+        # verification on DOWN, so skip that phase unless OVS is present.
+        if _has_ovs_iface(desire_state):
+            cleanup_down = {
+                Interface.KEY: _cleanup_ifaces(
+                    desire_state, InterfaceState.DOWN
+                )
+            }
```
jvaclav-rh, line 66: The DOWN step still includes all interfaces in the desired state whenever OVS is present. Could we limit it to the interfaces that need it?

tests/integration/testlib/genconf.py
```diff
@@ -32,22 +55,28 @@ def gen_conf_apply(desire_state):
...
+            libnmstate.apply(cleanup_down)
```
jvaclav-rh, line 67: If this DOWN apply fails, it will skip the ABSENT cleanup and file removal below. Could we use try/finally to make sure those are still attempted?

rust/src/lib/nm/query_apply/apply.rs
CrumblyBread: Yes, you are right. I checked `nm_api.rs`. I will addres it by adding a wait until active connections are gone.

rust/src/lib/nm/query_apply/connection.rs
```diff
@@ -82,15 +84,26 @@ pub(crate) async fn save_nm_connections(
...
     nm_conns: &[NmConnection],
-    conn_matcher: &NmConnectionMatcher,
 ) -> Result<(), NmstateError> {
     let mut nm_conns = nm_conns.to_vec();
     for i in 1..ACTIVATION_RETRY_COUNT + 1 {
         if !nm_conns.is_empty() {
+            // Refresh ACs after deactivate/delete so we do not reapply a
+            // connection that was intentionally deactivated (e.g. VLAN
+            // protocol change which requires full activate/recreate).
+            let nm_acs = nm_api
```
pfeifferj: refreshing the list doesn't mean deactivation has finished. i tried hitting this with a bounded pre-down hook that holds the vlan's deactivation until a replacement activation appears. `test_change_vlan_protoco`l fails with desired 802.1q, current 802.1ad.

the deactivating connection is still in this snapshot, so we try reapply, get rejected, and fall back to activation. that queues the replacement before teardown finishes. nm keeps the existing vlan device, including its old protocol.

we need to wait for deactivation and link teardown before requesting activation here. filtering out deactivating connections alone still queues activation too early. 

pls also cover this delayed-deactivation case in the regression test

tests/integration/ovs_rollback_test.py
```diff
@@ -0,0 +1,310 @@
...
+            libnmstate.show(include_secrets=True)[Interface.KEY],
+            name,
+            iface_type,
+        )
+        if current_iface is not None:
+            current = copy.deepcopy(current_iface)
+            _sort_ovs_bridge_ports(current)
+            # Interface state can flap briefly during rollback; assertlib also
+            # ignores it during verify.
+            expected.pop(Interface.STATE, None)
```
pfeifferj: I would keep state in the comparison

tests/integration/ovs_rollback_test.py
```diff
@@ -0,0 +1,310 @@
...
+    modified_state = copy.deepcopy(initial_state)
+    _trigger_post_apply_verification_failure(modified_state)
+
+    with pytest.raises(NmstateVerificationError):
+        libnmstate.apply(modified_state)
+
+    time.sleep(5)
+    # Compare desired topology only (by name+type). A full show() snapshot
+    # includes ephemeral runtime fields that often differ after rollback.
+    _assert_interfaces_state_match(topology_ifaces)
```
pfeifferj: the test doesn't check whether the `203.0.113.0/24` via br-ex we add to trigger the failure is removed by the rollback afaict

it looks to me like feeding the helper a restored interface snapshot with that route still configured passes the rollback assertions + we should probably check the configured routes too

tests/integration/vlan_test.py
```diff
@@ -348,11 +348,27 @@ def test_change_vlan_protocol(vlan_on_eth1):
...
     )
+    # Changing the VLAN protocol requires deactivation and activation.
+    # Check the logs because failed reapply also falls back to activation.
+    vlan_logs = [
+        message
+        for message in caplog.messages
+        if f": {VLAN_IFNAME}/vlan" in message
+    ]
+    assert any(
+        "Deactivating connection " in msg for msg in vlan_logs
```
pfeifferj: asserting on log messages is an anti-pattern

tests/integration/ovs_rollback_test.py
```diff
@@ -0,0 +1,310 @@
...
+            iface
+            for iface in ifaces
+            if iface[Interface.NAME] == name
+            and iface[Interface.TYPE] == iface_type
+        ),
+        None,
+    )
+
+
+def _sort_ovs_bridge_ports(iface):
```
pfeifferj: nit: `normalize()` already sorts these ports

pfeifferj (changes requested):

tests/integration/vlan_test.py
jvaclav-rh: any better ideas?

tests/integration/ovs_rollback_test.py
CrumblyBread: `normalize()` only sorts Linux-bridge ports, not OVS bridge ports

tests/integration/vlan_test.py
pfeifferj: why not record the kernel `ifindex` with `socket.if_nametoindex()` before applying the protocol change, keep the existing final-state assertion, then assert that the `ifindex` changed?

rust/src/lib/nm/query_apply/apply.rs
```diff
@@ -251,6 +257,36 @@ async fn delete_ifaces(
...
+    }
+    // DeactivateConnection only acknowledges the request; wait until the
+    // active connections are gone so OVS ports are detached before delete.
+    if !ovs_iface_uuids_to_deactivate.is_empty()
+        && let Err(e) = nm_api
+            .wait_connections_inactive(
+                ovs_iface_uuids_to_deactivate.as_slice(),
+                DEACTIVATE_WAIT_TIMEOUT_SEC,
+            )
+            .await
```
jvaclav-rh: If this times out while the connection is still deactivating, we proceed with deletion anyway. wouldn't that allow the same overlap this change is intended to prevent?

tests/integration/ovs_rollback_test.py
pfeifferj: it actually sorts both. it doesn't check the interface type

tests/integration/vlan_test.py
```diff
@@ -23,6 +24,10 @@
 
 VLAN_IFNAME = "eth1.101"
 VLAN2_IFNAME = "eth1.102"
+NM_DISPATCH_FOLDER = "/etc/NetworkManager/dispatcher.d"
```
pfeifferj: this needs to be `/etc/NetworkManager/dispatcher.d/pre-down.d` nm only looks there for pre-down hooks

